PR #463, opened 2017-04-10
Fix interop bugs

SvyatoslavScherbina:


INTEROP.md
```diff
@@ -245,6 +245,23 @@ In all cases the C string is supposed to be encoded as UTF-8.
 When C function takes or returns a struct `T` by value, the corresponding
 argument type or return type is represented as `CValue<T>`.
 
+`CValue<T>` is currently opaque, i.e. it's not yet possible to access struct's
```
olonho: "`CValue<T>` is currently opaque" -> `CValue<T>` is an opaque type, so structure fields cannot be accessed with appropriate Kotlin properties. 
It could be acceptable, if API uses structures as handles, but if field access is required,  there are following conversion methods available:

PR #488, opened 2017-04-14
Fix usage of new Kotlin command line arguments API

udalov:
This will be needed after an update to the new compiler because https://github.com/JetBrains/kotlin/commit/78c0111c6e85b54a2ba4ddd766099dabb4d633bd

~Please do not merge yet. There's `delimiter = " "` used in `linkerArguments`, which is not yet supported in the new API, I'll support it soon~ Fixed now


vvlevchenko:
applied as a part of #500

PR #491, opened 2017-04-15
Wrong workdir when use relative def file path like "-def interop.def"

enbandari:
When I was using cinterop in cmd line, 

```
cinterop -def kotliner.def -o build/kotliner/kotliner.kt.bc
```
 I got a error:

```
/Users/benny/Github/kotlin-native/dist/dependencies/clang-llvm-3.9.0-darwin-macos/bin/clang -Isrc/c -isystem /Users/benny/Github/kotlin-native/dist/dependencies/clang-llvm-3.9.0-darwin-macos/lib/clang/3.9.0/include -B/Users/benny/Github/kotlin-native/dist/dependencies/target-sysroot-1-darwin-macos/usr/bin --sysroot=/Users/benny/Github/kotlin-native/dist/dependencies/target-sysroot-1-darwin-macos -mmacosx-version-min=10.11 -emit-llvm -c build/kotliner/kotliner.kt.bc-build/natives/cstubs.c -o build/kotliner/kotliner.kt.bc-build/natives/cstubs.bc
clang-3.9: error: no such file or directory: 'build/kotliner/kotliner.kt.bc-build/natives/cstubs.c'
clang-3.9: error: no input files
Exception in thread "main" java.lang.Error: Process finished with non-zero exit code: 1
        at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.runExpectingSuccess(main.kt:112)
        at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.runCmd(main.kt:158)
        at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.processLib(main.kt:396)
        at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.main(main.kt:43)
```

It was obviously caused by a wrong "working dir". When I check the code in MainKt.processLib, I found that the workdir was retrieved from defFile passed from the command line and if the defFile was a relative file like File("interop.def") , its parentFile would be null thus making the workdir assigned by a temporary file.


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -17,12 +17,18 @@
 package org.jetbrains.kotlin.native.interop.gen.jvm
 
 import org.jetbrains.kotlin.native.interop.indexer.Language
-import org.jetbrains.kotlin.native.interop.indexer.NativeIndex
 import org.jetbrains.kotlin.native.interop.indexer.NativeLibrary
 import org.jetbrains.kotlin.native.interop.indexer.buildNativeIndex
 import java.io.File
+import java.lang.Class
```
olonho: All additional imports look irrelevant to the fix.

enbandari: Oh, sorry about that. It seems to be auto imported by IntelliJ. I'll check later.

PR #501, opened 2017-04-18
Turned conext.log("text") into context.log{"text"} so that the argument

alexander-gorshenev:
evaluated lazily. This is especially useful when the text is constructed from
large IR pieces using ir2stringWhole().


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LinkStage.kt
```diff
@@ -286,7 +286,7 @@ internal class LinkStage(val context: Context) {
     }
 
     fun linkStage() {
-        context.log("# Compiler root: ${distribution.konanHome}")
+        context.log{"# Compiler root: ${distribution.konanHome}"}
```
olonho: 20 век переходит в 21-й заменой круглых скобок на фигурные :).

PR #503, opened 2017-04-18
Add includedFunctions property to *.def files

AdrianVovk:
### Why?
This PR allows making smaller cstubs by only including a few functions (instead of excluding a few with `exlcudedFunctions`). This potentially decreases compilation times.

### Example
```
headers = time.h
includedFunctions = ctime time
```
This strips all of the functions from `time.h` except `ctime` and `time`

### Conflicting statements
```
headers = time.h
includedFunctions = a b
excludedFunctions = b c
```
In a situation like this, `includedFunctions` will be used, and `excludedFunctions` will be ignored

### Expansion?
Maybe this idea should be expanded for classes as well? The compiler could note down what classes are necessary (from return and parameter types of functions) and users can include extra classes with `includedClasses`

### DIFF
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/InteropConfiguration.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/InteropConfiguration.kt
@@ -25,6 +25,8 @@ class InteropConfiguration(
         val library: NativeLibrary,
         val pkgName: String,
         val excludedFunctions: Set<String>,
+        val includedFunctions: Set<String>,
+        val useIncludedFunctions: Boolean,
         val strictEnums: Set<String>,
         val nonStrictEnums: Set<String>
 )
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
@@ -54,6 +54,12 @@ class StubGenerator(
     val excludedFunctions: Set<String>
         get() = configuration.excludedFunctions
 
+	val includedFunctions: Set<String>
+	    get() = configuration.includedFunctions
+
+    val useIncludedFunctions: Boolean
+        get() = configuration.useIncludedFunctions
+
     val keywords = setOf(
             "as", "break", "class", "continue", "do", "else", "false", "for", "fun", "if", "in",
             "interface", "is", "null", "object", "package", "return", "super", "this", "throw",
@@ -150,7 +156,13 @@ class StubGenerator(
         get() = decl.kotlinName
 
 
-    val functionsToBind = nativeIndex.functions.filter { it.name !in excludedFunctions }
+    val functionsToBind = nativeIndex.functions.filter { 
+       if (useIncludedFunctions) {
+       	  it.name in includedFunctions
+       } else {
+          it.name !in excludedFunctions
+       }
+    }
 
     private val usedFunctionTypes = mutableMapOf<FunctionType, String>()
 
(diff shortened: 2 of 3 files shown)


AdrianVovk:
I wanted to clarify on the potential implementation of this for classes:

### Current
Users would have this as their def file:
```
headers = api.h
```
And this is the output:
```kotlin
fun func1(param: ClassA) : ClassB
fun func2(param: ClassA) : ClassB

class ClassA
class ClassB
class ClassC
class ClassD
class ClassE
```

### Using This PR
```
headers = api.h
includedFunctions = func1
```

Output:
```
fun func1(param: ClassA) : ClassB
// func2 excluded

class ClassA
class ClassB
class ClassC
class ClassD
class ClassE
```

### Potential Classes Implementation
This detects the classes necessary from parameters and return types

```
headers = api.h
includedFunctions = fun1
```

Output:
```
fun func1(param: ClassA) : ClassB

class ClassA // Required by func1
class ClassB // Required by func1
```

##### With custom included classes
This does the same as above but this shows off how the user can include extra classes

```
headers = api.h
includedFunctions = fun1
includedClasses = ClassE
```

Output:
```
fun func1(param: ClassA) : ClassB

class ClassA // Required by func1
class ClassB // Required by func1

class ClassE // In includedClasses
```

AdrianVovk:
Could you clarify @olonho

olonho:
Sure. Idea is as following, usually huge size of bindings comes from the fact that we build transitive closure of all headers, i.e.

a.h
 
    #include &lt;system.h&gt;
    #include "b.h"
    void a_fun1();

b.h
    #include "c.h"
    void b_fun1();
    void b_fun2();

...

As header files usually group functions naturally, by their functionality, I suggest instead of what you did implement syntax like

headersOnly = a.h b.h

For such cases functions and variables from c.h and system.h will not be emitted, and types defines outside a.h and b.h shall only be emitted if the are used in a.h and b.h.

olonho:
Adrian, your PR moved us to think a little bit more in the direction of this problem, so we what to redesign this part of .def file handling completely, and make it more easily usable.  So would you mind if we will not merge this PR as is, but instead work on more generic approach allowing to use library-level granularity.

AdrianVovk:
@olonho Of course I don't mind! Go ahead and improve it of course. Maybe I can contribute some

-------

Also I wanted to make sure I understood your idea correctly. Instead of including/excluding specific functions and classes, you want to include/exclude header files? 

#### Current
```
headers = a.h
```
Also pulls in `b.h` and `c.h`

#### Potential
```
onlyHeaders = a.h

```
Only pulls in `a.h` and dependencies from `b.h` and `c.h`

Did I get that correctly?

-------

I'd take a little bit of a different spin on the above idea. I would try this:

```
headers = gtk/gtk.h
modules = gtkwindow gtkapplication gtkbutton
```
This would be a better looking syntax.


Also I'd keep my pull request too for things like this:
```
headers = gtk/gtk.h
modules = gtkwindow gtkapplication gtkbutton
includedFunctions = funA funB funC
```

And then add:
```
headers = gtk/gtk.h
modules = gtkwindow gtkapplication gtkbutton
includedFunctions = funA funB funC
includedClasses = ClassA ClassB
```

And then string it together to end up with a system like this:
- `headers` imports everything
- `modules` limits what header files are imported from `headers`
- `excludedModules` excludes header files from `headers`
- `includedFunctions` includes specified functions from the included modules and excludes the rest
- `excludedFunctions` excludes specified functions from the included modules and includes the rest
- `includedClasses` includes specified classes from the included modules and excludes the rest, making sure not to remove classes that the included functions depend on
- `excludedClasses` excludes specified classes from the included modules and includes the rest, throwing an error if the user tries to exclude a class that the included functions depend on

Thoughts? What did your team think of?

olonho:
We are thinking about glob-based filters for header files.  How would you implement 'modules' keyword.

olonho:
See https://github.com/JetBrains/kotlin-native/pull/525 for current approach we took.

AdrianVovk:
It seems that this PR is no longer valid or needed. Closing

PR #515, opened 2017-04-20
GTK sample

olonho:


AdrianVovk:
Is it building now?

olonho:
Yes, but we had to do few changes in interoperability mechanisms to make it fully functional, that's why I didn't mentioned it yet in the issue. Also see https://github.com/JetBrains/kotlin-native/pull/514 - we got GTK stubs compilation down to 6 mins on my box.

PR #534, opened 2017-04-27
Unsigned types, step 0

olonho:


tools/generators/Operators.kt
```diff
@@ -0,0 +1,272 @@
...
+enum class Kind(val signed: Boolean = true,
+                val numeric: Boolean = true,
+                val integral: Boolean = true,
+                val floating: Boolean = false) {
+    UNSIGNED(signed = false),
+    SIGNED,
+    FLOATING(floating = true, integral = false),
+    // Yes, CHAR is numeric, but not integral.
+    CHAR(integral = false),
+    BOOLEAN(numeric = false)
```
ilmat192: Boolean считается целочисленным?

runtime/src/main/cpp/Types.h
```diff
@@ -22,14 +22,18 @@
 #include "TypeInfo.h"
 
 // Note that almost all types are signed.
```
ilmat192: Похоже, что это уже не так :)

tools/generators/Operators.kt
```diff
@@ -0,0 +1,272 @@
...
+        function == "ushr" && typePromotion[type] == Type.INT -> "return static_cast<uint32_t>(arg0) >> (arg1 & 31);"
+        function == "ushr" && typePromotion[type] == Type.LONG -> "return static_cast<uint64_t>(arg0) >> (arg1 & 63);"
+        function in setOf("toByte", "toShort", "toInt", "toLong", "toFloat", "toDouble", "toChar") -> "return arg0;"
+        else -> throw Error("Function $function is not implemented yet")
+    }
+    println("${retType.cTypeName} $symbol($arguments) { $body }")
+}
+
+fun makeSymbolName(type: Type, function: String, args: List<Type>) =
+        "Kotlin_${type}_${(listOf(function) + args.map { arg -> arg.toString() }).joinToString("_")}"
```
SvyatoslavScherbina: > `args.map { arg -> arg.toString() }`

Кстати, `joinToString` сам делает `.toString()` для каждого из элементов.

Arcnor:
Awesome that this is happening! Hopefully something similar can happen for JVM Kotlin 😃 

Also, I'm sure this is not the right place, and maybe it was even discussed already, but for a while I've tried to follow PRs on this repo, only to find that comments are sadly in Russian. Will it be possible (i.e. is it planned) to do future ones in English, so everybody can contribute, even if it's just to understand what's happening on them?

Again, sorry, I know this is not the right place, but this is the last PR I've tried to read with the same problem 😸 .

napperley:
As Edu Garcia has mentioned if the Kotlin community is to get involved then
all PR comments WILL need to be in English, which is the official language
for international communication in the IT industry. One should note that
there are some comments in English however the rest are in Russian, which
is VERY confusing for people to follow. Especially when more than one
language is used which is MUCH worse.

Some YouTube videos are like this unfortunately :( .

olonho:
This particular change will unlikely be integrated as is, and instead generic compiler support for unsigned types will be integrated. 

Regarding communication language: we fully understand your concerns, so we can make review process separate for internal team communications and publically discussed PRs.

Arcnor:
@olonho well, different people have different reasons for reading the PRs. In my case, I like to see what's happening inside the compiler to understand how it will work, so even if this change won't be integrated as it is, it has value for certain people.

Thanks for understanding, and hopefully we can all make Kotlin Native a success :).

PR #564, opened 2017-05-10
Mangle "*" as #STAR, rather than an explicit variance plus a bound.

alexander-gorshenev:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt
```diff
@@ -104,15 +104,18 @@ private fun acyclicTypeMangler(visited: MutableSet<TypeParameterDescriptor>, typ
...
+        return "#GENERIC" + upperBounds
     }
 
     var hashString = TypeUtils.getClassDescriptor(type)!!.fqNameSafe.asString()
     if (!type.arguments.isEmpty()) {
         hashString += "<${type.arguments.map {
-            val variance = it.projectionKind.label
-            val projection = if (variance == "") "" else "${variance}_" 
-            projection + acyclicTypeMangler(visited, it.type)
+            if (it.isStarProjection()) "#STAR"
```
olonho: Странное форматирование, может
 if (it.isStarProjection()) {
   "#STAR"
} else {
   ...
}

PR #566, opened 2017-05-11
Always use at least register-sized buffer for return value in ffi_call

SvyatoslavScherbina:
As required by the libffi docs.


olonho:
Тоже в v0.2-fixes

PR #640, opened 2017-06-06
Add the ko command

xyproto:
When I started programming Kotlin, I really missed the handy `go` command that comes with the Go project. The ability to enter a directory with source files and just type `go build`, `go fmt` or `go clean` is immensely useful.

This pull requests adds a small utility that provides the most basic functionality for a similar experience. Hopefully it can be rewritten, expanded upon and replaced by a more poweful tool in the future.

The included `ko` utility provides the following commands:

* `ko build` for building *.kt to an executable, using `kotlinc-native`
* `ko fmt` for formatting the source code using `ktlint`
* `ko clean` for cleaning the generated files after building
* `ko help` for a similar help output to the `go` command

The `ko` command aims to be to Kotlin what the `go` command is to Go. It's not perfect, but it's a start.

### DIFF
--- a/cmd/ko
+++ b/cmd/ko
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+command="$1"
+
+if [[ $command == "" ]] || [[ $command == "help" ]]; then
+  cat <<EOF
+Ko is a tool for managing Kotlin source code.
+
+Usage:
+
+        ko command [arguments]
+
+The commands are:
+
+        build       create an executable
+        clean       remove the build output
+        fmt         format sources
+        help        this help
+
+EOF
+  exit 0
+fi
+
+outputname=$(basename $(pwd))
+
+if [[ $command == "build" ]]; then
+  compiler=$(which kotlinc-native 2>/dev/null)
+  if [[ ! -f $compiler ]]; then
+    echo "Could not find the kotlinc-native executable."
+    exit 1
+  fi
+  mainfile=$(grep "fun main" *.kt -l)
+  if [[ ! -f $mainfile ]]; then
+    echo "Could not find a .kt file with fun main!"
+    exit 1
+  fi
+  package=$(grep package "$mainfile" | cut -d" " -f2)
+  if [[ $package == "" ]]; then
+    echo "Found no package declaration in $mainfile!"
+    exit 1
+  fi
+  firstname=${mainfile%.*}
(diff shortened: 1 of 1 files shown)


olonho:
Alexander, thanks for this change. Not sure if we can commit and support such a tool as part of Kotlin/Native now, as it requires a more design and polishing, however, you could consider suggesting this change to Kotlin repository at https://github.com/JetBrains/kotlin.

xyproto:
Thanks for reviewing the pull request.

I'm not sure I follow you. The `ko` tool is, in my opinion, increasing the amount of design and polish for the `kotlin-native` project, since it offers a way of using the `kotlin-native` compiler that is not currently available. If there are specific problems with the code, or a lack of features, I will be happy to look into it.

The utility in question is specifically for `kotlin-native` and not for `kotlin` in general. I was under the impression that [kotlin](https://github.com/JetBrains/kotlin) was only for Kotlin that used the JVM?

If the crux of the issue is that you don't primarily want users to use the command line, but rather would that they used (and/or bought) the IntelliJ IDE, that would be understandable too.

orangy:
@xyproto Kotlin is aiming to be a multiplatform language. We now feature JVM & JS targets, and Native is coming right here. Just like the language itself borrowed a lot of nice things from other languages, tools might do exactly the same and provide great out of the box experience for various scenarios – command line, IDE, web, remote, etc. However, in order to avoid confusion these tools must work across all the platforms and provide more or less similar experience regardless if it's native, JS or JVM. 

Your idea to make a tool similar to `go` is really great, and it is something community is already exploring on other platforms (see [kscript](https://github.com/holgerbrandl/kscript) for example of a different tool but similar in some cases it covers). Thank you for bringing it to our attention and we will definitely evaluate the power of `go` command in our design efforts to build a common infrastructure of tools. However we cannot accept it as is, because it is a very narrow feature for a single platform, and supporting it forever is not something we can commit to at this moment. It is not clear how dependency management is going to work, how C interop is generated with this tool, etc. As it is, it may be very suitable for small programs, but for larger projects not so much. 

How about setting up your own repository and developing it on your own, much like kscript I mentioned above? We will sure mention it whenever appropriate, and even can help with compiler API as needed, depending on the effort and priority.

xyproto:
That is understandable. Thank you for the thoughtful response. I will make sure to look into `kscript`.

PR #664, opened 2017-06-15
Cleanup (removed clean task).

Alefas:
Fixed 'getTask' naming (no side effects now).


Alefas:
According to comments from #663

PR #724, opened 2017-07-07
Don't create Progression objects in 'for' loops

ilmat192:
This patch optimizes the following pattern:
```
for (i in first..last step st) { ... }
```
In this case we need to create a Progression object and then call its iterator() method causing at least 2 allocation per loop. This change replaces such loops with the following construction:
```
var inductionVar = first
checkProgressionStep(step)  // check if step > 0, only if step is not a constant > 0.
last = getProgressionLastElement(first, last, step) // calculate the last element, only if step != 1.
if (first <= last) {
    do {
        i = inductionVar
        inductionVar += step
        ...
    } while(i != last)
}
```


olonho:
|`backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt:99`|
|---|
Для 4 элементов обычно спискок всегда лучше.

olonho:
|`backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ForLoopsLowering.kt:389`|
|---|
by the following way -> in such a manner

olonho:
А тесты на вложенные циклы?

PR #763, opened 2017-08-01
Add Kotlin object -> C string function to simplify debugging

olonho:


runtime/src/main/cpp/KDebug.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "Porting.h"
+#include "Types.h"
+
+#ifndef KONAN_NO_DEBUG_API
+
+extern "C" OBJ_GETTER(KonanObjectToUtf8Array, KRef object);
+
+namespace {
+
+char debugBuffer[1024];
```
alexander-gorshenev: А не мало? ;-)

PR #803, opened 2017-08-22
Rework dependency downloading

ilmat192:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Helper0.kt
```diff
@@ -18,11 +18,11 @@
 // Example startup helper for Kotlin N compiler. Compile to JAR and add it to
```
olonho: Kotlin N -> Kotlin/Native

PR #865, opened 2017-09-19
[debug] add Kotlin Object pretty printers for lldb

matklad:
Now lldb commands like `frame variable` print non-primitive Kotlin
objects nicely if you `command script import dist/tools/konan_lldb.py`

@vvlevchenko I've removed `show_variable` function altogether for now, because it seems to me that a custom type summary is strictly more powerful & convenient. If this is a way to go, I'll also update DEBUGGING.md, otherwise I could restore `show_variable`.


vvlevchenko:
nice, would you mind to sync DEBUGGING.md according your change?

matklad:
Updated the docs!

PR #868, opened 2017-09-19
Implement explicit and implicit object pinning for interop

SvyatoslavScherbina:
This allows passing Kotlin primitive arrays directly to native.

Also:
* Add `defer` support to `memScope`
* Improve `StableObjPtr` and rename it to `StableRef<T>` 
* Add other minor fixes and improvements


runtime/src/main/cpp/Arrays.cpp
```diff
@@ -350,4 +350,13 @@ KNativePtr Kotlin_ImmutableBinaryBlob_asCPointerImpl(KRef thiz, KInt offset) {
   return PrimitiveArrayAddressOfElementAt<KByte>(array, offset);
 }
 
+KNativePtr Kotlin_Arrays_getAddressOfElement(KRef thiz, KInt index) {
+  ArrayHeader* array = thiz->array();
+  if (index < 0 || index > array->count_) {
```
olonho: >=

PR #873, opened 2017-09-21
Easier adding library project as dependency

jiaminglu:
Currently if I want to declare dependency on another project, I have to write `dependsOn` on tasks, and then include the `klib` files manually. So I added a derivative to make declaring such dependency easier.

### DIFF
--- a/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanCompileTask.kt
+++ b/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanCompileTask.kt
@@ -201,6 +201,17 @@ open class KonanCompileConfig(
 
     // DSL. Libraries.
 
+    fun library(project: Project) = {
+        this.project.evaluationDependsOn(project.path)
+        compilationTask.dependsOn(project.tasks.findByPath("compileKonan"))
+        libraries(project.fileTree(project.konanCompilerOutputDir).include("*.klib"))
+    }
+    fun library(project: Project, artifactName: String) = {
+        this.project.evaluationDependsOn(project.path)
+        compilationTask.dependsOn(project.konanArtifactsContainer.getByName(artifactName).compilationTask)
+        library("${project.konanCompilerOutputDir}/$artifactName.klib")
+    }
+
     fun library(lib: Any) = libraries(lib)
     fun libraries(vararg libs: Any) = with(compilationTask) {
         libraries.add(project.files(*libs))
--- a/GRADLE_PLUGIN.md
+++ b/GRADLE_PLUGIN.md
@@ -230,6 +230,12 @@ For this project the task graph will be the following:
 
              // *.klib library for linking.
              library project.file('path/to/library')
+             
+             // library project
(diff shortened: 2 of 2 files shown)


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanCompileTask.kt
```diff
@@ -201,6 +201,15 @@ open class KonanCompileConfig(
 
     // DSL. Libraries.
 
+    fun library(project: Project) = run {
+        compilationTask.dependsOn(project.tasks.findByPath("compileKonan"))
+        libraries(project.fileTree(project.konanCompilerOutputDir).include("*.klib"))
+    }
+    fun library(project: Project, artifactName: String) = run {
+        compilationTask.dependsOn(project.tasks.findByPath("compileKonan${artifactName.capitalize()}"))
```
ilmat192: I think it's better to use a property of `KonanCompileConfig` instead of the task name:
```
compilationTask.dependsOn(project.konanArtifactsContainer.getByName(artifactName).compilationTask)
```
Also such access to a task of `project` will fail if `project` is not evaluated. So it's worth using the `evaluationDependsOn` method to be sure that `project` is already evaluated. It may look like this:
```
    fun library(project: Project, artifactName: String) {
        this.project.evaluationDependsOn(project.path)
        compilationTask.dependsOn(project.konanArtifactsContainer.getByName(artifactName).compilationTask)
        library("${project.konanCompilerOutputDir}/$artifactName.klib")
    }
```

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanCompileTask.kt
```diff
@@ -201,6 +201,17 @@ open class KonanCompileConfig(
...
 
+    fun library(project: Project) = {
+        this.project.evaluationDependsOn(project.path)
+        compilationTask.dependsOn(project.tasks.findByPath("compileKonan"))
+        libraries(project.fileTree(project.konanCompilerOutputDir).include("*.klib"))
+    }
+    fun library(project: Project, artifactName: String) = {
+        this.project.evaluationDependsOn(project.path)
+        compilationTask.dependsOn(project.konanArtifactsContainer.getByName(artifactName).compilationTask)
+        library("${project.konanCompilerOutputDir}/$artifactName.klib")
```
ilmat192: The output file name of the library may be changed by user via DSL and this path may become incorrect. The compilation task you got via the `artifactName` has a property `artifact` which can be used here.

ilmat192: Also we need to check if the artifact we're trying to add as a library is really klib or not (see the  `produce` property of the compilation task).

ilmat192:
I'll put a link to the new PR here.

ilmat192:
Here: https://github.com/JetBrains/kotlin-native/pull/876

PR #886, opened 2017-09-26
Fix Android build

olonho:


r4zzz4k:
Sorry to bother you, but I've noticed a small typo in the example def change: `linkerOpst` should probably be `linkerOpts`.

PR #899, opened 2017-09-30
Fix issue for gtk sample

francisl:
#### Problem
Using non standard include directory on mac os

```

Generating GTK stubs, may take few mins depending on the hardware...
Exception in thread "main" java.lang.Error: /var/folders/8p/x5dxkyrs4d78n96h43nldj7s2vmnqp/T/tmp2879105455459228259.c:1:10: fatal error: 'gtk/gtk.h' file not found
....

```

#### Solution
- Validate default brew directory and then `/opt`
- Add an option to `build.sh` -I=/path/to/include for OSX
- Return with an error

### DIFF
--- a/samples/gtk/README.md
+++ b/samples/gtk/README.md
@@ -3,18 +3,37 @@
  This example shows how one may use _Kotlin/Native_ to build GUI
  applications with the GTK toolkit.
 
-To build use `../gradlew build` or `./build.sh`.
+To build use `../gradlew build` or `./build.sh [-I=/include/path]`.
 
-Do not forget to install GTK3.
+Do not forget to install GTK3. See bellow.
 
 On Mac use `port install gtk3`, on Debian flavours of Linux - `apt-get install libgtk-3-dev`.
 To run on Mac also install XQuartz X server (https://www.xquartz.org/), and then
 
     ../gradlew run
-    
-Alternatively you can run artifact directly 
+
+Alternatively you can run artifact directly
 
     ./build/konan/bin/Gtk3Demo/Gtk3Demo.kexe
 
 Dialog box with the button will be shown, and application will print message
 and terminate on button click.
+
+
+#### GTK3 Install
+
+on Mac use
+
+    brew install gtk+3
+
+or
+
+    port install gtk3
+
+on Debian flavours of Linux
+
+    sudo apt-get install libgtk-3-dev
+
+on Fedora
+
+    sudo dnf install gtk3-devel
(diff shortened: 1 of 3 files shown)


samples/gtk/README.md
```diff
@@ -3,18 +3,33 @@
...
+
+Alternatively you can run artifact directly
 
     ./build/konan/bin/Gtk3Demo/Gtk3Demo.kexe
 
 Dialog box with the button will be shown, and application will print message
 and terminate on button click.
+
+
+#### GTK3 Install
```
olonho: This has to be combined with the text above: "On Mac use `port install gtk3`, on Debian flavours of Linux - `apt-get install libgtk-3-dev`." and use uppercased proper nouns.

proper noun |  
-- | --




proper noun |  
-- | --

samples/gtk/build.sh
```diff
@@ -3,18 +3,36 @@
...
+    I) IPREFIX="$OPTARG"
+    ;;
+  esac
+done
+
+if [ "$TARGET" == 'macbook' ] && [ ! -z $IPREFIX ]; then
+    IPREFIX_macbook=-I$IPREFIX
+elif [ "$TARGET" == 'macbook' ]; then
+    if [ -d /usr/local/include ]; then
+        IPREFIX_macbook=-I/usr/local/include
```
olonho: This better be fixed by adding more -I and -L statements in compilerOpts.osx and linkerOpts.osx in src/main/c_interop/gtk3.def, to avoid excessive logic here.

francisl: Can do that, but it will remove the options to manually add it. 
Also to support both `/usr` and `/opt` it will create a very long params.

    compilerOpts.osx = -I/usr/local/include/gtk-3.0 -I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include \
    -I/usr/local/include/pango-1.0 -I/usr/local/include/cairo -I/usr/local/include -I/usr/local/include/gdk-pixbuf-2.0 \
    -I/usr/local/include/atk-1.0 -I/opt/local/include/gtk-3.0 -I/opt/local/include/glib-2.0 \
    -I/opt/local/lib/glib-2.0/include -I/usr/local/include/pango-1.0 -I/opt/local/include/cairo -I/opt/local/include \
      -I/opt/local/include/gdk-pixbuf-2.0 -I/opt/local/include/atk-1.0

Other options is to include `/usr` in `compilerOpts.osx` and keep the cinterop command as-is.

francisl:
@olonho I pushed an update. I'm not 100% sure for the compilerOpts , because it need to support /usr for brew and /opt for ports. Also there is no option parameter equivalent for -L/-l

olonho:
Thanks, merged!

PR #913, opened 2017-10-05
Some tweaks to build platform libs on cross targets correctly.

alexander-gorshenev:


klib/platform.gradle
```diff
@@ -8,13 +8,15 @@ buildscript {
     }
 }
 
-def taskName = "${gradle.startParameter.projectProperties['name']}"
+def libraryName = "${gradle.startParameter.projectProperties['name']}"
```
olonho: Please rework to avoid collision with recent Slava's changes.

alexander-gorshenev: Yep. And Vasya's too.

PR #930, opened 2017-10-10
[build] crossDistPlatformLibs

vvlevchenko:


olonho:
And make bundle depend on it.

PR #941, opened 2017-10-13
[debug] fix line info for artificial IrElements

matklad:
Some IrElements have -1 as a startOffset (for example, generated methods
of data classes). To represent such offsets in debuginfo, we should use
0 instead of -1, because line/columns are one-based unsigneds.

closes KT-20443

The assembly before this change has the following entries: `.loc    1 4294967295 0`. The assembly after has `.loc    1 0 0` which seems more reasonable. Looks like the clag does use `0` for generated stuff too:

```C++
struct Foo {
    int x;
    Foo();
    Foo(Foo const&);
};

Foo::Foo() = default;

Foo::Foo(Foo const&) = default;
```

has some `!24 = !DILocation(line: 0, scope: !6)` in bitcode.


matklad:
сс @vvlevchenko

PR #947, opened 2017-10-14
update platform.windows to include most modern WIN32 APIs

msink:
А вообще наверно не совсем правильно называть его ```platform.windows``` - у винды теперь две разные платформы - Win32 и UWP (или Windows Universal). 
Так это Win32.

### DIFF
--- a/klib/src/platform/windows/windows.def
+++ b/klib/src/platform/windows/windows.def
@@ -1,3 +1,6 @@
 package = platform.windows
-headers = windows.h
-compilerOpts = -DUNICODE -Wno-incompatible-pointer-types -Wno-deprecated-declarations
+headers = windows.h commctrl.h dwmapi.h shlobj.h shlwapi.h shobjidl.h \
+    urlmon.h usp10.h uxtheme.h vfw.h wininet.h
+compilerOpts = -DUNICODE -DWINVER=0x0601 -D_WIN32_WINNT=0x0601 -DWINAPI_FAMILY=3 \
+    -Wno-incompatible-pointer-types -Wno-deprecated-declarations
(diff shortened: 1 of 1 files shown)


klib/src/platform/windows/windows.def
```diff
@@ -1,3 +1,5 @@
 package = platform.windows
-headers = windows.h
-compilerOpts = -DUNICODE -Wno-incompatible-pointer-types -Wno-deprecated-declarations
\ No newline at end of file
+headers = windows.h commctrl.h dwmapi.h shlobj.h shlwapi.h shobjidl.h \
+    urlmon.h usp10.h uxtheme.h wininet.h vfw.h
+compilerOpts = -DUNICODE -DWINVER=0x0A00 -D_WIN32_WINNT=0x0A00 -DWINAPI_FAMILY=3 \
```
olonho: 0x0A00 is Windows 10, maybe we need separate versions of def files for different windows APIs. Maybe 0x0601 (Win 7) is somewhat more universal common ground. Also we need to discuss separation of declarations between different libraries, as Windows API is not entirely monolithic.

msink: Downgraded ro 0x601
About separation - this is still not complete Win32 API, mostly GUI part.
Dont't know how to split it...

msink: Maybe ```urlmon.h``` and ```wininet.h``` can be separated into WinINet module.

msink: For GUI part of APIs - main missed part is gdiplus.h - it is c++ library, but many GUI functionalities cannot be implemented without it. :(

PR #950, opened 2017-10-16
 Represent Objective-C block pointers in methods as Kotlin functions

SvyatoslavScherbina:
Also 

* Enable method signature checks when overriding Objective-C method;
* Make linkage issues more understandable in Objective-C interop;
* Do some refactoring.

### DIFF
--- a/Interop/Runtime/src/native/kotlin/kotlinx/cinterop/ObjectiveCImpl.kt
+++ b/Interop/Runtime/src/native/kotlin/kotlinx/cinterop/ObjectiveCImpl.kt
@@ -124,6 +124,19 @@ annotation class InteropStubs()
 @Retention(AnnotationRetention.SOURCE)
 private annotation class ObjCMethodImp(val selector: String, val encoding: String)
 
+@konan.internal.ExportForCppRuntime("Kotlin_Interop_getObjCClass")
+private fun getObjCClassByName(name: NativePtr): NativePtr {
+    val result = objc_lookUpClass(name)
+    if (result == nativeNullPtr) {
+        val className = interpretCPointer<ByteVar>(name)!!.toKString()
+        val message = """Objective-C class '$className' not found.
+            |Ensure that the containing framework or library was linked.""".trimMargin()
+
+        throw RuntimeException(message)
+    }
+    return result
+}
+
 @konan.internal.ExportForCompiler
 private fun <T : ObjCObject> allocObjCObject(clazz: NativePtr): T {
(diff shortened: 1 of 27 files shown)


Interop/Runtime/src/native/kotlin/kotlinx/cinterop/ObjectiveCImpl.kt
```diff
@@ -124,6 +138,19 @@ annotation class InteropStubs()
 @Retention(AnnotationRetention.SOURCE)
 private annotation class ObjCMethodImp(val selector: String, val encoding: String)
 
+@konan.internal.ExportForCppRuntime("Kotlin_Interop_getObjCClass")
+private fun getObjCClassByName(name: NativePtr): NativePtr {
+    val result = objc_lookUpClass(name)
+    if (result == nativeNullPtr) {
+        val className = interpretCPointer<ByteVar>(name)!!.toKString()
+        val message = "Objective-C class '$className' not found.\n" +
```
olonho: Why not multiline literal?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/TypeUtils.kt
```diff
@@ -51,6 +51,7 @@ fun Type.getStringRepresentation(): String = when (this) {
         is ObjCClassPointer -> "Class$protocolQualifier"
         is ObjCObjectPointer -> "${def.name}$protocolQualifier*"
         is ObjCInstanceType -> TODO(this.toString()) // Must have already been handled.
+        is ObjCBlockPointer -> "id"
```
olonho: Is name conflict possible here?

runtime/src/main/cpp/ObjCInteropUtils.mm
```diff
@@ -103,6 +103,50 @@ id MissingInitImp(id self, SEL _cmd) {
   return nullptr;
 }
 
+// TODO: optimize
```
olonho: Optimize how?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/TypeUtils.kt
SvyatoslavScherbina: The names returned from this function are not to be used to declare types, but just to mention them. And `id` is the builtin Objective-C type, with meaning similar to `void*`.

PR #951, opened 2017-10-17
Manifest duplicates

alexander-gorshenev:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -142,6 +142,18 @@ private fun Properties.storeProperties(file: File) {
     }
 }
 
+private fun Properties.putAndRunOnReplace(key: Any, newValue: Any, beforeReplace: (Any, Any, Any) -> Unit) {
+    val oldValue = this[key]
+    if (oldValue != null && oldValue!! != newValue) {
+        beforeReplace(key, oldValue!!, newValue)
```
olonho: Do we need oldValue!! in then branch: DFA shall prove it's not needed.

PR #954, opened 2017-10-17
[compiletest]: add JUnit4-based infrastructure for testing compiler

matklad:
Here's my take on non-Gradle based testing infrastructure. It's far from polished at the moment: just something to get the ball rolling. 

`compiltest` Gradle project contains a new test framework (in `src/main`) and some actual tests (in `src/test`). 

The test framework is not tied to JUnit, so it is in theory possible to plug in another test runner. 

The actual tests are JUnit4 based (JUnit 5 looked significantly more complicated to setup).

cc @ilmat192 and Pavel (sorry, I don't know your GitHub handle :) )


backend.native/compiletest/src/main/kotlin/org/jetbrains/kotlin/compiletest/DistProperties.kt
```diff
@@ -0,0 +1,17 @@
...
+import java.nio.file.Paths
+
+
+object DistProperties {
+    private val isWindows: Boolean = requireProp("os.name").startsWith("Windows")
+
+    private val dist: Path = Paths.get(requireProp("konan.home"))
+    private val konancDriver = if (isWindows) "konanc.bat" else "konanc"
+    val konanc: Path = dist.resolve("bin/$konancDriver")
+    val lldb: Path = Paths.get("lldb")
```
matklad: This obviously should use something real, but I don't know at the moment how we are going to provide lldb for testing :) 

Currently, tests are skipped if there's no lldb.

backend.native/compiletest/src/test/kotlin/org/jetbrains/kotlin/compiletest/LldbTests.kt
```diff
@@ -0,0 +1,56 @@
...
+        [..] at main.kt:3, [..] stop reason = step over
+
+        > n
+        Process [..] stopped
+        [..] at main.kt:4, [..] stop reason = step over
+
+        > n
+        Process [..] stopped
+        [..] at main.kt:3, [..] stop reason = step over
+    """)
```
matklad: Here's how inline tests look like:

image

They are big, but you actually see what you are testing, and it's always possible to split a huge file into smaller test suites :) 

I am not married to this approach though, specifying test fixtures as files is fine by me!

matklad:
Yep, running a single test from gradle looks like: 

```
 ./gradlew :backend.native:compiletest:test --tests "*can step through code"
```

There's some support for "tags" in gradle as well, but not in the form of command-line options, you have to explicitly name tags in build.gradle: https://docs.gradle.org/current/userguide/java_plugin.html#test_grouping

backend.native/compiletest/src/main/kotlin/org/jetbrains/kotlin/compiletest/Matchers.kt
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        for ((pattern, body) in patterns.zip(bodies)) {
+            multilineMatch(pattern, body)
+        }
+    }
+
+    companion object {
+        fun parse(spec: String): LldbSessionSpecification {
+            val blocks = spec.trimIndent().split("(?=^>)".toRegex(RegexOption.MULTILINE))
```
PavelPunegov: Please add comments that describe that debugger commands are tied with the desired output

settings.gradle
```diff
@@ -25,6 +25,7 @@ include ':backend.native'
 include ':runtime'
 include ':common'
 include ':backend.native:tests'
+include ':backend.native:compiletest'
```
PavelPunegov: Not sure that the whole activity should be named as 'compiletest'. It's rather debugging or debug-ability

backend.native/compiletest/src/main/kotlin/org/jetbrains/kotlin/compiletest/DistProperties.kt
```diff
@@ -0,0 +1,17 @@
+package org.jetbrains.kotlin.compiletest
+
+import java.nio.file.Path
+import java.nio.file.Paths
+
+
+object DistProperties {
+    private val isWindows: Boolean = requireProp("os.name").startsWith("Windows")
```
PavelPunegov: I believe that "os.name" could be both "Windows" and "windows". It's widely used to check for "indows"

olonho: Please reuse same logic as in other places: 
TargetManager.host.family == Family.WINDOWS.

matklad: Resolved.

backend.native/compiletest/src/main/kotlin/org/jetbrains/kotlin/compiletest/Matchers.kt
matklad: Added some docs to the `lldbTest` function.

settings.gradle
matklad: In theory, we can use JUnit test runner to drive some other tests, and not only debugger-related ones, so I picked a more general name. Should I rename this to `debuggertests`?

PavelPunegov: DistProperties and Driver could be reused to other tests, and hence they could be called 'compiletest' or tests:shared' or harness/lib.
But Matchers.kt and LldbTest.kt are only debugger related.

So, I'd like to rename this

matklad: Made the module name more specific.

PR #959, opened 2017-10-18
[WIP] Win32 GDI+ bindings

msink:
Note:
"Windows GDI+ exposes a flat API that consists of about 600 functions,
which are implemented in Gdiplus.dll and declared in Gdiplusflat.h.
The functions in the GDI+ flat API are wrapped by a collection of
about 40 C++ classes. It is recommended that you do not directly call
the functions in the flat API.
Whenever you make calls to GDI+, you should do so by calling the methods
and functions provided by the C++ wrappers.
Microsoft Product Support Services will not provide support for code
that calls the flat API directly."

### DIFF
--- a/samples/gdiplus/build.bat
+++ b/samples/gdiplus/build.bat
@@ -1,0 +1,9 @@
+@echo off
+setlocal
+set DIR=.
+set "PATH=..\..\dist\bin;..\..\bin;%PATH%"
+if "%TARGET%" == "" set TARGET=mingw
+
+set "LFLAGS=-Wl,--subsystem,windows"
+
+call konanc -target "%TARGET%" "%DIR%\src\main\kotlin" -linkerOpts "%LFLAGS%" -o gdiplus || exit /b
--- a/samples/gdiplus/src/main/kotlin/main.kt
+++ b/samples/gdiplus/src/main/kotlin/main.kt
@@ -1,0 +1,6 @@
+import platform.windows.*
+
+fun main(args: Array<String>) {
+    MessageBoxW(null, "Konan говорит:\nЗДРАВСТВУЙ МИР!\n",
+            "Заголовок окна", MB_YESNOCANCEL or MB_ICONQUESTION)
+}
--- a/samples/gdiplus/src/main/kotlin/platform/gdiplus/Brush.kt
+++ b/samples/gdiplus/src/main/kotlin/platform/gdiplus/Brush.kt
@@ -1,0 +1,904 @@
+/*
+ * This file is part of the w32api package.
+ *
+ * Contributors:
+ *   Created by Markus Koenig <markus@stber-koenig.de>
+ *   Kotlin/Native port by Mike Sinkovsky <msink@permonline.ru>
+ *
+ * THIS SOFTWARE IS NOT COPYRIGHTED
+ *
+ * This source code is offered for use in the public domain. You may
+ * use, modify or distribute it freely.
+ *
+ * This code is distributed in the hope that it will be useful but
+ * WITHOUT ANY WARRANTY. ALL WARRANTIES, EXPRESS OR IMPLIED ARE HEREBY
+ * DISCLAIMED. This includes but is not limited to warranties of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ */
+
+package platform.gdiplus
+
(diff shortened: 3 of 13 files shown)


klib/src/platform/windows/GDIPlusFlat.def
```diff
@@ -0,0 +1,5 @@
+package = platform.GDIPlusFlat
```
olonho: You likely needs depends=windows

msink: Not sure, but OK - it's windows package anyway :)

msink: Hm, strange.
In generated ```GDIPlusFlat.kt``` there are bunch of inports like
```
import platform.posix.BITMAPINFO
import platform.posix.BOOL
import platform.posix.BOOLVar
import platform.posix.BYTEVar
import platform.posix.CLSID
...
```
Why ```posix```, shouldn't be ```windows```?

msink: Fixed - now MinGW platform.posix depends on platform.windows
And it really is just emulation of POSIX API over Windows.

msink:
I can port C++ part of GDI+ to kotlin by hand.
But - is it posdible somehow include hand-made Kotlin bindings to platform library, alongside with autogenerated?

klib/src/platform/windows/posix.def
```diff
@@ -1,3 +1,4 @@
+depends = windows
```
olonho: We usually do deps in another direction, see osx.def and linux.def. They depend on POSIX, not the opposite, so that common types are named same way across the platforms.

olonho:
Currently you may try to augment .def file with wrappers to C++ code, like we do here for couple important C macro-definitions:
https://github.com/JetBrains/kotlin-native/blob/master/klib/src/platform/linux/posix.def#L21
It may or may not work, depending on how C++ calls will look like.
Note that headers are parsed in C mode, so whatever you'll put under 
`#ifdef __cplusplus` will be ignored by interop tool.

msink:
Прототип сделал - примерно процентов 20 функций из GDI+
СтОит заканчивать?

olonho:
Я думаю лучше начать c простого gdiplus.def.

klib/src/platform/windows/posix.def
msink: Here posix.def includes winsock2.h, and winsock2.h includes windows.h
So all windows trash goes to posix.
Maybe it could be somewhat filterted, but it's not easy.

msink:
То есть пока убрать, и оставить только первые два коммита? Или только первый?

klib/src/platform/windows/posix.def
olonho: Please wait until https://github.com/JetBrains/kotlin-native/pull/966 will get merged. I tried to make deps more reasonable, so your change could be made easy.

msink:
Сделал процентов на 90.

Теперь при компиляции странная ошибка:
```
llvm-lto: error: Linking globals named 'kobjref:platform.windows.threadlocaleinfostruct.Companion': symbol multiply defined!
exception: org.jetbrains.kotlin.backend.konan.KonanExternalToolFailure: ...
```
Больше ошибок нет. Что это?

msink:
Как получить указатель на поле структуры, импортированной через cinterop?

olonho:
А для какой цели?

msink:
```
typedef DWORD ARGB;
struct Color { 
   ARGB Value; 
}
Status GdipGetHatchBackgroundColor(GpHatch*, ARGB*);
```

Теперь хочу сделать на Котлине аналог Сишного
```
Status GetBackgroundColor(Color *color)  {
    return GdipGetHatchBackgroundColor(nativeBrush,  color ? &color->Value : NULL);
}
```
Так не проходит:
```
fun GetBackgroundColor(color: Color?) = GdipGetHatchBackgroundColor(ptr, color?.Value)
```

msink:
Вот так компилируется:
```
fun GetBackgroundColor(color: Color?)
    = GdipGetHatchBackgroundColor(ptr, color?.memberAt<ARGBVar>(0)?.ptr)
```
Но как-то это криво.

olonho:
Seems no longer relevant, closing.

PR #961, opened 2017-10-18
Fix enableDebug() option not working in build.gradle

rnentjes:
The GRADLE_PLUGIN.md mentions enableDebug() option in the plugin dsl but that options doesn't work (the command line one does).

This pull request adds that option.


ilmat192:
Thank you for your pull request!
As you can see there is some inconvenience in DSL options representing compiler boolean flags: one of them has a boolean parameter (e.g. `noDefaultLibs true/false` or `enableDebug true/false`) while other just set the corresponding flag to true (e.g. `noMain()`). This pull request reworks the DSL and also fixes this problem by using a boolean parameter in all such methods: https://github.com/JetBrains/kotlin-native/pull/958

PR #980, opened 2017-10-25
Allow specifying top level target and flavor in nativeInteropPlugin.

alexander-gorshenev:


buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/NativeInteropPlugin.groovy
```diff
@@ -152,9 +152,11 @@ class NamedNativeInteropConfig implements Named {
         return new File(project.buildDir, "nativeInteropStubs/$name/kotlin")
     }
 
-    NamedNativeInteropConfig(Project project, String name) {
+    NamedNativeInteropConfig(Project project, String name, String target = null, String flavor = 'jvm') {
```
olonho: Why default is jvm?

alexander-gorshenev: That's the way the rest of the plugin is written. Just to minimize user astonishment by changing the default.

PR #988, opened 2017-10-27
Implement `-partialIncludeDir` and `-filteredIncludeDir` in cinterop

SvyatoslavScherbina:


olonho:
Also we need to port demos on that.

PR #992, opened 2017-10-30
[runtime] drop debug flags

vvlevchenko:


vvlevchenko:
mac build failed with missing libeate on agent

PR #994, opened 2017-10-31
[debug] add test for pretty-printers

matklad:
This adds tests for our Python debugger pretty printers. 

Note that these test **will** fail with the current master branch, because https://github.com/JetBrains/kotlin-native/commit/bbeefb0b561fdb5afe260cd5429f7a1c1bea7753 has broken the pretty printers.


matklad:
cc @vvlevchenko

matklad:
Ok, should be good to go now, thanks @vvlevchenko !

olonho:
@matklad please close the PR, if no longer relevant.

matklad:
This PR should be merged because it contains both a fix by @vvlevchenko and a regression test.

PR #1000, opened 2017-11-01
Fix object init/deinit bug, refactoring

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -295,27 +294,25 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
                     ret(null)
                 }
 
                 appendingTo(bbInit) {
                     context.llvm.fileInitializers
-                            .map { it as IrField }
-                            .filterNot { it.initializer is IrConst<*> }
-                            .forEach {
-                                val descriptor = it.descriptor
+                            .filterNot { it.initializer?.expression is IrConst<*> }
```
olonho: Можно.

PR #1008, opened 2017-11-03
new window title for objc sample

gildor:


msink:
Why? It's here for testing non-ascii symbols.
Just contrary, maybe add something Chinese.

gildor:
Okay, I see, just thought that it's probably accidental. Make sense to add some Unicode symbols than

olonho:
Yes, having non-ASCII symbols is intentional, we want to ensure string conversion is correct not only for Latin-derived languages.

PR #1009, opened 2017-11-03
[debug info][KT-20391] don't confuse lldb with array tag

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/DebugUtils.kt
```diff
@@ -152,11 +152,6 @@ internal fun generateDebugInfoHeader(context: Context) {
 internal fun KotlinType.dwarfType(context:Context, targetData:LLVMTargetDataRef): DITypeOpaqueRef {
     when {
         KotlinBuiltIns.isPrimitiveType(this) -> return debugInfoBaseType(context, targetData, this.getJetTypeFqName(false), llvmType(context), encoding(context).value.toInt())
-        KotlinBuiltIns.isArray(this) -> {
```
olonho: Why, how could we inspect arrays then?

vvlevchenko: With Konan_PrintDebug

PR #1029, opened 2017-11-13
Improve workers API, fix lost sync.

olonho:


runtime/src/main/kotlin/konan/worker/Worker.kt
```diff
@@ -100,18 +100,13 @@ class Future<T> internal constructor(val id: FutureId) {
...
 
-    override fun equals(other: Any?): Boolean {
-        return (other is Future<*>) && (id == other.id)
-    }
+    override fun equals(other: Any?) = (other is Future<*>) && (id == other.id)
 
-    override fun hashCode(): Int {
-        return id
-    }
+    override fun hashCode() = id as Int
```
SvyatoslavScherbina: What's the reason of adding `as Int`?

runtime/src/main/kotlin/konan/worker/Worker.kt
```diff
@@ -240,4 +235,12 @@ internal fun ThrowWorkerInvalidState(): Unit =
         throw IllegalStateException("Illegal transfer state")
 
 @ExportForCppRuntime
-internal fun WorkerLaunchpad(function: () -> Any?) = function()
+internal fun WorkerLaunchpad(function: () -> Any?): Any? {
+    try {
+        return function()
+    } catch (e: Throwable) {
+        println("worker thrown an exception: $e")
```
SvyatoslavScherbina: Has the worker thrown this exception, or it was the thread sending the job to the worker?

runtime/src/main/kotlin/konan/worker/Worker.kt
olonho: Nothing special, just code readability.

olonho: Removed.

runtime/src/main/kotlin/konan/worker/Worker.kt
olonho: Done

PR #1051, opened 2017-11-20
Update Readme.md to notify users to install JDK

itzurabhi:
Since tools.jar is required to build from source, I had to look around forums to have the answer. Hope this might help a few.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,7 @@ or where a developer is willing to produce a reasonably-sized self-contained pro
 without the need to ship an additional execution runtime.
 
 Prerequesties:
-
-Linux:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,11 @@ virtual machines are not desirable or possible (such as iOS, embedded targets),
 or where a developer is willing to produce a reasonably-sized self-contained program
 without the need to ship an additional execution runtime.
 
+Prerequesties:
+
+Linux:
```
olonho: It's true for all platforms, not just Linux.

itzurabhi: Should I revise and submit (remove the Linux tag ) ?

olonho:
That's OK, merged.

PR #1061, opened 2017-11-22
If inline function receives 'var' as argument - it should be evaluated

KonstantinAnisimov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -247,6 +244,12 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
             if (irCallableReference !is IrCallableReference) return false                   // Second statement of the block must be CallableReference.
             return true                                                                     // The expression represents lambda.
         }
+
+        private fun isGetVal(expression : IrExpression): Boolean {
```
olonho: Looks somewhat ugly.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -226,9 +223,9 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
                                       val argumentExpression : IrExpression) {
         val needsEvaluation: Boolean
             get() {
-                if (argumentExpression is IrGetValue) return false                         // Parameter is already GetValue - nothing to evaluate.
```
SvyatoslavScherbina: What is the purpose to have such an "optimization"?
For example, are you sure that storing `IrCallableReference` into variable would be less efficient than copying it into every usage?
And actually it is not correct at all:

```
inline fun call(block1: () -> Unit, noinline block2: () -> Int): Int {
    block1()
    return block2()
}

    var x = 5
    println(call({ x = 7 }, x::toInt)) 
}
```

Can you provide any case showing that this "optimization" has any effect?

PR #1072, opened 2017-11-24
gradle-plugin: generate CMake file

matklad:
This adds a minimal `generateCMake` task to kotlin-native-gradle plugin, which creates a CMakeLists.txt file, suitable for CLion. This is a minimal implementation, but it is sufficient to deal with all samples, which build on linux.

Notable omissions and problems:

* No support for multi-project builds. Proposed plan here is to generate a single CMakeLists.txt for the root project.

* No support for platform targets other then the host. Proposed plan is to handle targets at confiure time at CMake. That is, to create a `set(KONAN_TARGET "macbook")` variable and wrap all `konan_library` calls in `if (${KONAN_TARGET} == "macbook")`

* Listing source files is done on file-by-file bases, which means that `generateCMake` should be invoked when a new file is created. Note, however, that this is more or less true for pure cmake build scripts as well, because CMake will miss some new files even with recursive globing. Looks like the only two reliable solutions for CMake are 1) list each and every file in CMakeLists.txt 2) use custom command for building, which is always executed, and keeps track of modifications itself (aka, call gradle from CMake). We can make current solution prettier by listing only the directories and not the individual files. Unfortunately, neither FileCollection nor FileTree have a convenient interface to do it, so I've skipped this for now.

* The generated CMakeLists depends on `KotlinCMakeModule` which is provided by IDE.

cc @Alefas @ilmat192

### DIFF
--- a/tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/CMakeSpecification.groovy
+++ b/tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/CMakeSpecification.groovy
@@ -1,0 +1,68 @@
+package org.jetbrains.kotlin.gradle.plugin.test
+
+class CMakeSpecification extends BaseKonanSpecification {
+
+    def 'Plugin should generate CMake from project without additional settings'() {
+        when:
+        def project = KonanProject.createEmpty(projectDirectory) { KonanProject it ->
+            it.buildFile.write("""
+            plugins { id 'konan' }
+            konanArtifacts {
+                interop('stdio')
+                interop('sdl') {
+                    defFile 'src/main/c_interop/sdl.def'
+                    includeDirs '/usr/include/SDL2'
+                }
+                library('main_lib')
+                program('Main') {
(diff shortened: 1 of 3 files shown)


tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/CMakeSpecification.groovy
```diff
@@ -0,0 +1,68 @@
+package org.jetbrains.kotlin.gradle.plugin.test
+
+class CMakeSpecification extends BaseKonanSpecification {
```
olonho: А чего на Груви, на Котлине пока нельзя?

matklad: Это тесты на грейдл плагин, я их сделал также, как и остальные. Думаю, что тут Kotlin не поможет, потому что, как я понимаю, тут куча груву-специфичный рантайм интроспекций от грувейного [spockframework](http://spockframework.org/).

PR #1076, opened 2017-11-27
Add kotlin math

ilmat192:


runtime/src/main/kotlin/konan/internal/RuntimeUtils.kt
```diff
@@ -61,6 +61,11 @@ fun ThrowUninitializedPropertyAccessException(): Nothing {
     throw UninitializedPropertyAccessException()
 }
 
+@ExportForCppRuntime
+internal fun ThrowNotImplementedError(): Nothing {
```
PavelPunegov: Is it used anywhere?

runtime/src/main/cpp/Math.cpp
```diff
@@ -0,0 +1,345 @@
...
+KInt Kotlin_math_absi(KInt x) { return labs(x); }
+KLong Kotlin_math_absl(KLong x) { return llabs(x); }
+
+// endregion
+
+#else // if KONAN_WASM is defined
+
+// kotlin.math isn't supported for WASM.
+namespace {
+    void NotImplemented() { RuntimeAssert(false, "Not implemented for wasm"); }
```
olonho: ThrowNotImplemented()?

runtime/src/main/cpp/Math.cpp
```diff
@@ -0,0 +1,345 @@
...
+            }
+            return SQRT2 * result;
+        }
+    }
+}
+#endif
+
+extern "C" {
+
+#ifndef KONAN_WASM
```
olonho: Usually better avoid inverted condition. Better use #if FEATURE, not #if PLATFORM.
So maybe `#if !KONAN_NO_MATH`  here, as on most platforms math.h will be there.

runtime/src/main/cpp/Math.cpp
```diff
@@ -0,0 +1,345 @@
...
+ */
+
+#include <float.h>
+#include <math.h>
+#include <stdlib.h>
+
+#include "Exceptions.h"
+#include "Types.h"
+
+#if (__MINGW32__ || __MINGW64__)
```
olonho: Again, smth like `#if KONAN_NEED_ASINH_ACOSH`

PR #1078, opened 2017-11-28
[debug]: allow inspecting arrays of objects

matklad:
This enables synthetic children provider for arrays of objects.

### DIFF
--- a/llvmDebugInfoC/src/scripts/konan_lldb.py
+++ b/llvmDebugInfoC/src/scripts/konan_lldb.py
@@ -33,8 +33,6 @@ def lldb_val_to_ptr(lldb_val):
 def kotlin_object_type_summary(lldb_val, internal_dict):
     """Hook that is run by lldb to display a Kotlin object."""
     fallback = lldb_val.GetValue()
-    if str(lldb_val.type) != "struct ObjHeader *":
-        return fallback
 
     ptr = lldb_val_to_ptr(lldb_val)
     if ptr is None:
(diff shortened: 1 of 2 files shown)


llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -129,10 +137,22 @@ def get_child_at_index(self, index):
...
-        )
+
+        if self._element_type == KREF:
+            self._evaluate(KREF + " $retval;")
+            return self._evaluate(
+                "({self._element_type}){self._fn_prefix}_get({self._ptr}, {index}, &$retval)".format(
+                    self=self, index=index
+                ),
+                name=str(index)
+            )
```
matklad: I am not sure how well this branch of if interacts with runtime. Here, we synthesize a "local" variable slot in debugger, store a KObject in there, and increment some reference counts. Presumably, this leaks the object, because we never decrement the counter. I am not sure what to do here, and how is this supposed to work :) Related question: If I evaluate a C++ expression in C++, which gives me an `std::vector` back, is this vector destroyed eventually?

olonho:
Скоординируйтесь с Васей насчет этого ревью.

llvmDebugInfoC/src/scripts/konan_lldb.py
olonho: Это не должно так работать - мы не должны держать явных ссылок на объекты, особенно в виде ObjHeader.

olonho:
Obsolete.

PR #1080, opened 2017-11-28
All condition related to argument evaluation can be formulated as "ne…

KonstantinAnisimov:
…ver evaluate arguments corresponding to functional parameters"


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -331,7 +315,7 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
         parameterToArgumentOld.forEach {
             val parameterDescriptor = it.parameterDescriptor
 
-            if (it.isInlinableLambda) {                                                     // If argument is inlinable lambda.
+            if (parameterDescriptor.type.isFunctionType) {                                  // If parameter has functional type - it can not be evaluated.
```
SvyatoslavScherbina: Does the test below work?

```
inline fun call(block1: () -> Unit, noinline block2: () -> Int): Int {
    block1()
    return block2()
}

    var x = 5
    println(call({ x = 7 }, x::toInt)) 
}
```

PR #1081, opened 2017-11-28
Minor improvements in gradle plugin

ilmat192:
- Use new CDN paths for compiler bundles.
- Allow setting additional JVM arguments for compiler/cinterop.
- Fix `includeDir` processing in CMake project generation.

### DIFF
--- a/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
+++ b/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
@@ -50,9 +50,11 @@ internal abstract class KonanCliRunner(val toolName: String, val fullName: Strin
 
     override val jvmArgs = mutableListOf(
             "-ea",
-            "-Xmx3G",
             "-Dkonan.home=${project.konanHome}",
-            "-Djava.library.path=${project.konanHome}/konan/nativelib")
+            "-Djava.library.path=${project.konanHome}/konan/nativelib").apply {
+        addAll(project.konanExtension.jvmArgs)
+        addAll(project.jvmArgs)
+    }
 
     override val environment = mutableMapOf("LIBCLANG_DISABLE_CRASH_RECOVERY" to "1")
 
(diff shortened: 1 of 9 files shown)


tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/CMakeSpecification.groovy
```diff
@@ -1,17 +1,20 @@
 package org.jetbrains.kotlin.gradle.plugin.test
 
-class CMakeSpecification extends BaseKonanSpecification {
+import org.jetbrains.kotlin.konan.target.Family
+import org.jetbrains.kotlin.konan.target.TargetManager
 
+class CMakeSpecification extends BaseKonanSpecification {
     def 'Plugin should generate CMake from project without additional settings'() {
         when:
+        def sdlIncludePath = (TargetManager.host.family == Family.WINDOWS) ? "C:/SDL2/include" : "/usr/include/SDL2"
```
matklad: Hm, I think we can make this test more interesting if, insider of string interpolation in `build.gradle`, we write

```
interop('sdl') {
    target('mingw') { includeDirs "C:\\SDL2\\include" }
    target('linux') { includeDirs "/usr/include/SDL" }
}
```

and then, for CMake, use string interpolation as is already done.

ilmat192: Good idea, I'll try it.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
```diff
@@ -50,9 +50,11 @@ internal abstract class KonanCliRunner(val toolName: String, val fullName: Strin
 
     override val jvmArgs = mutableListOf(
             "-ea",
-            "-Xmx3G",
             "-Dkonan.home=${project.konanHome}",
-            "-Djava.library.path=${project.konanHome}/konan/nativelib")
+            "-Djava.library.path=${project.konanHome}/konan/nativelib").apply {
+        addAll(project.konanExtension.jvmArgs)
```
vvlevchenko: Классно :)

PR #1089, opened 2017-11-29
Supported recursive references in enums

homuroll:
Split enum entries initialization onto two steps:
1. Allocation of all instances and storing them in the array;
2. Calling constructors on each instance.

### DIFF
--- a/runtime/src/main/kotlin/konan/internal/RuntimeUtils.kt
+++ b/runtime/src/main/kotlin/konan/internal/RuntimeUtils.kt
@@ -105,6 +105,16 @@ fun <T: Enum<T>> valuesForEnum(values: Array<T>): Array<T>
     return result as Array<T>
 }
 
+@Intrinsic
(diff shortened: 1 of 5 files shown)


runtime/src/main/kotlin/konan/internal/RuntimeUtils.kt
```diff
@@ -105,6 +105,18 @@ fun <T: Enum<T>> valuesForEnum(values: Array<T>): Array<T>
...
 }
 
+@Intrinsic
+@PublishedApi
+internal fun <T> createUninitializedInstance(): T {
+    throw Exception("Call to this function should've been lowered")
+}
+
+@Intrinsic
+@PublishedApi
```
olonho: Why @PublishedApi?

homuroll: We need to call symbolName on it.

PR #1093, opened 2017-11-30
1) Do not inline in function declared as "inline"

KonstantinAnisimov:
2) Evaluate all arguments except lambdas


SvyatoslavScherbina:
```
inline fun foo(x: Any) {
    println(x === x)
}

    foo { args }
}
```

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -226,21 +236,35 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
...
+            if (this !is ValueParameterDescriptor) return false
+            if (this.isNoinline)                   return true                              // If parameter marked as "noinline" - do not inline.
+            return false
+        }
+
+        //-------------------------------------------------------------------------//
+
+        fun IrExpression.isLambda(): Boolean {
+            if (this !is IrBlock)                                     return false          // Lambda must be represented with IrBlock.
+            if (this.origin != IrStatementOrigin.LAMBDA &&                                  // Origin must be LAMBDA or ANONYMOUS.
```
KonstantinAnisimov: Ждем новый компилятор

PR #1132, opened 2017-12-12
fix minor typo in interop documentation

roamingthings:
The configuration property to specify the library path in a definition file is `libraryPaths` instead of `libraryPath`.


alexander-gorshenev:
Thanks!

PR #1182, opened 2017-12-21
Fix C API generation.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -64,12 +63,51 @@ private operator fun String.times(count: Int): String {
...
+        "else", "enum", "extern",
+        "float", "for", "goto",
+        "if", "int", "long",
+        "register", "return",
+        "short", "signed", "sizeof", "static", "struct", "switch",
+        "typedef", "union", "unsigned",
+        "void", "volatile", "while",
+        // Not exactly keywords, but reserved.
+        "and", "not", "or", "xor"
+)
```
alexander-gorshenev: Как-то мало. Не зватает inline и restrict, например.

msink: Also, each name that begins with a double underscore __ or an underscore followed by an uppercase letter is reserved.

PR #1194, opened 2017-12-26
Update stdlib and tests

PavelPunegov:
1. Adds several methods from Kotlin 1.2 and 1.1
2. Fix Kotlin version to 1.2
3. Fixes to stdlib from: https://github.com/JetBrains/kotlin/commit/599113b30f2ecc24b1a191532a336ab52379ee3d and https://github.com/JetBrains/kotlin/commit/934b0b8c386fb7af9fa36e961b13ee5f3dd3c063
4. Removes old tests and replaces with tests that run with TestRunner
5. Several test cases commented out because they test methods that depend on JDK classes
6. Fix whitespace detection to conform Unicode standard


PavelPunegov:
For the Unicode change. Here is a reference http://www.unicode.org/Public/UCD/latest/ucd/PropList.txt

runtime/src/main/cpp/KString.cpp
```diff
@@ -671,17 +671,17 @@ int iswalnum_Konan(KChar ch) {
 }
 
 int iswspace_Konan(KChar ch) {
```
PavelPunegov: Kotlin stdlib is also behaves differently than JDK: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/text/CharJVM.kt#L77

PR #1204, opened 2017-12-30
Resolve name conflicts when present (dyn lib generation)

### LINKED ISSUE #1199: Dynamic lib generation breaks on method with multiple receivers
Test case:

```kotlin
class Foo {
    fun String.hi() {
        print("hi")
    }
}
```

produces

```
/tmp/api3915368247962961211.cpp:34:56: error: redefinition of parameter 'thiz'
        void (*hi)(testproj_kref_Foo thiz, const char* thiz);
                                                       ^
/tmp/api3915368247962961211.cpp:34:38: note: previous declaration is here
        void (*hi)(testproj_kref_Foo thiz, const char* thiz);
                                     ^
1 error generated.
```

Tested against latest master (82ad1799)

kyonifer:
2e07b3f resolved #1199 by replacing the parameter name with generic `p0, p1, ...` names. However, the original parameter names are useful to have in the generated header for people using an IDE.

This commit instead looks for name duplicates and appends a number when it finds one. For the original test case in #1199 it produces 

```c
        void (*hi)(testproj_kref_Foo thiz, const char* thiz1);
```

If no name conflicts are present on a given signature then the parameter names are not modified.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
@@ -220,8 +220,12 @@ private class ExportedElement(val kind: ElementKind,
                     TypeUtils.getClassDescriptor(owner.context.builtIns.nullableAnyType)!!
             else -> uniqueName(original) to TypeUtils.getClassDescriptor(original.returnType!!)!!
         }
+        val namesRecorded = mutableMapOf<String, Int>()
         val params = ArrayList(original.explicitParameters.mapIndexed() { index, it ->
-            "p$index /* ${it.name.asString()} */" to TypeUtils.getClassDescriptor(it.type)!!
+            val name = owner.translateName(it.name.asString())
+            val count = namesRecorded.getOrDefault(name, 0)
+            namesRecorded[name] = count + 1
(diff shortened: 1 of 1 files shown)


kyonifer:
I pushed it out to a `CAdapterGenerator` property to avoid generation per-function. If there's a more preferred location (e.g. static storage) it should be relocatable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -382,6 +383,17 @@ internal class CAdapterGenerator(
...
     private lateinit var outputStreamWriter: PrintWriter
+    private val paramNamesRecorded = mutableMapOf<String, Int>()
+
+    internal fun paramsToUniqueNames(params: List<ParameterDescriptor>): List<String> {
+        paramNamesRecorded.clear()
+        return params.map {
+            val name = translateName(it.name.asString()) 
+            val count = paramNamesRecorded.getOrDefault(name, 0)
+            paramNamesRecorded[name] = count + 1
+            "$name${count.toString().takeUnless { it == "0" } ?: ""}"
```
olonho: Looks like it is somewhat optimized for the uncommon case of name collision.

kyonifer: Yes that code is somewhat golfed. I'll fix.

PR #1209, opened 2018-01-03
Add Torch example

JuliusKunze:
This example might be interesting for people familiar with machine learning. Again, please feel free to close this Pull Request if this is too specific.

Tested on OSX.


samples/torch/src/main/kotlin/HelloTorch.kt
```diff
@@ -0,0 +1,104 @@
...
+        THFloatTensor_addmm(result, 1f, added.raw, 1f, raw, matrix.raw)
+        added.dispose()
+        return FloatMatrix(result)
+    }
+
+    override fun toString() = "[${toArray().joinToString(",\n") { "[${it.joinToString { it.toString() }}]" }}]"
+}
+
+fun tensor(vararg values: Float): FloatVector {
+    val tensor = FloatVector(THFloatTensor_newWithSize1d(values.size.toLong())!!)
```
olonho: I would suggest to rework example a bit, to make tensor disposal more automated, see for example https://github.com/JetBrains/kotlin-native/blob/master/samples/videoplayer/src/main/kotlin/Disposable.kt and its usage.

olonho: Also using signExtend() instead of toLong() may make code more portable.

samples/torch/README.md
```diff
@@ -0,0 +1,47 @@
...
+To build use `../gradlew build` or `./build.sh`.
+
+    ./downloadMNIST.sh
+
+will download and unzip the [MNIST dataset](https://en.wikipedia.org/wiki/MNIST_database) of 
+[70000 labeled handwritten digits](http://yann.lecun.com/exdb/mnist/) for training and testing a classifier.
+
+Then run 
+
+    ../gradlew run
```
olonho: You need to update upper level settings.gradle with smth like `include ':torch'`

samples/torch/README.md
```diff
@@ -0,0 +1,47 @@
...
+    
+Now
+
+    ./downloadTorch.sh
+
+will install it into `$HOME/.konan/third-party/torch` (if not yet done). 
+
+To build use `../gradlew build` or `./build.sh`.
+
+    ./downloadMNIST.sh
```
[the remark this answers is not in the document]
JuliusKunze: Forgot to add it. Fixed!

samples/torch/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+task downloadTorch(type: Exec) {
+    workingDir getProjectDir()
+    commandLine './downloadTorch.sh'
+}
+
+konanArtifacts {
+    interop('TorchInterop') {
+        defFile "src/main/c_interop/torch.def"
+        includeDirs "${torchHome}/include"
```
olonho: For me `includeDirs "${torchHome}/include", "${torchHome}/include/TH"` is needed for sample to compile.

samples/torch/src/main/kotlin/HelloTorch.kt
```diff
@@ -0,0 +1,624 @@
+import kotlinx.cinterop.*
+import torch.*
+import platform.posix.*
+
+interface Disposable {
+    fun dispose()
+}
+
+abstract class FloatTensor(val raw: CPointer<THFloatTensor>) : Disposable {
```
olonho: I'd suggest to split FloatTensor, FloatVector and disposable handling into separate files for better readability.

samples/torch/src/main/kotlin/HelloTorch.kt
```diff
@@ -0,0 +1,624 @@
+import kotlinx.cinterop.*
+import torch.*
+import platform.posix.*
```
olonho: Which POSIX APIs are used here?

JuliusKunze: Methods for reading files, only loading the digit data (see MNIST class from line 517)

samples/torch/README.md
JuliusKunze: Fixed

samples/torch/build.gradle
JuliusKunze: Fixed. (Sorry for this, I always used build.sh over build.gradle)

samples/torch/src/main/kotlin/HelloTorch.kt
JuliusKunze: Done!

samples/torch/src/main/kotlin/Dataset.kt
```diff
@@ -0,0 +1,101 @@
...
+    private fun <T> read(fileName: String, action: (Int, CPointer<ByteVar>) -> T): T {
+        val filePath = directory + fileName
+        val file = fopen(fileName, "rb") ?: throw Error("Cannot read input file $filePath")
+
+        try {
+            memScoped {
+                fseek(file, 0, SEEK_END)
+                val fileSize = ftell(file)
+                println("Reading $fileSize bytes from $filePath...")
+                val buffer = allocArray<ByteVar>(fileSize)
```
olonho: Probably regular ByteArray like in here would be somewhat cleaner: 
https://github.com/JetBrains/kotlinconf-spinner/blob/b9d4550d7611e732f2c54a54a3feeb4d4e1d5f7b/kotlin-native/samples/fullstack/common/src/main/kotlin/kommon/Kommon.kt#L11

samples/torch/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Installation
+
+To build [ATen (Torch for C)](https://github.com/pytorch/pytorch/tree/master/aten), 
+make sure you have Python 2.X and pyyaml installed:
+
+    # macOS: if you don't have pip
+    sudo easy_install pip
+    # Linux: if you don't have pip
+    apt-get -y install python-pip
```
olonho: Why do we need Python at all here? K/N binary only depends on C code/headers.

JuliusKunze: That is true, but the build process of ATen requires it, as described here: https://github.com/pytorch/pytorch/tree/master/aten

samples/torch/downloadTorch.sh
```diff
@@ -0,0 +1,24 @@
...
+    make DESTDIR=$TH_TARGET_DIRECTORY install
+
+    cd $TH_TARGET_DIRECTORY
+
+    # remove 'usr/local' prefix produced by make:
+    mv usr/local/* .
+    rm -d usr/local usr
+
+    # hack to solve "fatal error: 'generic/THNN.h' file not found" when linking, -I$<DIR>/include/THNN did not work
+    cp include/THNN/generic/THNN.h include/TH/generic/THNN.h
```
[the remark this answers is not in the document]
JuliusKunze: Interesting. If I remove this hack, compilation works through build.gradle, but not through build.sh:

Exception in thread "main" java.lang.Error: /Users/juliuskunze/.konan/third-party/torch/include/TH/THGenerateFloatType.h:13:10: fatal error: 'generic/THNN.h' file not found

Not sure why. Any idea?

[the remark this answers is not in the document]
olonho: Because gradle adds two -I after review feedback fix (via includeDirs), while build.sh still one.

samples/torch/src/main/kotlin/Dataset.kt
JuliusKunze: Great, that made it much better.

JuliusKunze:
Thank you. Kotlin/Native is also neat btw! (:

I guess you are right, it could be a project on its own. For now, I don't have the intention to create a fully-fledged Torch client for Kotlin/Native though.

samples/torch/downloadTorch.sh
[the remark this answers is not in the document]
olonho: Use smth like -copt "-I$TH_TARGET_DIRECTORY/include -I$TH_TARGET_DIRECTORY/include/TH" with quotes, it shall work.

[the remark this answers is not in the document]
JuliusKunze: Turns out linking with konanc already works, what does not work is stub generation.
``cinterop -def $DIR/src/main/c_interop/torch.def \
    -copt "-I$TH_TARGET_DIRECTORY/include -I$TH_TARGET_DIRECTORY/include/TH" \
    -labels $TARGET -o $DIR/build/c_interop/torch || exit 1``

fails with 
`Exception in thread "main" java.lang.Error: /var/folders/2m/_whs043j3wb869nhv7wk6dv00000gn/T/tmp4064000223995487764.c:1:10: fatal error: 'TH/TH.h' file not found`

Issue #1119, opened 2017-12-07
Execution failed for task ':calculator:compileKonanKotlinArithmeticParserIphone'


Samartan:
Well I am try to run the Calculator sample from the kotlin-native repo. 

Followed the following steps: 

To build and run the sample do the following:

1. Open samples/calculator/calculator.xcodeproj with Xcode.

2. Open the project's target through project navigator, go to tab 'General'. In 'Identity' section change the bundle ID to the unique string in reverse-DNS format. Then select the team in 'Signing' section. See the Xcode documentation for more info.

3. Now build and run the application on a connected iPhone with Xcode.

But when I try to build the application in Xcode I get the following error and the log to it is below: 

...
...
...
The target is not enabled on the current host: linux
Target 'linux' of artifact 'libgit2' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
Target 'linux' of artifact 'libcurl' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: linux
The target is not enabled on the current host: raspberrypi
Target 'linux' of artifact 'sdl' is not supported on the current host
Target 'raspberrypi' of artifact 'sdl' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: raspberrypi
Target 'linux' of artifact 'tetris' is not supported on the current host
Target 'raspberrypi' of artifact 'tetris' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: mingw
Target 'linux' of artifact 'ffmpeg' is not supported on the current host
Target 'mingw' of artifact 'ffmpeg' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: mingw
Target 'linux' of artifact 'sdl' is not supported on the current host
Target 'mingw' of artifact 'sdl' is not supported on the current host
The target is not enabled on the current host: linux
The target is not enabled on the current host: mingw
Target 'linux' of artifact 'Player' is not supported on the current host
Target 'mingw' of artifact 'Player' is not supported on the current host
The target is not enabled on the current host: mingw
:calculator:checkKonanCompiler
:calculator:compileKonanKotlinArithmeticParserIphone FAILED

FAILURE: Build failed with an exception.

[...]

PR #1213, opened 2018-01-09
Update Calculator sample README

mchowning:
Just adding a bit more context about the required setup since the calculator sample will likely be the first place a number of people go to check out Kotlin native and it wasn't clear from the sample's README that you needed to build the compiler first (See discussion in #1119).

### DIFF
--- a/samples/calculator/README.md
+++ b/samples/calculator/README.md
@@ -5,15 +5,16 @@ an existing iOS project (e.g. written in Swift or Objective-C).
 
 To build and run the sample do the following:
 
-1.  Open `samples/calculator/calculator.xcodeproj` with Xcode.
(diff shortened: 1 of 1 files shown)


StefMa:
Since the samples/Readme already include that information (see https://github.com/JetBrains/kotlin-native/blob/master/samples/README.md) I would do it.
As you already can read, the compile shouldn't be built if you run it from a archive.

mchowning:
Fair point. My thinking was that the readme for the calculator seemed to be giving directions for building from source instead of running from an archive so the additional direction was useful, but I'm certainly not opposed to declining this if it's not helpful or too duplicative of what is in the samples/README.md file.

olonho:
We'll probably fix that by somewhat better dependency checking mechanism. Thanks for your PR.

PR #1248, opened 2018-01-22
Brought back the Standalone and Linker tests into backend.native:test…

alexander-gorshenev:
…s:run.


olonho:
Failing tests?

buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -390,6 +392,8 @@ class BuildKonanTest extends ExtKonanTest {
  * Runs test built with Konan's TestRunner
  */
 class RunKonanTest extends ExtKonanTest {
+    public def inDevelopersRun = true
```
PavelPunegov: I think it could have been better to use a single property in KonanTest itself. All other subclasses would only set it to desired value if it's necessary

PR #1254, opened 2018-01-24
Add basic support for multi-platform projects to the compiler

SvyatoslavScherbina:


udalov:
Please note that after https://youtrack.jetbrains.com/issue/KT-21913 is fixed, removing all expected declarations is likely going to break cases when some of those declarations have parameters with default values. In JVM and JS back-ends, we use `ExpectActualResolver` (not currently in master; to be introduced with the fix for KT-21913) to get the expected declaration for an actual declaration, and load default argument values from expected declaration's sources. (I'll attach the corresponding commits here, as soon as they're in master.) This happens in terms of descriptors though, so I'm not sure whether it's possible to do that in IR.

SvyatoslavScherbina:
@udalov 
Thank you for noticing this!
I suppose it could be possible to support such cases by attaching default arguments from `expect` declarations to `actual` ones, just before deleting the former.
Do I understand correctly that even if default argument body is attached to `expect` function parameter, it wouldn't reference any `expect` descriptors?

udalov:
> Do I understand correctly that even if default argument body is attached to `expect` function parameter, it wouldn't reference any `expect` descriptors?

It should be that way, yes. Expected descriptors should lose in resolution to the corresponding actuals everywhere.

udalov:
FYI here's the implementation for KT-21913 in JVM: https://github.com/JetBrains/kotlin/commit/db4ce703a6d4ad70ce1eddc413042a37b34cd68a
And an addition for JS:
https://github.com/JetBrains/kotlin/commit/e0245a9dc7bc2b71d13ba617cc6c338ef9c4770e

SvyatoslavScherbina:
@udalov Thank you!

PR #1264, opened 2018-01-26
Minor code cleanup.

olonho:


runtime/src/main/cpp/Memory.h
```diff
@@ -155,11 +169,11 @@ struct ObjHeader {
   // Unsafe cast to ArrayHeader. Use carefully!
   ArrayHeader* array() { return reinterpret_cast<ArrayHeader*>(this); }
   const ArrayHeader* array() const { return reinterpret_cast<const ArrayHeader*>(this); }
-};
 
-inline bool isPermanent(const ObjHeader* obj) {
-  return isPermanent(obj->container());
-}
+  inline bool permanent() const {
```
PavelPunegov: Why is isPermanent() was renamed to permanent()? It now looks like a getter

olonho:
It is getter of ‘permanent’ logical property.

PR #1273, opened 2018-01-30
Add support for custom Xcode location

SvyatoslavScherbina:


backend.native/konan.properties
```diff
@@ -50,35 +50,24 @@ osVersionMinFlagLd.osx = -macosx_version_min
 osVersionMinFlagClang.osx = -mmacosx-version-min
 osVersionMin.osx = 10.11
 entrySelector.osx = -alias _Konan_main _main
-dependencies.osx = target-sysroot-2-darwin-macos \
+dependencies.osx = \
     libffi-3.2.1-2-darwin-macos \
-    clang-llvm-5.0.0-darwin-macos \
-    target-toolchain-1-osx
+    clang-llvm-5.0.0-darwin-macos
```
alexander-gorshenev: А где, в итоге, написано название каталога `target-toolchain-1-osx` для макоса? Или у нас такой катаклог никогда больше не будет использоваться

SvyatoslavScherbina: Absolute path would be either `~/.konan/dependencies/target-toolchain-1-osx` (if downloaded from internal server) or path to toolchain from current Xcode.
The patch to `konan.properties` only removes such directories from default dependencies (they are added conditionally in `AppleConfigurablesImpl.kt`) and specifies these dependencies to be available from internal server only.

PR #1277, opened 2018-01-31
Refactor calculator samples build:

SvyatoslavScherbina:
* Use Gradle composite build to reduce number of dependencies
  (e.g. dependency of Kotlin/Native samples on Android SDK)

* Move parser lib to calculator/parser

* Enable Android app

* Update README.md


StefMa:
The jvmClient sample works without issues 👌 
But the android App have issues.
I can't use the `parser`... The jar aren't generated...

bildschirmfoto 2018-01-31 um 13 56 30

SvyatoslavScherbina:
@StefMa 
I'm able to reproduce the issue, which is strange because yesterday the same code worked for me.
I'm going to investigate this issue, thank you for the report.

SvyatoslavScherbina:
@StefMa 
Yes, this bug appears randomly. If I close and reopen the project in Android Studio, the bug sometimes disappears, and everything works properly. Could you please try to close and reopen the project a few times?

The bug seems to be related to the following assert in IDE:
`com.intellij.openapi.roots.impl.RootModelImpl.addModuleOrderEntry(RootModelImpl.java:284)`

SvyatoslavScherbina:
@StefMa
So it seems to be a bug in Android Studio, related to Gradle composite build support. As a workaround we've decided to disable composite build for androidApp, and include parser sources directly to it: https://github.com/JetBrains/kotlin-native/pull/1277/commits/20acd2694545fa59dc409b5d55b4036cf87ea9d9

StefMa:
Yes, your are right. It was totally randomly..
Should we report the issue somewhere? 

Just checked out your fix and the android and cli samples works perfectly 👌 
Haven't tested the iOS sample again...

SvyatoslavScherbina:
@StefMa 
> Should we report the issue somewhere?

I'm not sure. Actually, I haven't found any mention about support for Gradle composite build in Android Studio. This could mean that it's not officially supported yet. Also before reporting the issue the sample should probably be simplified, e.g. written in Java and mostly using default IDE-generated Gradle scripts. I am going to make some attempts, but later.

Thank you for your feedback!

PR #1280, opened 2018-02-02
WIP: [gradle-plugin] Support mutliplatform projects in current DSL.

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/tasks/KonanCompileTask.kt
```diff
@@ -106,21 +125,21 @@ abstract class KonanCompileTask: KonanBuildingTask(), KonanCompileSpec {
         addKey("-ea", enableAssertions)
         addKey("--time", measureTime)
         addKey("-nodefaultlibs", noDefaultLibs)
+        addKey("-Xmulti-platform", commonProject != null)
```
max-kammerer: Would it be a problem if 'Xmulti-platform' would be passed explicitly by user in gradle script?

ilmat192: Looks like it wouldn't but I'll double-check it.

PR #1283, opened 2018-02-06
Temporary files directory

aceisnotmycard:
Do not save intermediate compiler outputs (like `program.kt.bc`) by default. 
Use `--temporary_files_dir` if needed.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OutputFiles.kt
```diff
@@ -0,0 +1,33 @@
+package org.jetbrains.kotlin.backend.konan
```
alexander-gorshenev: copyright.

PR #1288, opened 2018-02-07
[samples] Fix the globalState sample for Linux

ilmat192:


samples/globalState/src/main/kotlin/Global.kt
```diff
@@ -29,9 +29,12 @@ data class SharedDataMember(val double: Double)
 
 data class SharedData(val string: String, val int: Int, val member: SharedDataMember)
 
+// pthread_t is Long on Linux and a pointer on other systems. Ensure we can print it correctly.
+fun Any?.asString() = if (this is CPointer<*>) rawValue.toString() else toString()
+
 fun dumpShared(prefix: String): Unit {
     println("""
-            $prefix: ${pthread_self().rawValue} x=${sharedData.x} f=${sharedData.f} s=${sharedData.string!!.toKString()}
+            $prefix: ${pthread_self().asString()} x=${sharedData.x} f=${sharedData.f} s=${sharedData.string!!.toKString()}
```
SvyatoslavScherbina: `CPointer` class has sane `toString` implementation, so it should be enough to use `${pthread_self()}`.

olonho:
Fixed with https://github.com/JetBrains/kotlin-native/pull/1289

PR #1295, opened 2018-02-08
[warnings] g/c some warnings and idea inspections

vvlevchenko:


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KLibInstall.kt
```diff
@@ -28,7 +28,8 @@ open class KlibInstall: Exec() {
 
     override fun configure(config: Closure<*>): Task {
         val result = super.configure(config)
-        val konanHome = project.rootProject.file(project.findProperty("konan.home") ?: "dist")
+        val konanHomePath = project.findProperty("konan.home") ?: "dist" // Makes kotlin analyzer happy.
```
alexander-gorshenev: the comment is not needed, I'd say.

PR #1298, opened 2018-02-08
Zephyr subtargets

alexander-gorshenev:


msink:
Btw, next Zephyr release, planned in March, should support building by clang/llvm
https://github.com/zephyrproject-rtos/zephyr/pull/5901

Maybe better help them finish it, instead if fixing bugs in mixed clang+gcc environment?

alexander-gorshenev:
Thanks for the heads up. We will take a look.

PR #1299, opened 2018-02-08
Fix init code name clash.

olonho:


olonho:
Guess we need to rework init somewhat, to make it work on all platforms.

olonho:
No longer relevant.

PR #1325, opened 2018-02-14
Fixed trailing newline in consoleReadUtf8

aceisnotmycard:


runtime/src/main/cpp/Porting.cpp
```diff
@@ -77,10 +77,21 @@ void consoleErrorUtf8(const void* utf8, uint32_t sizeBytes) {
 uint32_t consoleReadUtf8(void* utf8, uint32_t maxSizeBytes) {
 #ifdef KONAN_ZEPHYR
   return 0;
-#else 
+#else
   char* result = ::fgets(reinterpret_cast<char*>(utf8), maxSizeBytes - 1, stdin);
   if (result == nullptr) return 0;
-  return ::strlen(result);
+  uint32_t length = ::strlen(result);
```
olonho: Looks somewhat self-repeating. Why not

const char* current = result + length - 1;
bool trimming = true;
while (current != result && trimming) {
   switch (*current) {
     case '\r':  case '\n':
         *current = 0;
         break;
     default:
      trimming = false;
    }
    current--;
 }
This way we could account for other possible linefeeds. Alternative is to do that from Kotlin side with https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-end.html

aceisnotmycard: Yes, it makes sense

runtime/src/main/cpp/Porting.h
```diff
@@ -29,7 +29,8 @@ void consoleInit();
 void consolePrintf(const char* format, ...);
 void consoleWriteUtf8(const void* utf8, uint32_t sizeBytes);
 void consoleErrorUtf8(const void* utf8, uint32_t sizeBytes);
-uint32_t consoleReadUtf8(void* utf8, uint32_t maxSizeBytes);
+// Negative return value denotes that read wasn't successful
```
olonho: Dot (.) at the end of sentence.

runtime/src/main/cpp/Porting.cpp
```diff
@@ -74,13 +74,29 @@ void consoleErrorUtf8(const void* utf8, uint32_t sizeBytes) {
...
   return 0;
-#else 
+#else
   char* result = ::fgets(reinterpret_cast<char*>(utf8), maxSizeBytes - 1, stdin);
-  if (result == nullptr) return 0;
-  return ::strlen(result);
+  if (result == nullptr) return -1;
+  int32_t length = ::strlen(result);
+  // fgets reads until EOF or newline so we need to remove linefeeds
+  char* iter = result + length - 1;
```
olonho: iter -> current

runtime/src/main/cpp/Porting.cpp
```diff
@@ -74,13 +74,29 @@ void consoleErrorUtf8(const void* utf8, uint32_t sizeBytes) {
...
 #ifdef KONAN_ZEPHYR
   return 0;
-#else 
+#else
   char* result = ::fgets(reinterpret_cast<char*>(utf8), maxSizeBytes - 1, stdin);
-  if (result == nullptr) return 0;
-  return ::strlen(result);
+  if (result == nullptr) return -1;
+  int32_t length = ::strlen(result);
+  // fgets reads until EOF or newline so we need to remove linefeeds
```
olonho: fgets() reads until EOF or newline so we need to remove linefeeds.

PR #1343, opened 2018-02-20
 Support importing string literal macros with cinterop

SvyatoslavScherbina:
Also add other minor improvements

### DIFF
--- a/Interop/Indexer/build.gradle
+++ b/Interop/Indexer/build.gradle
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import org.jetbrains.kotlin.konan.target.HostManager
+
 buildscript {
     apply from: "$rootDir/gradle/kotlinGradlePlugin.gradle"
 }
@@ -151,7 +153,7 @@ task updatePrebuilt {
         }
 
         copy {
-            from("$buildDir/nativelibs") {
+            from("$buildDir/nativelibs/${HostManager.host.visibleName}") {
                 include 'clangstubs.c'
             }
             into 'prebuilt/nativeInteropStubs/c'
--- a/Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
+++ b/Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
@@ -110,10 +110,7 @@ private fun reparseWithCodeSnippet(library: NativeLibrary,
         val codeSnippetLines = when (library.language) {
             // Note: __auto_type is a GNU extension which is supported by clang.
             Language.C, Language.OBJECTIVE_C -> listOf(
-                    "const __auto_type KNI_INDEXER_VARIABLE = $name;",
-                    // Clang evaluate API doesn't provide a way to get a `long long` value yet;
-                    // so extract such values from the enum declaration:
-                    "enum { KNI_INDEXER_ENUM_CONST = (long long)KNI_INDEXER_VARIABLE };"
+                    "const __auto_type KNI_INDEXER_VARIABLE = $name;"
             )
         }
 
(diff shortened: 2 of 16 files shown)


Interop/Indexer/build.gradle
```diff
@@ -151,7 +153,7 @@ task updatePrebuilt {
         }
 
         copy {
-            from("$buildDir/nativelibs") {
+            from("$buildDir/nativelibs/${HostManager.host.visibleName}") {
```
olonho: To support different stubs on different hosts? technically speaking, unless same tree is shared on different hosts shalln't be an issue.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
```diff
@@ -133,53 +130,27 @@ private fun processCodeSnippet(
 ): ConstantDef? {
 
     var state = VisitorState.EXPECT_VARIABLE
-    var evalResultKind: CXEvalResultKind? = null
-    var type: Type? = null
-    var longValue: Long? = null
-    var doubleValue: Double? = null
+    var evalResultOrNull: CXEvalResult? = null
```
olonho: WDYT about lateinit and isInitialized here?

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
```diff
@@ -191,21 +162,40 @@ private fun processCodeSnippet(
...
+                IntegerConstantDef(name, type, clang_EvalResult_getAsLongLong(evalResult))
+
+            CXEvalResultKind.CXEval_Float ->
+                FloatingConstantDef(name, type, clang_EvalResult_getAsDouble(evalResult))
+
+            CXEvalResultKind.CXEval_ObjCStrLiteral,
+            CXEvalResultKind.CXEval_StrLiteral ->
+                StringConstantDef(name, type, clang_EvalResult_getAsStr(evalResult)!!.toKString())
+
+            CXEvalResultKind.CXEval_CFStr,
```
olonho: Why? From https://chromium.googlesource.com/chromiumos/third_party/clang/+/master/tools/c-index-test/c-index-test.c it seems clang_EvalResult_getAsStr() works on those strings as well.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/CodeUtils.kt
```diff
@@ -54,7 +54,11 @@ fun String.quoteAsKotlinLiteral(): KotlinExpression {
 
     this.forEach { c ->
         val escaped = when (c) {
-            in 'a' .. 'z', in 'A' .. 'Z', in '0' .. '9', '_', '@', ':', '{', '}', '=', '[', ']', '^', '#', '*' -> c.toString()
+            in 'a' .. 'z', in 'A' .. 'Z',
+            in '0' .. '9',
+            '_', '@', ':', ';', '.', ',', '{', '}', '=', '[', ']', '^', '#', '*',
```
olonho: What about adding all printables to set, as this loops seems to be pretty frequent and really slow now?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
```diff
@@ -761,13 +761,24 @@ class StubGenerator(
...
-        val kotlinType = mirror(constant.type).argType.render(kotlinFile)
+        val kotlinType = when (constant) {
+            is IntegerConstantDef,
+            is FloatingConstantDef -> mirror(constant.type).argType
+
+            is StringConstantDef -> KotlinTypes.string
+
+            else -> {
+                // Not supported yet, ignore:
+                return
```
olonho: Maybe null and let in outer context?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
```diff
@@ -776,7 +787,7 @@ class StubGenerator(
         // Also it provokes constant propagation which can reduce binary compatibility
         // when replacing interop stubs without recompiling the application.
 
-        out("val ${constant.name.asSimpleName()}: $kotlinType = $literal")
+        out("val ${constant.name.asSimpleName()}: $kotlinType get() = $literal")
```
olonho: Why add getter? To avoid init phase?

backend.native/tests/build.gradle
```diff
@@ -2415,6 +2419,12 @@ task interop_globals(type: RunInteropKonanTest) {
     interop = 'cglobals'
 }
 
+task interop_macros(type: RunInteropKonanTest) {
+    disabled = (project.testTarget == 'wasm32') // No interop for wasm yet.
```
olonho: Even when it will be available, it will be JS and unlikely similar to what we have  for C. Maybe just // C interop is irrelevant to WASM.

backend.native/tests/interop/basics/cmacros.def
```diff
@@ -0,0 +1,12 @@
+---
+#define ZERO 0
+#define ONE 1
+#define MAX_LONG 9223372036854775807
+#define FOO_STRING "foo"
```
olonho: What about wide strings and such, i.e. http://en.cppreference.com/w/cpp/language/string_literal?

backend.native/tests/interop/basics/macros.kt
```diff
@@ -0,0 +1,19 @@
...
+    assertEquals("foo", FOO_STRING)
+    assertEquals(0, ZERO)
+    assertEquals(1, ONE)
+    assertEquals(Long.MAX_VALUE, MAX_LONG)
+    assertEquals(42, FOURTY_TWO)
+
+    val seventeen: Long = SEVENTEEN
+    assertEquals(17L, seventeen)
+
+    val onePointFive: Float = ONE_POINT_FIVE
```
olonho: Why you need intermediate variable?

Interop/Indexer/build.gradle
SvyatoslavScherbina: No, to fix the `updatePrebuilt` task: C stubs have moved to this directory.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
SvyatoslavScherbina: I've tried to use it and discovered that local lateinit doesn't support `isInitialized`.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
SvyatoslavScherbina: Because I don't know what it is. Do you?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/CodeUtils.kt
SvyatoslavScherbina: It makes sense.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
SvyatoslavScherbina: It doesn't seem to be better. Could you explain why it should?

backend.native/tests/interop/basics/macros.kt
SvyatoslavScherbina: To ensure the type is `Float`.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
olonho: https://developer.apple.com/documentation/corefoundation/cfstring?language=objc ?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
olonho: I just had to thing where this return returns from ;).

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
SvyatoslavScherbina: It is not the answer required.
What an expression is treated by libclang as evaluatable with kind `CXEval_CFStr`?

PR #1380, opened 2018-03-02
Remove ignore tag for tests that don't fail any more

PavelPunegov:
Also:
* Check for execution correctness inside the test
* Fix run_external to include compileKotlinAgainstKotlin


backend.native/tests/external/codegen/box/constants/kt9532.kt
```diff
@@ -22,7 +20,10 @@ object B {
 fun box(): String {
     if (A.a !== B.a) return "Fail 1: A.a !== B.a"
 
-    if (A.b !== B.b) return "Fail 2: A.b !== B.b"
+    if (A.b !== B.b) {
+        println("${A.b} & ${B.b}")
```
olonho: println?

PavelPunegov: This part should be removed.  Was added accidentally when I was investigating the reason of failure of this test

PR #1389, opened 2018-03-06
Target aliases: make targets mostly pairs. Resolve some current simple names as aliases.

alexander-gorshenev:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt
```diff
@@ -39,17 +39,17 @@ enum class Architecture(val bitness: Int) {
...
-    object IPHONE :         KonanTarget( "iphone",          Family.IOS,         Architecture.ARM64,     "ios")
-    object IPHONE_SIM :     KonanTarget( "iphone_sim",      Family.IOS,         Architecture.X64,       "ios_sim")
-    object LINUX :          KonanTarget( "linux",           Family.LINUX,       Architecture.X64,       "linux")
-    object MINGW :          KonanTarget( "mingw",           Family.WINDOWS,     Architecture.X64,       "mingw")
-    object MACBOOK :        KonanTarget( "macbook",         Family.OSX,         Architecture.X64,       "osx")
-    object RASPBERRYPI :    KonanTarget( "raspberrypi",     Family.LINUX,       Architecture.ARM32,     "raspberrypi")
-    object LINUX_MIPS32 :   KonanTarget( "linux_mips32",    Family.LINUX,       Architecture.MIPS32,    "linux_mips32")
-    object LINUX_MIPSEL32 : KonanTarget( "linux_mipsel32",  Family.LINUX,       Architecture.MIPSEL32,  "linux_mipsel32")
-    object WASM32 :         KonanTarget( "wasm32",          Family.WASM,        Architecture.WASM32,    "wasm32")
+    object ANDROID_ARM32 :  KonanTarget( "android_arm32",   Family.ANDROID, Architecture.ARM32,     "android_arm32")
```
olonho: Why like that, not an alias list instead of detailedName?

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
```diff
@@ -81,7 +81,7 @@ internal class KonanInteropRunner(project: Project)
     : KonanCliRunner("cinterop", "Kotlin/Native cinterop tool", project)
 {
     init {
-        if (project.host == "mingw") {
+        if (project.host == "mingw_x64") {
```
olonho: Ideas for better fix?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt
alexander-gorshenev: We need to remove the "detailedName" field completely. It's a rudiment of early stages. That will be a separate followup PR.

olonho: Think list of short aliases like ios or linux is good, and I'm fine with keeping it.

PR #1390, opened 2018-03-06
Object freezing API

olonho:


ilya-g:
Extensions `freeze` and `isFrozen` have high chances to conflict with the same named members.

olonho:
Any ideas on better naming?

elizarov:
Please consider having `frozen` in addition to `freeze` and `isFrozen`. The semantics of `frozen` should be quite straightforward -- it should return the receive if it is frozen or copy the receiver and return a frozen copy of it, it it is not. 

We'll need it for a portable (cross-platform) approach to freezability in coroutines. Consider the code like this: 

```
val list = mutableListOf<SomeData>()
// populate the list
channel.send(list) // send this list to some coroutine, potentially in different thread, via channel
```

We'd like this code to safely work out-of-the box on JVM, JS, and Native. The idea is that implementation of `channel.send` would invoke `frozen` to get a frozen copy of the object and then send it. 

If we `freeze` in `channel.send` that it will introduce a side-effect on the element that we send, which is bad. Also, we cannot implement `freeze` on JVM collection (since they do not support freezing), however we can implement `frozen` on JVM collection which returns a copy.

Note, that is an invoker cares about performance and has a freezable object at hand, then it can invoke `freeze` before using `channel.send` in which case the extra invocation of `frozen` inside `send` implementation will not do any harm.

olonho:
Regarding frozen() API: do not think that we could create general deep copy of Any?, as this may have strange effects on objects where cloning is not desirable (i.e. objects, enums, etc).
I believe that such API could be built on top of provided APIs for classes with clear cloning semantics. And as this API currently is Native-specific anyway, some porting layer for cross-platform code shall exist anyway.

elizarov:
`frozen` should throw an exception if the object is neither frozen nor copyable (cloneable). It is the correct behavior to write a safe code based on coroutines. The basic idea that there are only two "kinds" of objects that you can send from one coroutine to another -- either frozen (immutable) objects or copyable (cloneable) objects. It is Ok if we start with support for frozen objects only, but it is quite limited and cannot be easily used from a portable code.

olonho:
In current K/N there's no explicit cloneability attribute (like Java's `implements Cloneable`), so we cannot write such code in universal manner. Guess library can handle proper cloning as needed.

PR #1398, opened 2018-03-12
Typed arrays API

olonho:


runtime/src/main/cpp/Arrays.cpp
```diff
@@ -131,6 +131,102 @@ KInt Kotlin_ByteArray_getArrayLength(KConstRef thiz) {
...
+  ArrayHeader* array = thiz->array();
+  if (static_cast<uint32_t>(index + 7) >= array->count_) {
+    ThrowArrayIndexOutOfBoundsException();
+  }
+  *reinterpret_cast<KLong*>(ByteArrayAddressOfElementAt(array, index)) = value;
+}
+
+void Kotlin_ByteArray_setFloatAt(KRef thiz, KInt index, KChar value) {
+  ArrayHeader* array = thiz->array();
+  if (static_cast<uint32_t>(index + 3) >= array->count_) {
```
PavelPunegov: Have you checked that conversion is correct in case of INFINITY, or other huge floating-point value?

The C++ standard says that such conversions are Undefined Behaviors

olonho: Guess here correct is any behavior that is fast :).

PR #1400, opened 2018-03-12
Add windows opengl32 platform library

soywiz:


platformLibs/src/platform/windows/opengl32.def
```diff
@@ -0,0 +1,5 @@
+depends = windows posix
+package = platform.opengl32
+headers = wtypes.h minwindef.h wtypes.h GL/gl.h GL/wglext.h
```
alexander-gorshenev: Why `wtypes.h` is specified twice?

alexander-gorshenev: I think we need some header filter here, so that we don't bind irrelevant declarations. Something similar to `headerFilter = GL/**`.

soywiz: I tired `gl/**` but `GL_` constants were not binded. `gl/** GL_/**` would include gl functions and GL_ constants?

olonho:
Following .def file seems to work fine:
```depends = windows posix
package = platform.opengl32
headers = GL/gl.h GL/wglext.h
headerFilter = GL/**
compilerOpts = -DUNICODE
linkerOpts = -lopengl32```

File: platformLibs/src/platform/windows/opengl32.def (hunk 1 above)
Comment (soywiz) [reply platformLibs/src/platform/windows/opengl32.def]: I didn't noticed. I think I copied the gdiplus.def one:
https://github.com/JetBrains/kotlin-native/blob/b34bfabe468c59fb43c6361ae67b8eb09a35ac21/platformLibs/src/platform/windows/gdiplus.def#L3
It is repeated there too

**Comment by soywiz** [comment]
I have updated it. Tried `GL/**` but included windows stuff. Why? Probably because in the GL folder there is a `wglext.h` that includes windows stuff. This one seems to work.

**Comment by soywiz** [comment]
Note that in windows "ext" functions have to be loaded dynamically (constants are available as normal). Something like this will still be needed to access things like shaders from GL >= 2.0:
```kotlin
val glCreateProgram by lazy { wglGetProcAddress("glCreateProgram").uncheckedCast<PFNGLCREATEPROGRAMPROC>() }
// ...
```

PR #1404, opened 2018-03-13
Better Xcode projects for calculator and uikit samples

stigger:


samples/uikit/konan/ViewController.h
```diff
@@ -1,20 +0,0 @@
...
-
-#import <UIKit/UIKit.h>
-
-@interface ViewController : UIViewController
-
-@property (weak, nonatomic) IBOutlet UILabel *label;
-@property (weak, nonatomic) IBOutlet UITextField *textField;
-@property (weak, nonatomic) IBOutlet UIButton *button;
-
-- (IBAction)buttonPressed;
```
SvyatoslavScherbina: How to use Xcode Interface Builder to specify event handler after removing this fake header?

stigger: By editing the storyboard manually. It sucks, but at least it is more or less logical. Having a dummy header that you need to manually keep in sync with .kt seems even worse to me. People will be confused, why they've added a new outlet to the .kt, but it the Interface Builder does not see it.

bagza: Well, studying the kotlinconf-app for the first time I was very confused with empty headers. Having the only source of truth is vital while storyboard usage is a matter of taste.

Albert-Gao: Now I get a little bit confused. Do we need to edit the storyboard manually every time if we try to set up the project? Previously I just create a sample app from Wizard of XCode and add Kotlin support then it works. Now this commit removed all the things. How to re-create this uikit sample? Any explanations for this? :(

PR #1412, opened 2018-03-14
Platform library for iconv.h on macos, ios and linux.

alexander-gorshenev:


platformLibs/src/platform/linux/iconv.def
```diff
@@ -0,0 +1,3 @@
+depends = posix
+package = platform.iconv
```
olonho: Does it work same way on MIPS and Android Linuxes?

Issue #1229, opened 2018-01-14
kotlin-native on arduino 


ghost:
[based on this](https://blog.jetbrains.com/kotlin/2017/04/kotlinnative-tech-preview-kotlin-without-a-vm/) 
- iOS applications (reusing code with Android)
- Embedded systems/IoT (e.g., Arduino and beyond)

I wondered if there is progress? 

Just read that [avr-llvm](https://github.com/avr-llvm/llvm) was merged into the main tree some months ago and [Kotlin-Native on iOS](https://blog.jetbrains.com/kotlin/2017/11/kotlinnative-v0-4-released-objective-c-interop-webassembly-and-more/) is already working.

thanks

StefMa:
Hi,

I started a very basic "mobile sample" here:
https://github.com/StefMa/kotlin-native/compare/master...StefMa:mobile_sample

There is a "shared library" with pure kotlin code. The iOS project will
create a framework on building of that shared lib and include it.
The android project will create a jar on bulking of that shared lib and
include it.

Its far from perfect. But I shows you how to start with both platforms and
a shared lib written in kotlin...

msink:
Currently Kotlin/Native aplication require at least 200-300 kilobytes of memory - will not fit in arduino, needs something like stm32f4.

olonho:
Yes, embedded targets are being worked on, however AVR-based Arduino are too low end from both CPU and memory standpoints.

MRezaNasirloo:
Golang is already on 34 different platform including arduino uno with 2KB of SRAM. I hope you guys figured out soon 👍

monouser7dig:
golang is not on Arduino, gobot only talks to Arduino but does not run on Arduino, otherwise that would be nice as well but does not seem to be feasible.

Arkuda:
So, if use esp8266/esp32 you can try join https://github.com/olonho/carkot/tree/master/translator and https://www.esp32.com/viewtopic.php?t=9226 
(for short: first its a kotlin to llvm translator, second its a llvm for esp8266/esp32)
But, its very hard.

thawkins:
Go on ardunio uno using llvm 

https://github.com/tinygo-org/tinygo

msink:
It supports ARM based microcontrollers and WASM, same as Kotlin/Native already do.

carlos4242:
[...]

PR #1418, opened 2018-03-16
Add nucleo_f412zg target and example

turbinenreiter:
This PR adds support for the NUCLEO-F412ZG board.

* `build.sh` in the zephyr example now takes BOARD as argument
* `build.sh` in the zephyr example now takes ZEPHYR_BASE as argument
* NUCLEO-F412ZG added to platforms
* NUCLEO-F412ZG added to c_interop in the zephyr example
* a separate `main.kt` for each target

I don't know how useful this PR is and how it relates to the plans for the zephyr targets, but it sure can act as a reference on which changes are needed at the moment to add a board target.
I'd be happy about feedback and also directions on how to go on, so I can make sure my efforts are in line with the projects plans.


alexander-gorshenev:
So, is it blinking?

turbinenreiter:
It is!

https://youtu.be/5QSA2teJM6A

jwoolston:
@turbinenreiter I know this has sat open for a bit and things have changed. Out of curiosity, to help me debug some of my own issues, if you use the latest kotlin-native/zephyr, does this still work? If not, what version of each were you using?

turbinenreiter:
I'm not sure if it works on the latest. I havn't been working on it in the meantime and I don't have the toolchain set up anymore. I used whatever the latest version was in March last year.

sbogolepov:
Hi. First of all, thank you for the PR!
Unfortunately, I have to close it for the following reasons:
1. Zephyr target has proven to be a bad extension point for arbitrary targets. For example, it's hard to provide support for these targets in tooling and kotlinx.libraries.
2. Another major problem is keeping target support up to date with changes in the compiler and runtime. It's already hard with the
   current set of targets, so we have to be quite conservative when deciding to support a new native target in Kotlin.

As I mentioned [here](https://github.com/JetBrains/kotlin-native/issues/1229#issuecomment-781844531), we have some ideas about a solution for these problems, but it is no easy task to implement them.

PR #1422, opened 2018-03-19
Concurrency model writeup.

olonho:


CONCURRENCY.md
```diff
@@ -0,0 +1,102 @@
...
+  int version;
+  void* kotlinObject;
+} SharedData;
+
+SharedData sharedData;
+```
+After running cinterop tool it allows sharing Kotlin data in versionized global structure,
+and interact with it from Kotlin transparently via autogenerated Kotlin like this:
+```kotlin
+class SharedData(rawPtr: NativePtr) : CStructVar(rawPtr) {
```
SvyatoslavScherbina: What's the purpose of including the whole generated stub for this `struct` instead of something like
```kotlin
class SharedData {
    var version: Int
    var kotlinObject: COpaquePointer?
}
```
?

CONCURRENCY.md
```diff
@@ -0,0 +1,102 @@
...
+   * Raw shared memory using C globals
+
+ ## Workers
+
+  Instead of threads Kotlin/Native runtime offers concept of workers: concurrently executing
+ control flow streams with an associated request queue. Workers are very similar to actors
+ in Actor Model. Worker can exchange Kotlin objects with other workers, so that at the moment
+ each mutable object is owned by the single worker, but ownership could be transferred.
+ See section [Object transfer and freezing](#transfer).
+
```
alexander-gorshenev: A link to `samples/workers`

CONCURRENCY.md
```diff
@@ -0,0 +1,102 @@
...
+  Once worker is started with `startWorker` function call, it can be uniquely addressed with an integer
+ worker id. Other workers, or non-worker concurrency primitives, such as OS threads, could send a message
+ to the worker with `schedule` call.
+ ```kotlin
+   val future = schedule(TransferMode.CHECKED, { SomeDataForWorker() }) {
+       // data returned by the second function argument comes to the
+       // worker routine as 'input' parameter.
+       input ->
```
SvyatoslavScherbina: Following Kotlin code formatting conventions would make this code more readable.

olonho: IDEA believes this code is properly formatted.

CONCURRENCY.md
olonho: OK

CONCURRENCY.md
olonho: OK

CONCURRENCY.md
SvyatoslavScherbina: Yes, you are right. Moreover, this formatting seems to be explicitly recommended for some cases: https://kotlinlang.org/docs/reference/coding-conventions.html#lambda-formatting

PR #1424, opened 2018-03-20
Force LLVM to inline basic math functions

aceisnotmycard:


runtime/src/main/cpp/Operator.cpp
```diff
@@ -36,28 +38,28 @@ extern "C" {
 
 //--- Boolean -----------------------------------------------------------------//
 
-KBoolean Kotlin_Boolean_not              (KBoolean a            ) { return     !a; }
-KBoolean Kotlin_Boolean_and_Boolean      (KBoolean a, KBoolean b) { return a && b; }
-KBoolean Kotlin_Boolean_or_Boolean       (KBoolean a, KBoolean b) { return a || b; }
-KBoolean Kotlin_Boolean_xor_Boolean      (KBoolean a, KBoolean b) { return a != b; }
+ALWAYS_INLINE KBoolean Kotlin_Boolean_not              (KBoolean a            ) { return     !a; }
```
olonho: Does it really work in the expected way?

aceisnotmycard: Yes, except boolean operations. Trying to fix it.

SvyatoslavScherbina: AFAIR, LLVM doesn't respect `always_inline` attribute during LTO.

aceisnotmycard: See https://github.com/llvm-mirror/llvm/blob/master/lib/Analysis/InlineCost.cpp#L1978. `llvm::getInlineCost` is called by `SimpleInliner` which is used by `LTOCodeGenerator`

SvyatoslavScherbina: However, as you mentioned, boolean operations aren't being inlined despite the attribute :)
And I've experienced the similar behaviour with other functions.

SvyatoslavScherbina: Is `SimpleInliner` enabled by `LTOCodeGenerator` when compiling with low optimization levels?

aceisnotmycard: Starting from -O2.

SvyatoslavScherbina: Yes, it seems to be the reason of the behaviour I've observed.
I mean, `clang` respects `always_inline` even with `-O0` specified.

aceisnotmycard: Got it :)
All callsites to Kotlin_Boolean_* looks kinda funny:
```
%231 = invoke i1 bitcast (i8 (i8, i8)* @Kotlin_Boolean_or_Boolean to i1 (i1, i1)*)(i1 %229, i1 %230)
```
Bitcasting prevents calls from inlining. If I change signature of these functions from `KBoolean` (aka `uint8_t`) to `bool` then inlinig is working.

olonho: Может нам в кодогенераторе как раз представлять Boolean как i8, а не i1?

olonho: Просто i1 невнятный тип с точки зрения ABI

SvyatoslavScherbina: This makes some sense. However, it would require an additional (bitcode) instruction on every conditional branch based on Kotlin boolean value.
AFAIR, @vvlevchenko is aware of some additional motivation behind representing `Boolean` as `i1`

PR #1440, opened 2018-03-23
Introduce environment variable to .konan parent dir

cy6erGn0m:
It is important to be able to change ~/.konan dir location as user's home directory could have limited capacity or could be mounted via NFS (or both)

The environment variable name is still open question

### DIFF
--- a/shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt
+++ b/shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt
@@ -53,6 +53,8 @@ sealed class KonanTarget(override val name: String, val family: Family, val arch
 
     // Tunable targets
     class ZEPHYR(val subName: String, val genericName: String = "zephyr") : KonanTarget("${genericName}_$subName", Family.ZEPHYR, Architecture.ARM32)
+
+    override fun toString() = name
 }
 
 fun hostTargetSuffix(host: KonanTarget, target: KonanTarget) =
--- a/shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/DependencyProcessor.kt
+++ b/shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/DependencyProcessor.kt
@@ -45,9 +45,6 @@ private val Properties.downloadingAttemptIntervalMs : Long
     get() = getProperty("downloadingAttemptPauseMs")?.toLong()
             ?: DependencyDownloader.DEFAULT_ATTEMPT_INTERVAL_MS
 
-private val Properties.homeDependencyCache : String
-    get() = getProperty("homeDependencyCache") ?: DependencyProcessor.DEFAULT_HOME_DEPENDENCY_CACHE
-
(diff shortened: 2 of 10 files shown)


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt
```diff
@@ -53,6 +53,8 @@ sealed class KonanTarget(override val name: String, val family: Family, val arch
 
     // Tunable targets
     class ZEPHYR(val subName: String, val genericName: String = "zephyr") : KonanTarget("${genericName}_$subName", Family.ZEPHYR, Architecture.ARM32, "${genericName}_$subName")
+
+    override fun toString() = name
```
cy6erGn0m: `./gradlew tasks`

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/DependencyProcessor.kt
```diff
@@ -212,11 +206,16 @@ class DependencyProcessor(dependenciesRoot: File,
...
+            File(System.getenv("KONAN_DATA_DIR") ?: (System.getProperty("user.home") + File.separator + ".konan"))
+        }
 
         @JvmStatic
-        val defaultDependenciesRoot
-            get() = Paths.get(System.getProperty("user.home")).resolve(".konan/dependencies").toFile()
+        val defaultDependenciesRoot: File
+            get() = localKonanDir.resolve("dependencies")
+
+        val dependencyCacheDir: File
```
ilmat192: I think `defaultDependencyCacheDir` would be better

ilmat192:
I've made a search in the project and `$HOME/.konan` is used in many places:
 - `org/jetbrains/kotlin/cli/klib/main.kt`, the `defaultRepository` property;
 - `dependencies/build.gradle`, the `rmDotKonan` task;
 - `samples/`, `.konan` is used in many shell scripts (`konan.sh`, scripts in torch, tensorflow and tetris samples etc).

Looks like we need to consume the `KONAN_DATA_DIR` variable in these places too.

Also we need to write a couple of words about this setting in some readme file. Looks like `HACKING.md` is the most suitable doc for this. Also `LIBRARIES.md` says about the default repository path so `KONAN_DATA_DIR` worth mentioning there too.

cy6erGn0m:
Fixed klib/main.kt and build.gradle, added docs in both hacking and libraries markdown pages. However I am not sure why should we apply these changes in samples?

ilmat192:
Some samples like tetris or zephyr explicitly access toolchains/clang and use the hardcoded path to `.konan/dependencies`. If the compiler downloaded these dependencies in another location, these samples may fail during building.
Note: the tetris sample uses this path not only in *.sh but also in the gradle build script.

Also tensorflow and torch samples use the `$HOME/.konan` dir to store libraries needed to build them. Having all Kotlin/Native stuff including these libraries in one place is convenient but we also may leave them in `$HOME/.konan/third-party`. @alexander-gorshenev @olonho WDYT?

PR #1453, opened 2018-03-29
Fix few problems in generation of C bindings.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -267,9 +291,11 @@ private class ExportedElement(val kind: ElementKind,
             else -> uniqueName(original, shortName) to TypeUtils.getClassDescriptor(original.returnType!!)!!
         }
         val uniqueNames = owner.paramsToUniqueNames(original.explicitParameters)
-        val params = ArrayList(original.explicitParameters.mapIndexed { idx, it ->
-            uniqueNames[idx] to TypeUtils.getClassDescriptor(it.type)!!
-        })
+        val params = ArrayList(original.explicitParameters
+                .filter { it.type.includeToSignature() }
+                .mapIndexed { idx, it ->
+                    uniqueNames[idx] to TypeUtils.getClassDescriptor(it.type)!!
```
SvyatoslavScherbina: Does it work correctly without adjusting `uniqueNames` above?

olonho: Right, will fix.

PR #1455, opened 2018-03-30
[build][version] generate KonanVersion using data from gradle.properties.

vvlevchenko:
- uses konanMetaVersion as meta version field

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,6 +22,6 @@ remoteRoot=konan_tests
 testDataVersion=1226829:id
 kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_CompilerAllPlugins),number:1.2.50-dev-515,tag:kotlin-native,pinned:true/artifacts/content/maven
 kotlinVersion=1.2.50-dev-515
-konanVersion=0.6.2
+konanVersion=0.7
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 
--- a/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/model/KonanModel.kt
+++ b/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/model/KonanModel.kt
@@ -23,7 +23,6 @@ import java.io.Serializable
  * This model is shared with the client processes such as an IDE.
  */
 interface KonanModel : Serializable {
-    val konanVersion: String
     val artifacts: List<KonanArtifact>
 }
 
--- a/GRADLE_PLUGIN.md
+++ b/GRADLE_PLUGIN.md
@@ -26,13 +26,8 @@ The Kotlin/Native plugin depends on `org.jetbrains.kotlin:kotlin-gradle-plugin`.
 plugins as buildscript dependencies, it's recommended to **declare them in the same `build.gradle`** to avoid issues with
 plugin classpath.
 
-The plugin downloads the compiler during its first run. You may specify a version of the compiler using `konan.version`
-project property:
-
-    konan.version=0.3
-
 If you already downloaded the compiler manually you may specify the path to its root directory using `konan.home`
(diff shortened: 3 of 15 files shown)


gradle.properties
```diff
@@ -21,7 +21,6 @@ remoteRoot=konan_tests
 # Download artifacts of https://teamcity.jetbrains.com/viewType.html?buildTypeId=Kotlin_120_Compiler
 testDataVersion=1226829:id
 kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_CompilerAllPlugins),number:1.2.50-dev-139/artifacts/content/maven
-kotlinVersion=1.2.50-dev-139
```
ilmat192: Looks like we still need it

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/VersionGenerator.kt
```diff
@@ -0,0 +1,45 @@
...
+                val meta = project.properties["konanMetaVersion"]?.let { "MetaVersion.${it.toString().toUpperCase()}" }
+
+                +""" package org.jetbrains.kotlin.backend.konan
+             class KonanVersion(val meta: MetaVersion?, val major: Int, val minor: Int, val maintenance: Int, val build:Int) {
+                 companion object {
+                     val CURRENT = KonanVersion($meta, $major, $minor, $maintenance, $build)
+                 }
+                 override fun toString() = if (meta != null) "${'$'}meta ${'$'}major.${'$'}minor.${'$'}maintenance-${'$'}build" else "${'$'}major.${'$'}minor.${'$'}maintenance-${'$'}build"
+             }
+            """.trimMargin()
```
ilmat192: It should be `trimIndent`.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/VersionGenerator.kt
```diff
@@ -0,0 +1,45 @@
...
+            val content = buildString {
+                operator fun String.unaryPlus() = this@buildString.append(this)
+                val version = project.properties["konanVersion"].toString().split(".")
+                val major = version[0].toInt()
+                val minor = version[1].toInt()
+                val maintenance = if (version.size > 2) version[2].toInt() else 0
+                val build = System.getenv("BUILD_NUMBER")?.toInt() ?: -1
+                val meta = project.properties["konanMetaVersion"]?.let { "MetaVersion.${it.toString().toUpperCase()}" }
+
+                +""" package org.jetbrains.kotlin.backend.konan
```
ilmat192: This string has strange indentation and is inconvenient to read. May be reformat it?
```
+"""
    package org.jetbrains.kotlin.backend.konan
    class KonanVersion(
            val meta: MetaVersion?,
            val major: Int,
            val minor: Int,
            val maintenance: Int,
            val build:Int
    ) {
        companion object {
            val CURRENT = KonanVersion($meta, $major, $minor, $maintenance, $build)
        }

        override fun toString() =
                if (meta != null) {
                    "${'$'}meta ${'$'}major.${'$'}minor.${'$'}maintenance-${'$'}build"
                } else {
                    "${'$'}major.${'$'}minor.${'$'}maintenance-${'$'}build"
                }
    }
""".trimIndent()
```

backend.native/build.gradle
```diff
@@ -39,6 +41,7 @@ allprojects {
         maven { url kotlinCompilerRepo }
     }
 }
+task compilerVersionGeneration(type: VersionGenerator){}
```
ilmat192: I think it's better to name it in the same manner as the protobuf generation task named. Something like `generateCompilerVersion`

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/model/KonanModel.kt
```diff
@@ -16,14 +16,14 @@
 
 package org.jetbrains.kotlin.gradle.plugin.model
 
+import org.jetbrains.kotlin.konan.KonanVersion
```
ilmat192: Unused?

PR #1456, opened 2018-03-30
Update toolchain.

olonho:


tools/scripts/update_xcode.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+KONAN_TOOLCHAIN_VERSION=4
+SDKS="macosx iphoneos iphonesimulator"
```
alexander-gorshenev: iphon-e-simulator or iphon-o-simulator?

PR #1466, opened 2018-04-04
[gradle-plugin] Pass system properties into the compiler

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
```diff
@@ -61,6 +59,11 @@ internal abstract class KonanCliRunner(val toolName: String, val fullName: Strin
         addAll(project.jvmArgs)
     }
 
+    override val additionalSystemProperties = mutableMapOf(
+            "konan.home" to project.konanHome,
+            "java.library.path" to "${project.konanHome}/konan/nativelib"
```
olonho: May we need to append user supplied  java.library.path and allow to override konan.home with CLI args?

olonho: Maybe not.

ilmat192: `konan.home` can be overridden using a project property. As for `java.library.path`, I don't sure that we need it.

PR #1471, opened 2018-04-05
Implement scope-local C pointer.

olonho:


INTEROP.md
```diff
@@ -337,6 +337,22 @@ manually:
 
 In all cases the C string is supposed to be encoded as UTF-8.
 
+### Scope-local pointers ###
+
+It is possible to create scope-stable pointer of C representation of `CValues<T>`
+instance using `MemScope.ptr` extension property. It allows to use APIs which requires C
```
SvyatoslavScherbina: It's better to refer this property as "`CValues<T>.ptr` extension property available under `memScoped { ... }`".

olonho: Done

samples/androidNativeActivity/src/main/kotlin/renderer.kt
```diff
@@ -306,10 +306,10 @@ class Renderer(val parentArena: NativePlacement, val nativeActivity: ANativeActi
         }
 
         glFrontFace(GL_CW)
-        glVertexPointer(4, GL_FLOAT, 0, vertices.toFloatArray().toCValues().getPointer(this))
-        glTexCoordPointer(2, GL_FLOAT, 0, texCoords.toFloatArray().toCValues().getPointer(this))
-        glNormalPointer(GL_FLOAT, 0, normals.toFloatArray().toCValues().getPointer(this))
-        glDrawElements(GL_TRIANGLES, triangles.size, GL_UNSIGNED_BYTE, triangles.toByteArray().toCValues().getPointer(this))
+        glVertexPointer(4, GL_FLOAT, 0, vertices.toFloatArray().toCValues().ptr)
```
abreslav: Looks like `toCValues().ptr` is a rather common operation. Do you think we could add a shortcut for it?

samples/uikit/src/main/kotlin/main.kt
```diff
@@ -5,7 +5,7 @@ import platform.UIKit.*
 fun main(args: Array<String>) {
     memScoped {
         val argc = args.size + 1
-        val argv = (arrayOf("konan") + args).map { it.cstr.getPointer(memScope) }.toCValues()
+        val argv = (arrayOf("konan") + args).map { it.cstr.ptr }.toCValues()
```
abreslav: What's the difference between `getPointer(memScope)` and `getPointer(this)`?

olonho: Those are synonymous: https://github.com/JetBrains/kotlin-native/blob/0a911ed5bbf5413f54facc0b37961cfc3fae0310/Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt#L413

samples/androidNativeActivity/src/main/kotlin/renderer.kt
olonho: Problem is that we have multiple toCValues() methods, so we must provides multiple shortcuts then, and semantics would be pretty magical.

PR #1478, opened 2018-04-06
Respect JAVA_OPTS env variable in CLI tools.

olonho:


anidotnet:
@olonho I see in [run_konan.bat](https://github.com/JetBrains/kotlin-native/blob/master/cmd/run_konan.bat#L71) also JAVA_OPTS is not honored. Could you please change it like 

```bat
set JAVA_OPTS=-ea ^
    -Xmx3G ^
    "-Djava.library.path=%NATIVE_LIB%" ^
    "-Dkonan.home=%_KONAN_HOME%" ^
    -Dfile.encoding=UTF-8 ^
	;%JAVA_OPTS%
```

PR #1481, opened 2018-04-10
 Unify interop with Objective-C in both directions

SvyatoslavScherbina:

PR #1482, opened 2018-04-10
Add very basic documentation for interop with Swift/Objective-C

SvyatoslavScherbina:


SvyatoslavScherbina:
To be merged after #1481

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
+# _Kotlin/Native_ interoperability with Swift/Objective-C
+
+This documents covers some details of Kotlin/Native interoperability with
+Swift/Objective-C.
+
+## Usage
+
+Kotlin/Native provides bidirectional interoperability with Objective-C.
+Objective-C frameworks and libraries can be used in Kotlin code if
+properly attached to the build (system frameworks are attached by default).
```
olonho: "attached to" -> "integrated with", attachment assumes somewhat asymmetrical relation.

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+See e.g. "Interop libraries" in
+[Gradle plugin documentation](GRADLE_PLUGIN.md#building-artifacts).
+Swift library can be used in Kotlin code if its API is exported to Objective-C
+with `@objc`.
```
olonho: Maybe explicitly mention "Pure Swift modules are not yet supported"?

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+
+Kotlin module can be used in Swift/Objective-C code if compiled into a
+[framework](GRADLE_PLUGIN.md#framework). See [calculator sample](samples/calculator)
+as an example.
+
+## Mappings
+
+The table below shows how Kotlin concepts are mapped to Swift/Objective-C and vice versa.
+
+| Kotlin | Swift | Objective-C | Notes |
```
olonho: +100!

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+Objective-C classes are imported into Kotlin with their original names.
+Protocols are imported as interfaces with `Protocol` name suffix,
+i.e. `@protocol Foo` -> `interface FooProtocol`.
+These classes and interfaces are placed into package [specified in build configuration](#usage)
+(`platform.*` packages for preconfigured system frameworks).
+
+Names of Kotlin classes and interfaces are prefixed when imported to Swift/Objective-C.
+The prefix is derived from the framework name.
+
+### Top-level functions and properties
```
olonho: Unit conversion paragraph?

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+
+Top-level Kotlin functions and properties are accessible as members of a special class.
```
olonho: Example?

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+### Category members
+
+Members of Objective-C categories and Swift extensions are imported to Kotlin
+as extensions. That's why these declarations can't be overridden in Kotlin.
+And extension initializers aren't available as Kotlin constructors.
+
+### Number
+
+`NSNumber` is not automatically translated to Kotlin primitive types
+when used as Swift/Objective-C parameter type or return value.
```
olonho: Maybe some rationale, or not yet implemented motivation?

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+
+### NSMutableString
+
+`NSMutableString` Objective-C class is not available from Kotlin.
+All instances of `NSMutableString` are copied when passed to Kotlin.
+
+### Mutable collections
+
+Every Kotlin `MutableSet` is `NSMutableSet`, however the opposite is not true.
```
olonho: Please mention explicitly that all other collections, such as Map and List are properly mapped.

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+val string = nsString as String
+val nsNumber = 42 as NSNumber
+```
+
+## Subclassing
+
+Kotlin classes and interfaces can be subclassed by Swift/Objective-C classes
+and protocols.
+Currently a class that adopts Kotlin protocol should inherit `NSObject`
+(either directly or indirectly). Note that all Kotlin classes do inherit `NSObject`.
```
olonho: What do you mean here?

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
...
+
+Swift/Objective-C classes can be subclassed with Kotlin `final` class.
```
olonho: Please mention, that non-final inheritance and complex object hierarchies are not yet supported.

OBJC_INTEROP.md
SvyatoslavScherbina: Instead of the last sentence or after it?

OBJC_INTEROP.md
SvyatoslavScherbina: Yes, it makes sense. What do you think, should it be related to function types mapping or to `Unit` type mapping?

OBJC_INTEROP.md
SvyatoslavScherbina: Actually while writing these paragraphs I was considering them to be read just as notes to the table, after following the corresponding links.
Now I see that they look very silly when being read as self-contained descriptions. So I suppose some of the paragraphs should be expanded.

OBJC_INTEROP.md
SvyatoslavScherbina: All Kotlin classes are visible to Swift/Objective-C as inheriting `NSObject`, so it is OK to adopt Kotlin protocol by a subclass of a Kotlin class.

OBJC_INTEROP.md
```diff
@@ -0,0 +1,124 @@
+# _Kotlin/Native_ interoperability with Swift/Objective-C
+
+This documents covers some details of Kotlin/Native interoperability with
+Swift/Objective-C.
```
SvyatoslavScherbina: Shouldn't we use some common name for Swift and Objective-C?

OBJC_INTEROP.md
olonho: After

OBJC_INTEROP.md
olonho: Maybe function types.

OBJC_INTEROP.md
olonho: Probably say that somehow more clear.

OBJC_INTEROP.md
olonho: Good idea, although I've no idea what could it be.

OBJC_INTEROP.md
SvyatoslavScherbina: But why this relation should be described as being symmetrical?
I mean, when importing a library to Kotlin code through `cinterop`, we don't import Kotlin code to the library.

We may use replace "attach" by "include" or "import".

OBJC_INTEROP.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Mappings
+
+The table below shows how Kotlin concepts are mapped to Swift/Objective-C and vice versa.
+
+| Kotlin | Swift | Objective-C | Notes |
+| ------ | ----- |------------ | ----- |
+| `class` | `class` | `@interface` | [note](#name-translation) |
+| `interface` | `protocol` | `@protocol` | |
+| `constructor` | Initializer | Initializer | |
```
r4zzz4k: I'm not sure if constructors map to initializers in a natural way for Kotlin developers, so there probably should be some note on `initBy(...)` and invocation ordering.

Background:
I'm playing around with Kotlin/Native based iOS app and finding initializers a bit confusing (having no background in Obj-C / Swift).
I'm trying out code-only approach, so I've removed storyboards and invoking this in `AppDelegate`:
```kotlin
val mainViewController = ViewController()
println("mainViewController ready [${mainViewController.hashCode()}]")
```
In my `ViewController` I've added logging to constructors and `initWith...` methods as following:
```kotlin
@ExportObjCClass
class ViewController: UIViewController {

    init {
        println("ViewController [${hashCode()}] ctor: common init")
    }

    constructor(): super() {
        println("ViewController [${hashCode()}] ctor: no-arg")
    }
    constructor(aDecoder: NSCoder): super(aDecoder) {
        println("ViewController [${hashCode()}] ctor: NSCoder")
    }
    constructor(nibNameOrNil: String?, bundle: NSBundle?): super(nibNameOrNil, bundle) {
        println("ViewController [${hashCode()}] ctor: nibNameOrNil")
    }

    override fun init(): UIViewController? {
        println("ViewController [${hashCode()}]: init")
        return initBy(ViewController())
    }
    override fun initWithCoder(aDecoder: NSCoder): UIViewController? {
        println("ViewController [${hashCode()}]: initWithCoder")
        return initBy(ViewController(aDecoder))
    }
    override fun initWithNibName(nibNameOrNil: String?, bundle: NSBundle?): UIViewController {
        println("ViewController [${hashCode()}]: initWithNibName")
        return initBy(ViewController(nibNameOrNil, bundle))
    }
    // ...
}
```

Here's the resulting output:
```
ViewController [1807837959]: initWithNibName
ViewController [1807837959] ctor: common init
ViewController [1807837959] ctor: nibNameOrNil
ViewController [1807837959] ctor: common init
ViewController [1807837959] ctor: no-arg
mainViewController ready [1807837959]
```

So there are several points I'm finding odd, which ask for being covered by interop documentation:
* No-arg constructor leads to `initWithNibName`. This is probably due to [`init(nibName:bundle:)`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621359-init) being [designated initializer](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID217).
* `initWithNibName` is being called before any constructor code. Looks like that's caused by the nature of initializers in Objective-C, so they are treated a bit differently from other methods. I see that `initBy` is intrinsic and [has a special handling](https://github.com/JetBrains/kotlin-native/blob/77139d58f822aaef72498570371c7448426a695d/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt#L1686) when invoked, but it's not documented in any way, so a note here might be helpful. And actual `initWith...` methods appear as usual methods in generated `.kt` stubs, so nothing reveals their real nature.
* Two different constructors being called on one object, with two calls to `init { ... }` block. I guess Objective-C base class is able to invoke another constructor of derived class, and I'm not sure if such behavior can be achieved with Kotlin/JVM, so this also is a surprise point.

Please let me know if I'm doing something wrong or if something should be filled as an issue -- I'll try to provide as much details as possible.

SvyatoslavScherbina: @r4zzz4k overriding of initializers is tricky and considered to be redesigned.
The problem with your example is that you call `super`-constructor that is based on non-designated initializer. The compiler should prohibit this, however it doesn't.

r4zzz4k: I see, so this is temporary state of thins. And so even if I want to use no-arg constructor without working with nibs it should still invoke `super(null, null)` to reference `UIViewController(nibNameOrNil: String?, bundle: NSBundle?)`. Is that correct? If so, something like this should be great to have for newcomers like me not to fail this trap:
> Note: When implementing constructors of Kotlin classes which are extending Objective-C ones make sure you are calling superclass constructor corresponding to it's [*designated initializer*](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID217) (i.e. [`initWithNibName:bundle:`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621359-initwithnibname) for `UIViewController`).

SvyatoslavScherbina: > if I want to use no-arg constructor without working with nibs it should still invoke `super(null, null)` to reference `UIViewController(nibNameOrNil: String?, bundle: NSBundle?)`. 

Yes. Swift also has this requirement.

We will either add such a warning to this doc or forbid this case in the compiler.

OBJC_INTEROP.md
olonho: Imported or used is fine with me.

OBJC_INTEROP.md
olonho: Maybe some high level description of initBy and current state of things around initializers makes sense indeed.

OBJC_INTEROP.md
```diff
@@ -0,0 +1,174 @@
...
+
+```
+func foo(block: (NSNumber) -> KotlinUnit)
+```
+
+and can be called like
+```
+foo {
+    bar($0 as! Int32)
+    return KotlinUnit()
```
olonho: Not too pretty, but we could live with it.

OBJC_INTEROP.md
```diff
@@ -0,0 +1,174 @@
...
+val nsNumber = 42 as NSNumber
+```
+
+## Subclassing
+
+Kotlin classes and interfaces can be subclassed by Swift/Objective-C classes
+and protocols.
+Currently a class that adopts Kotlin protocol should inherit `NSObject`
+(either directly or indirectly). Note that all Kotlin classes do inherit `NSObject`,
+so a subclass of Kotlin class can adopt Kotlin protocol.
```
olonho: Not entirely clear to me. Maybe "subclass of Kotlin class" -> "Objective-C subclass of Kotlin class"?

SvyatoslavScherbina: It makes sense.

OBJC_INTEROP.md
SvyatoslavScherbina: @olonho btw, I'm working on some improvements to Objective-C initializers support, which would make `initBy` (and also calling/overriding `init*` methods) obsolete. So I suggest waiting until this work is done and then completing the documentation about initializers.

SvyatoslavScherbina: And here it is: #1512

PR #1489, opened 2018-04-11
Fix debugger tests

PavelPunegov:
- Fix missed rename of WINDOWS to MINGW
- Fix matching of the debugger response


backend.native/debugger-tests/src/main/kotlin/org/jetbrains/kotlin/compiletest/DistProperties.kt
```diff
@@ -8,7 +8,7 @@ import java.nio.file.Paths
 
 object DistProperties {
     private val dist: Path = Paths.get(requireProp("konan.home"))
-    private val konancDriver = if (HostManager.host.family == Family.WINDOWS) "konanc.bat" else "konanc"
+    private val konancDriver = if (HostManager.host.family == Family.MINGW) "konanc.bat" else "konanc"
```
olonho: Maybe add HostManager.host.family.toolExtension?

PR #1492, opened 2018-04-12
[gradle][publish] main variant is per KonanBuildingConfig

vvlevchenko:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/tasks/KonanBaseTasks.kt
```diff
@@ -98,7 +98,7 @@ abstract class KonanArtifactTask: KonanTargetableTask(), KonanArtifactSpec {
 
         project.pluginManager.withPlugin("maven-publish") {
             platformConfiguration.artifacts.add(object: PublishArtifact {
-                override fun getName() = "${artifact.name}/${target.name}"
+                override fun getName() = "${artifact.name.removeSuffix("$artifactSuffix")}/${target.name}"
```
ilmat192: Should this name include an artifact prefix (e.g. `lib` for dynamic libraries)?

vvlevchenko: possibly, but not sure if it really needed, at least at this phase.

PR #1494, opened 2018-04-12
Mpp sample

StefMa:
I've created a sample implementation from the Multiplatform doc.
Don't know if you are interested in it or not. 
But I've just added it to the samples and refer it in the multiplatform doc.

Let me know if you want to use it (then I have to adjust the package names (guru.stefma to org.jetbrains) and stuff like that.
Otherwise feel free to just close it 😄


samples/multiplatform/androidApp/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
olonho: Better reuse shared gradlew in `../../`.

samples/multiplatform/androidApp/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
olonho: Same

StefMa:
@olonho I can' remove these two gradle wrapper files. But I can't remove the `androidApp/gradle/` (since it is needed for Android Studio. Otherwise it will download Gradle 4.0 which isn't working with the Android Plugin 3.0+).
Having just the gradle dir inside but not the wrappers looks very strange (and not common) IMO.
I would keep it like it is.

Beside of that and I wanted to be as close as possible with the documentation. And there stands nothing to use the wrapper from the root dir in the android project. 
If we really decided to use a architecture like that I would recommend that we update the documentation as well...?!

samples/multiplatform/androidApp/gradlew
ilmat192: Android Studio generates this wrapper while creating the project. I think it's better to not remove it because we have two separate builds in any case. The first one is the multiplatform library and is managed manually. The second one is the Android Studio build which compiles the Android app and is fully managed by Android Studio (including Gradle's version and its updates).

The Gradle guys also create a separate wrapper for each sample project in their samples (e.g. https://github.com/gradle/native-samples/tree/master/cpp/application)

StefMa:
@olonho @ilmat192 
Do you want to merge it?
Otherwise I would close it :D

ilmat192:
I think yes, adding such a sample makes sense. But I didn't check this PR on my machine yet so let me take a closer look at it first :)

@olonho What do you think about this MPP sample?

samples/multiplatform/androidApp/app/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
```
thevery: `jre` dependency is actually deprecated since 1.2.0 and you should use `jdk`:
http://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages

samples/multiplatform/androidApp/build.gradle
```diff
@@ -0,0 +1,27 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+    ext.kotlin_version = '1.2.31'
+    repositories {
+        google()
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.1.1'
```
thevery: 3.1.2 is actually latest version

ilmat192:
Hi @StefMa!

I want to bring some improvements in the sample (in particular rewrite the framework building script as shown in #1583 and fix the remarks above). What is the best way to do it? May you give me access rights to push in this branch in you repo? Or it's better to create a separate branch in our repo and move your commits in it?

StefMa:
Hi @ilmat192 

I've checked that little checkbox in this PR. So normally you have already push rights to this PR :)
bildschirmfoto 2018-05-15 um 13 17 06

See also [this article](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/) and [this one](https://help.github.com/articles/checking-out-pull-requests-locally/).

Feel free to checkout this branch out and push anything you like :) 👍

thevery:
@ilmat192 @StefMa maybe add multiplatform tests as well?

ilmat192:
@StefMa I've rebased the sample onto master and added some new changes. Please look at them and, if all is ok, I'll merge the PR.

MULTIPLATFORM.md
```diff
@@ -235,8 +235,14 @@ As well as `android`, this project contains platform-dependent implementations o
...
 
 Now we can create an Android application which will use the library we implemented on the previous step. Open Android
-Studio and create a new project in the `androidApp` directory. Android Studio generates all necessary files and
-directories so we only need to add a dependency on our library. There are 2 actions we need to do:
+Studio and create a new project in the `androidApp` directory. Android Studio will generate all necessary files and
+directories.
+
+Kotlin/Native requires Gradle 4.7 or higher so you need to make sure that the AS project uses the correct
+Gradle version. To do this, open `androidApp/gradle/gradle-wrapper.properties` and check the `distributionUrl`
+property. Fix it if it's necessary.
```
StefMa: Maybe 
> Upgrade if necessary
sounds a little bit better.
But its just a minor. Keep it if you like :)

samples/multiplatform/androidApp/app/build.gradle
```diff
@@ -21,8 +21,8 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'com.android.support:appcompat-v7:27.1.1'
-    implementation 'org.greeting:android:1.0'
+    implementation project(':greeting:android')
```
StefMa: I wanted this sample to reflect 100% of the documentation.
If we use this the documentation and the sample is not the same anymore.
Either change the docu or revert this...

## Update
Just seen that there is a big note in the documentation 🙃 
Maybe we don't "note" it anymore but used it as a general rule?!

samples/multiplatform/androidApp/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
```
StefMa: Have you upgraded via the `wrapper` task?
It's weird that the *jar haven't changed...

samples/multiplatform/greeting/build.gradle
```diff
@@ -11,7 +11,7 @@ buildscript {
     // Specify all the plugins used as dependencies
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "org.jetbrains.kotlin:kotlin-native-gradle-plugin:0.7-dev-1407"
+        classpath "org.jetbrains.kotlin:kotlin-native-gradle-plugin:${project.property('konan.plugin.version')}"
```
StefMa: Please update the documenation with somehting like `$kotlinNativeGradlePluginVersion` ...
Then it isn't hardcoded 👍

samples/multiplatform/settings.gradle
```diff
@@ -1,4 +1,9 @@
+enableFeaturePreview('GRADLE_METADATA')
```
StefMa: Is that needed for the sample?
Then we have to documented that as well

samples/multiplatform/settings.gradle
```diff
@@ -1,4 +1,9 @@
...
+
 include ':greeting'
 include ':greeting:common'
 include ':greeting:android'
 include ':greeting:ios'
+
+includeBuild '../../shared'
```
StefMa: I would recommend to add a comment here that this is not part "of the sample" and only used for **this** sample...
Know what I mean? 🙃

samples/multiplatform/androidApp/gradle/wrapper/gradle-wrapper.properties
ilmat192: Yes, just forgot about it. Thanks!

MULTIPLATFORM.md
ilmat192: Done

samples/multiplatform/settings.gradle
ilmat192: Yes, I see :)

The sample becomes a little bit complicated with all these composite builds. 
May be it's better to move it in a separate repo?

We use composite build to  allow a user to build all samples in the main repo with the compiler and the Gradle plugin built from sources right now. But in a separate repo we can use a fixed KN version and where will be no differences between the documentation and the sample itself. I can make a repo under JB organisation and move your changes into it. What do you think about it?

samples/multiplatform/.gitignore
```diff
@@ -0,0 +1,6 @@
+.iml
```
thevery: this one is not required

samples/multiplatform/androidApp/build.gradle
```diff
@@ -0,0 +1,27 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+    ext.kotlin_version = '1.2.31'
```
thevery: it `1.2.31` required? can we use latest `1.2.41`?

samples/multiplatform/greeting/android/build.gradle
```diff
@@ -0,0 +1,15 @@
+apply plugin: 'kotlin-platform-jvm'
+
+// Specify a group and a version of the library to access it in Android Studio.
+// By default the project directory name is used as an artifact name thus the full dependency
+// description will be 'org.greeting:android:1.0'
+group = 'org.greeting'
+version = 1.0
```
thevery: maven-style dependency is actually not used in this sample anymore, only project dependency

samples/multiplatform/iosApp/iosAppTests/iosAppTests.swift
```diff
@@ -0,0 +1,19 @@
...
+//  iosAppTests.swift
+//  iosAppTests
+//
+//  Created by jetbrains on 12/04/2018.
+//  Copyright © 2018 JetBrains. All rights reserved.
+//
+
+import XCTest
+import Greeting
+@testable import iosApp
```
thevery: unused

samples/multiplatform/androidApp/app/build.gradle
```diff
@@ -0,0 +1,28 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
```
thevery: since we don't have android app itself (only empty stub) is makes no sense to add extensions here

samples/multiplatform/androidApp/app/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+
+android {
+    compileSdkVersion 27
+    defaultConfig {
+        applicationId "org.konan.multiplatform"
+        minSdkVersion 24
```
thevery: imho sample should use either 15 or 21

samples/multiplatform/greeting/common/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+
+// Specify a group and a version of the library to access it in Android Studio.
+// By default the project directory name is used as an artifact name thus the full dependency
+// description will be 'org.greeting:common:1.0'
+group = 'org.greeting'
+version = 1.0
+
+dependencies {
+    // Set up compilation dependency on common Kotlin stdlib
+    compile "org.jetbrains.kotlin:kotlin-stdlib-common:$kotlin_version"
```
thevery: can we use `implementation` here?

samples/multiplatform/settings.gradle
StefMa: Indeed. The sample blows a little bit 🙃 
Yes, I think it make sense to move it in a new repository. 
The only thing we have to make sure is that the documentation and the new repo are always in sync.
We can also move the documentation into the new repository and link in *this* repo to the new repo?! 🤔 
Having the documentation here but the implementation in a new repo is a little bit strange...

Anyway.
I would welcome to have a separate repository for it.
The rest can be discussed then 👍

ilmat192: > Having the documentation here but the implementation in a new repo is a little bit strange
Having all docs in one place is more convenient though. But you're right too and I would rather move the MPP doc in the new repo. @olonho what do you think?

samples/multiplatform/.gitignore
ilmat192: Done

samples/multiplatform/androidApp/build.gradle
ilmat192: Done

samples/multiplatform/greeting/common/build.gradle
ilmat192: Done

samples/multiplatform/androidApp/app/build.gradle
ilmat192: Done

samples/multiplatform/greeting/android/build.gradle
ilmat192: We will use the maven-style dependency in a separate repo.

samples/multiplatform/iosApp/iosAppTests/iosAppTests.swift
ilmat192: Done

samples/multiplatform/androidApp/app/build.gradle
ilmat192: It may be useful if a user wants to write an application on basis of this sample.

thevery: AS will add this plugin automatically on creating new activity, but this is not a big deal anyway

ilmat192: Ok, removed this.

StefMa:
If I could I would approve that PR 👍 
Nice work. Thanks @ilmat192

samples/multiplatform/settings.gradle
ilmat192: The sample in a separate repo uses KN 0.7 and there is no need to add this line for that version.

ilmat192:
Thanks! The separate repo is here: https://github.com/JetBrains/kotlin-mpp-example. I've fixed all the remarks added to this PR there.

If there are no objections, I'll merge this PR.

CC @thevery @olonho

thevery:
Looks almost ok, but there are still some issues:
1) android build is not working - 
```
* What went wrong:
Execution failed for task ':app:transformResourcesWithMergeJavaResForDebug'.
> More than one file was found with OS independent path 'META-INF/main.kotlin_module'
```
Add 
```
    packagingOptions {
        exclude 'META-INF/main.kotlin_module'
    }

```
to fix
2) Why still `includeBuild` in https://github.com/JetBrains/kotlin-mpp-example/blob/master/androidApp/settings.gradle?

ilmat192:
@thevery Thanks, I've fixed the issue with packaging.

2) We add the Greeting build as an included one to allow resolving the Greeting library in the maven-like manner [here](https://github.com/JetBrains/kotlin-mpp-example/blob/master/androidApp/app/build.gradle#L28).

thevery:
2. ic
LGTM

PR #1496, opened 2018-04-13
Make weak reference work only on non-nullable types.

olonho:


runtime/src/main/kotlin/konan/ref/Weak.kt
```diff
@@ -21,13 +21,12 @@ package konan.ref
  * retrieve a strong reference to an object, or return null, if object was already destoyed by
  * the memory manager.
  */
-class WeakReference<T> {
+class WeakReference<T: Any> {
```
SvyatoslavScherbina: A space missing before `:` .

PR #1497, opened 2018-04-13
Objective-C names with '$' in them were blowing up

kpgalligan:
Not sure if this has test coverage anywhere. Didn't see it. Also, I'm not sure if other characters in Objc names are problematic. However, this fully solved my problem, and we're wrapping quite a lot of autogen Objc.


kpgalligan:
*were

PR #1508, opened 2018-04-17
Allow running compiler only for libraries

ilmat192:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -25,6 +25,7 @@ import org.jetbrains.kotlin.cli.common.CLICompiler
 import org.jetbrains.kotlin.cli.common.CLITool
 import org.jetbrains.kotlin.cli.common.ExitCode
 import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity.ERROR
```
olonho: `import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity.*`?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -48,15 +49,20 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
...
-        if (arguments.freeArgs.isEmpty() && !arguments.isUsefulWithoutFreeArgs) {
-            configuration.report(ERROR, "You have not specified any compilation arguments. No output has been produced.")
-        }
-
         val environment = KotlinCoreEnvironment.createForProduction(rootDisposable,
             configuration, EnvironmentConfigFiles.NATIVE_CONFIG_FILES)
         val project = environment.project
         val konanConfig = KonanConfig(project, configuration)
 
+        if (arguments.freeArgs.isEmpty() && !arguments.isUsefulWithoutFreeArgs && !konanConfig.linkOnly) {
```
olonho: Not really very readable, maybe somehow reorganize.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanDriver.kt
```diff
@@ -43,7 +43,7 @@ fun runTopLevelPhases(konanConfig: KonanConfig, environment: KotlinCoreEnvironme
         KonanPhases.list()
     }
 
-    if (config.kotlinSourceRoots.isEmpty()) return
+    if (config.kotlinSourceRoots.isEmpty() && !konanConfig.linkOnly) return
```
olonho: `if (!config.enoughArguments) return`?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
ilmat192: Done

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
ilmat192: Done

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanDriver.kt
ilmat192: Agree, but I think it's better to use another name, say `runPhases` or `infoArgsOnly`.

I mean technically `enoughArguments` should be true if something like `-list_targets` or `--list_phases` is specified. But this line checks if we need to execute the compiler phases (there are source files to compile and/or libraries to link) so it should return also if e.g. `-list_targets`  is passed (thus `enoughArguments` is true) but other args aren't.

PR #1509, opened 2018-04-17
[WIP] Improve codegen for boolean type

aceisnotmycard:
Use i8 type for field and local variables and i1 for the rest. This should match clang's behaviour.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
@@ -89,15 +89,13 @@ private fun ValueType.initCache(context: Context, cacheName: String,
             .setConstant(true)
     val staticData = context.llvm.staticData
     val values = (start..end).map { staticData.createInitializer(kotlinType, createConstant(it)) }
-    // TODO: use i8 for boolean
-    val llvmBoxType = structType(context.llvm.runtime.objHeaderType, this.llvmType)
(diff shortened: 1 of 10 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
```diff
@@ -89,13 +89,13 @@ private fun ValueType.initCache(context: Context, cacheName: String,
             .setConstant(true)
     val staticData = context.llvm.staticData
     val values = (start..end).map { staticData.createInitializer(kotlinType, createConstant(it)) }
-    val llvmBoxType = structType(context.llvm.runtime.objHeaderType, this.llvmType)
+    val llvmBoxType = structType(context.llvm.runtime.objHeaderType, getLLVMType(this, true))
```
olonho: Maybe instead of parameter just different getters? `this.llvmType` and `this.llvmStorageType`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -248,14 +248,16 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
 
 
     fun ret(value: LLVMValueRef?): LLVMValueRef {
+        val retval = cast(returnType, value)
```
olonho: Do not think it's a good idea.

PR #1516, opened 2018-04-18
Vtable fix

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/ClassVtablesBuilder.kt
```diff
@@ -82,26 +87,57 @@ internal class OverriddenFunctionDescriptor(
...
         val superVtableEntries = if (classDescriptor.isSpecialClassWithNoSupertypes()) {
             emptyList()
         } else {
             val superClass = classDescriptor.getSuperClassNotAny() ?: context.ir.symbols.any.owner
             context.getVtableBuilder(superClass).vtableEntries
         }
 
-        val methods = classDescriptor.sortedOverridableOrOverridingMethods
+        // TODO: extract method .isBridge()
+        val methods = classDescriptor.sortedOverridableOrOverridingMethods.filterNot { it.name.asString().contains("<bridge-") }
```
olonho: Hmm, not very reliable approach.

homuroll: Yeah, and because of this I placed this "TODO", also I has a fix for that in another branch about devirtualized bridges inlining.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/DataFlowIR.kt
```diff
@@ -71,8 +71,8 @@ internal object DataFlowIR {
 
         abstract class Declared(val isFinal: Boolean, val isAbstract: Boolean, val module: Module?, val symbolTableIndex: Int) : Type() {
             val superTypes = mutableListOf<Type>()
-            val vtable = mutableListOf<FunctionSymbol>()
-            val itable = mutableMapOf<Long, FunctionSymbol>()
+            val vtable = mutableListOf<FunctionSymbol?>()
```
SvyatoslavScherbina: Can `null` really get into this vtable and itable? Making these element types nullable is very inconvenient because it has observable impact on other code.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/ClassVtablesBuilder.kt
SvyatoslavScherbina: It makes some sense to apply the same fix when building itable below.

PR #1524, opened 2018-04-22
Make some test tasks depending on dist

ilmat192:


build.gradle
```diff
@@ -423,14 +423,9 @@ task performance {
...
 
 task clean {
+    dependsOn subprojects.collect { it.tasks.matching { it.name == "clean" } }
     doLast {
         delete distDir
         delete bundle.outputs.files
     }
 }
-
-// Don't run with other tasks (e.g. ./gradlew cleanAll bundle)
```
olonho: Have you updated all TC configs?

vvlevchenko: yes

PR #1525, opened 2018-04-22
Allow arch and family susbtitutions in manifest files.

alexander-gorshenev:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/TargetDefFile.kt
```diff
@@ -3,10 +3,4 @@ package org.jetbrains.kotlin.konan.util
...
 
-fun defaultTargetSubstitutions(target: KonanTarget) =
-        mapOf<String, String>(
-            "target" to target.visibleName,
-            "arch" to target.architecture.visibleName,
-            "family" to target.family.visibleName)
-
-fun DefFile(file: File?, target: KonanTarget) = DefFile(file, defaultTargetSubstitutions(target))
\ No newline at end of file
+fun DefFile(file: File?, target: KonanTarget) = DefFile(file, defaultTargetSubstitutions(target))
```
olonho: Extension constructor :)?

PR #1531, opened 2018-04-23
Support weak references to Objective-C objects

SvyatoslavScherbina:


runtime/src/main/cpp/ObjCInteropUtils.mm
```diff
@@ -165,6 +165,48 @@ KBoolean Kotlin_Interop_IsObjectKindOfClass(id obj, void* cls) {
   return [((id<NSObject>)obj) isKindOfClass:(Class)cls];
 }
 
+// Used as associated object for ObjCWeakReferenceImpl.
```
olonho: `Used as an  associated object`?

PR #1537, opened 2018-04-24
[gradle][publish] attempt to fix sample dependencies

vvlevchenko:


platformLibs/build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-native-gradle-plugin"
```
olonho: What about samples?

ilmat192: They use the version specified in `samples/gradle.properties`. It is`+` for building samples in a work tree and is replaced with correct plugin version when a bundle is built.

PR #1538, opened 2018-04-25
Add quick fix for unDCEd private function and classes tables

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -92,18 +92,29 @@ internal fun emitLLVM(context: Context) {
...
-                privateFunctions.map { constPointer(codegenVisitor.codegen.llvmFunction(it)).bitcast(int8TypePtr) },
-                isExported = true
-        )
+        privateFunctions.forEachIndexed { index, it ->
+            val function = codegenVisitor.codegen.llvmFunction(it)
+            LLVMAddAlias(
+                    context.llvmModule,
+                    function.type,
+                    function,
+                    irModule.descriptor.privateFunctionsTableSymbolName + "_$index"
```
olonho: Maybe `"${irModule.descriptor.privateFunctionsTableSymbolName}_$index"`?

SvyatoslavScherbina: Maybe. `privateFunctionsTableSymbolName` and `privateClassesTableSymbolName` must indeed be refactored, but the purpose of this PR is to fix regression as fast as possible.

SvyatoslavScherbina: Done.

PR #1541, opened 2018-04-26
Fix simulator tests

PavelPunegov:
- Make iPhone 6 to be default target to run during testing
- Move out codesign to utils to be used not only for framework tests

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -16,6 +16,7 @@
 
 import groovy.json.JsonOutput
 import org.jetbrains.kotlin.*
+import org.jetbrains.kotlin.konan.target.KonanTarget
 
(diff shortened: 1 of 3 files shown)


backend.native/tests/build.gradle
```diff
@@ -2612,6 +2612,9 @@ if (isMac()) {
                 args "-lobjc", '-fobjc-arc'
                 args '-fPIC', '-shared', '-o', "$buildDir/libobjcsmoke.dylib"
             }
+            if (project.testTarget == 'ios_x64') {
```
ilmat192: What if `iphone_sim` is specified instead of `ios_x64`? May be it's better to create a `KonanTarget` instance from this property and check it instead of just string?

PR #1551, opened 2018-05-02
samples/videoplayer: migrate from libavresample to libswresample

msink:
libavresample is deprecated, and even already removed in some distros.

### DIFF
--- a/samples/videoplayer/README.md
+++ b/samples/videoplayer/README.md
@@ -7,7 +7,7 @@ ffmpeg and SDL2 is needed for that to work, i.e.
      port install ffmpeg-devel
      brew install ffmpeg sdl2
      apt install libavcodec-dev libavformat-dev libavutil-dev libswscale-dev \
-                 libavresample-dev
+                 libswresample-dev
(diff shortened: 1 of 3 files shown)


samples/videoplayer/README.md
```diff
@@ -7,7 +7,7 @@ ffmpeg and SDL2 is needed for that to work, i.e.
      port install ffmpeg-devel
      brew install ffmpeg sdl2
      apt install libavcodec-dev libavformat-dev libavutil-dev libswscale-dev \
-                 libavresample-dev
+                 libswresample-dev
```
olonho: Thanks for this change, however it would be awesome, if you could detect in Gradle which ffmpeg is installed and support both old and new one with klib, for example by adding conditionally compiled wrapper functions into the .def file.

msink: Is it really needed? libswresample is copyrighted "2011-2014", so it's here at least five years.

olonho: OK, let's keep it as is.

PR #1561, opened 2018-05-04
Update KonanSharedVariablesManager to new SharedVariablesManager API

max-kammerer:
Merge commit on Kotlin compiler dependency update

PR #1567, opened 2018-05-07
 [kotlin compiler][update] 1.2.50-dev-1148

vvlevchenko:


olonho:
Is https://github.com/JetBrains/kotlin-native/pull/1561 included here? Also builds seem to fail.

PR #1570, opened 2018-05-08
Improve package name escaping in interop stubs

SvyatoslavScherbina:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
```diff
@@ -1026,6 +1026,6 @@ class StubGenerator(
             MappingBridgeGeneratorImpl(declarationMapper, simpleBridgeGenerator)
 
     companion object {
-        private val VALID_PACKAGE_NAME_REGEX = "[a-zA-Z1-9_.]".toRegex()
+        private val VALID_PACKAGE_NAME_REGEX = "[a-zA-Z1-9_.]+".toRegex()
```
olonho: What about `0`?

PR #1579, opened 2018-05-11
[gradle-plugin] Use dev plugin

StefMa:
I've just replaced the java plugin with the [gradle plugin development plugin](https://docs.gradle.org/current/userguide/java_gradle_plugin.html).

I've also replaced the konan.properties with the `gradlePlugins` block.
It seems the `default-konan-version` inside was a relict from old days and isn't used anymore.


olonho:
Thanks, we're evaluating this fix.

PR #1587, opened 2018-05-14
Fixed bug in Inliner + test

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/SuspendFunctionsLowering.kt
```diff
@@ -218,16 +220,15 @@ internal class SuspendFunctionsLowering(val context: Context): DeclarationContai
...
-            }
-        })
+    private fun removeReturnIfSuspendedCallAndSimplifyDelegatingCall(irFunction: IrFunction, delegatingCall: IrCall) {
+        val returnValue =
+                if (delegatingCall.descriptor.original == returnIfSuspendedDescriptor)
+                    delegatingCall.getValueArgument(0)!!
+                else delegatingCall
+        context.createIrBuilder(irFunction.symbol).run {
+            val statements = (irFunction.body as IrBlockBody).statements
+            statements[statements.size - 1] = irReturn(returnValue)
```
olonho: I would check what value is at `statements[statements.size - 1] ` now.

PR #1588, opened 2018-05-14
Support Xcode 9.3.1, allow version check override

olonho:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Apple.kt
```diff
@@ -56,7 +56,8 @@ class AppleConfigurablesImpl(
         properties.getProperty("useFixedXcodeVersion")?.let { requiredXcodeVersion ->
             val currentXcodeVersion = xcode.version
 
-            if (currentXcodeVersion != requiredXcodeVersion) {
+            if (properties.getProperty("ignoreXcodeVersionCheck") != "true" &&
+                    currentXcodeVersion != requiredXcodeVersion) {
```
SvyatoslavScherbina: Does it make some sense to add hint about `ignoreXcodeVersionCheck` to the error message below?

olonho: Done

PR #1591, opened 2018-05-14
Optimize important case of string appending to the builder.

olonho:


runtime/src/main/cpp/KString.cpp
```diff
@@ -840,6 +840,30 @@ OBJ_GETTER(Kotlin_String_plusImpl, KString thiz, KString other) {
   RETURN_OBJ(result->obj());
 }
 
+KInt Kotlin_StringBuilder_insertString(KRef builder, KInt position, KString fromString) {
+  auto toArray = builder->array();
+  RuntimeAssert(toArray->count_ >= fromString->count_ + position, "must be true");
+  memcpy(CharArrayAddressOfElementAt(toArray, position),
+         CharArrayAddressOfElementAt(fromString, 0),
+         fromString->count_ * sizeof(KChar));
+   return fromString->count_;
```
PavelPunegov: 3 spaces in indent

PR #1602, opened 2018-05-17
Improve tooling model + add checking Gradle version.

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanOutput.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import org.jetbrains.kotlin.konan.target.CompilerOutputKind
+
+enum class KonanOutput(val cliOption: String, val kind: CompilerOutputKind) {
+    PROGRAM("program", CompilerOutputKind.PROGRAM),
+    DYNAMIC("dynamic", CompilerOutputKind.DYNAMIC),
+    FRAMEWORK("framework", CompilerOutputKind.FRAMEWORK),
+    LIBRARY("library", CompilerOutputKind.LIBRARY),
+    BITCODE("bitcode", CompilerOutputKind.BITCODE)
+}
```
alexander-gorshenev: you think the toLowerCase() invariant should not be trusted, or something?

olonho: Yeah, unclear why we need another enum with same semantics.

ilmat192: Done.

PR #1607, opened 2018-05-18
[build][gradle] switch to gradle-4.7

vvlevchenko:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/tasks/KonanBaseTasks.kt
```diff
@@ -130,6 +143,20 @@ abstract class KonanArtifactTask: KonanTargetableTask(), KonanArtifactSpec {
     fun destinationDir(dir: Any) {
         destinationDir = project.file(dir)
     }
+
+    private fun KonanTarget.asOperatingSystemFamily(): OperatingSystemFamily {
```
olonho: Это гредловый класс, OperatingSystemFamily?

vvlevchenko: да

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/tasks/KonanBaseTasks.kt
```diff
@@ -130,6 +143,20 @@ abstract class KonanArtifactTask: KonanTargetableTask(), KonanArtifactSpec {
...
+        val name = when(this.family) {
+            Family.LINUX -> OperatingSystemFamily.LINUX
+            Family.OSX -> OperatingSystemFamily.MAC_OS
+            Family.MINGW -> OperatingSystemFamily.WINDOWS
+            Family.ANDROID -> "ANDROID"
```
ilmat192: Why not just `else -> family.name`?

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/tasks/KonanBaseTasks.kt
```diff
@@ -113,7 +116,17 @@ abstract class KonanArtifactTask: KonanTargetableTask(), KonanArtifactSpec {
...
             val linkUsage = objectFactory.named(Usage::class.java, Usage.NATIVE_LINK)
             val konanSoftwareComponent = config.mainVariant
-            konanSoftwareComponent.addVariant(NativeVariant(Names.of("${artifact.name.removeSuffix("$artifactSuffix")}_${target.name}"), linkUsage, null, linkUsage, platformConfiguration))
+            val variantName = "${artifact.name.removeSuffix("$artifactSuffix")}_${target.name}"
+            konanSoftwareComponent.addVariant(NativeVariantIdentity(
+                    variantName,
+                    project.provider{artifactName},
+                    project.provider{project.group.toString()},
+                    project.provider{ project.version.toString() },
+                    false,
```
ilmat192: We can determine if the artifact is debuggable/optimized for an artifact produced by the compiler (the `konanCompileTask` and its child tasks). I think we need to take this into account while creating a `NativeVariantIdentity`.

PR #1623, opened 2018-05-23
Add more features to JS <-> WASM logic

FloFri:
Adds the missing logic to use an idlInterfaceRef type as an method-argument in the WebAssembly stub generator. For example needed to add the context.drawImage method in the dom-idl.


alexander-gorshenev:
While the changes are okay, we don't really plan to extend the manually written bindings. rather we are working on `jsinterop` tool to create the bindings automatically from the webidl definitions.

FloFri:
Thanks for the answer. So there is already a new jsinterop-tool, I did not know about or is the tool still in development?

alexander-gorshenev:
It is still in the development.

FloFri:
Thanks, then I will wait :)

PR #1637, opened 2018-05-30
Move to XCode 9.4, add CoreML and ClassKit.

olonho:


platformLibs/src/platform/ios/CoreML.def
```diff
@@ -0,0 +1,9 @@
+depends = CoreFoundation Foundation darwin posix
```
SvyatoslavScherbina: + `CoreVideo`

platformLibs/src/platform/osx/CoreML.def
```diff
@@ -0,0 +1,9 @@
+depends = CoreFoundation Foundation darwin posix
```
SvyatoslavScherbina: \+ `CoreVideo`

PR #1638, opened 2018-05-30
Fix for enum equals method

aceisnotmycard:


backend.native/tests/runtime/basic/enum_equals.kt
```diff
@@ -0,0 +1,17 @@
+package runtime.basic.enum_equals
+
+import kotlin.test.*
+
+enum class EnumA {
```
olonho: Would be good to test that enums passed to a worker keep identity. Currently this is the case, as enum instance is shared between workers.

olonho: Like this one
```enum class A {
    A,
    B
}
data class Foo(val kind: A)

@Test fun runTest() {
    main(emptyArray())
}

    val result = startWorker().schedule(TransferMode.CHECKED,
            { Foo(A.B) },
            { input -> input.kind == A.B }).result()
    println(result)
}```

aceisnotmycard: Done.

PR #1640, opened 2018-05-31
Fix KT-24370. Make indexOf be consistent with Kotlin/JVM

PavelPunegov:


runtime/src/main/cpp/KString.cpp
```diff
@@ -1087,8 +1090,13 @@ KInt Kotlin_String_lastIndexOfChar(KString thiz, KChar ch, KInt fromIndex) {
 
 // TODO: or code up Knuth-Moris-Pratt.
 KInt Kotlin_String_indexOfString(KString thiz, KString other, KInt fromIndex) {
-  if (fromIndex < 0 || fromIndex > thiz->count_ ||
-      other->count_ > thiz->count_ - fromIndex) {
+  if (fromIndex < 0) {
+    fromIndex = 0;
+  }
+  if (fromIndex >= thiz->count_) {
```
olonho:
```
There is no restriction on the value of fromIndex. If it is negative, it has the same effect as if it were zero: this entire string may be searched. If it is greater than the length of this string, it has the same effect as if it were equal to the length of this string: -1 is returned.
```
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#indexOf(int,%20int)
I read that as -1 shall be always returned, no matter what other is.

PavelPunegov: "If it is negative, it has the same effect as if it were zero: this entire string may be searched."
This means that the substring should be searched anyway, while our implementation returned -1 meaning that there are no such substring.

PR #1645, opened 2018-06-03
Make String and box types frozen by default, more freeze checks

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1397,27 +1398,33 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
             assert (value.receiver == null)
             val ptr = context.llvmDeclarations.forStaticField(value.symbol.owner).storage
             return functionGenerationContext.loadSlot(ptr, value.descriptor.isVar())
         }
     }
 
     //-------------------------------------------------------------------------//
+    private fun needMutationCheck(expression: IrExpression): Boolean {
+        // For now we omit mutation checks on immutable types, as this allows initialization in constructor
+        // and it is assumed API doesn't allow to change them.
```
SvyatoslavScherbina: > it is

its?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1397,27 +1398,33 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
 
     private fun evaluateSetField(value: IrSetField): LLVMValueRef {
         context.log{"evaluateSetField               : ${ir2string(value)}"}
         val valueToAssign = evaluateExpression(value.value)
         if (value.descriptor.dispatchReceiverParameter != null) {
             val thisPtr = evaluateExpression(value.receiver!!)
-            functionGenerationContext.call(context.llvm.mutationCheck,
-                    listOf(functionGenerationContext.bitcast(codegen.kObjHeaderPtr, thisPtr)),
-                    Lifetime.IRRELEVANT, ExceptionHandler.Caller)
+            if (needMutationCheck(value.receiver!!)) {
```
SvyatoslavScherbina: The check based on receiver type doesn't look reliable to me. We can get `ClassDescriptor` from field descriptor instead, e.g.`value.descriptor.containingDeclaration`.

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -379,6 +404,7 @@ KFloat Kotlin_FloatArray_get(KConstRef thiz, KInt index) {
   if (static_cast<uint32_t>(index) >= array->count_) {
     ThrowArrayIndexOutOfBoundsException();
   }
+  mutabilityCheck(thiz);
```
SvyatoslavScherbina: Mutability check in array `get`?

runtime/src/main/kotlin/konan/Annotations.kt
```diff
@@ -61,6 +61,13 @@ public annotation class Used
 @Retention(AnnotationRetention.SOURCE)
 public annotation class Throws(vararg val exceptionClasses: KClass<out Throwable>)
 
+/**
+ * Class is immutable and is frozen by default.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.BINARY)
+annotation class Immutable
```
SvyatoslavScherbina: This annotation is unchecked and thus unsafe. Shouldn't it be internal yet?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: It does not have `findAnnotation`, it seems.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1397,27 +1398,33 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-        else {
+        } else {
             assert (value.receiver == null)
             val ptr = context.llvmDeclarations.forStaticField(value.symbol.owner).storage
             return functionGenerationContext.loadSlot(ptr, value.descriptor.isVar())
         }
     }
 
     //-------------------------------------------------------------------------//
+    private fun needMutationCheck(expression: IrExpression): Boolean {
```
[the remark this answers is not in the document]
olonho: Well, with object fields it got somewhat complicated, so let's keep this change simple.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: Hmm, not really.

runtime/src/main/cpp/Arrays.cpp
olonho: Mutability shall only be tested in setter, isn't it?

runtime/src/main/kotlin/konan/Annotations.kt
olonho: OK

runtime/src/main/cpp/Arrays.cpp
SvyatoslavScherbina: Exactly, and this function is named `Kotlin_FloatArray_get`. Is it ok?

backend.native/tests/runtime/workers/freeze2.kt
```diff
@@ -0,0 +1,155 @@
...
+
+import kotlin.test.*
+
+import konan.worker.*
+
+data class Data(var int: Int)
+
+/* An awful hack: we mark data as immutable, while it is not. */
+@konan.internal.Immutable
+data class SharedData(var field: Any)
```
SvyatoslavScherbina: What is the motivation behind adding test for an awful hack?

backend.native/tests/runtime/workers/freeze2.kt
```diff
@@ -0,0 +1,155 @@
...
+    ok = false
+    val a1 = ByteArray(2)
+    a1[1]++
+    a1.freeze()
+    try {
+        a1[1]++
+    } catch (e: InvalidMutabilityException) {
+        ok = true
+    }
+    if (!ok) throw Error("can mutate frozen ByteArray")
```
SvyatoslavScherbina: We have some `kotlin.test.*` methods to be used in this case, e.g. [`assertFails`](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails.html) and [`assertFailsWith`](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html).

backend.native/tests/runtime/workers/freeze2.kt
```diff
@@ -0,0 +1,155 @@
...
+    // Ensure that String and integral boxes are frozen by default.
+    val worker = startWorker()
+    val data = SharedData("Hello")
+    worker.schedule(TransferMode.CHECKED, { data } ) {
+        input -> println("Worker 1: $input")
+    }.result()
+
+    data.field = 42
+    worker.schedule(TransferMode.CHECKED, { data } ) {
+        input -> println("Worker2: $input")
```
SvyatoslavScherbina: How does this check that `Int` box is frozen by default?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
SvyatoslavScherbina: > and it is assumed API

Then "that" is missing between "assumed" and "API".

backend.native/tests/runtime/workers/freeze2.kt
olonho: Done

runtime/src/main/cpp/Arrays.cpp
olonho: OK

PR #1664, opened 2018-06-06
Improve code quality by explicit static container reference.

olonho:


olonho:
Average Ring score: absolute = 144.6232 us +- 1.1969 us, ratio = 8.40 +- 0.12

runtime/src/main/cpp/Assert.h
```diff
@@ -19,14 +19,20 @@
 
 #include "Common.h"
 
-RUNTIME_NORETURN void RuntimeAssertFailed(const char* location, const char* message);
+// To avoid cluttering optimized code with asserts, they could be turned off.
+#define KONAN_ENABLE_ASSERT 1
```
SvyatoslavScherbina: In some cases `RuntimeAssert` is used not to check an internal invariant, but to validate results of system APIs. So we seem to need another kind of assertions that aren't disabled with this flag.

olonho: Like where?

SvyatoslavScherbina: Like everywhere :)
For example, Exceptions.cpp:147
It would be one of the very first results of grepping our runtime for `RuntimeAssert`.

PR #1666, opened 2018-06-07
Rework Gradle plugin

ilmat192:


tools/kotlin-native-gradle-plugin/build.gradle
```diff
@@ -166,5 +166,9 @@ gradlePlugin {
             id = 'konan'
             implementationClass = 'org.jetbrains.kotlin.gradle.plugin.KonanPlugin'
         }
+        create("kotlin-native") {
+            id = 'kotlin-native'
```
StefMa: Gradle recommends to add a "package like" name.
Like you do with other plugins (https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm or https://plugins.gradle.org/plugin/org.jetbrains.dokka).
Should we use `org.jetbrains.kotlin-native` here? 🤔

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
```diff
@@ -100,5 +105,8 @@ internal class KonanInteropRunner(project: Project)
     }
 }
 
-internal class KonanCompilerRunner(project: Project) : KonanCliRunner("konanc", "Kotlin/Native compiler", project)
-internal class KonanKlibRunner(project: Project) : KonanCliRunner("klib", "Klib management tool", project)
+internal class KonanCompilerRunner(project: Project, additionalJvmArgs: List<String> = emptyList())
+    : KonanCliRunner("konanc", "Kotlin/Native compiler", project, additionalJvmArgs)
+
+internal class KonanKlibRunner(project: Project, additionalJvmArgs: List<String> = emptyList())
+    : KonanCliRunner("klib", "Klib management tool", project, additionalJvmArgs)
```
StefMa: Maybe named params look nicer here, what do you think? (same for the `KonanCompilerRunner` of course)

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeComponentImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.jetbrains.kotlin.konan.target.HostManager
+import org.jetbrains.kotlin.konan.target.KonanTarget
+import javax.inject.Inject
+
+open class KotlinNativeComponentImpl @Inject constructor(
+        private val name: String,
+        override val sources: KotlinNativeSourceSetImpl,
+        private val objectFactory: ObjectFactory,
+        fileOperations: FileOperations
+) : DefaultNativeComponent(fileOperations), KotlinNativeComponent, ComponentWithNames, PublicationAwareComponent {
```
StefMa: According to the [Kotlin style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting) interfaces should be located in a new line

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeExecutableImpl.kt
```diff
@@ -0,0 +1,67 @@
...
+        sources,
+        identity,
+        projectLayout,
+        CompilerOutputKind.PROGRAM,
+        objects,
+        componentImplementation,
+        configurations,
+        fileOperations),
+        KotlinNativeExecutable,
+    SoftwareComponentInternal
```
StefMa: Wow that looks super akward..
I totally oversee when the constructor was closed...
Maybe use named parameters here as well?
And indend the two `interfaces`  the same way

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeExecutableImpl.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.gradle.api.provider.Property
+import org.gradle.api.provider.Provider
+import org.gradle.language.cpp.internal.DefaultUsageContext
+import org.gradle.nativeplatform.Linkage
+import org.jetbrains.kotlin.gradle.plugin.experimental.KotlinNativeExecutable
+import org.jetbrains.kotlin.gradle.plugin.experimental.sourcesets.KotlinNativeSourceSet
+import org.jetbrains.kotlin.konan.target.CompilerOutputKind
+import javax.inject.Inject
+
+// TODO: SoftwareComponentInternal will be replaced with ComponentWithVariants by Gradle
```
StefMa: Maybe add this TODO below? Where the interface is used.

Do you know "when" or in which version will Gradle provide the `ComponentWithVariants`? Maybe we cann add this to the todo...

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeKLibraryImpl.kt
```diff
@@ -0,0 +1,65 @@
...
+        sources,
+        identity,
+        projectLayout,
+        CompilerOutputKind.LIBRARY,
+        objects,
+        componentImplementation,
+        configurations,
+        fileOperations),
+        KotlinNativeKLibrary,
+    SoftwareComponentInternal
```
StefMa: Same header issue like above

StefMa: Consider to add the TODO here as well..
Otherwise it may be unseen if someone refactore the other class...

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeVariantIdentity.kt
```diff
@@ -0,0 +1,31 @@
+package org.jetbrains.kotlin.gradle.plugin.experimental.internal
+
+import org.gradle.api.internal.component.UsageContext
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Provider
+import org.gradle.language.cpp.internal.NativeVariantIdentity
+import org.jetbrains.kotlin.konan.target.KonanTarget
+
+open class KotlinNativeVariantIdentity(name: String,
```
StefMa: According to the kotlin style guide (see link above) each property should be in a new line

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/plugins/KotlinNativeBasePlugin.kt
```diff
@@ -0,0 +1,75 @@
+package org.jetbrains.kotlin.gradle.plugin.experimental.plugins
+
+import org.gradle.api.Plugin
+import org.gradle.api.component.SoftwareComponentContainer
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.internal.FeaturePreviews
+import org.gradle.api.internal.project.ProjectInternal
+import org.gradle.api.internal.tasks.TaskContainerInternal
```
StefMa: ❌ You are using internal Apis here.
Is that expected?

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/plugins/KotlinNativePlugin.kt
```diff
@@ -0,0 +1,137 @@
+package org.jetbrains.kotlin.gradle.plugin.experimental.plugins
+
+import org.gradle.api.Plugin
+import org.gradle.api.attributes.Usage
+import org.gradle.api.internal.FactoryNamedDomainObjectContainer
+import org.gradle.api.internal.attributes.ImmutableAttributesFactory
+import org.gradle.api.internal.project.ProjectInternal
```
StefMa: Same here with internal APIs

tools/kotlin-native-gradle-plugin/src/test/kotlin/ExperimentalPluginTests.kt
```diff
@@ -0,0 +1,393 @@
...
+import org.junit.rules.TemporaryFolder
+import java.io.File
+import java.nio.file.Paths
+import java.util.concurrent.TimeUnit
+import kotlin.test.*
+
+class ExperimentalPluginTests {
+
+    val tmpFolder = TemporaryFolder()
+        @Rule get
```
StefMa: Hides us something to use junit5? 🤔

ilmat192:
@StefMa Thank you for the review! I'll address it in a separate commit :)

tools/kotlin-native-gradle-plugin/build.gradle
ilmat192: "Big" Kotlin provides both namings for its plugins ([repo](https://github.com/JetBrains/kotlin/tree/master/libraries/tools/kotlin-gradle-plugin/src/main/resources/META-INF/gradle-plugins)) but only plugins with package-like names are published to the Gradle plugin portal. I think we can use the same approach here + add the portal publication for our plugin.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
ilmat192: I'm not sure. Kotlin doesn't allow using names only for part of arguments so I cannot write something like
```
(toolName = "klib", fullName ="Klib management tool", project, additionalJvmArgs)
```
All the arguments must be passed with names so we get something like this:
```
( ... project = project, additionalJvmArgs = additionalJvmArgs)
```

Also IntelliJ IDEA can show hints for parameter names. It does so for this code too and the first two args look better when the file is opened in the IDE.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeComponentImpl.kt
ilmat192: Done

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeExecutableImpl.kt
ilmat192: > Do you know "when" or in which version will Gradle provide the ComponentWithVariants?

No, there is just a comment in Gradle's javadoc that the SoftwareComponentInternal will be replaced. I've moved the TODO below.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/KotlinNativeKLibraryImpl.kt
ilmat192: Fixed.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/plugins/KotlinNativeBasePlugin.kt
ilmat192: Unfortunately yes. I try to keep the plugin close to the new Gradle model for native projects (e.g. see [this blog post](https://blog.gradle.org/introducing-the-new-cpp-plugins)) and maximally reuse a code provided by it. This native support in Gradle is at a preview stage so there is no public wrappers for many of these APIs.

tools/kotlin-native-gradle-plugin/src/test/kotlin/ExperimentalPluginTests.kt
ilmat192: We can switch to JUnit5 and rework this part of tests. But I think it's better to do it in a separate PR.

tools/kotlin-native-gradle-plugin/build.gradle
StefMa: Don't know if we just follow the "big brother" instead of doing it "right"/"correct" from the beginning?!

PR #1667, opened 2018-06-07
[lldb][python] do not let to crash of python bindings on variables before local_inits basic block

vvlevchenko:


llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -47,6 +47,13 @@ def is_string(value):
 def is_array(value):
     return int(evaluate("(int)Konan_DebugIsArray({})".format(lldb_val_to_ptr(value))).GetValue()) == 1
 
+
+def check_type_info(addr):
+    """ This method checks self referencing of pointer of first member of TypeInfo including case when ObjHeader represents
```
olonho: This method checks self-referencing of pointer of first member of TypeInfo including case when object has an meta-object pointed by TypeInfo.

PR #1670, opened 2018-06-08
Use libm also for wasm target, because it is much faster

FloFri:
Use the libm math functions also for wasm, because that way, llvm can use the native instructions for them (like f64.sqrt or f64.floor). Using JS functions is about 3-4 times slower.

That way the following naive prim search function is now about 5% faster in WASM than in pure JS (even with all the WASM initializing). Without the change it is about 300+% slower than pure JS.

```
import kotlin.math.*

    val last = 10000000;

    var foundNum = 0;
    for (i in 3..last step 2) {
        val lastDiv = floor(sqrt(i.toDouble())).toInt();

        var found = false;
        for (j in 3..lastDiv step 2) {
            if (i % j == 0) {
                found = true;
                break;
            }
        }
        if (!found) {
            ++foundNum;
        }
    }
    println(foundNum);
}
```


SvyatoslavScherbina:
Kotlin/Native WASM target is not actively developed and maintained now.
Instead there is development of new separate WASM backend. cc @skuzmich.

PR #1677, opened 2018-06-09
Kotlin machine scheduling primitives.

olonho:


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -91,16 +112,54 @@ void AppendToInitializersTail(InitNode *next) {
...
+}
+
+RuntimeState* Kotlin_suspendRuntime() {
+    RuntimeCheck(::runtimeState != nullptr, "Runtime must be active on the current thread");
+    auto result = ::runtimeState;
+    RuntimeCheck(updateStatusIf(result, RUNNING, SUSPENDED), "Cannot transition state to SUSPENDED for suspend");
+    result->memoryState = SuspendMemory();
+    ::runtimeState = nullptr;
+#if !KONAN_NO_THREADS
+     __sync_synchronize();
```
SvyatoslavScherbina: Is it really required?

olonho: Maybe not.

PR #1688, opened 2018-06-13
[gradle][plugin] this patch adds pom entry to konan building config to re-evaluate it on each native publications

vvlevchenko:


samples/libcurl/build.gradle
```diff
@@ -13,6 +13,13 @@ konanArtifacts {
         target('macbook') {
             includeDirs.headerFilterOnly '/opt/local/include', '/usr/local/include'
         }
+        pom{
```
olonho: Could this one be written in Kotlin, please?

PR #1696, opened 2018-06-15
Fixed typo in Android calculator app layout

zsmb13:


samples/calculator/android/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,4 @@
 <resources>
     <string name="app_name">Konan Calculator</string>
-    <string name="enter_mathematic_exression">Enter mathematic exression:</string>
+    <string name="enter_mathematic_exression">Enter mathematic expression:</string>
```
olonho: Guess it makes sense to fix in resource name as well.

PR #1703, opened 2018-06-19
Tweaks to atomics implementation.

olonho:


runtime/src/main/cpp/Atomic.cpp
```diff
@@ -62,9 +82,38 @@ void Kotlin_AtomicReference_checkIfFrozen(KRef value) {
     }
 }
 
-KRef Kotlin_AtomicReference_compareAndSwap(KRef thiz, KRef expectedValue, KRef newValue) {
+OBJ_GETTER(Kotlin_AtomicReference_compareAndSwap, KRef thiz, KRef expectedValue, KRef newValue) {
     Kotlin_AtomicReference_checkIfFrozen(newValue);
-    return compareAndSwapImpl(thiz, expectedValue, newValue);
+    if (expectedValue == newValue) {
+        RETURN_OBJ(expectedValue);
```
SvyatoslavScherbina: Is it correct? What if current value is not `expectedValue`?

runtime/src/main/cpp/Atomic.cpp
```diff
@@ -30,6 +36,20 @@ template <typename T> T compareAndSwapImpl(KRef thiz, T expectedValue, T newValu
     return compareAndSwap(location, expectedValue, newValue);
 }
 
+inline AtomicReferenceLayout* asAtomicReference(KRef thiz) {
+    return reinterpret_cast<AtomicReferenceLayout*>(thiz + 1);
+}
+
+inline void lock(KRef thiz) {
```
SvyatoslavScherbina: It is not clear from the signature of this function that it is applicable only to AtomicReference.
It makes sense to change it name or parameter type (or even moving to `AtomicReferenceLayout` class).

runtime/src/main/cpp/Atomic.cpp
```diff
@@ -62,9 +82,38 @@ void Kotlin_AtomicReference_checkIfFrozen(KRef value) {
...
+OBJ_GETTER(Kotlin_AtomicReference_compareAndSwap, KRef thiz, KRef expectedValue, KRef newValue) {
     Kotlin_AtomicReference_checkIfFrozen(newValue);
-    return compareAndSwapImpl(thiz, expectedValue, newValue);
+    if (expectedValue == newValue) {
+        RETURN_OBJ(expectedValue);
+    }
+    // See Kotlin_AtomicReference_get() for explanations, why locking is needed.
+    lock(thiz);
+    // As we lock [thiz], atomicity here is not really important.
+    KRef old = compareAndSwapImpl(thiz, expectedValue, newValue);
```
SvyatoslavScherbina: Since we use lock, it is not necessary to perform atomic CAS here.

runtime/src/main/cpp/Atomic.cpp
```diff
@@ -62,9 +82,45 @@ void Kotlin_AtomicReference_checkIfFrozen(KRef value) {
...
+    unlock(thiz);
+    if (old == expectedValue) {
+        // CAS to the new value was successful, we transfer ownership of [expectedValue] to the caller.
+        // No need to update the reference counter, as before @AtomicReference was holding the reference,
+        // and now result holds it. Also note, that object referred by [expectedValue] cannot be freed,
+        // as it is guaranteed to be held by the caller (and by [thiz]).
+        *OBJ_RESULT = old;
+        return old;
+    } else {
+        // On this path we just create an additional reference to the [expectedValue], held by caller anyway.
```
SvyatoslavScherbina: Data race can also occur here, if other thread performs CAS between `unlock` and this line.
And the comment seems to be obsolete.

runtime/src/main/cpp/Atomic.cpp
```diff
@@ -62,9 +82,45 @@ void Kotlin_AtomicReference_checkIfFrozen(KRef value) {
...
+    // As we lock [thiz], atomicity here is not really important.
+    KRef old = compareAndSwapImpl(thiz, expectedValue, newValue);
+    unlock(thiz);
+    if (old == expectedValue) {
+        // CAS to the new value was successful, we transfer ownership of [expectedValue] to the caller.
+        // No need to update the reference counter, as before @AtomicReference was holding the reference,
+        // and now result holds it. Also note, that object referred by [expectedValue] cannot be freed,
+        // as it is guaranteed to be held by the caller (and by [thiz]).
+        *OBJ_RESULT = old;
+        return old;
```
SvyatoslavScherbina: Shouldn't we also make `AddRef(newValue)` here?

runtime/src/main/cpp/Atomic.cpp
olonho: Yep.

runtime/src/main/cpp/Atomic.cpp
olonho: Yes, as said in comment, it is just shorter.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1652,4 +1669,34 @@ void MutationCheck(ObjHeader* obj) {
...
+OBJ_GETTER(SwapRefLocked,
+    ObjHeader** location, ObjHeader* expectedValue, ObjHeader* newValue, int32_t* spinlock) {
+  lock(spinlock);
+  ObjHeader* oldValue = *location;
+  // We do not use UpdateRef() here to avoid having ReleaseRef() on return slot under lock.
+  if (oldValue == expectedValue) {
+    SetRef(location, newValue);
+  } else {
+    // We create an additional reference to the [oldValue] in the return slot.
+    if (!isArenaSlot(OBJ_RESULT) && oldValue != nullptr)
```
SvyatoslavScherbina: Is it correct? For example, `UpdateReturnRef` has special non-trivial handling for arena slot.

PR #1713, opened 2018-06-22
Make objects frozen by default

olonho:


runtime/src/main/kotlin/konan/Annotations.kt
```diff
@@ -101,18 +71,26 @@ public annotation class FixmeReflection
...
+/**
+ * Escape analysis annotations.
+ */
 public annotation class Escapes(val who: Int)
 
 public annotation class PointsTo(vararg val onWhom: Int)
 
 /**
- * Need to be fixed because of header/impl notation
+ * Top level variable or object is thread local, and so could be mutable.
```
SvyatoslavScherbina: This behaviour can't be implemented efficiently if we separate workers and OS threads.
I suppose we should support main-thread-only objects instead.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1188,28 +1188,27 @@ OBJ_GETTER(InitSharedInstance,
...
   ObjHeader* object = AllocInstance(type_info, OBJ_RESULT);
-  MEMORY_LOG("Calling UpdateRef from InitInstance\n")
+  RuntimeAssert(object->container()->normal() , "Shared object cannot be co-allocated");
+  MEMORY_LOG("Calling UpdateRef from InitSharedInstance\n")
   UpdateRef(localLocation, object);
 #if KONAN_NO_EXCEPTIONS
   ctor(object);
-  if (!object->container()->frozen())
-    ThrowFreezingException();
+  FreezeSubgraph(object);
```
SvyatoslavScherbina: So it makes sense to remove explicit freezing from enums, isn't it?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1550,85 +1549,63 @@ void traverseStronglyConnectedComponent(ContainerHeader* container,
...
+    reversedEdges.emplace(current, KStdVector<ContainerHeader*>(0));
+    traverseContainerReferredObjects(current, [current, &queue, &reversedEdges](ObjHeader* obj) {
+          ContainerHeader* objContainer = obj->container();
+          if (!objContainer->permanentOrFrozen()) {
+            if (objContainer->marked())
+              queue.push_back(objContainer);
+            reversedEdges.emplace(objContainer, KStdVector<ContainerHeader*>(0)).first->second.push_back(current);
+          }
+      });
+    }
```
SvyatoslavScherbina: Wrong indentation?

runtime/src/main/cpp/Memory.cpp
olonho: Yep.

runtime/src/main/cpp/Memory.cpp
olonho: Fixed.

olonho:
Abandon this one, https://github.com/JetBrains/kotlin-native/pull/1743 instead.

PR #1721, opened 2018-06-25
[gradle-plugin] Fix MPP in the new plugin after switching to 1.2.50

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/plugins/KotlinPlatformNativePlugin.kt
```diff
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlin.gradle.plugin.experimental.plugins
 
+import org.gradle.api.NamedDomainObjectContainer
```
vvlevchenko: unused

ilmat192: Fixed.

PR #1743, opened 2018-06-29
Freeze objects by default, stdlib improvements.

olonho:


runtime/src/main/kotlin/konan/Annotations.kt
```diff
@@ -101,18 +71,26 @@ public annotation class FixmeReflection
...
 public annotation class PointsTo(vararg val onWhom: Int)
 
 /**
- * Need to be fixed because of header/impl notation
+ * Top level variable or object is thread local, and so could be mutable.
  */
-public annotation class FixmeMultiplatform
+@Target(AnnotationTarget.FIELD, AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.BINARY)
+annotation class ThreadLocal
```
SvyatoslavScherbina: Is it still required?

runtime/src/main/kotlin/konan/worker/Freezing.kt
```diff
@@ -26,7 +26,8 @@ public class FreezingException() : RuntimeException()
 /**
  * Exception thrown whenever we attempt to mutate frozen objects.
  */
-public class InvalidMutabilityException() : RuntimeException()
+public class InvalidMutabilityException(where: Any) :
+        RuntimeException("mutation attempt of frozen $where (hash is 0x${(where as Any).hashCode().toString(16)})")
```
SvyatoslavScherbina: What's the purpose of casting `Any` to `Any`?

runtime/src/main/kotlin/konan/Annotations.kt
olonho: Well, without it we cannot locally disable freezing on objects. I can write big fat warning about instability of such annotation.

runtime/src/main/kotlin/konan/worker/Freezing.kt
olonho: Maybe none, I just didn't want virtual dispatch.

runtime/src/main/kotlin/konan/Annotations.kt
SvyatoslavScherbina: What's the purpose of introducing feature that is already deprecated and is not used?

PR #1744, opened 2018-06-29
Implement random and shuffle

### LINKED ISSUE #1675: Support for Collections.shuffled?
I'm using `implementation "org.jetbrains.kotlin:kotlin-stdlib-common:1.2.41"` in a common module of course, and there I'm able to call `.shuffled()` on a list object. This works fine in Android since it's a JVM implementation, but in iOS I'm having an `Unresolved reference: shuffled` error since this is [not implemented](https://github.com/JetBrains/kotlin-native/blob/670d05a4f35cbb3613d694406af515a13b796fb3/backend.native/tests/stdlib_external/collections/MutableCollectionsTest.kt#L108) in kotlin-native, so I'm wondering if this should be implemented in kotlin-native too or if this should be removed from `kotlin-stdlib-common` or what's the expected behavior in this scenario. The same with `.fill()` and other collection apis.

### LINKED ISSUE #1699: Generating a random number in Kotlin -> WebAssembly
Hello,

When compiling Kotlin to WebAssembly, I can see that in the generated JavaScript file there is `knjs__Math_random`, however when trying to import `kotlin.math.random` in Kotlin, I get a compilation error. Doesn't the `random` function reside in `kotlin.math` or I am missing something? Other math functions work fine.

PavelPunegov:
- Implements a simple random using rand/srand() methods and random() if available. Wasm uses JS Math.random. Fixes #1699
- Fix shuffle method(). Fixes #1675

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -493,6 +493,15 @@ task runtime_basic_exit(type: RunStandaloneKonanTest) {
     expectedExitStatus = 42
 }
 
+task runtime_random(type: RunKonanTest) {
+    source = "runtime/basic/random.kt"
+}
+
+task runtime_worker_random(type: RunKonanTest) {
+    disabled = (project.testTarget == 'wasm32')
+    source = "runtime/basic/worker_random.kt"
+}
+
 task hello0(type: RunKonanTest) {
     goldValue = "Hello, world!\n"
     source = "runtime/basic/hello0.kt"
--- a/backend.native/tests/runtime/basic/random.kt
+++ b/backend.native/tests/runtime/basic/random.kt
@@ -1,0 +1,65 @@
+package runtime.basic.random
+
+import kotlin.collections.*
+import kotlin.random.*
+import kotlin.system.*
+import kotlin.test.*
+
+/**
+ * Tests that setting the same seed make random generate the same sequence
+ */
+private inline fun <reified T> testReproducibility(seed: Long, generator: () -> T) {
+    // Reset seed. This will make Random to start a new sequence
+    Random.seed = seed
+    val first = Array<T>(50, { i -> generator() }).toList()
+
+    // Reset seed and try again
+    Random.seed = seed
+    val second = Array<T>(50, { i -> generator() }).toList()
+    assertTrue(first == second, "FAIL: got different sequences of generated values " +
+            "first: $first, second: $second")
+}
+
+/**
+ * Tests that setting seed makes random generate different sequence.
+ */
+private inline fun <reified T> testDifference(generator: () -> T) {
+    Random.seed = 12345678L
+    val first = Array<T>(100, { i -> generator() }).toList()
+
+    Random.seed = 87654321L
+    val second = Array<T>(100, { i -> generator() }).toList()
+    assertTrue(first != second, "FAIL: got the same sequence of generated values " +
+            "first: $first, second: $second")
+}
+
+@Test
+fun testInts() {
+    testReproducibility(getTimeMillis(), { Random.nextInt() })
+    testReproducibility(Long.MAX_VALUE, { Random.nextInt() })
+}
+
+@Test
+fun testLong() {
+    testReproducibility(getTimeMillis(), { Random.nextLong() })
+    testReproducibility(Long.MAX_VALUE, { Random.nextLong() })
+}
+
+@Test
+fun testDiffInt() = testDifference { Random.nextInt() }
+
+@Test
+fun testDiffLong() = testDifference { Random.nextLong() }
+
+@Test
+fun testBoundInt() {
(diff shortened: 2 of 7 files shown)


backend.native/tests/stdlib_external/collections/MutableCollectionsTest.kt
```diff
@@ -94,17 +94,13 @@ class MutableCollectionTest {
...
     @Test fun listFill() {
         val list = MutableList(3) { it }
         list.fill(42)
         assertEquals(listOf(42, 42, 42), list)
     }
-*/
 
-/*
-    // shuffle is Kotlin/JVM only method
+    // shuffle is not Kotlin/JVM only method
```
olonho: Not sure if this comment makes much sense.

backend.native/tests/build.gradle
```diff
@@ -489,6 +489,11 @@ task runtime_basic_exit(type: RunStandaloneKonanTest) {
     expectedExitStatus = 42
 }
 
+task runtime_random(type: RunKonanTest) {
+    disabled = (project.testTarget == 'wasm32')
```
olonho: Why, if you have WASM implementation below?

runtime/src/main/cpp/Natives.h
```diff
@@ -136,6 +136,10 @@ OBJ_GETTER0(Kotlin_konan_internal_undefined);
 void Kotlin_konan_internal_GC_suspend(KRef);
 void Kotlin_konan_internal_GC_resume(KRef);
 
+// Random operations.
+KInt Kotlin_random();
```
olonho: I'd say that you'd better avoid using system's RNG and have consistent cross-platform implementation instead.

backend.native/tests/runtime/basic/random.kt
```diff
@@ -0,0 +1,95 @@
+package runtime.basic.random
+
+import konan.worker.*
+import kotlin.collections.*
+import kotlin.random.*
+import kotlin.system.*
+import kotlin.test.*
+
+private inline fun <reified T> getRandomNumbers(amount: Int = 10, producer: () -> T): List<T> {
```
olonho: Not sure if this function got anything to do with random numbers, maybe smth like
`Array<T>(amount, producer).toList()` would be a shorter implementation.

backend.native/tests/runtime/basic/random.kt
```diff
@@ -0,0 +1,95 @@
...
+    for (i in 0 .. amount - 1) {
+        list.add(producer())
+    }
+    return list
+}
+
+/**
+ * Tests that setting the same seed make random generate the same sequence
+ */
+private inline fun <reified T> testReproducibility(seed: Int, generator: () -> T) {
```
olonho: Why reified is needed here and above?

runtime/src/main/kotlin/kotlin/random/Random.kt
```diff
@@ -0,0 +1,77 @@
...
+     * This implementation is a singleton generator. A sequence of generated numbers
+     * and seed is shared between all workers and native threads.
+     */
+    companion object : Random() {
+        init { updateSeed() }
+
+        /**
+         * Random generator seed value.
+         */
+        var seed: Int = getTimeNanos().toInt()
```
olonho: Please don't, it will kill object freezing. Do it differently:
make your RNG keep state in an AtomicInt/AtomicLong and have system RNG
keeping such an object as a frozen val.

PavelPunegov:
Fixed review issues and implemented the same LCD random as in Apache Harmony

olonho:
Hmm, it seems 64-bit CAS is not present on MIPS, may make sense to replace it with locked operation for MIPS, or just add missing symbol there.
.

runtime/src/main/kotlin/kotlin/random/Random.kt
```diff
@@ -0,0 +1,78 @@
...
+    abstract fun nextInt(bound: Int): Int
+
+    abstract fun nextLong(): Long
+
+    /**
+     * A default pseudo-random LCD-generator.
+     */
+    companion object : Random() {
+        private const val MULTIPLIER = 0x5deece66dL
+        private val _seed = AtomicLong(mult(getTimeNanos()))
```
olonho: Maybe make seed 32-bit to avoid MIPS CAS problems?

runtime/src/main/kotlin/konan/internal/RuntimeUtils.kt
```diff
@@ -157,4 +157,10 @@ internal fun <T> listOfInternal(vararg elements: T): List<T> {
     for (i in 0 until elements.size)
         result.add(elements[i])
     return result
-}
\ No newline at end of file
+}
```
olonho: Why it is still needed?

Issue #1716, opened 2018-06-23
Add Serverless Sample


napperley:
In the Kotlin Native samples there isn't a **Serverless** sample. With Serverless there is a good case for the use of Kotlin Native. The JVM isn't optimised/designed for [Serverless applications](https://blog.travelex.io/blazing-fast-microservice-with-go-and-lambda-d30d95290f28) due to their performance characteristics/requirements:

- Quick start up time
- Low memory usage
- Small code size
- Minimal software dependencies
- Quick response time

There is a valuable opportunity for Kotlin Native to enable Kotlin to have a significant Serverless presence. One key part of taking advantage of this opportunity requires having Serverless samples available so that Kotliners can get started quickly with Serverless. Already some organisations like Pivotal (heavy adopter of Kotlin for server-side development) are [exploring](https://github.com/projectriff/riff/issues/283#issuecomment-390734154) the use of Kotlin Native over Kotlin JVM as the preferred/best way to develop Serverless applications for the **"long term"**. IBM is another organisation that is [considering](https://github.com/ibm-functions/runtime-kotlin/issues/12) providing Kotlin Native support (also likely to be **"long term"**).

One recommended way to provide a Serverless sample would be to have one based on the [OpenFaaS Weather sample](https://gitlab.com/napperley/openfaas-weather-sample) (unofficial OpenFaaS sample). [OpenFaaS](https://github.com/openfaas/faas) would be a good Serverless platform to start with for the following reasons:

- OpenFaaS is a open source project that uses a permissive license (MIT)
- Isn't tied into a proprietary FaaS (Function As A Service) provider
- Uses Docker to easily manage deployment (via Docker images)
- Automatic scaling with Docker Swarm or Kubernetes
- Uses Prometheus for performance metrics and monitoring
- Some major business's have adopted or are experimenting with OpenFaaS (Intel, VMWare, T-Mobile, HashiCorp etc)
- Has a basic Web tool to test Serverless Functions in the development phase
- Provides reasonable quality [documentation](https://docs.openfaas.com/)

olonho:
Nick, generally it sounds like a noble goal, would you be interested in providing PR with this sample?

napperley:
Sorry about the late response. Ended up with so many **kotlin-native** repo notifications in the Inbox that I created a filter to move them into a folder (one of the hazards of email filtering - important emails get lost in the system). Would be happy to do a PR for the sample.

olonho:
Merged.

PR #1746, opened 2018-07-01
Weather Function Sample

napperley:
This **new** Kotlin Native sample covers how to do a [Serverless Function](https://github.com/JetBrains/kotlin-native/issues/1716), which uses [OpenFaaS](https://www.openfaas.com/) as the target Serverless platform.


samples/weather_function/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
olonho: Why not do as other samples and use Gradle wrapper from ..?

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM openjdk:8u171-jdk-stretch as builder
+WORKDIR /opt
+RUN wget "https://github.com/JetBrains/kotlin-native/releases/download/v0.7.1/kotlin-native-linux-0.7.1.tar.gz"
+RUN tar -xzf kotlin-native-linux-0.7.1.tar.gz && rm kotlin-native-linux-0.7.1.tar.gz
```
StefMa: Can we extract a "variable" or something for the version?
Would be great - if we update the sample to a newer version - to just change the variable...

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get update && apt-get install libcurl4-openssl-dev
+
+ENV KOTLIN_NATIVE_HOME "/opt/kotlin-native-linux-0.7.1"
+ENV GRADLE "/opt/gradle-4.7/bin/gradle"
```
StefMa: Gradle 4.8.1 is already relased 👼

samples/weather_function/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+konanArtifacts {
+    interop("curl") {
+        defFile("curl.def")
+    }
+
+    interop("cjson") {
+        defFile("cjson.def")
+    }
+
+	program("weather") {
```
StefMa: Can you please fix the indentation here?

samples/weather_function/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+    	entryPoint("org.example.weather_func.main")
+        libraries {
```
StefMa: Same here with the indentation.

samples/weather_function/readme.md
```diff
@@ -0,0 +1,36 @@
+# Weather Function Sample
+
+This is a Serverless Function that fetches the current weather information. Part of the Function's output includes HTTP response headers, and the HTTP response body. This sample is designed to be deployed on OpenFaaS via a Docker image.
+
+This document assumes that you are using a PC running Debian (v9 or later), Ubuntu (v14.04 or later), or Linux Mint (v17 or later).
```
StefMa: Since it is docker, why is it "needed" to have a Debian, Ubuntu or Linux Mint OS? 🤔

samples/weather_function/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+# Building Docker Image
+
+It will be assumed that Docker is already installed.
+
+1. Clone the Kotlin Native Git repository: ```git clone https://github.com/JetBrains/kotlin-native ~/repos/kotlin-native```
```
StefMa: Is that path still correct?
Why should it be cloned to this path?

samples/weather_function/readme.md
```diff
@@ -0,0 +1,36 @@
+# Weather Function Sample
+
+This is a Serverless Function that fetches the current weather information. Part of the Function's output includes HTTP response headers, and the HTTP response body. This sample is designed to be deployed on OpenFaaS via a Docker image.
```
StefMa: Please add some information from where it fetched the weather information

samples/weather_function/readme.md
```diff
@@ -0,0 +1,36 @@
...
+4. Select **MANUALLY** tab
+5. For **Docker image** enter in **weather**
+6. For **Function name** enter in **weather**
+7. Click on **DEPLOY** button
+8. Select the **weather** function
+9. For **Request body** enter in the location: 
+**-l="christchurch,nz"**
+10. Click on **INVOKE** button
+
+Do note that the program (weather) can print weather information from a JSON file, eg: ```./weather -f="current_weather.json"```, but this functionality isn't available in the Serverless Function.
```
StefMa: Please change it to:
```
> **Note:** The program (weather) ....
```

This looks a little bit nicer

samples/weather_function/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Building Docker Image
+
+It will be assumed that Docker is already installed.
+
+1. Clone the Kotlin Native Git repository: ```git clone https://github.com/JetBrains/kotlin-native ~/repos/kotlin-native```
+2. Change working directory to the sample:
+```cd ~/repos/kotlin-native/samples/weather_function```
+3. Create the **openweathermap_key.txt** file and append your [Open Weather Map API key](https://openweathermap.org/appid) to the file
+4. Build Docker image: ```docker build --t weather .```
```
StefMa: inline code can be used with just a single \`.
Eg:
```
This is not code. But `this is`.
```

samples/weather_function/src/main/kotlin/org/example/weather_func/Weather.kt
```diff
@@ -0,0 +1,14 @@
+package org.example.weather_func
+
+/** Models the current weather for a location. */
```
StefMa: The correct doc should be styled like this:
```kotlin
/**
* Some doc
*/
```

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.cinterop.readBytes
+
+/**
+ * Provides basic HTTP client functionality through the libCurl library.
+*/
+internal class CUrl(val url: String) {
+	private val stableRef = StableRef.create(this)
+	private val curlObj = initCurl()
+	val header = Event<String>()
+    val body = Event<String>()
```
StefMa: fix the intention here

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,86 @@
...
+        setCurlOption(curlObj, CURLOPT_HEADERDATA, stableRef.asCPointer())
+        setCurlOption(curlObj, CURLOPT_WRITEFUNCTION, writeData)
+        setCurlOption(curlObj, CURLOPT_WRITEDATA, stableRef.asCPointer())
+    }
+
+    fun fetch() {
+        // Have Curl do a HTTP/HTTPS request and store the response (status).
+        val response = curlFileSync(curlObj)
+        // length in bytes.
+        val length = 8192 * 50
```
StefMa: What are these for magic numbers?

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,86 @@
...
+        // Print the error message if the Curl status code isn't OK (CURL_ERROR_OK).
+        if (response != CURL_ERROR_OK) println("curlFileSync() failed: ${curlErrorMessage(response)?.toKString(length)}")
+    }
+
+	fun close() {
+		cleanupCurl(curlObj)
```
StefMa: intention

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,86 @@
...
+        val data = buffer.toKString((size * totalItems).toInt()).trim()
+        val curlRef = userData.asStableRef<CUrl>().get()
+
+        curlRef.body(data)
+        responseSize = size * totalItems
+    }
+    return responseSize
+}
+
+private fun CPointer<ByteVar>.toKString(length: Int) = readBytes(length).stringFromUtf8()
```
StefMa: I think that is not a good naming.
Maybe it make sense here to have a normal (not extension) method which sounds like `byteBarToString(CPointer, length)`

samples/weather_function/src/main/kotlin/org/example/weather_func/json.kt
```diff
@@ -0,0 +1,162 @@
...
+	addNumberToObject(tmpObj, "temp", temp.toDouble())
+	addNumberToObject(tmpObj, "minTemp", minTemp.toDouble())
+	addNumberToObject(tmpObj, "maxTemp", maxTemp.toDouble())
+	addItemToObject(rootObj, "tempInfo", tmpObj)
+}
+
+private fun addConditionsArray(rootObj: CPointer<cJSON>?, conditions: Array<Pair<String, String>>) {
+	val tmpArr = addArrayToObject(rootObj, "conditions")
+
+	conditions.forEach { c ->
```
StefMa: what is `c`? Can be ommited and we can use `it` here...

samples/weather_function/src/main/kotlin/org/example/weather_func/main.kt
```diff
@@ -0,0 +1,120 @@
...
+import kotlin.system.exitProcess
+
+private val API_KEY by lazy { fetchApiKey() }
+
+fun main(args: Array<String>) {
+	val input = readLine()
+	val location = fetchLocationArg(input)
+	val jsonFile = fetchJsonFileArg(input)
+
+	if (location.isNotEmpty()) {
```
StefMa: Could be a `when {}`

samples/weather_function/src/main/kotlin/org/example/weather_func/main.kt
```diff
@@ -0,0 +1,120 @@
...
+
+private fun createUrl(location: String): String {
+	val baseUrl = "http://api.openweathermap.org/data/2.5/weather"
+	return "$baseUrl?q=$location&units=metric&appid=$API_KEY"
+}
+
+private fun fetchApiKey(): String {
+	var result = ""
+	// bufferLength in bytes.
+	val bufferLength = 50 * 8
```
StefMa: What are these for magic numbers?

samples/weather_function/readme.md
napperley: The path provides a central place users can easily follow in the documentation.

samples/weather_function/src/main/kotlin/org/example/weather_func/Weather.kt
napperley: According to the Kotlin Coding Conventions [short comments](https://kotlinlang.org/docs/reference/coding-conventions.html#documentation-comments) can be placed on a single line.

samples/weather_function/src/main/kotlin/org/example/weather_func/json.kt
napperley: The **c** parameter in the lambda stands for condition which is fine in this case. Since a [multi line lambda](https://kotlinlang.org/docs/reference/coding-conventions.html#lambda-formatting) is being used it wouldn't be appropriate to use **it** as a parameter name.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
napperley: Looking to change line 51 so it has the following:
```kotlin
// length in bytes (maximum characters is 50).
// Biggest sized UTF character is 4 bytes (https://stackoverflow.com/questions/9533258/what-is-the-maximum-number-of-bytes-for-a-utf-8-encoded-character/9533324#9533324)
val length = 4 * 50
```

samples/weather_function/src/main/kotlin/org/example/weather_func/main.kt
napperley: Looking to change lines 107-108 so it has the following:
```kotlin
// Largest UTF character size is 4 bytes (https://stackoverflow.com/questions/9533258/what-is-the-maximum-number-of-bytes-for-a-utf-8-encoded-character/9533324#9533324).
// length in bytes (maximum characters is 50).
val length = 4 * 50
```

napperley:
@olonho - Docker doesn't allow files/directories to be accessed outside of the build environment which means the sample has to have its own copy of the Gradle Wrapper. Using a symbolic link doesn't work with Docker's **COPY** command in the **Dockerfile**.

samples/weather_function/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
olonho: Please avoid having separate Gradle distro for this sample.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,89 @@
+package org.example.weather_func
+
+import org.example.weather_func.Event
+// size_t is a alias for Long.
```
olonho: It is not on 32-bit platforms.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,89 @@
...
+import platform.posix.size_t
+
+import curl.curl_easy_setopt as setCurlOption
```
olonho: Not sure if `import smth as smth_else` is generally a good practise, as introduces uneeded level of indirection.

samples/weather_function/src/main/kotlin/org/example/weather_func/events.kt
```diff
@@ -0,0 +1,31 @@
+package org.example.weather_func
+
+internal typealias EventHandler<T> = (T) -> Unit
+
+/**
+ * Covers event handling for the program.
+*/
+internal class Event<T : Any> {
```
olonho: Why non-nullable types only here?

samples/weather_function/src/main/kotlin/org/example/weather_func/main.kt
```diff
@@ -0,0 +1,123 @@
...
+		| Usage examples (pass one of the following as input to the function):
+		|   * Location (uses the Open Weather Map service), eg:
+		|       -l="christchurch,nz"
+		|   * JSON File, eg: -f="weather.json"
+		""".trimMargin()
+	)
+	exitProcess(0)
+}
+
+private fun checkFile(file: String) {
```
olonho: In theory this is racy, as file may go away between check and actual open, maybe an open with error check would be better.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
olonho: This extension already exists in kotlinx.cinterop, so just remove it and shall work.

samples/weather_function/src/main/kotlin/org/example/weather_func/main.kt
```diff
@@ -0,0 +1,123 @@
...
+    val utfCharSize = 4
+	// bufferLength in bytes.
+	val bufferLength = utfCharSize * maxChars
+	// Open the file using the fopen function and store the file handle.
+	val file = fopen("openweathermap_key.txt", "r")
+
+	memScoped {
+		val buffer = allocArray<ByteVar>(bufferLength)
+		// Read a line from the file using the fgets function.
+		result = fgets(buffer, bufferLength, file)?.toKString()?.replace("\n", "") ?: ""
```
olonho: Isn't `trim()` better fit here? Also you could get away without `memScoped`, with smth like
```
val array = ByteArray(1024)
val nextLine = fgets(buffer.refTo(0), buffer.size, file)?.toKString()?.trim()
```
Also it's better check if `file != null` before reading.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
```diff
@@ -0,0 +1,89 @@
...
+        curl_easy_setopt(curlObj, CURLOPT_URL, url)
+        curl_easy_setopt(curlObj, CURLOPT_HEADERFUNCTION, header)
+        curl_easy_setopt(curlObj, CURLOPT_HEADERDATA, stableRef.asCPointer())
+        curl_easy_setopt(curlObj, CURLOPT_WRITEFUNCTION, writeData)
+        curl_easy_setopt(curlObj, CURLOPT_WRITEDATA, stableRef.asCPointer())
+    }
+
+    fun fetch() {
+        // Have Curl do a HTTP/HTTPS request and store the response (status).
+        val response = curl_easy_perform(curlObj)
```
alexellis: I'm curious (and new to Kotlin).. is there no better way to do this than linking to curl? I've never really seen that done in a higher-level language - there is normally some kind of http package in the stdlib or ecosystem.

samples/weather_function/readme.md
```diff
@@ -0,0 +1,33 @@
...
+# Usage
+
+Make sure that you have completed the *Building Docker Image* section, and that OpenFaaS is [installed](https://docs.openfaas.com/deployment/) along with the [OpenFaaS CLI](https://github.com/openfaas/faas-cli) before proceeding.
+
+1. Start OpenFaaS
+2. Goto the following to access the OpenFaaS Portal:
+http://127.0.0.1:8080
+3. Click on **Deploy New Function** button
+4. Select **MANUALLY** tab
+5. For **Docker image** enter in **weather**
```
alexellis: I do not recommend this approach, the UI is less than ideal for deployments.

I recommend you create a YAML file here and use that to deploy from. For inspiration please see the way we deploy the samples: https://github.com/openfaas/faas/blob/master/stack.yml

samples/weather_function/readme.md
```diff
@@ -0,0 +1,33 @@
...
+4. Build the *weather* Docker image by running the shell script: `./build_image`
+
+
+# Usage
+
+Make sure that you have completed the *Building Docker Image* section, and that OpenFaaS is [installed](https://docs.openfaas.com/deployment/) along with the [OpenFaaS CLI](https://github.com/openfaas/faas-cli) before proceeding.
+
+1. Start OpenFaaS
+2. Goto the following to access the OpenFaaS Portal:
+http://127.0.0.1:8080
```
alexellis: This URL is only valid if the user follows the Swarm deployment guide, otherwise it's port `:31112` for Kubernetes (when using a local environment). Might be better to ask them to install on Swarm if you want to assume a quick/happy-path.

samples/weather_function/readme.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+# Building Docker Image
+
+It will be assumed that Docker is already installed.
+
+1. Clone the Kotlin Native Git repository: `git clone https://github.com/JetBrains/kotlin-native ~/repos/kotlin-native`
+2. Change working directory to the sample: `cd ~/repos/kotlin-native/samples/weather_function`
+3. Create the **openweathermap_key.txt** file and append your [Open Weather Map API key](https://openweathermap.org/appid) to the file
+4. Build the *weather* Docker image by running the shell script: `./build_image`
```
alexellis: Once you convert to the YAML file format for OpenFaaS you can use the CLI do to this build instead (which is preferred)

```
faas-cli build -f weather-example.yml
```

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+COPY gradle/wrapper/gradle-wrapper.properties /app/gradle/wrapper
+COPY gradle/wrapper/gradle-wrapper.jar /app/gradle/wrapper
+COPY gradlew /app
+COPY .konan/ /root/.konan
+COPY *.def /app/
+COPY *.kts /app/
+COPY openweathermap_key.txt /app
+COPY src/ /app/src
+COPY lib/cJSON-1.7.7/include/cjson/cJSON.h /app/lib/cJSON-1.7.7/include/cJSON.h
```
alexellis: This is very verbose and each step will take additional time. I suggest you add the whole folder that you want, since this uses a multi-stage build there is no advantage to picking individual files out.

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM openjdk:8u171-jdk-stretch as builder
+WORKDIR /opt
+ENV KOTLIN_NATIVE_VER "0.7.1"
+RUN wget "https://github.com/JetBrains/kotlin-native/releases/download/v0.7.1/kotlin-native-linux-$KOTLIN_NATIVE_VER.tar.gz"
```
alexellis: These lines should all be one instruction to enable caching, this is an inefficient usage of a Dockerfile

```
RUN wget ... \
  && tar .... \
  && apt-get 
```

For your `apt-get` you need to specify `-qy` as an argument to suppress unexpected prompting behaviour.

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN tar -xzf kotlin-native-linux-$KOTLIN_NATIVE_VER.tar.gz && rm kotlin-native-linux-$KOTLIN_NATIVE_VER.tar.gz
+RUN apt-get update && apt-get install libcurl4-openssl-dev
+
+ENV KOTLIN_NATIVE_HOME "/opt/kotlin-native-linux-$KOTLIN_NATIVE_VER"
+RUN mkdir -p /app/src/main/kotlin/org/example/weather_func
```
alexellis: Again all these lines should be one instruction.

alexellis:
@napperley Given the public nature of this sample, I've given some feedback from an OpenFaaS perspective. As much as you want this to be Kotlin-idiomatic, it needs to be OpenFaaS-idiomatic too. Hope the feedback is helpful.

samples/weather_function/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM openjdk:8u171-jdk-stretch as builder
+WORKDIR /opt
+ENV KOTLIN_NATIVE_VER "0.7.1"
```
olonho: Maybe 0.8?

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
napperley: No, at least before Kotlin Native 0.8. However with the recent Kotlin Native [release](https://blog.jetbrains.com/kotlin/2018/07/kotlinnative-v0-8-released/) there is a opportunity to use the Kotlin Native HTTP client library (a port of the Ktor ver from the JVM side) instead of Curl. Kotlin JVM doesn't have this issue since there are some well established HTTP client libraries available ([OkHttp](http://square.github.io/okhttp/), [Retrofit](http://square.github.io/retrofit/), [Apache HttpClient](https://hc.apache.org/httpcomponents-client-ga/) etc).

If there are some people involved in the development of the new Kotlin Native HTTP Client library, who are reading this then please get in contact. Interested in migrating from Curl to the new library.

samples/weather_function/Dockerfile
napperley: Noticed that the Docker image build takes a while to copy over the Konan cache (does it every time Docker is building a image), and a user would need to compile a basic Linux amd64 program with Kotlin Native **before** the Konan cache can be copied over. Is there a easier way to manage the Konan cache?

samples/weather_function/Dockerfile
napperley: Looking to move to 0.8 provided it is feasible to migrate from Curl to the new Kotlin Native HTTP Client library.

samples/weather_function/src/main/kotlin/org/example/weather_func/curl.kt
napperley: @alexellis - Received a response about the Ktor Linux port [issue](https://github.com/ktorio/ktor/issues/479). The port is being taken under consideration for the near future. In the meantime Curl remains the HTTP client library to use for this sample even though it isn't a ideal fit  :(.

PR #1752, opened 2018-07-02
[kotlin compiler][update] 1.2.70-dev-56

vvlevchenko:
- Revert "Use local Kotlin jars built from 372e2c5c"

This reverts commit fd016db33af2b2639a0ba272650e88e5d0b102de.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -144,8 +144,16 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
     }
 
     //-------------------------------------------------------------------------//
-
-    private fun inlineFunction(callee: IrCall,                                              // Call to be substituted.
+    /**
+     * TODO: JVM inliner crushed on attempt inline this function from transform.kt
```
olonho: Crashed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -144,8 +144,16 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
...
+     */
+    private fun <T : IrElement> MutableList<T>.transform(transformation: (T) -> IrElement) {
```
SvyatoslavScherbina: Will JVM inliner crash if this function is `inline` with `reified T`?

vvlevchenko: No. It seems information for inline is missed.

PR #1753, opened 2018-07-02
Dispatch Any methods called on interface via vtable instead of itable

qwwdfsad:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -629,14 +631,37 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+        assert(typeInfoPtr.type == codegen.kTypeInfoPtr) { LLVMPrintTypeToString(typeInfoPtr.type)!!.toKString() }
+        val declaration = descriptor.containingDeclaration as ClassDescriptor
+        var actualDescriptor = descriptor as SimpleFunctionDescriptor
+        val callDescriptor = descriptor.descriptor
+
+        /*
+         * Resolve owner of the call with special handling of Any methods:
+         * if toString/eq/hc is invoked on an interface instance, we resolve
+         * owner as Any and dispatch it via vtable
+         */
```
SvyatoslavScherbina: The code below completely differs from the description above.
It should be
1) More straightforward
2) Extracted to the method, e.g.
```
fun IrSimpleFunction.findOverriddenMethodOfAny(): IrSimpleFunction?
```

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -629,14 +631,37 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
             load(typeInfoPtrPtr)
         }
 
-        assert (typeInfoPtr.type == codegen.kTypeInfoPtr) { LLVMPrintTypeToString(typeInfoPtr.type)!!.toKString() }
+        assert(typeInfoPtr.type == codegen.kTypeInfoPtr) { LLVMPrintTypeToString(typeInfoPtr.type)!!.toKString() }
+        val declaration = descriptor.containingDeclaration as ClassDescriptor
+        var actualDescriptor = descriptor as SimpleFunctionDescriptor
+        val callDescriptor = descriptor.descriptor
```
SvyatoslavScherbina: Is it possible to write this code without using descriptors?

PR #1769, opened 2018-07-05
Atomic lazy getter.

olonho:


PavelPunegov:
|`runtime/src/main/kotlin/konan/worker/Atomics.kt:154-157`|
|---|
Why not an enum or a sealed class?

PavelPunegov:
|`runtime/src/main/kotlin/konan/worker/Atomics.kt:167`|
|---|
I'd replace this and `else` with
`val ctor = initializer_.get() ?: throw Error()`

PavelPunegov:
|`runtime/src/main/kotlin/konan/worker/Atomics.kt:180`|
|---|
`!==` ?

Issue #1791, opened 2018-07-10
Bound checking elimination


soywiz:
For this kotlin code:

```kotlin
fun main(args: Array<String>) {
        val a = IntArray(100_000_000)
        for (n in 0 until a.size) {
                a[n] = 1
        }
        var sum = 0
        for (n in 0 until a.size) {
                sum += a[n]
        }
        println(sum)
}
```

```
kotlinc -opt demo.kt
```

-->

This seems to be the C decompiled code after generating an optimized binary:

![bound_checking](https://user-images.githubusercontent.com/570848/42531353-0db0c0ac-8484-11e8-836e-f494837e9fb2.png)

So I guess something can be done here. If LLVM knows that a value is going to be constant for the whole loop, it would move it outside the loop, and would match the symbolic range check eliminating the if check completely.

Maybe the -1 there doesn't help the optimizer either to do the symbolic range checking for the dead code elimination.

---

This C code seems to do bound checking elimination because LLVM knows that `n` won't never be greater than `array.size` and knows that `array.size` is constant:

```
// clang -S -O3 -emit-llvm test.c && cat test.ll # no abort appears in the .ll file

#include <stdio.h>

typedef struct Array {
        char* ptr;
        int size;
} Array;

inline void arrayCheck(Array array, int n) {
        if (n < 0 || n >= array.size) abort();
}

int main(int argc, char **argv) {
        int size = atoi(argv[1]);
        Array array = {alloca(size), size};
        memset(array.ptr, 0, array.size);
        for (int n = 0; n < array.size; n++) {
                arrayCheck(array, n);
                array.ptr[n] = n * n;
        }
        for (int m = 0; m < array.size; m++) {
                printf("%d\n", m);
        }
        return 0;
}
```

---

Would put here const would help? Though I guess LLVM should be able to guess it if you are only setting it at construction. Are you setting the _count in other place than the construction?

https://github.com/JetBrains/kotlin-native/blob/da7e31a10a590452dc8ee39dfff0226fb7228494/runtime/src/main/cpp/Memory.h#L240

Would help marking this as pure and inline?
https://github.com/JetBrains/kotlin-native/blob/master/runtime/src/main/cpp/Natives.h#L110

---

If can't instruct LLVM to do it. I guess you can store just after loading an array as local its size in other local and use it instead as intrinsics, there is a comment suggesting to do that in the future.

[...]

Issue #1780, opened 2018-07-09
EXC_BAD_ACCESS (code=EXC_I386_GPFLT) crash on GC using MacOS native API


soywiz:
![screen shot 2018-07-09 at 02 49 35](https://user-images.githubusercontent.com/570848/42425903-0f2d0696-8323-11e8-8b90-39e1b6764a70.png)

This branch:
https://github.com/korlibs/korlibs/tree/native.error.on.gc.using.macos.native.api

To reproduce:
```
./gradlew clean install
cd samples
../gradlew :sample1-native:compileDebugMacos_x64KotlinNative
sample1/native/build/exe/main/debug/macos_x64/sample1-native.kexe
```

olonho:
Problem look similar to https://github.com/JetBrains/kotlin-native/issues/1748 in this case on instance of com.soywiz.korag.AGNative. When walking over fields of AGNative instance field 16th, onReady we get
```
(lldb) p (void*)Konan_DebugGetFieldAddress(0x101a51738, 16)
(void *) $42 = 0x0000000101a517c8
(lldb) p (char*)Konan_DebugGetFieldName(0x101a51738, 16)
(char *) $43 = 0x00000001003cc641 "onReady"
(lldb) x/2g 0x0000000101a517c8
0x101a517c8: 0x3ff0000000000000 0x0000000101a51908
```
Looks like some other module wrote 0x3ff == 1023 on place, where `onReady` callback is expected. This likely happens due to fact that we see mismatching object layout in different libraries.

homuroll:
Fixed in master.

PR #1797, opened 2018-07-12
Bound check elimination using array.size

soywiz:
Proposal. Might have bugs. Let's wait for tests to run :)

Tested with this snippet:
```
fun loop(out: IntArray) {
        for (n in 0 until out.size) {
                out[n] = 10
        }
}

        val out = IntArray(1024)
        loop(out)
        println(out[10])
}
```

This should fix: https://github.com/JetBrains/kotlin-native/issues/1791

The idea is that K/N was generating for a:

```
for (n in 0 until array.size) X
```

a loop with (simplified):

```
val last = array.size - 1
var n = 0
do {
   X
} (n++ != last)
```

and LLVM doesn't like that.

Now I'm generating:

```
val last = array.size 
var n = 0
do {
   X
} (n++ < last)
```

And LLVM likes it and can detect that last matches the check in the bounds checking.
It still checks if the array size <= 0 before the loop starts. It should be fixed if the code is changed to:

```
val last = array.size 
var n = 0
while (n < last) {
   X
   n++
}
```

Also calling the `Kotlin_*_getArrayLength` directly + const was required for LLVM to like this.

### DIFF
--- a/runtime/src/main/kotlin/kotlin/Arrays.kt
+++ b/runtime/src/main/kotlin/kotlin/Arrays.kt
@@ -43,12 +43,12 @@ public final class ByteArray {
     // TODO: What about inline constructors?
     @InlineConstructor
     public constructor(size: Int, init: (Int) -> Byte): this(size) {
-        for (i in 0..size - 1) {
+        for (i in 0 until this.size) {
             this[i] = init(i)
         }
     }
 
-    public val size: Int
+    inline public val size: Int
         get() = getArrayLength()
 
     @SymbolName("Kotlin_ByteArray_get")
@@ -58,7 +58,8 @@ public final class ByteArray {
     external public operator fun set(index: Int, value: Byte): Unit
 
     @SymbolName("Kotlin_ByteArray_getArrayLength")
-    external private fun getArrayLength(): Int
+    @PublishedApi
+    external internal fun getArrayLength(): Int
 
     /** Creates an iterator over the elements of the array. */
     public operator fun iterator(): ByteIterator {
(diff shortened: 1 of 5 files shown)


runtime/src/main/kotlin/kotlin/Arrays.kt
```diff
@@ -146,12 +148,12 @@ public final class ShortArray {
...
     @InlineConstructor
     public constructor(size: Int, init: (Int) -> Short): this(size) {
-        for (i in 0..size - 1) {
+        for (i in 0 until this.size) {
             this[i] = init(i)
         }
     }
 
-    public val size: Int
+    inline public val size: Int
```
olonho: Better be handled by codegen, I guess, and actually may make optimizer understand what's going on easier.

sbogolepov:
Sadly, it is not working in ugly corner-cases like this one:
```kotlin
    for (i in Int.MAX_VALUE..Int.MAX_VALUE) {
        println(i)
    }
}
```
Still, the proposal is great and we will investigate ways to improve it. Suggestions are welcome :)

runtime/src/main/kotlin/kotlin/Arrays.kt
soywiz: Before doing this PR I tried to some intrinsics but I had some strange errors about a symbol not found, so maybe I did it wrong.

soywiz:
@sbogolepov Yeah :( , I have just checked and understood the problem. Though for until + step 1, this problem shouldn't happen. I have tried to do a quick change but would require some adjustments.
Maybe it could be optimized for some cases only where you know there is not going to be integer overflow.

So I'm going to close this for now.

Right now I'm waiting for this: https://github.com/JetBrains/kotlin-native/issues/1780 to be fixed and I will be able to run a performance-critical app, after that I will be able to check again and fine-tune, and I'll revisit this with more information :) and will make another PR if I manage to cover all the cases and improves the performance and code size. Next time I will also try edge cases at the bounds with several steps beforehand. :+1:

sbogolepov:
BTW, running `:backend.native:tests:run_external` is a good way to detect mistakes :)

PR #1808, opened 2018-07-17
[gradle][experimental] publish the per component pom into variant generated publications

vvlevchenko:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/experimental/internal/AbstractKotlinNativeComponent.kt
```diff
@@ -66,5 +69,8 @@ abstract class AbstractKotlinNativeComponent @Inject constructor(
         extraOpts.addAll(values.map { it.toString() })
     }
 
+    override fun pom(action: Action<MavenPom>) {
```
ilmat192: It's better to add this method in `KotlinNativeMainComponent` instead of `AbstractKotlinNativeComponent` because the main component is intended to be published unlike the abstract one.

vvlevchenko: There's `KotlinNativeTestSuite` which also required implementation of this method. I believe this part `KotlinNativeTestSuite` will change soon so I've decided to skip designing this part for test publishing and make it common for both hierarchical directions :)

PR #1820, opened 2018-07-19
CLI loading of compiler lowering plugins

sandwwraith:


sandwwraith:
Corresponding changes in upstream compiler: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2136

I also need some kind of help from Native team to load plugins from Gradle

sandwwraith:
Moved to ExpressionHelpers.kt in ir builders

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanBackendContext.kt
```diff
@@ -29,11 +29,15 @@ import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.incremental.components.NoLookupLocation
 import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.util.SymbolTable
 import org.jetbrains.kotlin.name.Name
 
 abstract internal class KonanBackendContext(val config: KonanConfig) : CommonBackendContext {
     abstract override val builtIns: KonanBuiltIns
 
+    override val symbolTable: SymbolTable
```
sandwwraith: What's the problem with it? I need it to get references to properties of class for which I generate serializer, and to reference methods and classes I have in the runtime library. (e.g. here: https://github.com/jetbrains/kotlin/blob/f0a6688dac6a8f6baa5bd2b492036d83230c7520/plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/backend/ir/SerializerIrGenerator.kt#L207). Is there better way to do this?

olonho:
Could you please rebase it on the current master to properly review it?

PR #1825, opened 2018-07-23
Gradle plugin improvements

ilmat192:


ilmat192:
We use 4.5 only to test that the plugin throws an error if a Gradle version older than a required one is used. But currently the minimal required version is 4.7 so it makes sense to change the version in this test.

PR #1827, opened 2018-07-25
RoboRIO target for FIRST Robotics Competition

kunalsheth:
Here is a video of the example in action: https://youtu.be/40QiPb88Hgw

Note: While the example directly uses the WPILib [HAL](https://github.com/wpilibsuite/allwpilib#structure-and-organization), actual FRC teams would need a "frontend" like [WPILibJ](https://github.com/wpilibsuite/allwpilib#structure-and-organization) as well as the [NetworkTables](https://github.com/wpilibsuite/ntcore) and [Phoenix](https://github.com/CrossTheRoadElec/Phoenix-frc-lib) libraries.

### DIFF
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -200,6 +200,44 @@ abiSpecificLibraries.linux_arm32_hfp = \
     lib/arm-linux-gnueabihf \
     usr/lib/arm-linux-gnueabihf
 
+# FIRST Robotics Competition
+targetToolchain-macos_x64-frc_arm32.default = \
+  /usr/local/arm-frc-linux-gnueabi
+libffi-3.2.1-frc_arm32.default = \
+  /Users/kunalsheth/Documents/konanRIO/libffi-3.2.1-frc_arm32
+target-sysroot-frc_arm32.default = \
+  /usr/local/arm-frc-linux-gnueabi
+target-gcc-toolchain-macos_x64.default = \
+  /usr/bin/gcc
+
+gccToolchain.macos_x64 = target-gcc-toolchain-macos_x64
+targetToolchain.macos_x64-frc_arm32 = targetToolchain-macos_x64-frc_arm32
+dependencies.macos_x64-frc_arm32 = \
+    clang-llvm-5.0.0-darwin-macos \
+    targetToolchain-macos_x64-frc_arm32 \
+    libffi-3.2.1-frc_arm32 \
+    target-sysroot-frc_arm32 \
+    target-gcc-toolchain-macos_x64 \
+    libffi-3.2.1-2-darwin-macos
+
+quadruple.frc_arm32 = armv7a-frc-linux-gnueabi
+entrySelector.frc_arm32 = --defsym main=Konan_main
+linkerNoDebugFlags.frc_arm32 = -S
+linkerDynamicFlags.frc_arm32 = -shared
+linkerOptimizationFlags.frc_arm32 = --gc-sections
+targetSysRoot.frc_arm32 = target-sysroot-frc_arm32
+# We could reuse host toolchain here.
+libffiDir.frc_arm32 = libffi-3.2.1-frc_arm32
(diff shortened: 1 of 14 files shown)


konan/konan.properties
```diff
@@ -200,6 +200,44 @@ abiSpecificLibraries.linux_arm32_hfp = \
     lib/arm-linux-gnueabihf \
     usr/lib/arm-linux-gnueabihf
 
+# FIRST Robotics Competition
+targetToolchain-macos_x64-frc_arm32.default = \
+  /usr/local/arm-frc-linux-gnueabi
+libffi-3.2.1-frc_arm32.default = \
+  /Users/kunalsheth/Documents/konanRIO/libffi-3.2.1-frc_arm32
```
kunalsheth: How do I fix this?

konan/konan.properties
```diff
@@ -200,6 +200,44 @@ abiSpecificLibraries.linux_arm32_hfp = \
     lib/arm-linux-gnueabihf \
     usr/lib/arm-linux-gnueabihf
 
+# FIRST Robotics Competition
+targetToolchain-macos_x64-frc_arm32.default = \
+  /usr/local/arm-frc-linux-gnueabi
```
kunalsheth: Normally, students download this from here: http://first.wpi.edu/FRC/roborio/toolchains/

konan/konan.properties
```diff
@@ -200,6 +200,44 @@ abiSpecificLibraries.linux_arm32_hfp = \
...
+libffi-3.2.1-frc_arm32.default = \
+  /Users/kunalsheth/Documents/konanRIO/libffi-3.2.1-frc_arm32
+target-sysroot-frc_arm32.default = \
+  /usr/local/arm-frc-linux-gnueabi
+target-gcc-toolchain-macos_x64.default = \
+  /usr/bin/gcc
```
kunalsheth: I am not sure why this is necessary, but it seems like [`ClangArgs.kt`](https://github.com/JetBrains/kotlin-native/blob/0da110a7869676ed0a4b18597f4c55a04d7246f0/shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt#L162) depends on it.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -270,6 +269,51 @@ open class LinuxBasedLinker(targetProperties: LinuxBasedConfigurables)
     }
 }
 
+open class FrcLinker(targetProperties: FrcConfigurables)
```
kunalsheth: It seems unnecessary to have a separate class for this, but modifying the `LinuxBasedLinker` to also support the FRC target got very messy.

olonho:
Thanks Kunal, looks interesting, however we try to avoid having too exotic platforms in the baseline. Maybe it may make sense to support it in the separate repo? Also, hardcoding absolute toolchain paths doesn't look like a good idea.

Shusek:
@olonho
Kotlin native support at this moment custom targets (like in this topic) as a plugin or some like that?  If it would be possible  you can define own platforms without touch kotlin native code.

kunalsheth:
Thanks Nikolay for taking a look at this. I understand that FRC is not Kotlin-Native's main audience, but I do think this may be a good case-study to popularize Kotlin to the almost 100,000 participating students.

Regarding the hardcoded toolchain path, I’m guessing that would be solved by uploading the toolchain to the `dependenciesUrl` ([download.jetbrains.com/kotlin/native](url)) but I’m not sure how I could do it otherwise.

kunalsheth:
Correct me if I’m wrong, but I don’t think Kotlin-Native has custom targets right now. So I’ld probably have to make users clone my fork, `./gradlew bundle`, and then set the `konan.home` gradle property in their projects.
Not sure how accessible this would be.

PR #1833, opened 2018-07-30
Add libiconv to mingw platform

cy6erGn0m:
There is libiconv in mingw 

```
# unzip -l msys2-mingw-w64-x86_64-gcc-7.2.0-clang-llvm-5.0.0-windows-x86-64.zip \
    | grep -E '.*\.(so|h|a)$' | grep ......
     9292  2017-11-14 10:37   /include/iconv.h
  1085382  2017-11-14 10:37   /lib/libiconv.a
     8964  2017-11-14 10:37   /lib/libiconv.dll.a
```

### DIFF
--- a/platformLibs/src/platform/mingw/iconv.def
+++ b/platformLibs/src/platform/mingw/iconv.def
@@ -1,0 +1,5 @@
+depends = posix
+package = platform.iconv
+headers = iconv.h
+headerFilter = **
+linkerOpts = -liconv


msink:
It is defined in POSIX.1-2008, maybe just add it to `posix.def`?

cy6erGn0m:
Yes, it is according to the man page. However it provides quite specialized feature set so at least this need to be in a separate package: I wouldn't like to see it in `platform.posix` package and there are already too many declarations in it.

msink:
Ok, I was wrong - for consistency with other platforms it should be in separate `iconv.def`.
But then there should be `zlib.def`, for same reasons.

msink:
Test failed:
```
> Task :backend.native:tests:interop_libiconv FAILED
F:\src\kotlin-native-msink\test.output\local\interop_libiconv\libiconv.kt:28:26: error: unresolved reference: iconv_open
        val conversion = iconv_open("UTF-8", "LATIN1")
                         ^
F:\src\kotlin-native-msink\test.output\local\interop_libiconv\libiconv.kt:30:9: error: unresolved reference: iconv
        iconv(conversion, sourcePtr.ptr, sourceLength.ptr, destPtr.ptr, destLength.ptr)
        ^
F:\src\kotlin-native-msink\test.output\local\interop_libiconv\libiconv.kt:37:9: error: unresolved reference: iconv_close
        iconv_close(conversion)
        ^


FAILURE: Build failed with an exception.
```

Test passed if:
- rename `inconv_open` to `libiconv_open`, and so on
- link libiconv statically: `linkerOpts = -Wl,-Bstatic -liconv -Wl,-Bdynamic`

So there should be
```
---
#undef iconv_open
static inline iconv_t iconv_open(const char* tocode, const char* fromcode) {
    return libiconv_open(tocode, fromcode);
}
```
And so on for all functions.

I can do it if you want.
AFAIK mingw support in [kotlinx-io](https://github.com/Kotlin/kotlinx-io) was disabled because of that, do it is important.

olonho:
Yeah, please make such a fix, as original author is on vacations.

msink:
#1861

olonho:
Closing, merged as 1861.

PR #1844, opened 2018-08-06
[experiment] windows 32-bit target

msink:
Started exploring possibilites to add 32-bit windows target.

First, currently we have on windows one big toolchain, combining clang/llvm and gcc.
On other hosts `clang` is separated from toolchain - it's host tool, used for all targets.
So on windows it should be separated too?

Ok, but then another issue - if `gcc` toolchain separated from `clang` - we cannot use `clang++` as linker.
And how to replace it? I see possibilites:
- replace to `g++`
- replace to `ld`, as on all other platforms
- replace to `lld` from llvm.

I tried first option, but get error when compiling `tetrts`:

```
error: f:\Work\.konan\dependencies\msys2-mingw-w64-x86_64-gcc-7.3.0-clang-llvm-lld-6.0.1\bin\g++ invocation reported errors
```

And no other info. How to debug what it want?
Or maybe I'm on wrong way?


msink:
will wait for `llvm_c_api` finished

Issue #1811, opened 2018-07-18
exception: java.lang.AssertionError: Assertion failed


chinacharis:
fun String.Companion.format(format: String, args: Array<Any>): String {
    var buffer = StringBuilder()

    return buffer.toString()
}

fun String.Companion.format(format: String, value: String): String {
    var buffer = StringBuilder()

    return buffer.toString()
}

val sql = String.format("SELECT updatetime from %s where account=\'%s\'", arrayOf<Any>("uinfo", "zhua"))

lead this exception:
Information:Building target 'testsdk' for architecture 'arm64' using 'iOS 11.4' sdk
Information:Running script 'Compile Kotlin/Native'
> Configure project :
feature GRADLE_METADATA is not enabled: publication is disabled
> Task :checkKonanCompiler
> Task :compileKonanTestsdkIos_arm64
Error:compilation failed: Assertion failed
 * Source files: AObserve.kt, AuthServiceObserver.kt, Base64.kt, Decoder.kt, Encoder.kt, Observer.kt, StringEx.kt, UcSTARSystem.kt
 * Compiler version info: Konan: 0.8 / Kotlin: 1.2.70
 * Output kind: FRAMEWORK
exception: java.lang.AssertionError: Assertion failed
	at org.jetbrains.kotlin.backend.konan.llvm.LlvmUtilsKt.importGlobal(LlvmUtils.kt:219)
	at org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator.getObjectInstanceStorage(CodeGenerator.kt:64)
	at org.jetbrains.kotlin.backend.konan.llvm.FunctionGenerationContext.getObjectValue(CodeGenerator.kt:698)
	at org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportCodeGeneratorKt.createObjectInstanceAdapter(ObjCExportCodeGenerator.kt:1048)
	at org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportCodeGeneratorKt.createTypeAdapter(ObjCExportCodeGenerator.kt:950)
	at org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportCodeGeneratorKt.access$createTypeAdapter(ObjCExportCodeGenerator.kt:1)
[...]

PR #1851, opened 2018-08-08
Fix for #1811

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -1045,7 +1045,9 @@ private fun ObjCExportCodeGenerator.createObjectInstanceAdapter(
     return generateObjCToKotlinSyntheticGetter(selector) {
         initRuntimeIfNeeded() // For instance methods it gets called when allocating.
 
-        val value = getObjectValue(context.ir.get(descriptor), shared = false, locationInfo = null, exceptionHandler = ExceptionHandler.Caller)
+        val irClass = context.ir.get(descriptor)
+        val shared = irClass.symbol.objectIsShared && context.config.threadsAreAllowed
```
SvyatoslavScherbina: What's the reason to have `shared` parameter in `getObjectValue` function instead of computing this property in the function's body?

PR #1853, opened 2018-08-08
[installation] docker image with kotlin native compiler

vvlevchenko:


tools/docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:18.04
+RUN apt-get update
+RUN apt-get install -y wget
+RUN apt-get install -y default-jdk
+ENV version 0.8.1
+ENV kotlin_native kotlin-native-linux-${version}
+RUN wget https://download-cf.jetbrains.com/kotlin/native/builds/releases/${version}/linux/${kotlin_native}.tar.gz
```
olonho: Is `wget` always there?

StefMa: What do you mean with `always there`?
In Line 3 `wget` will be installed (if not already shipped by default in ubuntu what I guess).

StefMa:
I don't know what is planned with that Dockerfile.
But wouldn't it be good to add a small "README" to it?

PR #1854, opened 2018-08-09
Support unsigned types

SvyatoslavScherbina:
To be merged after #1840 and updating to 1.3.
Migrating C interop to unsigned types is to be added later as a separate PR.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -1091,11 +1104,15 @@ private fun objCFunctionType(context: Context, methodBridge: MethodBridge): LLVM
 
 private val ObjCValueType.llvmType: LLVMTypeRef get() = when (this) {
     ObjCValueType.BOOL -> int8Type
+    ObjCValueType.UNICHAR -> int16Type
```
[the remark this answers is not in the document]
SvyatoslavScherbina: Kotlin compiler appears to optimize `when` over enum entries.

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
```diff
@@ -237,6 +237,22 @@ external private fun copyImpl(array: IntArray, fromIndex: Int,
 external private fun copyImpl(array: LongArray, fromIndex: Int,
                               destination: LongArray, toIndex: Int, count: Int)
 
+@SymbolName("Kotlin_ByteArray_copyImpl")
```
olonho: Maybe add comment that operation is bitwise identical to signed types?

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
```diff
@@ -279,6 +295,30 @@ fun LongArray.copyRangeTo(destination: LongArray, fromIndex: Int, toIndex: Int,
     copyImpl(this, fromIndex, destination, destinationIndex, toIndex - fromIndex)
 }
 
+@SinceKotlin("1.3")
+@ExperimentalUnsignedTypes
```
olonho: What's the point of this annotation?

runtime/src/main/kotlin/kotlin/collections/Arrays.kt
```diff
@@ -972,6 +1060,54 @@ public actual fun LongArray.contentHashCode(): Int {
...
 
+/**
+ * Returns a hash code based on the contents of this array as if it is [List].
+ */
+@SinceKotlin("1.3")
+@ExperimentalUnsignedTypes
+public fun UByteArray.contentHashCode(): Int {
+    var result = 1
+    for (element in this)
+        result = 31 * result  + element.hashCode()
```
olonho: Why call `element.hashCode()` and not just `element`?

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
SvyatoslavScherbina: It is recommended to have one on any declaration that uses unsigned types. That's how Kotlin experimental features are organized.

runtime/src/main/kotlin/kotlin/collections/Arrays.kt
SvyatoslavScherbina: 1) Kotlin doesn't have mixed signed-unsigned operations.
2) All other variants of this function call hashCode, including one for ints.

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
olonho: Not sure if it makes any sense for Native.

SvyatoslavScherbina: I'm sure that it makes sense for Native.

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
SvyatoslavScherbina: Done

PR #1862, opened 2018-08-10
[codegen] added konan.internal.NoReorderFields to disable field reordering

vvlevchenko:


runtime/src/main/kotlin/konan/worker/Atomics.kt
```diff
@@ -123,6 +124,7 @@ external private fun checkIfFrozen(ref: Any?)
  * Otherwise memory leak could happen.
  */
 @Frozen
+@NoReorderFields
```
olonho: Also add this on WeakReferenceCounter in WeakPrivate.kt, it follows the same pattern.

vvlevchenko: done

PR #1875, opened 2018-08-13
[kotlin compiler][update] 1.2.70-dev-1218

vvlevchenko:


runtime/src/main/kotlin/konan/internal/KonanRuntimeTypes.kt
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package konan.internal
+
+
+internal class FunctionReference
```
olonho: Why it is needed?

vvlevchenko: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt#L184

PR #1878, opened 2018-08-14
Make lazy freezing-aware.

olonho:


runtime/src/main/kotlin/konan/worker/Lazy.kt
```diff
@@ -0,0 +1,71 @@
...
+                } else {
+                    do {
+                        result = valueFrozen_.get()
+                    } while (result === INITIALIZING)
+                }
+                assert(result !== UNINITIALIZED && result !== INITIALIZING)
+                @Suppress("UNCHECKED_CAST")
+                return result as T
+            } else {
+                if (result === UNINITIALIZED_VALUE) {
```
SvyatoslavScherbina: What is `UNINITIALIZED_VALUE`? Where is it declared? How is it different from `UNINITIALIZED`?

runtime/src/main/kotlin/konan/worker/Lazy.kt
```diff
@@ -0,0 +1,71 @@
...
+                    result = initializer_!!()
```
SvyatoslavScherbina: What happens if `initializer_()`  freezes this object?

runtime/src/main/kotlin/konan/worker/Lazy.kt
```diff
@@ -0,0 +1,71 @@
...
+            if (isFrozen) {
+                if (result !== UNINITIALIZED) {
+                    assert(result !== INITIALIZING)
+                    @Suppress("UNCHECKED_CAST")
+                    return result as T
+                }
+                // Barrier.
+                do {
+                    result = valueFrozen_.get()
+                } while (result === INITIALIZING)
```
SvyatoslavScherbina: In case of recursive initialization we seem to get infinite loop here.

olonho: Yes, as designed.

runtime/src/main/kotlin/konan/worker/Lazy.kt
olonho: Typo.

runtime/src/main/kotlin/konan/worker/Lazy.kt
olonho: Next statement will throw mutability exception. What do you suggest - go frozen path?

runtime/src/main/kotlin/konan/worker/Lazy.kt
SvyatoslavScherbina: Hung program is harder to debug than crashed one.

runtime/src/main/kotlin/konan/worker/Lazy.kt
SvyatoslavScherbina: At least we can throw more understandable error for now.

olonho: OK.

runtime/src/main/kotlin/konan/worker/Lazy.kt
olonho: I have quite an opposite opinion, if debugger is available. Anyway, what do you suggest?

SvyatoslavScherbina: I suggest to think a little about user-friendly handling for this case.

Btw, spin lock doesn't seem to be most suitable mutex implementation for this class. On the other hand, we are still missing proper lock implementations in our stdlib, which would be useful here and in many other cases.

PR #1890, opened 2018-08-15
move Kotlin/Native metadata (.proto file + generated classes) to a separate module

ddolovov:
move Kotlin/Native metadata (.proto file + generated classes) to a separate module

### DIFF
--- a/extracted/konan.metadata/src/ext_options.proto1
+++ b/extracted/konan.metadata/src/ext_options.proto1
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2010-2015 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


extracted/konan.metadata/src/ext_options.proto1
```diff
@@ -0,0 +1,29 @@
+/*
```
SvyatoslavScherbina: I believe this file is copied from `backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/`.
Is the original version still required to be in place?

ddolovov: I am planning to drop all *.proto1 files in Kotlin/Native project (both that lay in `backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/` directory and those that were copied to `extracted/konan.metadata/src`) after transferring `extracted` to Big Kotlin repo.

Because all these files are actually copies of the corresponding *.proto files in Big Kotlin project. And I believe they should not remain in Kotlin/Native repo.

PR #1891, opened 2018-08-16
[mingw] samples/tetris: static link

msink:


samples/tetris/build.bat
```diff
@@ -9,12 +9,11 @@ set MINGW=\msys64\mingw64
 
 set "CFLAGS=-I%MINGW%\include\SDL2"
 rem Add -Wl,--subsystem,windows for making GUI subsystem application.
-set "LFLAGS=%DIR%\Tetris.res -L%MINGW%\lib -lSDL2"
+set "LFLAGS=%DIR%\Tetris.res -L%MINGW%\lib -Wl,-Bstatic -lstdc++ -static -lSDL2 -limm32 -lole32 -loleaut32 -lversion -lwinmm -mwindows"
```
olonho: Does Gradle build produce dynamically linked binary?

msink: Gradle part was merged as part of LLVM6 - so now it produces static binary, but still copy unused SDL2.dll

PR #1892, opened 2018-08-16
Add ContainsRangeLowering

soywiz:
This optimises (containing in `Byte/Char/Short/Int/Long` ranges):
* `a in (b until c)`
* `a in (b .. c)`

Based on this profiling:
screen shot 2018-08-16 at 15 11 24

Notes:

I think that the higher part is not evaluated (short-circuited) if the first part is not true. Should I put low and high in temps too so they get computed always?
Is there a `irAnd` intrinsic for "&&"? (couldn't find one)


olonho:
What are the performance improvements?

olonho:
I mean what are the actual changes in benchmarks, like `./gradlew perf`, or if there are none, please extend perf with benchmark showing the benefits.

soywiz:
```
MASTER:
Euler.problem9 : absolute = 1.0358 us +- 0.0097 us, ratio = 1.41 +- 0.02
Euler.problem8 : absolute = 477.3414 us +- 3.1223 us, ratio = 47.16 +- 0.59

BRANCH:
Euler.problem9 : absolute = 0.8347 us +- 0.0042 us, ratio = 1.14 +- 0.01
Euler.problem8 : absolute = 295.1179 us +- 2.1550 us, ratio = 29.37 +- 0.39
```

Though with several runs yielded several results, the branch seems to be faster always.

Are you plotting this somewhere per commit?

It probably optimize these:
* https://github.com/JetBrains/kotlin-native/blob/ac3cdd5c8d1900a4c8ad512d4d5e35a90975e896/performance/src/main/kotlin/org/jetbrains/ring/EulerBenchmark.kt#L77
* https://github.com/JetBrains/kotlin-native/blob/ac3cdd5c8d1900a4c8ad512d4d5e35a90975e896/performance/src/main/kotlin/org/jetbrains/ring/EulerBenchmark.kt#L119

The other thing that I have noticed that is not optimized is: `for (item in array_or_list)` that could be converted into `for (n in 0 until array_or_list.size) { val item = array_or_list[n]; ... }` and not sure if `for (n in array_or_list.indices)` is optimized either.

backend.native/tests/build.gradle
```diff
@@ -2027,6 +2027,11 @@ task range0(type: RunKonanTest) {
     source = "runtime/collections/range0.kt"
 }
 
+task range1(type: RunKonanTest) {
+    goldValue = "ABCfalse\nABCtrue\n"
```
sbogolepov: Order of side-effects should be "BCA".

soywiz: Makes sense. Fixed.

soywiz:
I'm closing this, since this optimization should be done in a way that works for all the targets

PR #1901, opened 2018-08-18
Fix kotlin.system.getTimeMillis()

soywiz:
- high_resolution_clock cannot be used with time_since_epoch

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -513,6 +513,11 @@ task runtime_worker_random(type: RunKonanTest) {
     source = "runtime/basic/worker_random.kt"
 }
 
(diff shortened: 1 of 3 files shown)


soywiz:
```
➜  bin cat demo.cpp
#include <chrono>
#include <iostream>

using namespace std::chrono;

int main() {
	std::cout << duration_cast<milliseconds>(high_resolution_clock::now().time_since_epoch()).count() << "\n";
	std::cout << duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count() << "\n";
	return 0;
}
➜  bin g++ demo.cpp && ./a.out
108959069
1534593679776
```

olonho:
Providing guaranteed base for counters is still not something we could guarantee. Closing for now.

PR #1906, opened 2018-08-21
Freeze top-level val.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -292,8 +292,11 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
         LLVMBuildStore(builder, value, ptr)
     }
 
-    fun storeAny(value: LLVMValueRef, ptr: LLVMValueRef) {
+    fun storeAny(value: LLVMValueRef, ptr: LLVMValueRef, shallFreeze: Boolean = false) {
         if (isObjectRef(value)) {
+            if (shallFreeze) {
+                call(context.llvm.freezeSubgraph, listOf(value),  Lifetime.IRRELEVANT, ExceptionHandler.Caller)
```
SvyatoslavScherbina: Proper context-dependent `ExceptionHandler` must be used here instead.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -292,8 +292,11 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
         LLVMBuildStore(builder, value, ptr)
     }
 
-    fun storeAny(value: LLVMValueRef, ptr: LLVMValueRef) {
+    fun storeAny(value: LLVMValueRef, ptr: LLVMValueRef, shallFreeze: Boolean = false) {
```
SvyatoslavScherbina: `shallFreeze` seems to be not that closely related to storing a value to memory. It makes some sense to provide `freeze` function here and call it where required.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -51,6 +51,9 @@ private val threadLocalAnnotationFqName = FqName("kotlin.native.ThreadLocal")
 val IrClassSymbol.objectIsShared get() =
     !descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName)
 
+val IrFieldSymbol.isShared get() =
```
SvyatoslavScherbina: We should avoid using `Ir*Symbol` classes in signatures and use IR declaration classes instead.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -398,7 +404,21 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+                                }
+                            }
+                    ret(null)
+                }
+
+                appendingTo(bbLocalInit) {
+                    context.llvm.fileInitializers
+                            .forEach {
+                                if (it.initializer?.expression !is IrConst<*>?) {
+                                   if (!it.symbol.isShared) {
```
SvyatoslavScherbina: I believe the same check is required when generating `bbLocalDeinit`.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -78,10 +79,14 @@ RuntimeState* initRuntime() {
...
+  bool firstRuntime = atomicAdd(&aliveRuntimesCount, 1) == 1;
   // Keep global variables in state as well.
-  InitOrDeinitGlobalVariables(INIT_GLOBALS);
-  konan::consoleInit();
-  atomicAdd(&aliveRuntimesCount, 1);
+  if (firstRuntime) {
+    konan::consoleInit();
+    InitOrDeinitGlobalVariables(INIT_GLOBALS);
+  } else {
+    InitOrDeinitGlobalVariables(INIT_THREAD_LOCAL_GLOBALS);
```
SvyatoslavScherbina: `INIT_THREAD_LOCAL_GLOBALS` appears to be asymmetric with `DEINIT_THREAD_LOCAL_GLOBALS`:
the former is not called for the first runtime, while the latter is called for the last runtime.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: Yep.

PR #1914, opened 2018-08-21
[compiler][update] 1.3-M2-eap-82 / build: 1.3-M2-eap-35

vvlevchenko:


runtime/src/main/cpp/KString.cpp
```diff
@@ -737,6 +737,26 @@ KInt Kotlin_String_compareTo(KString thiz, KString other) {
   return diff < 0 ? -1 : 1;
 }
 
+KInt Kotlin_String_compareToIgnoreCase(KString thiz, KConstRef other) {
+  RuntimeAssert(thiz->type_info() == theStringTypeInfo &&
+                other->type_info() == theStringTypeInfo, "Must be strings");
+  // Important, due to literal internalization.
+  KString otherString = other->array();
+  if (thiz == otherString) return 0;
+  if (thiz->count_ > otherString->count_) return 1;
```
olonho: Comparison is lexicographic, length is not so important if common prefix is different.

PR #1917, opened 2018-08-22
KLIB reader: better isolation of implementation details

ddolovov:


utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/InteropCompiler.kt
```diff
@@ -72,12 +74,11 @@ fun invokeInterop(flavor: String, args: Array<String>): Array<String> {
     )
 
     val importArgs = allLibraries.flatMap {
-        val manifestProperties = it.manifestProperties
         // TODO: handle missing properties?
-        manifestProperties["package"]?.let {
-            val pkg = it as String
-            val headerIds = (manifestProperties["includedHeaders"] as String).split(' ')
-            val arg = "$pkg:${headerIds.joinToString(";")}"
+        val libraryReader = it
```
SvyatoslavScherbina: Maybe specify lambda parameter name explicitly?
```
val importArgs = allLibraries.flatMap { libraryReader ->
```

ddolovov: Will fix

extracted/konan.serializer/src/org/jetbrains/kotlin/konan/library/KonanLibrary.kt
```diff
@@ -1,44 +1,64 @@
...
+    val bitcodePaths: List<String>
+    val includedPaths: List<String>
+
+    val targetList: List<String>
+
+    val dataFlowGraph: ByteArray?
+    val moduleHeaderData: ByteArray
+    fun packageMetadata(fqName: String): ByteArray
+
+    // FIXME: ddol: to be refactored into some global resolution context
```
olonho: Better use `FIXME(ddol): what` syntax.

ddolovov: OK. I will fix this with a separate commit.

PR #1920, opened 2018-08-23
Fix inlining of local objects.

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DeepCopyIrTreeWithDescriptors.kt
```diff
@@ -784,6 +784,34 @@ internal class DeepCopyIrTreeWithDescriptors(val targetDescriptor: FunctionDescr
 
 class SubstitutedDescriptor(val inlinedFunction: FunctionDescriptor, val descriptor: DeclarationDescriptor)
 
+/**
+ * Searches recursively for value that is corresponds to [key],
+ * then to typeMapper(this[key]) and so on until typeMapper or Map.get returns null.
+ * Returns null if there is no corresponding value for given [key] inside the map
+ * Gives control and latest value to [cycleHandler] if recursion led to cycle.
+ */
+private fun <K, V> Map<K, V>.deepSearch(key:K?, typeMapper: (V) -> K?,
```
SvyatoslavScherbina: Why is this function not `inline`?

sbogolepov: Compiler will fail:
```
e: java.lang.IllegalStateException: Backend Internal error: Exception during code generation
Cause: Back-end (JVM) Internal error: Couldn't inline method call 'deepSearch' into...
```
Yeah, Kotlin/JVM Inliner will fail to inline part of Kotlin/Native's inliner :)

PR #1922, opened 2018-08-23
Top level values can be only accessed from the main thread.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -49,12 +50,27 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.classId
...
+    descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName)
+
 val IrField.isShared get() =
-    !descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName) && !descriptor.isVar
+    descriptor.annotations.hasAnnotation(sharedAnnotationFqName)
+
+val IrField.isMainOnly get() =
+    !descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName) &&
+    !descriptor.annotations.hasAnnotation(sharedAnnotationFqName) &&
+    !descriptor.isDelegated
```
olonho: Is such a check actually correct?

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -82,6 +83,7 @@ RuntimeState* initRuntime() {
   bool firstRuntime = atomicAdd(&aliveRuntimesCount, 1) == 1;
   // Keep global variables in state as well.
   if (firstRuntime) {
+    isMainThread = 1;
```
SvyatoslavScherbina: What if Kotlin runtime gets initialized on non-main thread first?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -49,12 +50,27 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.classId
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameUnsafe
 
 private val threadLocalAnnotationFqName = FqName("kotlin.native.ThreadLocal")
+private val sharedAnnotationFqName = FqName("kotlin.native.SharedImmutable")
 
+// TODO: maybe unannotated singleton objects shall be accessed from main thread only as well?
 val IrClass.objectIsShared get() =
     !descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName)
 
+val IrField.isThreadLocal get() =
```
SvyatoslavScherbina: Are this predicates supposed to be mutually exclusive?
If so, then it makes sense to make a enum-returning function instead.

runtime/src/main/cpp/Runtime.cpp
olonho: It is impossible with this definition of the main thread - main thread *is* the first thread which initialized the runtime. It's up to platform-specific code to ensure that Kotlin/Native main thread matches to GUI thread or whatever is the the main thread in other senses.

SvyatoslavScherbina: > It's up to platform-specific code to ensure that Kotlin/Native main thread matches to GUI thread

How do you suggest to implement this on iOS then?

olonho: Call properly in launcher for direct interop, let's discuss how to do for reverse interop. My gut feeling is that somewhere near https://github.com/JetBrains/kotlin-native/blob/9b2fc6c6dddfa9e749eea21dce8f3fb79a34b1a4/runtime/src/main/cpp/ObjCExport.mm#L418 may help.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: Hmm, not sure if they must be mutually exclusive, they are more like computations of expected policy of accessing the field.

PR #1928, opened 2018-08-24
 [compiler][update] 1.3-M2-eap-105

vvlevchenko:


vvlevchenko:
I've just included for checking, now PRs could be applied separately.

PR #1933, opened 2018-08-25
Fix widening signed integer to unsigned implicitly and with `kotlinx.cinterop.convert`

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/Ir.kt
```diff
@@ -149,6 +149,16 @@ internal class KonanSymbols(context: Context, val symbolTable: SymbolTable, val
 
     val allIntegerClasses = signedIntegerClasses + unsignedIntegerClasses
 
+    val unsignedToSignedOfSameBitWidth = unsignedIntegerClasses.associate {
+        it to when (it) {
+            uByte -> byte
+            uShort -> short
```
olonho: Hope we don't need chat -> short conversion, do we?

SvyatoslavScherbina: We don't.

PR #1954, opened 2018-08-28
Fix for KT-26164

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -376,40 +373,40 @@ private class Inliner(val globalSubstituteMap: MutableMap<DeclarationDescriptor,
...
-                        parameterDescriptor = parameterDescriptor,
+                        parameter = parameter,
                         argumentExpression  = emptyArray
                     )
                 }
 
                 else -> {
-                    val message = "Incomplete expression: call to $functionDescriptor " +
-                        "has no argument at index ${parameterDescriptor.index}"
+                    val message = "Incomplete expression: call to $irFunction " +
```
SvyatoslavScherbina: `IrFunction.toString()` is useless, so it is better to include `functionDescriptor` into error message.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DeepCopyIrTreeWithDescriptors.kt
```diff
@@ -150,7 +150,9 @@ internal class DeepCopyIrTreeWithDescriptors(val targetDescriptor: FunctionDescr
                     /* name                  = */ generateCopyName(oldDescriptor.name),
                     /* kind                  = */ oldDescriptor.kind,
                     /* source                = */ oldDescriptor.source
-            )
+            ).also {
```
SvyatoslavScherbina: Is it still necessary? I'm not sure whether it is more correct to copy `isActual` value or not to copy.

PR #1965, opened 2018-08-29
 Enable implicit integer coercion for interop libraries

SvyatoslavScherbina:
Also:
* Improve code using interop
* Don't use fields for non-strict interop enum constants

### DIFF
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
@@ -518,15 +518,15 @@ class StubGenerator(
             it.name !in macroConstantsByName
         }
 
-        val typeName: String
+        val kotlinType: KotlinType
 
-        val baseKotlinType = mirror(e.baseType).argType.render(kotlinFile)
+        val baseKotlinType = mirror(e.baseType).argType
         if (e.isAnonymous) {
             if (constants.isNotEmpty()) {
                 out("// ${e.spelling}:")
             }
 
-            typeName = baseKotlinType
+            kotlinType = baseKotlinType
         } else {
             val typeMirror = mirror(EnumType(e))
             if (typeMirror !is TypeMirror.ByValue) {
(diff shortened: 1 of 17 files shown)


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
```diff
@@ -732,32 +732,31 @@ class StubGenerator(
...
-            val signedLiteral = integerLiteral(size, true, value)
-            val converter = when (size) {
-                1 -> "toUByte"
-                2 -> "toUShort"
-                4 -> "toUInt"
-                8 -> "toULong"
-                else -> return null
+        return if (isSigned) {
+            if (value == Long.MIN_VALUE) {
+                return "${value + 1} - 1" // Workaround for "The value is out of range" compile error.
```
olonho: :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -674,6 +674,36 @@ internal class InteropLoweringPart1(val context: Context) : IrBuildingTransforme
         }
     }
 
+    override fun visitProperty(declaration: IrProperty): IrStatement {
+        val backingField = declaration.backingField
+        return if (declaration.isConst && backingField?.isStatic == true && context.config.isInteropStubs) {
```
olonho: Maybe comment why this is needed?

backend.native/tests/interop/basics/opengl_teapot.kt
```diff
@@ -45,13 +45,13 @@ fun display() {
 
 fun initialize() {
     // select projection matrix
-    glMatrixMode(GL_PROJECTION.convert())
+    glMatrixMode(GL_PROJECTION)
```
olonho: Cool!

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
SvyatoslavScherbina: And Kotlin still don't provide a way to define `Long.MIN_VALUE` with a single literal, so this workaround is still required!

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
SvyatoslavScherbina: What do you mean by "this"? The entire transformation or the check above?

olonho: The whole transformation.

SvyatoslavScherbina: It is just an optimization and shouldn't be required by our memory model. Does it still need any comments?

olonho: It better should be commented anyway, as this is pretty nontrivial transformation.

SvyatoslavScherbina: Done?

olonho: Yep

Issue #1973, opened 2018-08-30
obj-c generics are not included in klib


ildarsharafutdinov:
I've compiled a lib which contains the following h-file: https://github.com/TextureGroup/Texture/blob/master/Source/ASViewController.h

However, there is no `ASViewController` in the klib.
Are obj-c generics supported?

SvyatoslavScherbina:
Objective-C classes with generics are generally imported as Kotlin classes without generics.
Are you sure that you have included this header into the library? Does it contain e.g. `ASDisplayTraitsForTraitCollectionBlock`?

ildarsharafutdinov:
@SvyatoslavScherbina,

I've just cleaned the cache and rebuilt the klib. Also I checked my def file and klib.

There is `ASViewController.h` in my def file, however, the klib misses both `ASViewController` and `ASDisplayTraitsForTraitCollectionBlock`.

This is pretty strange since e.g. `ASDisplayNode`, which is a part of the same lib is available in klib.

Any idea of what could go wrong?

ildarsharafutdinov:
Here is what I've found.

If I move `ASViewController.h` to the beginning of my headers list in def file, `ASViewController` is added to the klib.

Here is the def file, which results in missing `ASViewController`
```
depends = UIKit CoreText
language = Objective-C
package = pods.Texture
headers =  Texture-prefix.pch \
    ASAbsoluteLayoutElement.h ASAbsoluteLayoutSpec.h ASAbstractLayoutController.h ASAsciiArtBoxCreator.h \
    ASAssert.h ASAvailability.h ASBackgroundLayoutSpec.h ASBaseDefines.h ASBasicImageDownloader.h \
    ASBatchContext.h ASBatchFetchingDelegate.h ASBlockTypes.h ASButtonNode.h ASCGImageBuffer.h \
    ASCellNode.h ASCenterLayoutSpec.h ASCollectionElement.h ASCollectionFlowLayoutDelegate.h \
    ASCollectionGalleryLayoutDelegate.h ASCollectionLayoutContext.h \ 
    ASCollectionLayoutDelegate.h ASCollectionLayoutState.h ASCollectionNode+Beta.h ASCollectionNode.h \
    ASCollectionView.h ASCollectionViewLayoutController.h ASCollectionViewLayoutFacilitatorProtocol.h \
    ASCollectionViewLayoutInspector.h ASCollectionViewProtocols.h ASConfiguration.h ASConfigurationDelegate.h \
[...]

PR #1988, opened 2018-09-01
Mark KClass for interop classes unimplemented explicitly

SvyatoslavScherbina:
Also fix https://github.com/JetBrains/kotlin-native/issues/1973#issuecomment-417634981

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DelegationLowering.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DelegationLowering.kt
@@ -9,11 +9,14 @@ import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irBlock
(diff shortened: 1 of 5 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DelegationLowering.kt
```diff
@@ -325,22 +330,43 @@ internal fun IrBuilderWithScope.irKType(context: KonanBackendContext, type: IrTy
...
     }
 }
+
+internal fun IrBuilderWithScope.irKClass(context: KonanBackendContext, symbol: IrClassifierSymbol): IrExpression {
+    val symbols = context.ir.symbols
+    return when {
+        symbol !is IrClassSymbol -> // E.g. for `T::class` in a body of an inline function itself.
+            irCall(symbols.ThrowNullPointerException.owner)
+
+        symbol.descriptor.isObjCClass() ->
```
olonho: What will happen if we subclass Obj-C base class with Kotlin, shall it still fail like that?

SvyatoslavScherbina: Yes, this branch covers the case with Kotlin subclass of Objective-C class, and this behaviour is intentional as current implementation of `::class` for these classes is incorrect.

PR #1996, opened 2018-09-03
[gradle-plugin] Support custom entry points

ilmat192:


StefMa:
Why do you provide a public var and a setter method? 🤔
People can set the var directly...

ilmat192:
@StefMa Just to allow doing this in both ways and make switching from the old DSL easier (a setter method is used there) :)

PR #2001, opened 2018-09-04
Close zip filesystems on exit.

alexander-gorshenev:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/file/ZipUtil.kt
```diff
@@ -0,0 +1,48 @@
...
+        FileSystems.getFileSystem(zipUri)
+    }
+}
+
+fun File.mutableZipFileSystem() = this.zipFileSystem(mutable = true)
+
+fun File.zipPath(path: String): Path
+    = this.zipFileSystem().getPath(path)
+
+val File.asZipRoot: File
```
ddolovov: `asZipRoot` may lead to a very long IO-section, because you don't know how far from `asZipRoot` will `close()` happen. I would suggest to replace ```val File.asZipRoot:
    get() = File(this.zipPath("/"))``` by ```fun File.withZipRoot(action: (File) -> Unit) {
	val zipRoot = File(this.zipPath("/"))
	try {
	    action(zipRoot)
	} finally {
        zipRoot.javaPath.fileSystem.close()
    }
}```. This way it could be possible to have IO-section strictly limited to the time of `action` execution.

Similarly for `asWritableZipRoot`. What do you think?

alexander-gorshenev: good idea

PR #2002, opened 2018-09-04
Fixed Torch and Python samples

sbogolepov:


samples/torch/src/main/kotlin/Dataset.kt
```diff
@@ -28,7 +28,7 @@ object MNIST {
 
         val file = fopen(path, "rb") ?: fail()
         try {
-            ByteArray(size).also { fread(it.refTo(0), 1, size.signExtend(), file) }
+            ByteArray(size).also { fread(it.refTo(0), 1, size.toULong(), file) }
```
olonho: Maybe better use `convert` here, as on some platforms fread may accept 32-bit value, in theory.

PR #2016, opened 2018-09-06
more fixed for documentation files

jonnyzzz:
More fixes to make kotlin-web-site compile
those changes breaks links between `.md` files on GitHub


OBJC_INTEROP.md
```diff
@@ -237,5 +237,5 @@ this library would disable these compiler checks.
 
 ## C features
 
-See [INTEROP.md](INTEROP.md) for the case when library uses some plain C features
+See [INTEROP.md](c_interop.html) for the case when library uses some plain C features
```
olonho: There's no c_interop.html and other HTML files. .md files shall be self-contained, and links better be inserted on site side.

jonnyzzz:
I made the web-site generator smarter to allow links to local md files that are included into the `_nav_reference.yml`
Reverted changes to add the `---` header as well

PR #2029, opened 2018-09-10
Mark few JVM-compatibility declarations deprecated.

olonho:


runtime/src/main/kotlin/kotlin/native/Compatibility.kt
```diff
@@ -11,12 +11,15 @@ package kotlin
 
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.FIELD)
 @Retention(AnnotationRetention.SOURCE)
+@Deprecated("@Volatile is meaningless in Native", level = DeprecationLevel.WARNING)
```
SvyatoslavScherbina: Btw, we already have these annotations in common stdlib with `@OptionalExpectation`:
https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/common/src/kotlin/JvmAnnotationsH.kt#L113

This is supposed to make these declarations redundant: common code with such an annotation would be properly compiled with Native backend. And these declarations can only be required to compile Kotlin/JVM code (which has default `kotlin.jvm.*` import) with Kotlin/Native, which doesn't seem to be a case we'd like to support, in light of Kotlin Multiplatform.

ilya-g:
Consider unifying wording with the similar deprecations in JS stdlib: https://github.com/JetBrains/kotlin/commit/50f75ade6405af33a0cb7f22d5364a9d6c10c26f

runtime/src/main/kotlin/kotlin/native/Compatibility.kt
ilya-g: If you don't pursue the full deprecation cycle, you can raise the deprecation level straight to ERROR for this annotation.

PR #2033, opened 2018-09-10
Add basic checks for illegal object sharing in Objective-C interop

SvyatoslavScherbina:


runtime/src/main/kotlin/kotlin/native/concurrent/Internal.kt
```diff
@@ -69,11 +70,22 @@ internal fun ThrowFreezingException(toFreeze: Any, blocker: Any): Nothing =
 
 @ExportForCppRuntime
 internal fun ThrowInvalidMutabilityException(where: Any): Nothing {
-    val kClass = where::class
+    val description = debugDescription(where::class, where.identityHashCode())
+    throw InvalidMutabilityException("mutation attempt of frozen $description")
+}
+
+@ExportForCppRuntime
+internal fun ThrowIllegalObjectSharingException(typeInfo: NativePtr, address: NativePtr) {
```
olonho: `: Nothing`

SvyatoslavScherbina: No, because with `: Nothing` its C++ signature becomes too complex.

PR #2034, opened 2018-09-11
Split serialized package fragments metadata

semoro:
This mainly required to reduce IDE footprint
Also, this change improves size of Kotlin Native Dist
as it recreates string & names table per each package part


olonho:
How does it affect compilation performance and bundle size?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanSerializationUtil.kt
```diff
@@ -30,34 +31,56 @@ import org.jetbrains.kotlin.serialization.KonanDescriptorSerializer
...
  */
 
-internal class KonanSerializationUtil(val context: Context, metadataVersion: BinaryVersion) {
+internal class KonanSerializationUtil(val context: Context, private val metadataVersion: BinaryVersion) {
 
-    val serializerExtension = KonanSerializerExtension(context, metadataVersion)
+    /*val serializerExtension = KonanSerializerExtension(context, metadataVersion)
     val topSerializer = KonanDescriptorSerializer.createTopLevel(context, serializerExtension)
-    var classSerializer: KonanDescriptorSerializer = topSerializer
+    var classSerializer: KonanDescriptorSerializer = topSerializer*/
```
alexander-gorshenev: This is garbage.

olonho:
Merged as Dima's PR.

PR #2036, opened 2018-09-11
Lazy stacktrace

sbogolepov:


runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -57,20 +57,33 @@ class AutoFree {
   }
 };
 
+// Backtrace is an array of pointers.
+// So we store pointers inside IntArray on 32-bit platforms
+// and inside LongArray on 64-bit platforms.
+template <class T>
+const TypeInfo* getStackTraceTypeInfo() {
+  if (sizeof(T) == sizeof(KInt)) {
```
olonho: Please don't, better always use LongArray.

runtime/src/main/kotlin/kotlin/Throwable.kt
```diff
@@ -47,5 +52,10 @@ public open class Throwable(open val message: String?, open val cause: Throwable
     }
 }
 
+// We don't care about return type because it's only matter to
+// `Kotlin_getStackTraceStrings` implementation.
 @SymbolName("Kotlin_getCurrentStackTrace")
-private external fun getCurrentStackTrace(): Array<String>
\ No newline at end of file
+private external fun getCurrentStackTrace(): Any
```
olonho: No longer `Any`.

PR #2037, opened 2018-09-11
 Update Kotlin with tests to 1.3.0-dev-391

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/DeepCopyIrTreeWithDescriptors.kt
```diff
@@ -366,7 +368,7 @@ internal class DeepCopyIrTreeWithDescriptors(val targetDescriptor: FunctionDescr
                         /* outType                   = */ substituteType(oldDescriptor.type)!!,
                         /* typeParameters            = */ oldDescriptor.typeParameters,
                         /* dispatchReceiverParameter = */ (containingDeclaration as ClassDescriptor).thisAsReceiverParameter,
-                        /* receiverType              = */ substituteType(oldDescriptor.extensionReceiverParameter?.type))
+                        /* extensionReceiverParamter = */ substituteType(oldDescriptor.extensionReceiverParameter?.type).createExtensionReceiver(this))
```
olonho: Maybe in Kotlin style, without comments?

SvyatoslavScherbina: The called method is written in Java, so this doesn't seem to be possible.

PR #2040, opened 2018-09-11
Support OptionalExpectation

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KonanToolRunner.kt
```diff
@@ -22,6 +22,7 @@ import org.gradle.api.file.FileCollection
 import org.jetbrains.kotlin.konan.target.HostManager
 import org.jetbrains.kotlin.konan.target.KonanTarget
 import org.jetbrains.kotlin.konan.util.DependencyProcessor
+import java.io.File
```
olonho: Why not org.jetbrains.kotlin.konan.file.File, and createTempFile from there?

PR #2041, opened 2018-09-11
Limit number of Gradle workers

ilmat192:


StefMa:
Why do you do this? 🤔
What issues are you facing to limit the workers?

Issue #1841, opened 2018-08-06
Imported Swift framework can’t be used in Kotlin Multiplatform even with @objc attribute.


VitMichal:
I'm trying Kotlin Multiplatform: https://kotlinlang.org/docs/reference/multiplatform.html
In platform module I would like to use custom made swift framework. 

I've created swift framework with class `WrapperCompletable`. 
Swift framework source code:
```swift
import Foundation

@objc public class WrapperCompletable: NSObject { }
```
Regarding to [https://github.com/JetBrains/kotlin-native/blob/master/OBJC_INTEROP.md#kotlinnative-interoperability-with-swiftobjective-c](url) documentation: “Swift library can be used in Kotlin code if its API is exported to Objective-C with @objc”. 

This framework is included in gradle script.

Gradle script:
```gradle
konanArtifacts {
    def frameworksDir = "${projectDir}/Frameworks"
    def productsDir = frameworksDir
    def helperLibDir = frameworksDir

    interop("RxSwiftWrapperFramework") {
        defFile "src/main/c_interop/RxSwiftWrapperFramework.def"

        compilerOpts "-F${productsDir}", "-F${helperLibDir}"
        linkerOpts "-F${productsDir}", "-F${helperLibDir}"

        includeDirs "${frameworksDir}/RxSwiftWrapperFramework.framework/Headers"
    }

    // Declare building into a framework.
    framework('Multiweather') {
        // The multiplatform support is disabled by default.
        enableMultiplatform true
        extraOpts '-module_name', ''

        libraries {
            artifact "RxSwiftWrapperFramework"
        }

        linkerOpts '-rpath', "${frameworksDir}"
    }
}
```

RxSwiftWrapperFramework.def  file:
```
depends = Foundation
package = platform.RxSwiftWrapperFramework
language = Objective-C
linkerOpts = -framework RxSwiftWrapperFramework
compilerOpts = -fmodules -framework RxSwiftWrapperFramework
headers = RxSwiftWrapperFramework.h

```

There is no error at importing of platform.RxSwiftWrapperFramework. But When I try to make an instance of class from this framework I'll get error: Unresolved reference: WrapperCompletable

 source code:
```kotlin
package jh.shared.rx.infrastructure

import platform.RxSwiftWrapperFramework.*

var xxx = WrapperCompletable()
```

What should I do to make content of this framework visble in Kotlin platform module ?

SvyatoslavScherbina:
[...]

PR #2072, opened 2018-09-17
Improve interop from Kotlin to `@objc` Swift

SvyatoslavScherbina:
See https://github.com/JetBrains/kotlin-native/issues/1841#issuecomment-411346727

### DIFF
--- a/Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
+++ b/Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
@@ -122,10 +122,6 @@ internal class NativeIndexImpl(val library: NativeLibrary) : NativeIndex() {
         return library.headerToIdMapper.getHeaderId(filePath)
     }
 
-    private fun getContainingFile(cursor: CValue<CXCursor>): CXFile? {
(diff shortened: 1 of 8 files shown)


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -737,12 +743,8 @@ internal class NativeIndexImpl(val library: NativeLibrary) : NativeIndex() {
                 }
             }
 
-            CXIdxEntity_ObjCClass -> {
-                if (isAvailable(cursor) &&
-                        cursor.kind != CXCursorKind.CXCursor_ObjCClassRef /* not a forward declaration */) {
-
-                    getObjCClassAt(cursor)
-                }
+            CXIdxEntity_ObjCClass -> if (cursor.kind != CXCursorKind.CXCursor_ObjCClassRef /* not a forward declaration */) {
```
olonho: Maybe `isForwardDeclaration(cursor)`?

SvyatoslavScherbina: This is not correct, because cursor can have `CXCursor_ObjCClassRef` kind in different cases.

PR #2086, opened 2018-09-18
Allow sharing top level vals with types marked as frozen.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -68,6 +68,10 @@ internal val IrField.storageClass: FieldStorage get() {
     return when {
         descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName) -> FieldStorage.THREAD_LOCAL
         descriptor.annotations.hasAnnotation(sharedAnnotationFqName) -> FieldStorage.SHARED
+        // TODO: simplify, once IR types are fully there.
+        // TODO: check if we do have backing field indeed.
+        !descriptor.isVar && type is IrSimpleType && ((type as IrSimpleType).
```
SvyatoslavScherbina: `IrType` has suitable extension `.classifierOrNull`.

SvyatoslavScherbina: `isVar` can (and must) be checked on `IrField` itself instead.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -68,6 +68,10 @@ internal val IrField.storageClass: FieldStorage get() {
     return when {
         descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName) -> FieldStorage.THREAD_LOCAL
         descriptor.annotations.hasAnnotation(sharedAnnotationFqName) -> FieldStorage.SHARED
+        // TODO: simplify, once IR types are fully there.
+        // TODO: check if we do have backing field indeed.
```
SvyatoslavScherbina: This code already operates on `IrField`, there is no need to check whether it exists.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
SvyatoslavScherbina: By the way, how should we handle `@SharedImmutable var`?

backend.native/tests/runtime/workers/worker10.kt
```diff
@@ -0,0 +1,57 @@
+package runtime.workers.worker10
+
+import kotlin.test.*
+
+import kotlin.native.concurrent.*
+
+class Data(val x: Int)
+
+val topInt = 1
+val topString = "string"
```
SvyatoslavScherbina: It also makes sense to test for `var` `String`.

olonho: Done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: Ok.

olonho: Yeah, shared immutable var shall be disabled - done.

olonho: `type.classifierOrNull?.let {
            descriptor.annotations.hasAnnotation(frozenAnnotationFqName)} == true` doesn't work, while this code -does.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -65,13 +65,13 @@ val IrClass.objectIsShared get() =
 
 internal val IrField.storageClass: FieldStorage get() {
     val descriptor = propertyDescriptor
+    val isVal = this.isFinal
     return when {
         descriptor.annotations.hasAnnotation(threadLocalAnnotationFqName) -> FieldStorage.THREAD_LOCAL
-        descriptor.annotations.hasAnnotation(sharedAnnotationFqName) -> FieldStorage.SHARED
+        isVal && descriptor.annotations.hasAnnotation(sharedAnnotationFqName) -> FieldStorage.SHARED
```
SvyatoslavScherbina: So `isVal` check can be extracted to the separate branch above, and I believe that the entire strategy would sound closer to what we'd like to say.

PR #2093, opened 2018-09-19
Don't expose interface nested types as Swift nested types

SvyatoslavScherbina:
Since this turns out to be unsupported by Swift.

Also don't do this for nested interfaces as their status is unclear.


DrewCarlson:
Was this change absolutely necessary?  I am not very familiar with Swift but k/n 0.9.1 created great Swift names for nested interfaces in the sense that they were identical in every language that consumed my MPP library.  0.9.2 now requires mangling names because in Kotlin I have for example the interface `MobiusLoop.Controller` and an implementation `MobiusLoopController` resulting in the Swift name `MobiusLoopController_` for the Interface which isn't great from an API perspective.

SvyatoslavScherbina:
@DrewCarlson 
Well, it appears that you have interface nested into the class. This case seemed to work, however it was completely unclear whether it was indented `swift_name` attribute feature or just wasn't forbidden properly.
Btw, Swift compiler doesn't allow to nest protocol into a class in Swift code, and this fact was the motivation to forbid this when importing Kotlin declarations to Swift.

PR #2094, opened 2018-09-19
Update Xcode version to 10 on README

kitasuke:
When I run `./gradlew dependencies:update`, I got error below. Would be nice to update Xcode version from 9 to 10 since it's major update.

```
* What went wrong:
A problem occurred evaluating root project 'kotlin-native'.
> expected Xcode version 10.0, got 9.4.1, consider updating Xcode or use "ignoreXcodeVersionCheck" variable in konan.properties
```


olonho:
Fixed in 3b4b0728efb2b68d79a534bea8a6400dfb5a046e

PR #2110, opened 2018-09-20
Make native gradle plugin compatible with the MPP one

ilmat192:


ilmat192:
> Does this mean both plugins needs to be applied? In the sample only the multiplatform plugin is used. https://github.com/h0tk3y/k-new-mpp-samples/blob/master/lib-and-app/sample-lib/build.gradle

No, it means that the MPP plugin will be able to consume artifacts published by this one.

PR #2121, opened 2018-09-23
Publish kotlin-stdlib-native sources

jonnyzzz:
Added several tasks to zip and publish stdlib sources
I use TeamCity service message to publish library from the build
It is not clear if it makes sense to include same sources into the compiler or kotlin-stdlib packages

### DIFF
--- a/backend.native/build.gradle
+++ b/backend.native/build.gradle
@@ -156,6 +156,26 @@ task unzipStdlibSources(type: Copy) {
     destinationDir commonSrc
 }
 
+final List<File> stdLibSrc = [
+        project(':Interop:Runtime').file('src/main/kotlin'),
+        project(':Interop:Runtime').file('src/native/kotlin'),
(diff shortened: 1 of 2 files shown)


backend.native/build.gradle
```diff
@@ -156,6 +156,33 @@ task unzipStdlibSources(type: Copy) {
     destinationDir commonSrc
 }
 
+final List<File> stdLibSrc = [
```
PavelPunegov: I don't think it's a good idea to always print `"##teamcity..."`, specially if you build it locally

PavelPunegov: The task or print could be surrounded by the check for env or property existence "TEAMCITY_BUILD_PROPERTIES_FILE"

jonnyzzz: Added the `if`.

jonnyzzz:
Thank you. I do not see sources zip in the published artifacts. It looks like I added the wrong dependency of the service message is being lost. I'll investigate

PR #2132, opened 2018-09-25
Publish kotlin-stdlib-native sources take 2

jonnyzzz:
Few more commits to finally fix the intent of the 
https://github.com/JetBrains/kotlin-native/pull/2121
pull request. 

Patched `bundle` task transitive dependencies to make sources are included too

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -296,6 +296,7 @@ task commonDistRuntime(type: Copy) {
 task crossDistRuntime(type: Copy) {
     dependsOn.addAll(targetList.collect { "${it}CrossDistRuntime" })
     dependsOn('commonDistRuntime')
+    dependsOn(':backend.native:teamcityPublishStdLibSources')
 }
 
 task crossDistPlatformLibs {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -296,6 +296,7 @@ task commonDistRuntime(type: Copy) {
 task crossDistRuntime(type: Copy) {
     dependsOn.addAll(targetList.collect { "${it}CrossDistRuntime" })
     dependsOn('commonDistRuntime')
+    dependsOn(':backend.native:teamcityPublishStdLibSources')
```
PavelPunegov: May it be better to make stdlib build task be `finalizedBy` the zip task that will print the desired line in its `doLast` part?
And wherever the stdlib build is triggered it will publish the sources, while top level tasks could be changed independently without the need to add such dependence.

jonnyzzz:
Makes sense. I'll update.

msink:
Will it ever be finalized?
Not having stdlib sources in IDEA navigation is very inconvenient.

ddolovov:
It has been already done in a separate pull request: https://github.com/JetBrains/kotlin-native/commit/1dfca08269a89059972c2eca30f80589a3e7c340

PR #2141, opened 2018-09-26
Update and exclude failing tests

PavelPunegov:
* Update tests to a latest build
* Create an exclude list for external tests
* Exclude -ReleaseCoroutines

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -164,6 +164,19 @@ task resultsTask() {
     }
 }
 
+boolean isExcluded(String task) {
+    // List of tests that fail due to unresolved comiler bugs
+    def excluded = [ "codegen/box/functions/bigArity", "codegen/box/functions/functionExpression" ]
+
+    boolean result = false
+    excluded.forEach {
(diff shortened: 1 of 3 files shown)


backend.native/tests/build.gradle
```diff
@@ -164,6 +164,19 @@ task resultsTask() {
     }
 }
 
+boolean isExcluded(String task) {
```
ilmat192: I think it's better to use another name for the parameter, e.g.`testDirectory`.

backend.native/tests/build.gradle
```diff
@@ -164,6 +164,19 @@ task resultsTask() {
...
+    // List of tests that fail due to unresolved comiler bugs
+    def excluded = [ "codegen/box/functions/bigArity", "codegen/box/functions/functionExpression" ]
```
olonho: Can we exclude only individual test which shows FE bug in functionExpression?

PavelPunegov: Only by adding the "IGNORE" keyword in the test. Otherwise it will turn on to maintaining of two different exclude strategies

PR #2149, opened 2018-09-27
Add fat jars for gradle plugin

ilmat192:


samples/README.md
```diff
@@ -5,7 +5,7 @@ built using either command line tools (via `build.sh` script presented in each s
 See `README.md` in sample directories to learn more about specific samples and the building process.
 
 **Note**: If the samples are built from a source tree (not from a distribution archive) the compiler and the gradle
-plugin built from the sources are used. So one must build the compiler by running `./gradlew cross_dist` from the 
+plugin built from the sources are used. So one must build them by running `./gradlew cross_dist` from the
```
olonho: Will cross_dist install to local Maven?

samples/settings.gradle
```diff
@@ -22,6 +22,5 @@ include ':uikit'
 include ':win32'
 include ':videoplayer'
 include ':workers'
+include ':serialization'
```
olonho: Are you sure we shall include dependency on serialization by default?

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanGenerateCMakeTask.kt
```diff
@@ -51,10 +51,10 @@ open class KonanGenerateCMakeTask : DefaultTask() {
     private val host = HostManager.host
 
     private fun generateCMakeLists(
```
olonho: Do we need that/does it work?

samples/README.md
ilmat192: Yes

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanGenerateCMakeTask.kt
ilmat192: It works but it doesn't look like we really need it now.

PR #2155, opened 2018-09-28
Filter java.endorsed.dirs in compiler runner

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanToolRunner.kt
```diff
@@ -51,6 +51,9 @@ internal abstract class KonanCliRunner(
         } ?: emptyList<String>()
     }
 
+    protected val blacklistProperties: List<String> =
+        listOf("java.endorsed.dirs")
```
olonho: setOf?

PR #2161, opened 2018-10-01
Remove jvm annotations, make NativeRandom internal

ilya-g:


runtime/src/main/kotlin/kotlin/random/Random.kt
```diff
@@ -7,7 +7,7 @@ package kotlin.random
 import kotlin.native.concurrent.AtomicLong
 import kotlin.system.getTimeNanos
 
-public abstract class NativeRandom {
+internal class NativeRandom {
```
ilya-g: Should we keep default implementation as companion object or we can make it just a top-level object?

runtime/src/main/kotlin/kotlin/random/Random.kt
```diff
@@ -28,39 +28,12 @@ public abstract class NativeRandom {
...
         }
 
-        /**
-         * Returns a pseudo-random Int value between 0 and specified value (exclusive)
-         */
-        override fun nextInt(bound: Int): Int {
-            if (bound <= 0) throw IllegalArgumentException("Incorrect bound: $bound")
-
-            if (bound and (bound - 1) == 0) {
-                return ((bound * (nextInt() ushr 1).toLong()) ushr 31).toInt();
```
ilya-g: I don't think this marginal implementation difference deserves its own reimplementation that has to be maintained and tested.

runtime/src/main/kotlin/kotlin/random/Random.kt
```diff
@@ -28,39 +28,12 @@ public abstract class NativeRandom {
...
-                var r = nextInt() ushr 1
-                m = r % bound
-            } while (r - m + (bound - 1) < 0)
-            return m
-        }
-
-        /**
-         * Returns a pseudo-random Long number.
-         */
-        override fun nextLong(): Long = (nextInt().toLong() shl 32) + nextInt().toLong()
```
olonho: Why you removed this one?

runtime/src/main/kotlin/kotlin/random/Random.kt
olonho: Could you please clarify how it shall affect clients of RNG?

runtime/src/main/kotlin/kotlin/random/Random.kt
ilya-g: The same implementation is inherited from the base class.

runtime/src/main/kotlin/kotlin/random/Random.kt
ilya-g: Perhaps getting a random number in [0..bound) range where bound is 2^n will be not so fast as in JVM, until we make `fastLog2` (or `Int.numberOfLeadingZeroes` operation) really fast.

PavelPunegov:
|`runtime/src/main/kotlin/kotlin/random/Random.kt:49`|
|---|
I would rather have this removed also as there are no need to have such seeding available in native stdlib. Tests that use it could be just moved to common stdlib tests or removed

runtime/src/main/kotlin/kotlin/random/Random.kt
PavelPunegov: I think we can do make it a top level

runtime/src/main/kotlin/kotlin/random/Random.kt
PavelPunegov: we have https://github.com/JetBrains/kotlin-native/blob/79755d8fbfaf495e642fc41348b99029c4f6fbde/runtime/src/main/kotlin/kotlin/native/internal/HexStringParser.kt#L272
It looks the same as in JVM so we can make `Int.numberOfLeadingZeroes` be as fast as in the JVM (without taking jit into account)

PavelPunegov:
Than this method should be explicitly intrinsified

PavelPunegov:
Then this method should be explicitly intrinsified

PR #2165, opened 2018-10-02
Do not depend on ReflectionTypes from upstream Kotlin

udalov:
ReflectionTypes was incorrectly copied to common backend in
https://github.com/JetBrains/kotlin/commit/27365dc4bed8833880dc7bd0fc0189fed5ff6fed.
In fact, the copied class references `kProperty*Impl` types which are
only available in kotlin-native's runtime [1], thus using them in the
common Kotlin would lead to all kinds of confusion.

The next step is to remove ReflectionTypes in the main Kotlin project
(will be done later), remove the unneeded override
Context.reflectionTypes, and rename reflectionTypes0 -> reflectionTypes.

[1] Types with the same name are also declared in kotlin-reflect, but
the JVM backend still can't reference them because generated JVM
bytecode can only depend on kotlin-stdlib


udalov:
Thanks!

PR #2166, opened 2018-10-03
Remove the ability to change seed of the default Random

ilya-g:


backend.native/tests/runtime/basic/random.kt
```diff
@@ -13,51 +13,51 @@ import kotlin.test.*
 /**
  * Tests that setting the same seed make random generate the same sequence
  */
-private inline fun <reified T> testReproducibility(seed: Long, generator: () -> T) {
+private inline fun <reified T> testReproducibility(seed: Long, generator: Random.() -> T) {
```
ilya-g: I've edited this test so that it compiles, but at this point it duplicates SeededRandomSmokeTest from common stdlib.

backend.native/tests/runtime/basic/worker_random.kt
```diff
@@ -19,7 +19,6 @@ fun testRandomWorkers() {
     val attempts = 3
     val results = Array(attempts, { ArrayList<Int>() } )
     for (attempt in 0 until attempts) {
-        Random.seed = seed
```
ilya-g: Nothing depends on seed in this test.

backend.native/tests/runtime/basic/random.kt
ilya-g: So would you like this test to be removed then?

PR #2186, opened 2018-10-09
Implement -Xembed-bitcode and -Xembed-bitcode-marker

SvyatoslavScherbina:
These options currently make the produced framework bitcode-friendly
in release and debug environments correspondigly without embedding
real bitcode.

Also rebuild libffi for iOS and macOS to make it bitcode-friendly.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeEmbedding.kt
```diff
@@ -0,0 +1,41 @@
+package org.jetbrains.kotlin.backend.konan
```
olonho: Any way to test bitcode embedding? I.e. compile some code to framework and link it with code with options forcing bitcode usage.

PR #2198, opened 2018-10-10
Turn on bridges inlining

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt
```diff
@@ -994,20 +993,32 @@ internal object Devirtualization {
...
+    private inline fun <reified T : IrElement> MutableList<T>.transform(transformation: (T) -> IrElement) {
+        forEachIndexed { i, item ->
+            set(i, transformation(item) as T)
+        }
+    }
+
+    // TODO: do it more reliably.
+    private fun IrExpression.isBoxOrUnboxCall() =
+            (this is IrCall && symbol.owner.name.asString().let { it == "<box>" || it == "<unbox>" })
+                    || (this is IrPrivateFunctionCall && dfgSymbol.name?.let { it.contains("<box>") || it.contains("<unbox>") } == true)
```
olonho: Somewhat ugly logic. Can we list all box/unbox function is set and check with `contains`?

homuroll: I left "TODO" exactly because of this.

PR #2233, opened 2018-10-18
Link INTEROP.md

StefMa:
.. and add a "dot" at the end of the file 🙃

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -63,7 +63,7 @@ See the samples coming with the distribution.
 
 To see the list of available flags, run `kotlinc -h`.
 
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -63,7 +63,7 @@ See the samples coming with the distribution.
 
 To see the list of available flags, run `kotlinc -h`.
 
-For documentation on C interoperability stubs see INTEROP.md.
+For documentation on C interoperability stubs see [INTEROP.md](INTEROP.md).
```
olonho: Please use full URL, like below.

StefMa: Oh yeah, thanks.
It's part of the binaries. Should be the full url, you are right 👍

PR #2255, opened 2018-10-22
[kotlin compiler][update] 1.3.20-dev-1048

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLower.kt
```diff
@@ -93,6 +93,13 @@ internal class KonanLower(val context: Context, val parentPhaser: PhaseManager)
         phaser.phase(KonanPhase.LOWER_ENUMS) {
             EnumClassLowering(context).run(irFile)
         }
+
+        /**
+         * TODO:  this is workaround for issue of unitialized parents in IrDeclaration,
+         * the last one detected in [EnumClassLowering]. The issue appears in [DefaultArgumentStubGenerator].
+         */
+        irFile.patchDeclarationParents()
```
SvyatoslavScherbina: Parents are supposed to be initialized properly from the very beginning of `lowerFile`.
So there must be a bug, and any lowering may fail under non-trivial circumstances.
That's why I believe that it is better to find the cause of the problem than to add this workaround.

vvlevchenko: Not in this case. `EnumLovering` produces contractors having arguments without initialised `.parent` field, similar to `LocalDeclarationsLowering`, owners of mentioned lowerings can fix them, especially that easy to reproduce just with removing patching line.

PR #2262, opened 2018-10-24
Implement '-Xexport-library=<path>' compiler argument

SvyatoslavScherbina:
Produced framework includes the entire API of exported libraries

### DIFF
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
@@ -103,6 +103,14 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-Xenable", deprecatedName = "--enable", valueDescription = "<Phase>", description = "Enable backend phase")
     var enablePhases: Array<String>? = null
 
+    @Argument(
+            value = "-Xexport-library",
+            valueDescription = "<path>",
+            description = "Path to the library to be included into produced framework API\n" +
+                    "Must be the path of a library passed with '-library'"
+    )
+    var exportedLibraries: Array<String>? = null
+
     @Argument(value=  "-Xlist-phases", deprecatedName = "--list_phases", description = "List all backend phases")
     var listPhases: Boolean = false
 
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ExportedLibraries.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ExportedLibraries.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.kotlin.backend.konan
+
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.konan.file.File
+
+internal fun validateExportedLibraries(configuration: CompilerConfiguration, libraryFiles: List<File>) {
+    val exportedLibraries = getExportedLibraries(configuration)
+
+    val missingExportedLibraries = exportedLibraries - libraryFiles
+
+    if (missingExportedLibraries.isNotEmpty()) {
+        val message = buildString {
+            appendln("Following libraries are specified to be exported with -Xexport-library, but not included to the build:")
+            missingExportedLibraries.forEach { appendln(it) }
+            appendln()
+            appendln("Included libraries:")
+            libraryFiles.forEach { appendln(it) }
+        }
+
+        configuration.report(CompilerMessageSeverity.STRONG_WARNING, message)
+    }
+}
+
+internal fun getExportedLibraries(configuration: CompilerConfiguration) =
+        configuration.getList(KonanConfigKeys.EXPORTED_LIBRARIES).map { File(it) }.toSet()
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ExportedLibraries.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ExportedLibraries.kt
@@ -7,25 +7,78 @@ package org.jetbrains.kotlin.backend.konan
 
 import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
 import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.descriptors.konan.CurrentKonanModuleOrigin
+import org.jetbrains.kotlin.descriptors.konan.DeserializedKonanModuleOrigin
+import org.jetbrains.kotlin.descriptors.konan.SyntheticModulesOrigin
+import org.jetbrains.kotlin.descriptors.konan.konanModuleOrigin
 import org.jetbrains.kotlin.konan.file.File
+import org.jetbrains.kotlin.konan.library.KonanLibrary
+import org.jetbrains.kotlin.konan.library.isInterop
+import org.jetbrains.kotlin.konan.library.resolver.KonanLibraryResolveResult
 
-internal fun validateExportedLibraries(configuration: CompilerConfiguration, libraryFiles: List<File>) {
-    val exportedLibraries = getExportedLibraries(configuration)
+internal fun validateExportedLibraries(configuration: CompilerConfiguration, resolvedLibraries: KonanLibraryResolveResult) {
+    getExportedLibraries(configuration, resolvedLibraries, report = true)
+}
 
-    val missingExportedLibraries = exportedLibraries - libraryFiles
+internal fun Context.getExportedDependencies(): List<ModuleDescriptor> {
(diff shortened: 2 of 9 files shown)


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -103,6 +103,14 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-Xenable", deprecatedName = "--enable", valueDescription = "<Phase>", description = "Enable backend phase")
     var enablePhases: Array<String>? = null
 
+    @Argument(
+            value = "-Xexport-library",
+            valueDescription = "<path>",
+            description = "Path to the library to be included into produced framework API\n" +
+                    "Must be the path of a library passed with '-library'"
```
olonho: Can platform libs or stdlib be exported this way?

olonho: Can it be the name, not path?
Gradle plugin support for that shall be done as well.

SvyatoslavScherbina: > Can platform libs or stdlib be exported this way?

Yes.

> Can it be the name, not path?

No, because this would require the feature to be tightly coupled with library resolution machinery.

> Gradle plugin support for that shall be done as well.

AFAIK, Gradle provides absolute paths for libraries when invoking the compiler.

olonho:
`./dist/bin/konanc a.kt -Xexport-library=/Users/nikolay.igotti/kotlin/kotlin-native/dist/klib/common/stdlib -p framework -o a` produces a lot of unresolved symbols like
`_kfun:kotlinx.cinterop.convert@kotlin.Byte.()Generic`

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
```diff
@@ -128,10 +138,15 @@ internal class ObjCExportNamerImpl(
 
     override fun getFileClassName(file: SourceFile): ObjCExportNamer.ClassOrProtocolName {
         val baseName by lazy {
-            val psiSourceFile = file as? PsiSourceFile ?: error("SourceFile '$file' is not PsiSourceFile")
-            val psiFile = psiSourceFile.psiFile
-            val ktFile = psiFile as? KtFile ?: error("PsiFile '$psiFile' is not KtFile")
-            PackagePartClassUtils.getFilePartShortName(ktFile.name)
+            val fileName = when (file) {
```
olonho: What if we'll try to export interop library once it will be generated from descriptors?

SvyatoslavScherbina:
> ./dist/bin/konanc a.kt -Xexport-library=/Users/nikolay.igotti/kotlin/kotlin-native/dist/klib/common/stdlib -p framework -o a produces a lot of unresolved symbols like
_kfun:kotlinx.cinterop.convert@kotlin.Byte.()Generic

Well, there are a lot of reasons for that, e.g.:
* intrinsics
* target-specific declarations

Do you suggest to forbid exporting stdlib explicitly?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
SvyatoslavScherbina: 1) It doesn't make much sense to export interop library.
2) No one knows how will everything work when interop is generated from descriptors. For example, such a module can have origin different from `DeserializedKonanModuleOrigin`.

olonho: I afraid for use cases like SDL or OpenGL, where there are many declarations potentially important for consumer (such as constants), it is important to have ability to import interop libs.

olonho:
Idea was to properly ignore such cases in imported. Also declarations like
`@Deprecated("Unable to import this declaration", level = DeprecationLevel.ERROR)
external fun atomic_flag_clear(arg0: CValuesRef<atomic_flag>?): Unit` from interop libs shall be filtered.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
SvyatoslavScherbina: What prevents user from importing "interop library" directly and idiomatically by importing the original C/Objective-C library?

SvyatoslavScherbina:
stdlib is too special to support this case easily and too large to make exporting it to have any sense.

olonho:
Maybe so, but it also a great testcase for covering various language features.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
olonho: Potential version mismatch and need to install original framework, i.e. if Kotlin wraps version A, while user needs to make sure (by side channels) that  he/she installs matching version. If Kotlin framework was self-containing - it would make things easier.

SvyatoslavScherbina: Double API conversion would produce ugly API.
I believe that this goal should be achieved by other means, e.g. by including the original headers into the framework.
So I don't consider it as a part of this feature.

olonho: Maybe so, but why not make import work on interop libs anyway, it gives us significant testing coverage, if nothing else.

SvyatoslavScherbina: Because it
* Requires a lot of efforts, since interop libraries are built with magic
* Testing coverage is not that significant, because interop libraries usually have quite simple language constructs. For testing coverage it makes sense to take Kotlin libraries like kotlinx.coroutines, ktor etc.

olonho:
We also need diagnostics if library was not found, and maybe, if option is used in non framework mode.

olonho:
When using coroutines .klib nothing is exported at all (is everything suspend?).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
olonho: From my experiment not much is needed on actual C platform libs:

SvyatoslavScherbina:
> When using coroutines .klib nothing is exported at all (is everything suspend?).

I can't reproduce this. My headers are full of kotlinx.coroutines declarations.
Maybe your compiler skipped the library due to incompatible ABI/compiler version?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
SvyatoslavScherbina: This work is completely unrelated to the feature being developed in this PR.
The result looks quite useless for our users.

> and generating code in interop libs which leads to linkage errors due to missing definition looks questionable anyway, so this change is generally good

This behaviour is unrelated too.

olonho: You (or me) may fix it in separate PR, but producing libraries which only link when DCE is in place is not correct.

SvyatoslavScherbina: > You (or me) may fix it in separate PR, but producing libraries which only link when DCE is in place is not correct.

1. I am pretty sure that the library can be linked properly to any program even if DCE is disabled. Program can't call a function that is not callable, and the library itself doesn't have any related unresolved references in its code.
2. Previously we (you and me) had a discussion and concluded that it is OK to rely on DCE to avoid link errors because it is unfortunately pretty common for native libraries to declare functions that don't have definitions.

olonho: This discussion is out of scope of this PR, indeed, so let's discuss in https://github.com/JetBrains/kotlin-native/pull/2265.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ExportedLibraries.kt
```diff
@@ -7,25 +7,78 @@ package org.jetbrains.kotlin.backend.konan
...
+    val libraries = resolvedLibraries.getFullList(null)
+
+    for (library in libraries) {
+        val libraryFile = library.libraryFile
+        if (libraryFile in exportedLibraries) {
+            remainingExportedLibraries -= libraryFile
+            if (library.isInterop || library.isDefault) {
+                if (report) {
+                    val kind = if (library.isInterop) "Interop" else "Default"
+                    configuration.report(
```
olonho: Not sure if this limitation makes sense for interop libs. I guess it shall be limited to stdlib.

PR #2263, opened 2018-10-24
Move file and hide Regex.Mode enum

ilya-g:
Please **do not squash** these commits.


runtime/src/main/kotlin/kotlin/text/Strings.kt
```diff
@@ -7,6 +7,185 @@ package kotlin.text
 
 import kotlin.comparisons.*
 
+/**
+ * Returns the index within this string of the first occurrence of the specified character, starting from the specified offset.
+ */
+@SymbolName("Kotlin_String_indexOfChar")
+external internal actual inline fun String.nativeIndexOf(ch: Char, fromIndex: Int): Int
```
ilya-g: `external inline` — I wonder does it make sense?

runtime/src/main/kotlin/kotlin/text/Strings.kt
```diff
@@ -7,6 +7,185 @@ package kotlin.text
...
+
+/**
+ * Returns `true` if the specified range in this char sequence is equal to the specified range in another char sequence.
+ * @param thisOffset the start offset in this char sequence of the substring to compare.
+ * @param other the string against a substring of which the comparison is performed.
+ * @param otherOffset the start offset in the other char sequence of the substring to compare.
+ * @param length the length of the substring to compare.
+ */
+@SymbolName("Kotlin_CharSequence_regionMatches")
+external public actual fun CharSequence.regionMatches(
```
ilya-g: Both `Kotlin_CharSequence_regionMatches` and `Kotlin_String_regionMatches` take `KString` parameters. How does this magic works for CharSequences? Or does `KString` correspond to `CharSequence` interface?

runtime/src/main/kotlin/kotlin/text/Strings.kt
olonho: Unlikely so.

runtime/src/main/kotlin/kotlin/text/Strings.kt
olonho: This is likely a bug, this code doesn't seem to work on non kotlin.String classes.

runtime/src/main/kotlin/kotlin/text/Strings.kt
ilya-g: I've found why they are `inline` — because they are `inline` in the corresponding expect declaration. I'll remove it there and then here.

PR #2265, opened 2018-10-25
Avoid hanging external functions.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/DescriptorTable.kt
```diff
@@ -244,13 +246,16 @@ private val FunctionDescriptor.uniqueName: String
             throw AssertionError(this.toString())
         }
 
-        this.annotations.findAnnotation(symbolNameAnnotation)?.let {
-            if (this.isExternal) {
+        if (this.isExternal) {
```
SvyatoslavScherbina: This code doesn't look to be related to bitcode generation.
AFAIK, it is used for inline functions serialization.

The similar code for bitcode generation can be found here:
https://github.com/JetBrains/kotlin-native/blob/daf1a5fb8cdba2aea97d3a0cf53ad919fdb02b13/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt#L203

olonho: Shall be fixed.

PR #2270, opened 2018-10-26
FAQ: Examples in MPP DSL, not the first Native DSL.

ddolovov:


FAQ.md
```diff
@@ -15,7 +15,14 @@ garbage.
 
 ### Q: How do I create a shared library?
 
-A: Use the `-produce dynamic` compiler switch, or `konanArtifacts { dynamic('foo') {} }` in Gradle.
+A: Use the `-produce dynamic` compiler switch, or `compilations.main.outputKinds 'DYNAMIC'` in Gradle, i.e.
+```groovy
+targets {
+    fromPreset(presets.iosArm64, 'mylib') {
```
SvyatoslavScherbina: What's the purpose of using `mylib` for target name? Is it the approach we are proposing?

ddolovov: > What's the purpose of using `mylib` for target name?

Target produces source sets and all related lifecycle Gradle tasks (compile this, build that, etc). `mylib` was chosen by analogy with `myapp` mentioned in this document below.

PR #2279, opened 2018-10-29
WIP: Pointers in cinterop

ilmat192:


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
```diff
@@ -110,7 +114,7 @@ private fun reparseWithCodeSnippet(library: NativeLibrary,
         val codeSnippetLines = when (library.language) {
             // Note: __auto_type is a GNU extension which is supported by clang.
             Language.C, Language.OBJECTIVE_C -> listOf(
-                    "const __auto_type KNI_INDEXER_VARIABLE = $name;"
+                    "void kni_indexer_function() { __auto_type KNI_INDEXER_VARIABLE = $name; }"
```
SvyatoslavScherbina: Is it required to define a wrapping function?

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
```diff
@@ -127,34 +131,41 @@ private fun processCodeSnippet(
...
-
+                evalResultOrNull = clang_Cursor_Evaluate(cursor)
+                state = VisitorState.EXPECT_VARIABLE_VALUE
                 CXChildVisitResult.CXChildVisit_Recurse
             }
 
             state == VisitorState.EXPECT_VARIABLE_VALUE && clang_isExpression(kind) != 0 -> {
                 typeOrNull = typeConverter(clang_getCursorType(cursor))
-                state = VisitorState.EXPECT_END
+                state = if (typeOrNull != null) {
```
SvyatoslavScherbina: `typeConverter` doesn't produce `null` values, AFAIK.

backend.native/tests/interop/basics/cmacros.def
```diff
@@ -10,5 +24,19 @@
 #define ONE_POINT_ZERO 1.0
 #define ONE_POINT_FIVE 1.5f
 
+#define NULL_PTR (void*)0
```
SvyatoslavScherbina: It makes sense to add brackets to this and following definitions to make them less ambiguous, i.e.
`#define NULL_PTR ((void*)0)`

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/StubGenerator.kt
```diff
@@ -848,6 +848,16 @@ class StubGenerator(
             }
         }
 
+        nativeIndex.wrappedMacros.forEach {
+            try {
+                stubs.add(
+                        GlobalVariableStub(GlobalDecl(it.name, it.type, isConst = true), this)
```
SvyatoslavScherbina: This is ok, but `GlobalVariableStub` has special handling for struct- and array-typed C globals, and may generate incorrect stubs for such macros.
I suppose `GlobalVariableStub` could distinguish C macro and C global cases, and don't do special handling for the first one.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
ilmat192: Looks like it's required. Without the wrapping function the variable is global and we cannot use a function to initialize it. So we cannot expand macros like `#define FOO foo()`.

backend.native/tests/interop/basics/cmacros.def
ilmat192: Fixed.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
SvyatoslavScherbina: Btw, is `errno` supported now? On macOS it is declared like
```
#define errno (*__error())
```

ilmat192: Yes, it's supported now.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/MacroConstants.kt
ilmat192: Fixed

Issue #2291, opened 2018-11-01
runtime assert: Recursive GC is disallowed


brettwillis:
Running Kotlin 1.3.0 on iOS, I sometimes get this error at this line of Kotlin code:

    actual fun logDebug(tag: String, message: String) {
        NSLog("DEBUG: $tag: $message") // <-- SIGABRT here
    }

Error message:

    /opt/teamcity-agent/work/4d622a065c544371/runtime/src/main/cpp/Memory.cpp:1372: runtime assert: Recursive GC is disallowed

Stacktrace:

<img width="274" alt="screen shot 2018-11-01 at 8 41 34 pm" src="https://user-images.githubusercontent.com/6925974/47839145-94afdc00-de16-11e8-91d7-f3dc647b12af.png">

The error is very intermittent, I'd say it happens about 1 time out of 1000.

olonho:
Could you please help us get the self-contained reproducer?

brettwillis:
Ok, so I've learnt some more. Previously, around Kotlin/Native 0.8\~0.9 and Coroutines 0.26\~0.27, this happened very occasionally and I only remember it in `NSLog(...)` as described above. 

Now as I am experimenting more with Kotlin 1.3.0 and Coroutines 1.0.0, I've realised it's happening much more often, almost like it happens after a certain period of time. And the `SIGABRT` actually happens anywhere and everywhere, with no relation. Currently I don't know what part is causing the error so I can give you a reproducer. Here are some more stack traces:

<img width="505" alt="screen shot 2018-11-01 at 8 53 41 pm" src="https://user-images.githubusercontent.com/6925974/47875029-40d6de80-de7a-11e8-9828-896d33ea2106.png">

<img width="505" alt="screen shot 2018-11-01 at 8 50 01 pm" src="https://user-images.githubusercontent.com/6925974/47875033-43393880-de7a-11e8-8309-85c55c55fc9f.png">

I'm not explicitly using any `dealloc` hooks, but I don't know what's happening in the third party libraries I'm using. However I'm not seeing and obvious dealloc frames in the stack traces.

This may or may not be related, but I'm using Coroutines with my own Dispatcher wrapper for the main thread. The dispatcher looks like below. I've noticed that the errors often happen when there are a few levels of recursion in the Dispatcher (i.e. it dispatches a block, which dispatches a block, which dispatches a block, etc... around 5 levels).

	@UseExperimental(InternalCoroutinesApi::class)
[...]

PR #2302, opened 2018-11-07
Fix for https://github.com/JetBrains/kotlin-native/issues/2291

homuroll:
ObjC deallocation hooks might trigger recursive GC - forbid that.

### DIFF
--- a/runtime/src/main/cpp/Memory.cpp
+++ b/runtime/src/main/cpp/Memory.cpp
@@ -709,10 +709,14 @@ void ScanRoots(MemoryState* state) {
 }
 
 void CollectRoots(MemoryState* state) {
+  // Here we might free some objects and call deallocation hooks on them,
+  // which in turn might call DecrementRC and trigger new GC - forbid that.
+  state->gcSuspendCount++;
   for (auto container : *(state->roots)) {
     container->resetBuffered();
     CollectWhite(state, container);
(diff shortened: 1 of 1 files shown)


runtime/src/main/cpp/Memory.cpp
```diff
@@ -577,7 +578,7 @@ inline void DecrementRC(ContainerHeader* container, bool useCycleCollector) {
         container->setBuffered();
         auto state = memoryState;
         state->toFree->push_back(container);
-        if (state->gcSuspendCount == 0 && freeableSize(state) >= state->gcThreshold) {
+        if (state->gcSuspendCount == 0 && !state->finalizationFromGCInProgress && freeableSize(state) >= state->gcThreshold) {
```
olonho: Why not reuse gcSuspendCount for this purpose?

homuroll: Good point.

SvyatoslavScherbina: As far as I understand, the issue occurs during `CollectWhite` phase of `CollectCycles`.

homuroll: Indeed.

SvyatoslavScherbina:
Do we have a reproducer?

PR #2311, opened 2018-11-12
Sharing the mutable binary data

olonho:


runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public fun append(data: ByteArray, start: Int = 0, count: Int = data.size - start): Unit = locked(lock) {
+        val where = resizeDataLocked(this.size + count)
+        data.copyRangeTo(buffer, start, count, where)
+    }
+
+    /**
+     * Appends C data to the buffer.
+     */
+    public fun append(data: COpaquePointer?, count: Int): Unit = locked(lock) {
```
qwwdfsad: why is it nullable if a call with null doesn't do anything?

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+external private fun Any.share()
+
+@SymbolName("Kotlin_CPointer_CopyMemory")
+external private fun CopyMemory(to: COpaquePointer?, from: COpaquePointer?, count: Int)
+
+/**
+ * Mutable concurrently accessible data buffer. Could be accessed from several workers simulteniously.
+ */
+@Frozen
+public class MutableData {
```
qwwdfsad: Should it be `@Experimental`? I am not sure this API is going to be stable in a nearest future

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+    public fun copyRangeTo(output: ByteArray, fromIndex: Int, toIndex: Int, destinationIndex: Int): Unit = locked(lock) {
+        buffer.copyRangeTo(output, fromIndex, toIndex, destinationIndex)
+    }
+
+    /**
+     * Get a byte from the mutable data.
+     *
+     * @Throws IndexOutOfBoundsException if index is beyound range.
+     */
+    public operator fun <R> get(index: Int): Byte = locked(lock) {
```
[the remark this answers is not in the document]
qwwdfsad: Why is it parametrized by `<R>` btw?

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+        // index < 0 is checked below by array access.
+        if (index > size)
```
qwwdfsad: `MutableData().get(0)` will return `0`, though it should throw `IOOBE`.
It should be `>=`

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+            throw IndexOutOfBoundsException()
```
cy6erGn0m: Would be nice to format error message with `index` and `size`

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+        buffer[index]
+    }
+
+    /**
+     * Executes provided block under lock with raw pointer to the data stored in the buffer.
+     * Block is executed under the spinlock, and must be short.
+     */
+    public fun <R> asPointerLocked(block: (COpaquePointer, dataSize: Int) -> R) = locked(lock) {
```
cy6erGn0m: `withPointerLocked`? `as` prefix makes impression that it will be converted and returned as result

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+        data.copyRangeTo(buffer, start, count, where)
+    }
+
+    /**
+     * Appends C data to the buffer.
+     */
+    public fun append(data: COpaquePointer?, count: Int): Unit = locked(lock) {
+        if (data == null || count == 0) return
+        val where = resizeDataLocked(this.size + count)
+        var index = 0
```
cy6erGn0m: unused?

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+        val pointer = data.reinterpret<ByteVar>()
+        buffer.usePinned {
+            it -> CopyMemory(it.addressOf(where), pointer, count)
```
cy6erGn0m: Are we corrupting memory here if `count` is negative as it as interpreted by `memcpy` as unsigned (size_t)?

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+        val toCopy = data.size
+        val where = resizeDataLocked(size + toCopy)
+        data.copyRangeTo(buffer, 0, size, where)
+    }
+
+    /**
+     * Appends byte array to the buffer.
+     */
+    public fun append(data: ByteArray, start: Int = 0, count: Int = data.size - start): Unit = locked(lock) {
+        val where = resizeDataLocked(this.size + count)
```
cy6erGn0m: negative `count` need to be prohibited before to avoid buffer size reduction

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+ * Mutable concurrently accessible data buffer. Could be accessed from several workers simulteniously.
+ */
+@Frozen
+public class MutableData {
+    init {
+        // Instance of MutableData is shared.
+        share()
+    }
+
+    private var buffer = ByteArray(16).apply { share() }
```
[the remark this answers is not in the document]
cy6erGn0m: Please note that buffer of size `0` is potentially dangerous since `buffer.usePinned { it.addressOf(0) }` will crash AFAIK. So zero capacity need to be forbidden or need to be checked everywhere

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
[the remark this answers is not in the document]
olonho: > It's ok for a prototype, but volatile read is enough for `get`

Not sure if we can have volatile read.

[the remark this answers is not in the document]
olonho: > Why is it parametrized by `<R>` btw?

Typo.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+    private var size_ = 0
+    private val lock = Lock()
+
+    private fun resizeDataLocked(newSize: Int): Int {
+        if (newSize > buffer.size) {
+            val actualSize = maxOf(buffer.size * 3 / 2 + 1, newSize)
```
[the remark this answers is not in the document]
olonho: Added an assert.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
olonho: Just for the ease of use with C/Objective-C APIs.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
olonho: Everything in kotlin.native.concurrent is experimental, no need to clutter it with uneeded annotations.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
olonho: Added check.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,123 @@
...
+ */
+@Frozen
+public class MutableData {
+    init {
+        // Instance of MutableData is shared.
+        share()
+    }
+
+    private var buffer = ByteArray(16).apply { share() }
+    private var size_ = 0
```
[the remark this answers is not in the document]
ilya-g: Leading and trailing underscores both are used in stdlib, but leading ones occur more often.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,124 @@
...
+    public fun append(data: MutableData) = locked(lock) {
+        val toCopy = data.size
+        val where = resizeDataLocked(size + toCopy)
+        data.copyRangeTo(buffer, 0, toCopy, where)
+    }
+
+    /**
+     * Appends byte array to the buffer.
+     */
+    public fun append(data: ByteArray, start: Int = 0, count: Int = data.size - start): Unit = locked(lock) {
```
ilya-g: Subranges are usually defined by start + end (exclusive) rather than start + count

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,124 @@
...
+        buffer.usePinned {
+            it -> CopyMemory(it.addressOf(where), data, count)
+        }
+    }
+
+    /**
+     * Copies range of mutable data to the byte array.
+     */
+    public fun copyRangeTo(output: ByteArray, fromIndex: Int, toIndex: Int, destinationIndex: Int): Unit = locked(lock) {
+        buffer.copyRangeTo(output, fromIndex, toIndex, destinationIndex)
```
ilya-g: Isn't `copyRangeTo` deprecated in favor of `copyInto`?
For consistency it may have sense to rename this function as well.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
ilya-g: `@Experimental` annotations are to make consumer opt in for experimental api, so they are surely not unneeded.

olonho: As I mentioned, we do not expect much stability in kotlin.native.concurrent.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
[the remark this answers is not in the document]
olonho: I personally find tailing underscore more readable.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
ilya-g: `copyInto`  has parameters in the following order: `destination: BooleanArray, destinationOffset: Int = 0, startIndex: Int = 0, endIndex: Int = size`

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
```diff
@@ -0,0 +1,124 @@
...
+        val toCopy = data.size
+        val where = resizeDataLocked(size + toCopy)
+        data.copyInto(buffer, 0, toCopy, where)
+    }
+
+    /**
+     * Appends byte array to the buffer.
+     */
+    public fun append(data: ByteArray, fromIndex: Int = 0, toIndex: Int = data.size): Unit = locked(lock) {
+        if (toIndex <= fromIndex) return
```
ilya-g: I'd expect the contract of `Appendable.append` here, so that IOOBE is thrown when from > to

runtime/src/main/cpp/Memory.h
```diff
@@ -83,10 +85,18 @@ struct ContainerHeader {
     refCount_ = (refCount_ & ~CONTAINER_TAG_MASK) | CONTAINER_TAG_FROZEN;
   }
 
+  inline void makeShareable() {
+      refCount_ = (refCount_ & ~CONTAINER_TAG_MASK) | CONTAINER_TAG_ATOMIC;
+  }
+
   inline bool permanentOrFrozen() const {
```
SvyatoslavScherbina: Is it needed? For example, is it ok to use it here?
https://github.com/JetBrains/kotlin-native/blob/dfbafd420bafdfe3954da56609b29efbdca8f81f/runtime/src/main/cpp/Memory.cpp#L1817

runtime/src/main/kotlin/kotlin/native/concurrent/Continuation.kt
```diff
@@ -0,0 +1,105 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.native.concurrent
+
+import kotlin.native.internal.*
+import kotlinx.cinterop.*
+
+public class Continuation0(block: () -> Unit,
+                    private val invoker: CPointer<CFunction<(COpaquePointer?) -> Unit>>,
```
SvyatoslavScherbina: What's the purpose of using pointers in API here? I believe this class can take frozen lambda as invoker.
So invoker can be like `(() -> Unit) -> Unit` or even
```
interface Invoker {
    fun invoke(block: () -> Unit)
}
```

The invoker gets shared wrapper for the original `block` with its arguments bound, and can invoke it directly from appropriate context. This eliminates the requirement to have complicated work-flow with `callContinuation*`.

runtime/src/main/kotlin/kotlin/native/concurrent/Continuation.kt
```diff
@@ -0,0 +1,105 @@
...
+                    private val singleShot: Boolean = false): Function0<Unit> {
+
+    private var stable = StableRef.create(block)
```
SvyatoslavScherbina: AFAIR, `StableRef` has direct object reference to the object, so `block` will be frozen too.

SvyatoslavScherbina: `var`?

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
SvyatoslavScherbina: I bet suppressing errors is not the ease we are looking to.

runtime/src/main/cpp/Memory.h
olonho: I left that code intentionally, as not sure that with current GC we can use shared objects in Konan_ensureAcyclicAndSet (and also it is useless, as cycles may appear concurrently after the check).

runtime/src/main/kotlin/kotlin/native/concurrent/Continuation.kt
olonho: Interesting point. It was done that way as practical invokers I was planning for are ones that call C/Obj-C APIs and pass C callbacks and `void*` argument, so API kind of modelled around this pattern.

runtime/src/main/kotlin/kotlin/native/concurrent/Continuation.kt
olonho: Oh, that's unfortunate, and maybe shall be changed in StableRef impl. WDYT?

SvyatoslavScherbina: I totally agree. `StableRef` shall be reimplemented as `inline class` over a pointer type. So it would become shareable by default.

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/StableRef.kt
```diff
@@ -66,12 +64,11 @@ data class StableRef<out T : Any> @PublishedApi internal constructor(
     /**
      * Returns the object this handle was [created][StableRef.create] for.
      */
-    fun get() = this.ref
+    fun get() = derefStablePointer(this.stablePtr) as T
```
SvyatoslavScherbina: It would be nice to have checked cast (as before).
I bet it is possible to return back a cast in `asStableRef`, but ignore its result.

runtime/src/main/kotlin/kotlin/native/concurrent/Continuation.kt
SvyatoslavScherbina: Ok, now I see. I suppose we can keep this base version, but return back later to improve support for block-style APIs.

runtime/src/main/kotlin/kotlin/native/concurrent/MutableData.kt
olonho: Hmm, it really depends on what kind of operations we expect here. IMO, null are legit common case here, which shall be tolerated.

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/StableRef.kt
olonho: Not sure if I understand what do you mean, could you please hint what kind of code you'd like to see?

SvyatoslavScherbina: `asStableRef` could be like
```
inline fun <reified T : Any> CPointer<*>.asStableRef(): StableRef<T> = StableRef<T>(this)
    .also { it.get as T }
```

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/StableRef.kt
```diff
@@ -26,17 +26,16 @@ typealias StableObjPtr = StableRef<*>
  *
  * Any [StableRef] should be manually [disposed][dispose]
  */
-data class StableRef<out T : Any> @PublishedApi internal constructor(
-        private val stablePtr: COpaquePointer,
-        private val ref: T // Note: storing the reference itself to avoid unchecked casts.
+inline class StableRef<out T : Any> constructor(
```
SvyatoslavScherbina: The constructor can be made internal again using `@Suppress("NON_PUBLIC_PRIMARY_CONSTRUCTOR_OF_INLINE_CLASS")`
(it is often applied to stdlib classes, e.g. `Result`, `UInt` etc.)

PR #2322, opened 2018-11-13
Avoid recursion in GC traversals.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -627,51 +627,66 @@ void ScanRoots(MemoryState*);
...
+    } else {
+      container->mark();
+    }
+    traverseContainerReferredObjects(container, [&toVisit](ObjHeader* ref) {
+      auto childContainer = ref->container();
+      RuntimeAssert(!isArena(childContainer), "A reference to local object is encountered");
+
+      if (!childContainer->permanentOrFrozen()) {
+        childContainer->decRefCount<false>();
+        toVisit.push_front(childContainer);
```
olonho: Is it important we do DFS, not BFS here?

sbogolepov: Maybe check that `childContainer` is marked before adding to the queue?

sbogolepov: Maybe check that `childContainer` is marked before adding it to the queue?

PR #2332, opened 2018-11-15
curl sample enhancements

ddolovov:


samples/build.gradle
```diff
@@ -36,6 +36,7 @@ task buildSamplesWithPlatformLibs() {
     dependsOn ':workers:assemble'
 
     if (MPPTools.isMacos() || MPPTools.isLinux()) {
+        dependsOn ':curl:assemble'
```
olonho: What about mingw support?

PR #2353, opened 2018-11-20
Make crash with non-converted Kotlin exception debugger-friendly

SvyatoslavScherbina:
* Use `abort` instead of `exit` to make debugger recognize it as crash
* Terminate from C++ to prevent debugger from showing Kotlin handlers

 #KT-28066 Fixed

### DIFF
--- a/runtime/src/main/cpp/ObjCExportErrors.mm
+++ b/runtime/src/main/cpp/ObjCExportErrors.mm
@@ -19,6 +19,7 @@
 #import <Foundation/NSDictionary.h>
 #import <Foundation/NSError.h>
 #import <Foundation/NSString.h>
+#import <stdio.h>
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/cpp/ObjCExportErrors.mm
```diff
@@ -19,6 +19,7 @@
 #import <Foundation/NSDictionary.h>
 #import <Foundation/NSError.h>
 #import <Foundation/NSString.h>
+#import <stdio.h>
```
olonho: Hmm, why not rely on konan::consolePrintf, adding new logging mechanism is really not desirable.

SvyatoslavScherbina: Fixed.

PR #2354, opened 2018-11-20
Small MM refactor, ignore acyclic and frozen objects in cycle collector.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,27 +6,53 @@
...
+        if (KotlinBuiltIns.isPrimitiveArray(type.toKotlinType()))
+            return true
+        return false
+    }
+
     private fun flagsFromClass(classDescriptor: ClassDescriptor): Int {
         var result = 0
         if (classDescriptor.isFrozen)
            result = result or TF_IMMUTABLE
+        // TODO: maybe perform deeper analysis to find surely acyclic types.
```
homuroll: Why not to do it right now? Seems quite simple to implement.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,27 +6,53 @@
...
+import org.jetbrains.kotlin.ir.util.defaultType
 import org.jetbrains.kotlin.name.FqName
 
-
 internal class RTTIGenerator(override val context: Context) : ContextUtils {
 
+    private val acyclicCache = mutableMapOf<IrType, Boolean>()
+
+    private fun checkAcyclicFieldType(type: IrType): Boolean = acyclicCache.getOrPut(type) {
+        val type = type
```
homuroll: What's this for?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: How do you envision that exactly?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: Typo.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,27 +6,57 @@
...
+        val classifier = type.classifierOrNull
+        if (classifier != null && classifier in safeAcyclicFieldTypes)
+            return true
+        return false
+    }
+
+    private fun checkAcyclicClass(classDescriptor: ClassDescriptor): Boolean = when {
+        classDescriptor.symbol == context.ir.symbols.array -> false
+        classDescriptor.isArray -> true
+        context.llvmDeclarations.forClass(classDescriptor).fields.all {
```
sbogolepov: Maybe `Context.getFields(classDescriptor: ClassDescriptor)` instead?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,27 +6,57 @@
...
 import org.jetbrains.kotlin.ir.declarations.IrField
 import org.jetbrains.kotlin.ir.declarations.IrProperty
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.util.isAnnotationClass
 import org.jetbrains.kotlin.name.FqName
 
-
 internal class RTTIGenerator(override val context: Context) : ContextUtils {
 
+    private val acyclicCache = mutableMapOf<IrType, Boolean>()
```
SvyatoslavScherbina: Btw, `IrType` doesn't have `equals` implementation checking type equality, so types are compared by object identity.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,27 +6,57 @@
...
+    private val safeAcyclicFieldTypes = setOf(context.irBuiltIns.stringClass) +
+            context.ir.symbols.primitiveArrays.values +
+            context.ir.symbols.unsignedArrays.values
+
+    private fun checkAcyclicFieldType(type: IrType): Boolean = acyclicCache.getOrPut(type) {
+        if (type.computePrimitiveBinaryTypeOrNull() != null) return true
+        val classifier = type.classifierOrNull
+        if (classifier != null && classifier in safeAcyclicFieldTypes)
+            return true
```
SvyatoslavScherbina: `return` returns from `checkAcyclicFieldType`, and the result doesn't seem to get into the cache.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
SvyatoslavScherbina: I guess `context.llvmDeclarations.forClass(classDescriptor).fields` is more appropriate here.
`ContextUtils.getFields` computes the list of fields every time.

runtime/src/main/cpp/Memory.h
```diff
@@ -153,9 +165,17 @@ struct ContainerHeader {
   }
 
   inline void setColor(unsigned color) {
```
SvyatoslavScherbina: As far as I understand, it makes sense to make the "default" coloring function ensuring that it doesn't recolor green object. I.e. to have `setColor` (asserting that current color is not green), `setColorUnlessGreen` and `setColorEvenIfGreen`. Btw, is the latter actually needed?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -561,17 +561,19 @@ inline uint32_t freeableSize(MemoryState* state) {
...
-inline void DecrementRC(ContainerHeader* container, bool useCycleCollector) {
+template <bool Atomic, bool UseCycleCollector>
+inline void DecrementRC(ContainerHeader* container) {
   if (container->decRefCount<Atomic>() == 0) {
     FreeContainer(container);
-  } else if (!Atomic && useCycleCollector) { // Possible root.
-    // Do not use cycle collector for frozen objects, as we already detected possible cycles during
-    // freezing.
-    if (container->color() != CONTAINER_TAG_GC_PURPLE) {
+  } else if (!Atomic && UseCycleCollector) { // Possible root.
```
SvyatoslavScherbina: Is it correct to check for `Atomic` here? May be check only for `UseCycleCollector` and add an assertion that `!Atomic` in this case?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -635,13 +637,11 @@ void MarkGray(ContainerHeader* start) {
...
-      if (container->color() == CONTAINER_TAG_GC_GRAY) continue;
-    } else {
-       if (container->marked()) continue;
-    }
-    if (useColor) {
+      int color = container->color();
+      if (color == CONTAINER_TAG_GC_GRAY || color == CONTAINER_TAG_GC_GREEN) continue;
       container->setColor(CONTAINER_TAG_GC_GRAY);
     } else {
+      if (container->marked()) continue;
```
SvyatoslavScherbina: Do we have to traverse green objects here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: And you believe it is incorrect?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: Oh, guess I already made this mistake ;)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
SvyatoslavScherbina: No, I've just mentioned the fact.

runtime/src/main/cpp/Memory.cpp
olonho: Seems we shall.

PR #2369, opened 2018-11-23
WIP: use linked lists instead of external collections in memory manager.

olonho:


olonho:
Shows significant speedup, but not too concurrency friendly.

PR #2417, opened 2018-11-25
Add toCValues() extension functions to Arrays of unsigned types

martinoconnor:


msink:
Very inefficient, `UByteArray.refTo` is much better.

martinoconnor:
This matches the extension functions already available on the signed types such as `ByteArray.toCValues()`

martinoconnor:
Sometimes you want byref semantics whereby the called function can mutate your array and `refTo()` will work in this case. Sometimes you want immutability and ensure that the called function cannot modify the array. `toCValues()` effectively provides byval semantics for arrays.

PR #2355, opened 2018-11-20
Contracts support in K/N

PavelPunegov:
- Support contracts in native
- Refactor and cleanup `KonanDescriptorSerializer`
- Add missing contracts to `kotlin.test`

Fixes [KT-27770](https://youtrack.jetbrains.com/issue/KT-27770) and #2239

PR #2425, opened 2018-11-27
Update compiler to 1.3.20-dev-2050

PavelPunegov:
Should be merged into master with https://github.com/JetBrains/kotlin-native/pull/2355
Also excludes test

### DIFF
--- a/buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
+++ b/buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
@@ -889,7 +889,8 @@ fun runTest() {
 
     static def excludeList = [
             "build/external/compiler/codegen/box/functions/functionExpression/functionExpressionWithThisReference.kt", // KT-26973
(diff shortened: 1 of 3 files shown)


olonho:
All tests fail with "/opt/buildAgent/work/4d622a065c544371/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanDescriptorSerializer.kt: (159, 83): No value passed for parameter 'childSerializer'". Maybe combine two commits into one, so that we never have red tree.

PavelPunegov:
> All tests fail with "/opt/buildAgent/work/4d622a065c544371/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanDescriptorSerializer.kt: (159, 83): No value passed for parameter 'childSerializer'". Maybe combine two commits into one, so that we never have red tree.

I'll merge it and combine these changes then. It would be definitely better having a consistent tree

PavelPunegov:
Fixed with https://github.com/JetBrains/kotlin-native/commit/441bf924d4eca8d39f3d59432a4dd01634960ada

PR #2450, opened 2018-12-09
Make launcher.js work in Web Workers / Worklets

chicoxyzzy:
Since WebAssembly object should be exposed to Window/Worker/Worklet according to [it's WebIDL](https://webassembly.github.io/spec/js-api/index.html#webassembly-namespace)

### DIFF
--- a/runtime/src/launcher/js/launcher.js
+++ b/runtime/src/launcher/js/launcher.js
@@ -19,7 +19,7 @@ let heap;
 let global_arguments;
 
 function isBrowser() {
-    return typeof window !== 'undefined';
+    return typeof self !== 'undefined';
 }
 
 let runtime;


chicoxyzzy:
also `WebAssembly.compileStreaming` should be used instead of `WebAssembly.complie` + `WebAssembly.instantiate` when presented for better performance

bashor:
lgtm
cc @alexander-gorshenev

bashor:
> also WebAssembly.compileStreaming should be used instead of ...
Agree, PR welcome ;)

@alexander-gorshenev could you please write some instruction how to run tests for wasm?

alexander-gorshenev:
We have run the tests by ourselves.

chicoxyzzy:
@alexander-gorshenev so it's not possible to run test locally? I think tests could be useful for 
https://github.com/JetBrains/kotlin-native/pull/2477

PR #2451, opened 2018-12-10
More fault-tolerance in the Worker API.

olonho:


runtime/src/main/cpp/Exceptions.h
```diff
@@ -55,10 +55,11 @@ void RUNTIME_NORETURN ThrowNotImplementedError();
 // Throws illegal character conversion exception (used in UTF8/UTF16 conversions).
 void RUNTIME_NORETURN ThrowIllegalCharacterConversionException();
 void RUNTIME_NORETURN ThrowIllegalArgumentException();
+void RUNTIME_NORETURN ThrowIllegalStateException();
```
SvyatoslavScherbina: `ThrowIllegalStateException` is not `void`, it returns an object reference.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -24,15 +24,24 @@ import kotlinx.cinterop.*
...
          * better to use non-blocking IO combined with more lightweight coroutines.
          */
-        public fun start(): Worker = Worker(startInternal())
+        public fun start(errorReporting: Boolean = true): Worker = Worker(startInternal(errorReporting))
+
+        /**
+         * Return current worker, if known, null otherwise, for example, in the main thread or in platform
+         * thread without an associated worker.
+         */
+        public val current: Worker? get() {
```
SvyatoslavScherbina: The documentation is not clear enough to make a conclusion that `Worker.current != null` only in a thread that was created with `Worker.start()`.

olonho: Somewhat reworded.

runtime/src/main/cpp/Exceptions.h
olonho: Offline.

PR #2452, opened 2018-12-10
 Disable incorrect optimization for varargs and list of string literals

SvyatoslavScherbina:


benasher44:
Are there plans to re-enable this? I ended up offloading some calls to C functions with varargs into an inline C function in a .def file, which got me a pretty significant performance boost

SvyatoslavScherbina:
@benasher44 this patch shouldn't affect performance of C calls, it is mostly related to pure Kotlin function calls.
Performance of calls to C variadic functions is likely to become significantly better in Kotlin 1.3.30.

benasher44:
Ah awesome! Looking forward to that. Is that a PR that has landed already?

SvyatoslavScherbina:
This one: #2528 (merged month ago).

benasher44:
Thanks!

benasher44:
Wow libffi is no longer needed??

SvyatoslavScherbina:
Exactly!

benasher44:
👏👏👏👏

PR #2453, opened 2018-12-10
Fixes in kotlin.test

ilmat192:


backend.native/tests/build.gradle
```diff
@@ -2760,6 +2766,14 @@ task testing_custom_main(type: RunStandaloneKonanTest) {
             'Testing finished\n'
 }
 
+task testing_library_usage(type: RunStandaloneKonanTest) {
```
PavelPunegov: Do not run it on wasm32. It doesn't have exceptions

PR #2454, opened 2018-12-10
[kotlin compiler][update] 1.3.30-dev-610

vvlevchenko:


olonho:
Maybe we shall update to next 1.3.30 next Monday.

PR #2458, opened 2018-12-10
Avoid using Nothing in codegen as if it was an object.

olonho:


SvyatoslavScherbina:
```
open class A {
    open fun foo() = Any()
}

class B : A() {
    override fun foo() = throw Error()
}
```

Compilation with `-p framework` produces invalid module.

SvyatoslavScherbina:
Kotlin bridges for the following classes
```
open class A {
    open fun foo() = Any()
}

class B : A() {
    override fun foo() = throw Error()
}
```

are incorrect.

`void`-returning `B.foo` takes the same slot in vtable as reference-returning `A.foo`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -113,16 +114,19 @@ internal class ObjCExportCodeGenerator(
     fun FunctionGenerationContext.kotlinToObjC(
             value: LLVMValueRef,
             typeBridge: TypeBridge
-    ): LLVMValueRef = when (typeBridge) {
-        is ReferenceBridge -> kotlinReferenceToObjC(value)
-        is ValueTypeBridge -> kotlinToObjC(value, typeBridge.objCValueType)
-    }
+    ): LLVMValueRef = when {
+            LLVMTypeOf(value) == voidType -> kNullInt8Ptr
```
SvyatoslavScherbina: This is incorrect, because Objective-C method can have non-pointer return type, e.g. if `Nothing`-returning Kotlin method overrides `Int`-returning method.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -919,13 +922,8 @@ private fun ObjCExportCodeGenerator.createTypeAdapter(
...
         } else {
             // Mark it as non-overridable:
-            baseMethods.distinctBy { namer.getSelector(it) }.forEach { base ->
-                reverseAdapters += KotlinToObjCMethodAdapter(
-                        namer.getSelector(base),
-                        -1,
-                        -1,
-                        NullPointer(int8Type)
-                )
+            baseMethods.distinctBy { namer.getSelector(it) }.forEach { baseMethod ->
```
SvyatoslavScherbina: This is wrong.
The following test should fail:
```
// Kotlin
open class A {
    fun foo() {}
}

// Swift
class B : A {
    override func foo() {}
}

B()
```

However with this patch it doesn't fail.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -962,7 +960,7 @@ private fun ObjCExportCodeGenerator.createTypeAdapter(
 
     when (descriptor.kind) {
         ClassKind.OBJECT -> {
-            classAdapters += if (descriptor.isUnit()) {
+            classAdapters += if (descriptor.isUnit() || descriptor.isNothing()) {
```
SvyatoslavScherbina: This code handles `descriptor.kind == ClassKind.OBJECT` case.
Is there `object Nothing`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -1056,7 +1054,7 @@ private fun ObjCExportCodeGenerator.createObjectInstanceAdapter(
         descriptor: ClassDescriptor
 ): ObjCExportCodeGenerator.ObjCToKotlinMethodAdapter {
     assert(descriptor.kind == ClassKind.OBJECT)
-    assert(!descriptor.isUnit())
+    assert(!descriptor.isUnit() && !descriptor.isNothing())
```
SvyatoslavScherbina: This code handles `descriptor.kind == ClassKind.OBJECT` case.
Is there `object Nothing`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
olonho: Why shall it fail?

olonho: It seems, it fails in the runtime with ` 'NSGenericException', reason: '[op.B foo] can't be implemented'` while better fail in the compile time. Anyway, restored old behavior.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -974,7 +978,8 @@ private fun ObjCExportCodeGenerator.createTypeAdapter(
             }
         }
         else -> {
-            // Nothing special.
+            if (descriptor.isNothing())
```
SvyatoslavScherbina: This code adds `unit` class method to `Nothing` Objective-C wrapper class, and that method returns `Unit` instance.
What's the purpose of doing this?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: Unfortunately, this doesn't seem to be possible in the compile time as Objective-C doesn't have a concept of `final` methods.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
olonho: Maybe none, then, removed.

PR #2461, opened 2018-12-11
Common kotlin.test

PavelPunegov:
See the first commit. The second one is to reformat code.


PavelPunegov:
This PR requires https://github.com/JetBrains/kotlin/commit/0ac74ba32305aa7a9fad79bc85a92a0b8f591f9b

runtime/src/main/kotlin/kotlin/test/Annotation.kt
```diff
@@ -46,7 +46,7 @@ public annotation class AfterEach
  */
 @Retention(AnnotationRetention.SOURCE)
 @Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
-public annotation class Ignore
+public actual annotation class Ignore
 
-public typealias AfterTest = AfterEach
-public typealias BeforeTest = BeforeEach
+public actual typealias AfterTest = AfterEach
+public actual typealias BeforeTest = BeforeEach
```
ilya-g: Do we want to unify names of `BeforeTest`/`AfterTest` annotations by deprecating `BeforeEach`/`AfterEach`?

runtime/src/main/kotlin/kotlin/test/DefaultAsserter.kt
```diff
@@ -8,7 +8,7 @@ package kotlin.test
 /**
  * Default [Asserter] implementation for Kotlin/Native.
  */
-public object DefaultAsserter : Asserter {
+public object DefaultObjectAsserter : Asserter {
```
ilya-g: Perhaps we can have an instance of common `DefaultAsserter` in a `@SharedImmutable` top-level val instead?

runtime/src/main/kotlin/kotlin/test/Annotation.kt
ilmat192: It sounds reasonable.

runtime/src/main/kotlin/kotlin/test/DefaultAsserter.kt
PavelPunegov: There are no need actually to have `DefaultAsserter` under this annotation. Maybe it could be better to fix todo here and use this asserter https://github.com/JetBrains/kotlin/blob/2e12834b6a529f63e7a7955f4b5c4db24ee7adf9/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/DefaultAsserter.kt#L12

runtime/src/main/kotlin/kotlin/test/Annotation.kt
PavelPunegov: This is a good point. I can rename `After/BeforeEach` into `After/BeforeTest` and both `*Each`' make a typealiases with a deprecation annotation.

PavelPunegov:
Already in master

PR #2469, opened 2018-12-14
Fix memory leak, improve logging.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -639,11 +640,47 @@ inline void initThreshold(MemoryState* state, uint32_t gcThreshold) {
...
+  if (relativeName) konan::free(relativeName);
+}
+
+void dumpContainerContent(ContainerHeader* container) {
+  if (isAggregatingFrozenContainer(container)) {
+    MEMORY_LOG("%s aggregating container %p with %d objects rc=%d\n",
+               colorNames[container->color()], container, container->objectCount(), container->refCount());
+    ContainerHeader** subContainer = reinterpret_cast<ContainerHeader**>(container + 1);
+    for (int i = 0; i < container->objectCount(); ++i) {
+      ObjHeader* obj = reinterpret_cast<ObjHeader*>(subContainer + 1);
```
SvyatoslavScherbina: It this correct?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -639,11 +640,47 @@ inline void initThreshold(MemoryState* state, uint32_t gcThreshold) {
...
+      MEMORY_LOG("    object %p of type %p: ", obj, obj->type_info());
+      dumpObject(obj, 4);
+    }
+  } else {
+    MEMORY_LOG("%s regular %s%scontainer %p with %d objects rc=%d\n",
```
SvyatoslavScherbina: Missing spaces.

runtime/src/main/cpp/Memory.cpp
olonho: Why not, supercontainer contains references to frozen objects containers.

runtime/src/main/cpp/Memory.cpp
olonho: Not really, we add spaces to strings.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: `obj` seems to be the same on all loop iterations, doesn't it?
Is it really located at `container + sizeof(ContainerHeader) + sizeof(ContainerHeader*)`?

PR #2471, opened 2018-12-17
Rewrote application with benchmarks using MPP plugin and added collec…

LepilkinaElena:
Changes in application with benchmarks:

- MPP plugin
- Json format of report


performance/build.gradle
```diff
@@ -137,8 +187,51 @@ class Results {
...
+    }
+    
+    Object getInstance() {
+        return instance
+    }
+}
+
+class Machine extends JsonSerializable {
+    Machine() {
+        instance.metaClass.os =  System.getProperty("os.name") // OperatingSystem.current().getName()
```
LepilkinaElena: Decided to do in one way as other found ways don't cover all needed information.  Saved possible wrappers in comments.

performance/src/main/kotlin/org/jetbrains/ring/launcher.kt
```diff
@@ -24,12 +24,36 @@ val BENCHMARK_SIZE = 100
 
 //-----------------------------------------------------------------------------//
 
-class Launcher(val numWarmIterations: Int, val numberOfAttempts: Int) {
+class Launcher(val numWarmIterations: Int, val numberOfAttempts: Int, val jsonReport: String?) {
     class Results(val mean: Double, val variance: Double)
 
+    enum class Status(val value: String) {
```
LepilkinaElena: I plan to replace these classes later to use in app for analysis.

performance/src/main/kotlin/org/jetbrains/ring/launcher.kt
```diff
@@ -24,12 +24,36 @@ val BENCHMARK_SIZE = 100
...
+        """
+
+        }
+    }
+
     val results = mutableMapOf<String, Results>()
+    val bechmarkResults = mutableListOf<BenchmarkResult>()
 
-    fun launch(benchmark: () -> Any?): Results {                          // If benchmark runs too long - use coeff to speed it up.
+    fun launch(benchmark: () -> Any?, name: String): Unit {                          // If benchmark runs too long - use coeff to speed it up.
```
sbogolepov: It's better to move writing to JSON out from `Launcher` to separate class because the only responsibility of `Launcher` is benchmark running.

performance/src/main/kotlin/org/jetbrains/ring/launcher.kt
```diff
@@ -24,12 +24,36 @@ val BENCHMARK_SIZE = 100
...
 
+    enum class Status(val value: String) {
+        PASSED("PASSED"),
+        FAILED("FAILED")
+    }
+
+    data class BenchmarkResult(val name: String, val status: Status,
+                          val score: Double, val runtimeInUs: Double,
+                          val repeat: Int, val warmup: Int) {
+        override fun toString(): String {
```
sbogolepov: What if we will want to dump `BenchmarkResult` as CSV?

performance/src/main/kotlin/main.kt
```diff
@@ -27,6 +28,11 @@ fun main(args: Array<String>) {
             numWarmIterations = args[0].toInt()
             numberOfAttempts = args[1].toInt()
         }
+        3 -> {
+            numWarmIterations = args[0].toInt()
+            numberOfAttempts = args[1].toInt()
+            jsonReport = args[2].toString()
```
sbogolepov: `jsonReport` should be also noted in a help message below

performance/src/main/kotlin-native/org/jetbrains/ring/Utils.kt
```diff
@@ -46,3 +48,19 @@ class Random() {
         }
     }
 }
+
+//-----------------------------------------------------------------------------//
+
+class FileWriter() {
```
sbogolepov: Maybe use `expect`/`actual`?

LepilkinaElena: Ok, thank you.

LepilkinaElena:
Fixed comments from Sergey

performance/src/main/kotlin-native/org/jetbrains/ring/Utils.kt
```diff
@@ -46,3 +48,19 @@ class Random() {
         }
     }
 }
+
+//-----------------------------------------------------------------------------//
+
+actual class FileWriter() {
+    actual companion object {
+        actual fun writeToFile(fileName: String, text: String): Unit {
+            val file = fopen(fileName, "wt")
```
homuroll: Here the elvis operator could be used:
`val file = fopen(...) ?: error("...")`

performance/src/main/kotlin/org/jetbrains/ring/JsonReportCreator.kt
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.ring
+
+class JsonReportCreator(val data: Iterable<BenchmarkResult>) {
+    fun printJsonReport(jsonReport: String): Unit {
+        val reportText = data.joinToString(prefix = "[", postfix = "]") { it ->
```
homuroll: `it ->` could be omitted.

performance/src/main/kotlin/org/jetbrains/ring/launcher.kt
```diff
@@ -96,6 +102,7 @@ class Launcher(val numWarmIterations: Int, val numberOfAttempts: Int) {
         runOctoTest()
 
         printResultsNormalized()
```
sbogolepov: Maybe move `printResultsNormalized` from `Launcher` just like you moved JSON dumping before? And implement following logic: if user provides `jsonReport` in `main` we write results to file and to stdout otherwise. 
Also it makes sense to remove `Results` in favor for `BenchmarkResult`.

performance/src/main/kotlin/org/jetbrains/ring/BenchmarkResult.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class BenchmarkResult(val name: String, val status: Status,
+                      val score: Double, val runtimeInUs: Double,
+                      val repeat: Int, val warmup: Int) {
+    enum class Status(val value: String) {
+        PASSED("PASSED"),
+        FAILED("FAILED")
+    }
+
+    fun toJson(): String {
```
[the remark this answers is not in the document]
LepilkinaElena: I a bit disagree in this point. In my opinion, it's resposibility of entity to know how it should be converted to another representation.

performance/src/main/kotlin/org/jetbrains/ring/launcher.kt
LepilkinaElena: I stay current flow working as is until all analysis isn't placed in separate project. When it's ready I'll remove extra logic from benchmark application.

performance/src/main/kotlin/org/jetbrains/ring/BenchmarkResult.kt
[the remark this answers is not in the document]
sbogolepov: I won't insist because in this trivial case it's fine, but in more complicated cases it will lead to bloated code with unnecessary cyclic dependencies between classes.

performance/build.gradle
```diff
@@ -137,8 +187,51 @@ class Results {
     }
 }
 
+class JsonSerializable {
```
olonho: Complex logic better be in Kotlin and interoped with.

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
```diff
@@ -0,0 +1,66 @@
...
+    var workingDir: Any = project.projectDir
+    var outputFileName: String? = null
+    private var curArgs: List<String> = emptyList()
+    private val curEnvironment: MutableMap<String, Any> = mutableMapOf()
+
+    fun args(vararg args: Any) {
+        curArgs = args.map { it.toString() }
+    }
+
+    fun environment(map: Map<String, Any>) {
```
olonho: `Map<String, String>`?

performance/src/main/kotlin-native/org/jetbrains/ring/Utils.kt
```diff
@@ -46,3 +48,15 @@ class Random() {
...
     }
 }
+
+//-----------------------------------------------------------------------------//
+
+
+actual fun writeToFile(fileName: String, text: String) {
+    val file = fopen(fileName, "wt") ?: error("Cannot write file '$fileName'")
+    try {
+        fputs(text, file)
```
olonho: `if (fputs() == EOF) throw Error("File write error")`?

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
LepilkinaElena: As I understood this type is predefined in plugin. I need such type because project.exec.environment has such type.

PR #2477, opened 2018-12-18
Use WebAssembly.instantiateStreaming when presented in browser

chicoxyzzy:
As I proposed in 
https://github.com/JetBrains/kotlin-native/pull/2450#issuecomment-445585728

Still not sure where the tests are and how to run them.

### DIFF
--- a/runtime/src/launcher/js/launcher.js
+++ b/runtime/src/launcher/js/launcher.js
@@ -210,15 +210,27 @@ function instantiateAndRunSync(arraybuffer, args) {
     return invokeModule(instance, args)
 }
 
+
+// Instantiate module in Browser as a promise of streaming instantiation.
+function instantiateAndRunStreaming(filename) {
+    linkJavaScriptLibraries();
+    WebAssembly.instantiateStreaming(fetch(filename), konan_dependencies)
+      .then(resultObject => invokeModule(resultObject.instance, [filename]));
+}
+
 konan.moduleEntry = function (args) {
     if (isBrowser()) {
         if (!document.currentScript.hasAttribute("wasm")) {
             throw new Error('Could not find the wasm attribute pointing to the WebAssembly binary.');
         }
         const filename = document.currentScript.getAttribute("wasm");
-        fetch(filename)
-            .then(response => response.arrayBuffer())
-            .then(arraybuffer => instantiateAndRun(arraybuffer, [filename]));
+        if (typeof WebAssembly.instantiateStreaming === 'function') {
+            instantiateAndRunStreaming(filename);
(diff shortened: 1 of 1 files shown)


runtime/src/launcher/js/launcher.js
```diff
@@ -195,11 +195,17 @@ function invokeModule(inst, args) {
 
 // Instantiate module in Browser in a sequence of promises.
 function instantiateAndRun(arraybuffer, args) {
-    WebAssembly.compile(arraybuffer)
-        .then(module => {
-            linkJavaScriptLibraries();
-            return WebAssembly.instantiate(module, konan_dependencies);
-        }).then(instance => invokeModule(instance, args));
+    if (typeof WebAssembly.compileStreaming === 'function') {
+        linkJavaScriptLibraries();
```
chicoxyzzy: should it be moved to line before if statement?

chicoxyzzy:
@olonho yes, this instantiates and compiles wasm module using streams so resulting instance will be ready as far as fetch finishes streaming.

chicoxyzzy:
Do you want me to explain these changes in more details?

runtime/src/launcher/js/launcher.js
```diff
@@ -216,9 +216,15 @@ konan.moduleEntry = function (args) {
             throw new Error('Could not find the wasm attribute pointing to the WebAssembly binary.');
         }
         const filename = document.currentScript.getAttribute("wasm");
-        fetch(filename)
-            .then(response => response.arrayBuffer())
-            .then(arraybuffer => instantiateAndRun(arraybuffer, [filename]));
+        if (typeof WebAssembly.instantiateStreaming === 'function') {
+            linkJavaScriptLibraries();
+            WebAssembly.instantiateStreaming(fetch(filename), konan_dependencies)
+              .then(resultObject => invokeModule(resultObject.instance, [filename]));
```
alexander-gorshenev: May be extract this scenarion as a third function, similar to instantiateAndRun and instantiateAndRunSync.

chicoxyzzy: Should it take only filename as an argument? Like
```js
// Instantiate module in Browser as a promise of streaming instantiation.
function instantiateAndRunStreaming(filename) {
    linkJavaScriptLibraries();
    WebAssembly.instantiateStreaming(fetch(filename), konan_dependencies)
      .then(resultObject => invokeModule(resultObject.instance, [filename]));
}
```

alexander-gorshenev: that's okay, I suppose.

PR #2480, opened 2018-12-19
Update OBJC_INTEROP.md

JThramer:
`NSMutableString` is an Objective-C type, not a Swift type. Mutability is a core concept in Swift with `let` vs `var`.

Example: 

```swift
var test: String = "Hello!"
print(test)
test = "World!"
print(test)
```

Note that no `NSMutableString` type was used, and yet I could mutate the string. Magic!


SvyatoslavScherbina:
Swift has `NSMutableString` type.
Objective-C `NSMutableString*` type is mapped to `NSMutableString` type in Swift (when importing Objective-C headers to Swift).
The opposite is also true: `NSMutableString` type in Swift is mapped to `NSMutableString*` type in Objective-C (when generating Objective-C header from Swift module).

This row in the table seems correct to me: it declares that Objective-C and Swift `NSMutableString`  types (mapped to each other) are translated to `String` when imported to Kotlin.

olonho:
No response, closing.

PR #2488, opened 2018-12-21
Improve symbolicator, generate better debug info

olonho:


samples/symbolication/src/symbolicationMain/kotlin/main.kt
```diff
@@ -21,6 +21,32 @@ fun archToCpuName(arch: String): cpu_type_t = when (arch) {
     else -> TODO("unsupported $arch")
 }
 
+fun sourceInfoString(owner: CValue<CSSymbolOwnerRef>, address: ULong): String? {
+    var sourceInfo: CValue<CSTypeRef>? = CSSymbolOwnerGetSourceInfoWithAddress(owner, address)
+    if (sourceInfo == null || CSIsNull(sourceInfo)) return null
```
SvyatoslavScherbina: I guess `CSSymbolOwnerGetSourceInfoWithAddress` can't return `null`, so it makes sense to use `!!` here.

PR #2489, opened 2018-12-21
Add line numbers to stacktraces in macos_x64/ios_x64 debug binaries

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -106,17 +108,33 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
...
-    internal val defaultNativeLibraries = 
-        if (produce == CompilerOutputKind.PROGRAM) 
-            File(distribution.defaultNatives(target)).listFiles.map { it.absolutePath } 
-        else emptyList()
+    internal val defaultNativeLibraries: List<String> = mutableListOf<String>().apply {
+        add(if (debug) "debug.bc" else "release.bc")
+
+        if (produce == CompilerOutputKind.PROGRAM) {
+            add("start.bc")
+            add("launcher.bc")
```
alexander-gorshenev: We'd better keep the details of distribution file names and locations in the Distribution, I think. The way we reference runtime.bc there.

runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -171,7 +172,17 @@ OBJ_GETTER(GetStackTraceStrings, KConstRef stackTrace) {
     RuntimeCheck(symbols != nullptr, "Not enough memory to retrieve the stacktrace");
     AutoFree autoFree(symbols);
     for (int index = 0; index < size; ++index) {
-      CreateStringFromCString(symbols[index], ArrayAddressOfElementAt(strings->array(), index));
+      auto sourceInfo = Kotlin_getSourceInfo(*PrimitiveArrayAddressOfElementAt<KNativePtr>(stackTrace->array(), index));
+      const char* symbol = symbols[index];
+      const char* result;
+      char line[512];
```
olonho: Is buffer sufficiently big.

PR #2490, opened 2018-12-22
Zephyr ARM sysroot

jwoolston:
Adds script for generating sysroot tarball from zephyr sdk. I have successfully built using this sysroot by modifying my platform konan properties and manually adding the sysroot to `.konan/dependencies` and adding it to the extracted file. I will submit a separate PR for updating the konan properties. This PR assumes you have the Zephyr SDK directory installed at either the default location (`/opt/zephyr-sdk`) or an environment variable (`ZEPHYR_SDK_INSTALL_DIR`) set to the correct location.

As part of reviewing this I would like to point out some of the broader issues with this sysroot. First, zephyr provides an sdk (what this sysroot is generated from) that provides more than simply arm support and their CMake scripts already determine the correct sysroot (it is available in their cmake). Kotlin native LLVM could access these and then support would be very universally broadened to all of Zephyr. 

I have successfully configured the zephyr code with a kotlin native project to build using the Zephyr SDK with the existing arm toolchain/wasm sysroot for kotlin. At present zephyr's C++ support is most notably missing `new` which I believe is a problem for kotlin native's runtime code at present.

Alternatively, zephyr has full support for being built with GNU ARM tools (assuming an ARM target) and in this situation if I understand correctly will use the GNU ARM provided sysroot. Kotlin native already distributes this and with a small modification, the separate sysroot could be distributed in a way that fits with existing model. If additional architectures were desired, they could be added like any other assuming LLVM compatible toolchains (i.e. GNU MIPS). 

In my opinion, it would be best to use the Zephyr SDK with Zephyr sysroot or GNU ARM toolchain with GNU ARM sysroot, either of which would mean this PR is invalid. Since both are GNU licensed, I don't see that it makes much of a difference in that regard, the issue would be more the desired path by those in charge/the community.

### DIFF
--- a/tools/scripts/update_zephyr.sh
+++ b/tools/scripts/update_zephyr.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+KONAN_TOOLCHAIN_VERSION=1
+TARBALL_zephyr_arm=target-sysroot-$KONAN_TOOLCHAIN_VERSION-zephyr-arm
+OUT=`pwd`
+
+if [ -z "ZEPHYR_SDK_INSTALL_DIR" ]; then
+    echo "Using default Zephyr SDK install location"
+    export ZEPHYR_SDK_INSTALL_DIR="/opt/zephyr-sdk"
+fi
+
+sdk=armv5-zephyr-eabi
+p=$ZEPHYR_SDK_INSTALL_DIR/sysroots/$sdk/usr
+echo "Packing SDK $sdk as $OUT/$TARBALL_zephyr_arm.tar.gz..."
+tar -czvf $OUT/$TARBALL_zephyr_arm.tar.gz -C $p .


msink:
Well, maybe it simplifies setup on Linux, but please do not drop support for Mac and Windows hosts.
Zephyr SDK works only on Linux, AFAIK.

msink:
My setup on Windows is simple:
- `git`, `cmake` and `python3-x64` installed as usual on Windows, from their sites (and should be already installed on any sane developers computer)
- a bit tricky are `ninja`, `dtc` and `gperf` - I installed them in msys2-mingw64 and copyied to directory listed in system PATH  (zipped and attached, just in case: [windows-ninja-dtc-gperf.zip](https://github.com/JetBrains/kotlin-native/files/2705902/windows-ninja-dtc-gperf.zip))

jwoolston:
I have not yet tried to use zephyr under windows so I will take your word for it. With that in mind would it not make sense to just use the sysroot provided by GNU ARM, especially since Zephyr has full support for using that toolchain and their CMAKE scripts properly handle this possibility?

msink:
Maybe, but there can be subtle incompatibilities between Kotlin runtime compiled by clang and sysroot compiled by gcc.
Ideally everything should be compiled by clang.

msink:
There even is long standing issue in Zephyr repo for LLVM support: zephyrproject-rtos/zephyr#3689
But seems that nobody cares about it, as it moves on roadmap plans from release to release.

PR #2493, opened 2018-12-24
Write benchmarks analyzer to compare benchmarks runs.

LepilkinaElena:
Written tool allows to compare json reports and get flexible comparasion report (example attached).
[report.txt](https://github.com/JetBrains/kotlin-native/files/2707151/report.txt)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -462,10 +462,11 @@ task uploadBundle {
     }
 }
 
-task performance {
+// TODO return when problems with HostManagers will be solved
+/*task performance {
     dependsOn 'dist'
     dependsOn ':performance:bench'
-}
+}*/
 
 task teamcityKonanVersion {
     doLast {


tools/benchmarks/shared/src/main/kotlin/org/jetbrains/report/BenchmarksReport.kt
```diff
@@ -107,14 +107,28 @@ data class BenchmarksReport(val env: Environment, val benchmarks: List<Benchmark
...
+        private fun structBenchmarks(benchmarksList: List<BenchmarkResult>): Map<String, List<BenchmarkResult>> {
+            val result = mutableMapOf<String, MutableList<BenchmarkResult>>()
+            for (bench in benchmarksList) {
+                if (bench.name !in result) {
+                    result[bench.name] = mutableListOf<BenchmarkResult>()
+                }
+                result[bench.name]?.add(bench)
+            }
+            return result
+        }
```
olonho: More Kotlin'ish would be
```benchmarksList.forEach {
   result.getOrPut(it.name) { mutableListOf() } += it
}```

File: tools/benchmarksAnalyzer/src/main/kotlin-jvm/org/jetbrains/analyzer/Utils.kt [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.jetbrains.benchmarksAnalyzer
+
+import java.io.File
+import java.io.InputStream
+
+actual fun readFile(fileName: String): String {
+    val inputStream: InputStream = File(fileName).inputStream()
```
olonho: No need to specify explicit types.

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
```diff
@@ -0,0 +1,53 @@
...
+    } finally {
+        fclose(file)
+    }
+    return text
+}
+
+actual fun format(number: Double, decimalNumber: Int): String {
+    var str = ByteArray(1024)
+    val formatStr = "%.${decimalNumber}f"
+    sprintf(str.refTo(0), formatStr, number)
```
olonho: snprintf, no need for formatStr variable.

tools/benchmarksAnalyzer/src/main/kotlin/main.kt
```diff
@@ -0,0 +1,86 @@
...
+}
+
+fun main(args: Array<String>) {
+    var mainReport: String? = null
+    var compareToReport: String? = null
+    var outputFile: String? = null
+    var eps: Double = 0.5
+    var shortReport = false
+
+    // Parse args
```
olonho: Please dot at the end of comments.

tools/benchmarksAnalyzer/src/main/kotlin/main.kt
```diff
@@ -0,0 +1,86 @@
...
+fun printError(message: String) {
+    val usageDesc = "Usage: benchComparator main_report report_to_compare_to [-o output_file] [-eps eps_value] [-short]\n"
+    error("${usageDesc}${message}")
+}
+
+fun main(args: Array<String>) {
+    var mainReport: String? = null
+    var compareToReport: String? = null
+    var outputFile: String? = null
+    var eps: Double = 0.5
```
olonho: No need for Double type, https://github.com/JetBrains/kotlinconf-spinner/blob/master/getopt/src/main/kotlin/kliopt/KliOpt.kt could be reused.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -0,0 +1,84 @@
...
+
+// Entity to describe avarage values which conssists of mean and variance values.
+data class MeanVariance(val mean: Double, val variance: Double) {
+    override fun toString(): String {
+        val format = { number: Double -> format(number, 2)}
+        return "${format(mean)}+-${format(variance)}"
+    }
+}
+
+// Composite benchmark which descibe avarage result for several runs and contains mean and variance value
```
olonho: describes

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -0,0 +1,84 @@
...
+                        abs(other.meanBenchmark.score - other.varianceBenchmark.score)
+
+        val variance = abs(abs(mean) - max(minValueChange, maxValueChange))
+        return MeanVariance(mean * 100, variance * 100)
+    }
+
+    // Calculate ratio value compare to another.
+    fun calcRatio(other: MeanVarianceBench): MeanVariance {
+        val mean = meanBenchmark.score / other.meanBenchmark.score
+        val minRatio = (meanBenchmark.score - varianceBenchmark.score) / (other.meanBenchmark.score + other.varianceBenchmark.score)
```
olonho: Need to check for zero divisor?

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -0,0 +1,84 @@
...
+        val maxRatio = (meanBenchmark.score + varianceBenchmark.score) / (other.meanBenchmark.score - other.varianceBenchmark.score)
+        val ratioConfInt = min(abs(minRatio - mean), abs(maxRatio - mean))
+        return MeanVariance(mean, ratioConfInt)
+    }
+
+    override fun toString(): String {
+        val format = { number: Double -> format(number, 4)}
+        return "${format(meanBenchmark.score)}+-${format(varianceBenchmark.score)}"
```
olonho: Why not just use default argument of 4 to two argument format?

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+// Class with different statistical operations
+class Statistics {
+    companion object {
+        private val zStar = 1.96
+
+        fun geometricMean(values: List<Double>): Double {
```
olonho: `fun geometricMean(values: List<Double>) = values.map { it.pow(1.0 / values.size) }.reduce { a, b -> a * b }` feels more functional.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -0,0 +1,84 @@
...
+            val iPow = 1.0 / values.size
+            return values.map { it.pow(iPow) }.reduce { a, b -> a * b }
+        }
+
+        fun getMeanVariance(samples: List<Double>): MeanVariance {
```
olonho: Not sure it's `getSmth` kind of function, either `computeMeanVariance` or just `meanVariance`.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,418 @@
...
+typealias BenchTable = Map<String, MeanVarianceBench>
+typealias SummaryBenchTable = Map<String, SummaryBench>
+typealias ScoreChange = Pair<MeanVariance, MeanVariance>
+
+// Summary report with comparasion of separate benchmarks results
+class SummaryBenchmarksReport (val currentReport: BenchmarksReport,
+                               val previousReport: BenchmarksReport? = null,
+                               val meaningfulChangesValue: Double = 0.5) {
+
+    // Lists of benchmarks in different status
```
olonho: Dots at end of comments here and in other places.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,418 @@
...
+    private val failedBenchs = mutableListOf<String>()
```
olonho: Why not lateinit immutable lists?

LepilkinaElena: I changed where it is possible, but somewhere buckets are intersected. Separate filling of them added extra looping over results.

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
```diff
@@ -0,0 +1,60 @@
...
+    val file = fopen(fileName, "wt") ?: error("Cannot write file '$fileName'")
+    try {
+        if (fputs(text, file) == EOF) throw Error("File write error")
+    } finally {
+        fclose(file)
+    }
+}
+
+actual fun exitProcess(status: Int) {
+    kotlin.system.exitProcess(status)
```
olonho: What is the point of having two identical actuals?

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
```diff
@@ -0,0 +1,60 @@
...
+}
+
+actual fun assert(value: Boolean, lazyMessage: () -> Any) {
+    kotlin.assert(value, lazyMessage)
+}
```
olonho: Same here.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+        val geoMean = Statistics.geometricMean(benchTable.toList().map { (_, value) -> value.meanBenchmark.score })
+        val varianceGeoMean = Statistics.geometricMean(benchTable.toList().map { (_, value) -> value.varianceBenchmark.score })
+        val meanBench = BenchmarkResult(geoMeanBenchName, geoMean)
+        val varianceBench = BenchmarkResult(geoMeanBenchName, varianceGeoMean)
+        return MeanVarianceBench(meanBench, varianceBench)
+    }
+
+    // Merge current and compare to report.
+    private fun createMergedReport(curBenchs: BenchTable, prevBenchs: BenchTable?) {
+        curBenchs.forEach { (name, current) ->
```
olonho: Please no shortened variable names, use `currentBenchmarks` here.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+                    if (curBench.status == BenchmarkResult.Status.FAILED) {
+                        newFailures.add(statusChange)
+                    } else {
+                        newPasses.add(statusChange)
+                    }
+                }
+            }
+        }
+
+        // Find removed becnhmarks.
```
olonho: benchmarks

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+                    curStatus = result.status
+                }
+                repeatSeq[index] = result.repeat
+                if (currentWarmup != -1)
+                    if (result.warmup != currentWarmup)
+                        println("Check data consistency. Warmup value for benchmark '${result.name}' differs.")
+                currentWarmup = result.warmup
+            }
+
+            repeatSeq.sort()
```
olonho: Same, `repeatedSequence`?

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+                append("Performance summary")
+                append(headerSeparator)
+                printPerformanceBucket(regressions, "Regressions")
+                printPerformanceBucket(improvements, "Improvements")
+                append()
+            }
+        }
+
+        private fun formatValue(number: Double, isPercent: Boolean = false): String {
+            if (isPercent) {
```
olonho: `private fun formatValue(number: Double, isPercent: Boolean = false) = if (isPercent) format(number, 2) + "%" else format(number)`

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+                return format(number, 2) + "%"
+            }
+            return format(number)
+        }
+
+        private fun formatColumn(content:String, isWide: Boolean = false): String {
```
olonho: Same

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+            // Print geometric mean.
+            val geoMeanChangeMap = geoMeanScoreChange?.
+                                    let { mapOf(geoMeanBench.first!!.meanBenchmark.name to geoMeanScoreChange!!) }
+            printBenchmarksDetails(mutableMapOf(geoMeanBench.first!!.meanBenchmark.name to geoMeanBench),
+                                         geoMeanChangeMap)
+            printTableLineSeparator()
+            printBenchmarksDetails(fullSet, regressions)
+            printBenchmarksDetails(fullSet, improvements)
+            if (!onlyChanges) {
+                // Print all left results.
```
olonho: left -> remaining

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -0,0 +1,225 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.jetbrains.kliopt
+
+import org.jetbrains.benchmarksAnalyzer.exitProcess
```
olonho: Nope, better not use exitProcess in relatively general purpose library code.

tools/benchmarksAnalyzer/src/tests/AnalyzerTests.kt
```diff
@@ -0,0 +1,102 @@
...
+    fun testGeoMean() {
+        val numbers: List<Double> = listOf(4.0, 6.0, 9.0)
+        val value = Statistics.geometricMean(numbers)
+        val expected = 6.0
+        assertTrue(abs(value - expected) < eps)
+    }
+
+    @Test
+    fun testComputeMeanVariance() {
+        val numbers: List<Double> = listOf(10.1, 10.2, 10.3)
```
olonho: Type inference shall work here.

tools/benchmarksAnalyzer/src/tests/AnalyzerTests.kt
```diff
@@ -0,0 +1,102 @@
...
+        val ratio = inputs.first.calcRatio(inputs.second)
+        val expectedMean = 0.9
+        val expectedVariance = 0.00001899
+
+        assertTrue(abs(ratio.mean - expectedMean) < eps)
+        assertTrue(abs(ratio.variance - expectedVariance) < eps)
+    }
+
+    /* Possible tests
+    fun testStableBenchmarks() {
```
olonho: Not sure what is meant to be here.

olonho: Either fill it with smth meaningful or remove.

tools/benchmarks/shared/src/main/kotlin/org/jetbrains/report/BenchmarksReport.kt
```diff
@@ -107,14 +107,25 @@ data class BenchmarksReport(val env: Environment, val benchmarks: List<Benchmark
                 error("Benchmarks field is expected to be an array. Please, check origin files.")
             }
         }
+
+        // Made a map of becnhmarks with name as key from list.
+        private fun structBenchmarks(benchmarksList: List<BenchmarkResult>): Map<String, List<BenchmarkResult>> {
```
olonho: Can be even shorter:
```
private fun structBenchmarks(benchmarksList: List<BenchmarkResult>) =
   mutableMapOf<String, MutableList<BenchmarkResult>>.apply {
      benchmarksList.forEach {
                 getOrPut(it.name) { mutableListOf() } += it
             }
      }

File: tools/benchmarksAnalyzer/src/main/kotlin-jvm/org/jetbrains/analyzer/Utils.kt [hunk 25]
```diff
@@ -0,0 +1,45 @@
...
+import java.io.InputStream
+
+actual fun readFile(fileName: String): String {
+    val inputStream = File(fileName).inputStream()
+    val inputString = inputStream.bufferedReader().use { it.readText() }
+    return inputString
+}
+
+actual fun format(number: Double, decimalNumber: Int): String {
+    val formatStr = "%.${decimalNumber}f"
```
olonho: No need in excessive temporary.

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
```diff
@@ -0,0 +1,60 @@
...
+
+package org.jetbrains.benchmarksAnalyzer
+
+import platform.posix.*
+import kotlinx.cinterop.CValuesRef
+import kotlinx.cinterop.*
+
+actual fun readFile(fileName: String): String {
+    val file = fopen(fileName, "r") ?: error("Cannot write file '$fileName'")
+    var str = ByteArray(1024)
```
olonho: str -> buffer

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
LepilkinaElena: Because this function exists only for JVM and Native as written in docs https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/exit-process.html

To use it in common part we need to use expect/actual mechanism because it is absent in common part of stdlib.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/kliopt/KliOpt.kt
LepilkinaElena: Sorry, I a bit don't understand. What then is better way to exit program after using help flag? All known by me arg parsers added help flag and behaviour for it by themselves.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+            bucket?.let {
+                // There are changes in performance.
+                // Output changed benchmarks.
+                for ((name, change) in bucket) {
+                    append("${formatColumn(name, true)} " +
+                            "${formatColumn(fullSet[name]!!.first!!.toString())} " +
+                            "${formatColumn(fullSet[name]!!.second!!.toString())} " +
+                            "${formatColumn(change.first.toString() + " %")} " +
+                            "${formatColumn(change.second.toString())}")
+                    fullSet.remove(name)
```
sbogolepov: It is really unintuitive that printing function is also removing entries from its argument.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+        // Sort regressions and improvements.
+        regressions = performanceChanges["regressions"]
+                ?.sortedByDescending { it.second.first.mean }?.map { it.first to it.second }
+                ?.toMap() ?: mapOf<String, ScoreChange>()
+        improvements = performanceChanges["improvements"]
+                ?.sortedBy { it.second.first.mean }?.map { it.first to it.second }
+                ?.toMap() ?: mapOf<String, ScoreChange>()
+
+        // Calculate change for geometric mean.
+        geoMeanScoreChange = Pair(geoMeanBench.first!!.calcPercentageDiff(geoMeanBench.second!!),
```
sbogolepov: Big number of `!!` is a bad sign :)

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+        val content = renderInstance.render(onlyChanges)
+        outputFile?.let {
+            writeToFile(outputFile, content)
+        } ?: run {
+            println(content)
+        }
+    }
+
+    // Common interface for printing report in different formats.
+    interface Render {
```
sbogolepov: Maybe move `Render` and `TextRender` to separate file? It will simplify reasoning about `SummaryBenchmarksReport`. For example, `addedBenchs` could be expressed as `mergedReport.filter { it.value.first == null }.map { it.key }` and be moved to `TextRender` from `SummaryBenchmarksReport`.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+                    currentWarmup)
+            val varianceBench = BenchmarkResult(name, curStatus, scoreMeanVariance.variance,
+                    runtimeInUsMeanVariance.variance, repeatSeq[resultsSet.size - 1],
+                    currentWarmup)
+             name to MeanVarianceBench(meanBench, varianceBench)
+        }.toMap()
+    }
+
+    // Print report using render.
+    fun print(renderInstance: Render, onlyChanges: Boolean = false, outputFile: String? = null) {
```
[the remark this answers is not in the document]
LepilkinaElena: Idea was to print current report using some render instance. Now it's text render, but it's possible to add csv, html, etc.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,423 @@
...
+// Summary report with comparasion of separate benchmarks results.
+class SummaryBenchmarksReport (val currentReport: BenchmarksReport,
+                               val previousReport: BenchmarksReport? = null,
+                               val meaningfulChangesValue: Double = 0.5) {
+
+    // Lists of benchmarks in different status.
+    private val failedBenchs = mutableListOf<String>()
+    private val addedBenchs = mutableListOf<String>()
+    private val newFailures = mutableListOf<FieldChange<BenchmarkResult.Status>>()
+    private val newPasses = mutableListOf<FieldChange<BenchmarkResult.Status>>()
```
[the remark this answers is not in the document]
sbogolepov: Yeah, something like `benchmarksWithChangedStatus` and filter in `renderStatusSummary`. It will make code modification easier in case of new `BenchmarkResult.Status`.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
LepilkinaElena: I don't want to open inner structure of SummaryReport, it doesn't seems right design. At the same time I want possibility to add new forms of output easily. In C++ I can make render classes as friends, I was looking for similar in Kotlin, and the most close is inner class.

About counting with lambdas - we discussed this yesterday, but that adds extra 4 looping over all benchmarks, that doesn't seem to be a great idea too. I changed to lamdas everywhere where it doesn't influence on operations number a lot.

sbogolepov: It will help to decouple two separate processes (analysis of results and rendering of the report). Explicit data passing them will make code really easier.
You can also make `TextRender` a private class in `SummaryBenchmarksReport.kt` so it won't violate incapsulation.

I think in such tool readability is more important. 4 extra O(n) operations on collection of size 100-1000 is really nothing.

BTW, class friendship in C++ is frequently considered as bad design choice :)

tools/benchmarksAnalyzer/src/main/kotlin-native/org/jetbrains/analyzer/Utils.kt
olonho: It looks like an oversight in MPP design, but outside of this PR  scope.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/kliopt/KliOpt.kt
olonho: Argument parser shall not exit an application on its own, API consumer shall decide what to do depending on available args.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -0,0 +1,222 @@
...
+                            parsedValues[name] = ParsedArg(descriptor, args[index + 1])
+                            index++
+                        } else {
+                            // An error, option with value without value.
+                            printError("No value for ${descriptor.textDescription}")
+                        }
+                    } else {
+                        if (name == "help") {
+                            println(makeUsage())
+                            exitProcess(0)
```
olonho: Just do println and return, no need to exit process.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/renders/Render.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    private fun formatValue(number: Double, isPercent: Boolean = false): String =
+            if (isPercent) format(number, 2) + "%" else format(number)
+
+    private fun formatColumn(content:String, isWide: Boolean = false): String =
+            content.padEnd(if (isWide) wideColumnWidth else standardColumnWidth, ' ')
+
+    private fun printBenchmarksDetails(fullSet: Map<String, SummaryBenchmark>,
+                                       bucket: Map<String, ScoreChange>? = null) {
+        bucket?.let {
```
sbogolepov: Plain if-else is cleaner in this case.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
```diff
@@ -0,0 +1,204 @@
...
+        val geoMeanBenchmarkName = "Geometric mean"
+        val geoMean = geometricMean(benchTable.toList().map { (_, value) -> value.meanBenchmark.score })
+        val varianceGeoMean = geometricMean(benchTable.toList().map { (_, value) -> value.varianceBenchmark.score })
+        val meanBenchmark = BenchmarkResult(geoMeanBenchmarkName, geoMean)
+        val varianceBenchmark = BenchmarkResult(geoMeanBenchmarkName, varianceGeoMean)
+        return MeanVarianceBenchmark(meanBenchmark, varianceBenchmark)
+    }
+
+    // Merge current and compare to report.
+    private fun createMergedReport(currentBenchmarks: BenchmarksTable, previousBenchmarks: BenchmarksTable?) {
```
sbogolepov: You can split this method into `m1` and `m2` returning merged bucket and geomean, respectively. Then in `init` block 
```
mergedReport = m1(..)
geoMeanBenchmark = m2(...)
```
Which will allow to mark these variables as `val` as it should be.

StefMa:
You use at least 4 times the StringBuilder constructor instead of the "buildString" extensions method.
Are there reasons why to do so?
From my point of view there is not need to create a instance of them explizite...

LepilkinaElena:
I've looked at buildString, but renders used several methods for creating report so having one instance of String builder for everything seems ok , buildString isn't applicable there. In other places it also doesn't help to simplify, so I haven't found good places for this function.

PR #2497, opened 2018-12-25
Make Threadlocal and SharedImmutable optional common annotations

PavelPunegov:
This change makes these annotations be available in common code as optional annotations.

### DIFF
--- a/runtime/src/main/kotlin/kotlin/native/Annotations.kt
+++ b/runtime/src/main/kotlin/kotlin/native/Annotations.kt
@@ -44,19 +44,21 @@ public annotation class Throws(vararg val exceptionClasses: KClass<out Throwable
  * Top level variable or object is thread local, and so could be mutable.
  * One may use this annotation as the stopgap measure for singleton
  * object immutability.
+ *
  * PLEASE NOTE THAT THIS ANNOTATION MAY GO AWAY IN UPCOMING RELEASES.
  */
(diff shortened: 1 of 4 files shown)


runtime/src/main/kotlin/kotlin/native/Annotations.kt
```diff
@@ -40,23 +40,15 @@ public annotation class Retain
...
-
-/**
- * Top level variable is immutable and so could be shared.
- * PLEASE NOTE THAT THIS ANNOTATION MAY GO AWAY IN UPCOMING RELEASES.
- */
-@Target(AnnotationTarget.PROPERTY)
-@Retention(AnnotationRetention.BINARY)
-public annotation class SharedImmutable
+@Deprecated("ThreadLocal was moved to kotlin.native.concurrent package " +
+        "and became available in Kotlin Common as optional annotation",
```
olonho: "and became" -> "and is"

runtime/src/main/kotlin/kotlin/native/Annotations.kt
```diff
@@ -40,23 +40,15 @@ public annotation class Retain
...
+        ReplaceWith("kotlin.native.concurrent.ThreadLocal"))
+public typealias ThreadLocal = kotlin.native.concurrent.ThreadLocal
+
+@Deprecated("SharedImmutable was moved to kotlin.native.concurrent package " +
+        "and became available in Kotlin Common as optional annotation",
```
olonho: Same

Issue #2513, opened 2019-01-07
cannot setup master branch(with kotlinVersion 1.3.30-dev-532)


furyu-john:
I run below task but get one error. 
```
$ ./gradlew dependencies:update
w: /Users/satohjohn/sandbox/kotlin/git-kotlin-native/kotlin-native/buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt: (125, 52): Unnecessary non-null assertion (!!) on a non-null receiver of type ExecResult
w: /Users/satohjohn/sandbox/kotlin/git-kotlin-native/kotlin-native/shared/src/library/kotlin/org/jetbrains/kotlin/konan/library/impl/KonanLibraryImpl.kt: (53, 34): The corresponding parameter in the supertype 'KonanLibrary' is named 'fqName'. This may cause problems when calling this function with named arguments.
w: /Users/satohjohn/sandbox/kotlin/git-kotlin-native/kotlin-native/shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/WithLogger.kt: (9, 17): Parameter 'message' is never used

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/satohjohn/sandbox/kotlin/git-kotlin-native/kotlin-native/backend.native/build.gradle' line: 152

* What went wrong:
A problem occurred evaluating project ':backend.native'.
> Could not resolve all files for configuration ':backend.native:kotlin_common_stdlib_src'.
   > Could not find org.jetbrains.kotlin:kotlin-stdlib-common:1.3.30-dev-532.
     Searched in the following locations:
       - https://cache-redirector.jetbrains.com/maven-central/org/jetbrains/kotlin/kotlin-stdlib-common/1.3.30-dev-532/kotlin-stdlib-common-1.3.30-dev-532.pom
       - https://cache-redirector.jetbrains.com/maven-central/org/jetbrains/kotlin/kotlin-stdlib-common/1.3.30-dev-532/kotlin-stdlib-common-1.3.30-dev-532-sources.jar
       - https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_CompilerAllPlugins),number:1.3.30-dev-532,pinned:false/artifacts/content/maven/org/jetbrains/kotlin/kotlin-stdlib-common/1.3.30-dev-532/kotlin-stdlib-common-1.3.30-dev-532.pom
[...]

PR #2514, opened 2019-01-07
cannot setup master branch(with kotlinVersion 1.3.30-dev-532)

furyu-john:
please see #2513.

1.3.30-dev-532 seem not published so i change kotlin version to 1.3.30-dev-588. 

Is there any problem with this change?


furyu-john:
fix by #2454

PR #2531, opened 2019-01-14
[kotlin compiler][update] 1.3.30-dev-761

vvlevchenko:


runtime/src/main/kotlin/kotlin/native/Annotations.kt
```diff
@@ -66,4 +48,10 @@ public annotation class SharedImmutable
  */
 @Target(AnnotationTarget.FUNCTION)
 @Retention(AnnotationRetention.BINARY)
-public annotation class CName(val externName: String = "", val shortName: String = "")
\ No newline at end of file
+public annotation class CName(val externName: String = "", val shortName: String = "")
+
+@Deprecated("please use kotlin.native.concurrent")
```
olonho: Upper case.

PR #2542, opened 2019-01-15
Use dlopen/dlsym for CoreSymbolication

SvyatoslavScherbina:
Linking it directly seems to be inconvenient when producing static lib


runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -41,39 +42,57 @@ typedef unsigned long long CSArchitecture;
 
 #define kCSNow LONG_MAX
 
-extern "C" {
+CSSymbolicatorRef (*CSSymbolicatorCreateWithPid)(pid_t pid);
```
olonho: Maybe put those globals to `namespace { }`?

PR #2557, opened 2019-01-18
Add compiler_rt to linking flags on Apple platforms.

olonho:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -117,7 +117,15 @@ open class MacOSBasedLinker(targetProperties: AppleConfigurables)
 
     private val libtool = "$absoluteTargetToolchain/usr/bin/libtool"
     private val linker = "$absoluteTargetToolchain/usr/bin/ld"
-    internal val dsymutil = "$absoluteLlvmHome/bin/llvm-dsymutil"
+    private val dsymutil = "$absoluteLlvmHome/bin/llvm-dsymutil"
+    private val compilerRtLibrary: String get() {
+            val suffix = when (configurables.target) {
```
SvyatoslavScherbina: Why not check `configurables.target.family` instead?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -117,7 +117,15 @@ open class MacOSBasedLinker(targetProperties: AppleConfigurables)
...
+                KonanTarget.MACOS_X64 -> "osx"
+                KonanTarget.IOS_ARM32, KonanTarget.IOS_ARM64, KonanTarget.IOS_X64 -> "ios"
+                else -> TODO()
+            }
+            return "$absoluteTargetToolchain/usr/lib/clang/10.0.0/lib/darwin/libclang_rt.$suffix.a"
```
SvyatoslavScherbina: 10.0.0 is likely to be Xcode version. So this wouldn't work on any Xcode version but 10.0.0.

PR #2561, opened 2019-01-18
Filter reflect tests and one test that fails due to a bug

PavelPunegov:


buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -703,7 +703,10 @@ class RunExternalTestGroup extends RunStandaloneKonanTest {
         def languageSettings = findLinesWithPrefixesRemoved(text, "// !LANGUAGE: ")
         if (languageSettings.size() != 0) {
             languageSettings.forEach { line ->
-                line.split(" ").toList().forEach { flags.add("-XXLanguage:$it") }
+                line.split(" ").toList().forEach {
+                    def s = it.replace('-DataClassInheritance', '+DataClassInheritance')
```
olonho: What is it intended to do?

PavelPunegov: It just replaces the `-DataClassInheritance` option with a `+` one. The test has this option should work with + in native.

PR #2566, opened 2019-01-21
Improve async wasm instantiation in browsers

chicoxyzzy:
Perform both compilation and instantiation in one step.

Also added some neats:
- fix indentation to be consistent (4 spaces)
- update comments

### DIFF
--- a/runtime/src/launcher/js/launcher.js
+++ b/runtime/src/launcher/js/launcher.js
@@ -193,13 +193,11 @@ function invokeModule(inst, args) {
     return exit_status;
 }
 
-// Instantiate module in Browser in a sequence of promises.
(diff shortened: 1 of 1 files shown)


chicoxyzzy:
At least this change removes redundand tick of event loop. Also (depending on implementation) it improves performance by removing unnecessary memory allocations in VMs.

chicoxyzzy:
Basically one only needs to call WebAssembly.compile only if it’s necessary to pass compiled module or cache it before instantiation.

chicoxyzzy:
Just in case, caching in IndexedDB was removed from wasm spec draft. New cahing will be based on Cache API (which is used in ServiceWorker), but related changes in both spec won’t land any soon.

PR #2575, opened 2019-01-24
Support Linux x64 and RaspberryPi cross-compilation on Windows and macOS hosts.

olonho:


msink:
Maybe first cleanup Linux target a little?
At least split toolchain and sysroot to separate artifacts.

msink:
Also - can be `llvm-ar` used instead of plain `ar` from platform binutils?

msink:
BTW, I have canadian-cross-compiled `target-gcc-toolchain-3-linux-x86-64` for mingw host, but not sure it is needed.

olonho:
Nope, cross-compiler in that direction (mingw->Linux) not needed so far, but maybe mingw compiler hosted on Linux would make sense eventually.

PR #2581, opened 2019-01-25
Inlined ir-as-descriptors typedefs.

alexander-gorshenev:


SvyatoslavScherbina:
The original idea behind "`irasdescriptors`" was to inline `typealias`es (done in this PR) **and** rename correspondingly typed entities which have `descriptor` in their names. I guess it would be much easier to perform the second part in conjunction with the first one instead of doing the rename later.
WDYT? Do you agree or find putting it off more appropriate?

alexander-gorshenev:
Ah, I see. Makes a lot of sense. I am going to try that.

PR #2582, opened 2019-01-27
Sample showing conversion from Kotlin Long to C Long

TheIndifferent:
The operations seems simple and obvious, but was surprisingly hard to figure out, and there are no good samples around.

The code is mostly based on the response from community:
https://discuss.kotlinlang.org/t/long-to-cvaluesref-time-tvar-segfaults/11312


olonho:
Not sure if sample like that is sufficiently complex to be included in the Kotlin/Native repo. Generally, it manifests general pointer operations shown by other samples.

PR #2604, opened 2019-01-31
Custom exception reporter.

olonho:


backend.native/tests/build.gradle
```diff
@@ -2036,6 +2036,12 @@ task extend_exception(type: RunKonanTest) {
     source = "runtime/exceptions/extend0.kt"
 }
 
+task custom_hook(type: RunStandaloneKonanTest) {
```
SvyatoslavScherbina: Testing Swift -> Kotlin call case may also be useful.

runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -206,10 +213,24 @@ void ThrowException(KRef exception) {
...
+    ReportUnhandledException(throwable);
+  } else {
+    ExceptionReporterLaunchpad(handler, throwable);
+  }
+}
 
-RUNTIME_NORETURN void TerminateWithUnhandledException(KRef e) {
-  ReportUnhandledException(e);
+RUNTIME_NORETURN void TerminateWithUnhandledException(KRef throwable) {
+  OnUnhandledException(throwable);
```
SvyatoslavScherbina: It makes a lot of sense to prevent an exception to be thrown out of `OnUnhandledException`.

runtime/src/launcher/kotlin/konan/start.kt
```diff
@@ -21,9 +21,7 @@ private fun Konan_start(args: Array<String>): Int {
         return 0
 
     } catch (e: Throwable) {
-        // TODO: may be add some more info.
-        print("Uncaught exception from Kotlin's main: ")
-        e.printStackTrace()
+        kotlin.native.internal.OnUnhandledException(e)
```
SvyatoslavScherbina: Maybe use `TerminateWithUnhandledException` here?
Then such a termination would be detected as a crash by a debugger. Do we want application to behave like that?

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -86,11 +86,21 @@ internal fun PrintThrowable(throwable: Throwable) {
...
-    e.printStackTrace()
+    throwable.printStackTrace()
 }
 
+@ExportForCppRuntime
+internal fun ExceptionReporterLaunchpad(reporter: (Throwable) -> Unit, throwable: Throwable)
+        = reporter(throwable)
+
+/**
+ * Called once to report unhandled expection. To be used in launcher only.
```
SvyatoslavScherbina: Maybe place this Kotlin function to `start.kt` instead?

backend.native/tests/build.gradle
olonho: Manually tested.

runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -206,10 +213,24 @@ void ThrowException(KRef exception) {
...
 
-void ReportUnhandledException(KRef e);
+OBJ_GETTER(Kotlin_setUnhandledExceptionHook, KRef hook) {
+  RETURN_RESULT_OF(SwapRefLocked,
+    &currentUnhandledExceptionHook, currentUnhandledExceptionHook, hook, &currentUnhandledExceptionHookLock);
+}
+
+void OnUnhandledException(KRef throwable) {
+  ObjHolder handlerHolder;
+  auto* handler = SwapRefLocked(&currentUnhandledExceptionHook, currentUnhandledExceptionHook, nullptr,
```
benasher44: This makes it so the exception handler only activates once (i.e. only activates once if you decide in your implementation to rethrow inside the handler)?

SvyatoslavScherbina: Rethrowing exception inside the handler would only interrupt the handler itself: https://github.com/JetBrains/kotlin-native/pull/2604/files#diff-e305300924774d54efaaded633649b52R99
Generally the hook is not supposed and can't prevent program termination: https://github.com/JetBrains/kotlin-native/pull/2604/files#diff-f7907f299b4df441dc0eb2cf56d6cbf1R234

benasher44: Ah okay gotcha. The app will continue terminating when this function returns. That's what I was looking for. Thanks!

PR #2606, opened 2019-01-31
Fix warnings in both compiler and stdlib

PavelPunegov:


runtime/src/main/kotlin/kotlin/collections/Arrays.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.util.sortArray
 public actual inline fun <reified T> Array<out T>?.orEmpty(): Array<out T> = this ?: emptyArray<T>()
 
 
-@PublishedApi internal inline fun checkCopyOfRangeArguments(fromIndex: Int, toIndex: Int, size: Int) {
```
olonho: Not sure if removing inline here is a good idea.

runtime/src/main/kotlin/kotlin/collections/Collections.kt
```diff
@@ -11,11 +11,9 @@ import kotlin.random.*
 
 /** Copies typed varargs array to an array of objects */
 internal actual fun <T> Array<out T>.copyToArrayOfAny(isVarargs: Boolean): Array<out Any?> =
-        if (isVarargs)
-            // if the array came from varargs and already is array of Any, copying isn't required.
-            @Suppress("UNCHECKED_CAST") (this as Array<out Any?>)
-        else
-            @Suppress("UNCHECKED_CAST") (this.copyOfUninitializedElements(this.size) as Array<out Any?>)
+        // if the array came from varargs and already is array of Any, copying isn't required.
+        if (isVarargs) this
```
olonho: Same line?

runtime/src/main/kotlin/kotlin/collections/Maps.kt
```diff
@@ -6,7 +6,7 @@
 package kotlin.collections
 
 // creates a singleton copy of map, if there is specialization available in target platform, otherwise returns itself
-internal inline actual fun <K, V> Map<K, V>.toSingletonMapOrSelf(): Map<K, V> = toSingletonMap()
+internal actual fun <K, V> Map<K, V>.toSingletonMapOrSelf(): Map<K, V> = toSingletonMap()
```
olonho: Why remove inline here?

runtime/src/main/kotlin/kotlin/collections/Arrays.kt
PavelPunegov: Yes, it should not be removed. But I found that kotlin-compiler issues a warning here stating that inlining in this case makes performance impact insignificant.

runtime/src/main/kotlin/kotlin/collections/Maps.kt
PavelPunegov: Here is the same as previous

runtime/src/main/kotlin/kotlin/collections/Collections.kt
PavelPunegov: yep

olonho: I mean whole statement on same line.

PR #2608, opened 2019-01-31
Move compiler version to 1.3.20, simplify interop runtime.

olonho:


Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt
```diff
@@ -392,19 +391,20 @@ fun List<String>.toCStringArray(autofreeScope: AutofreeScope): CPointer<CPointer
...
-            override val size get() = 2 * (chars.size + 1)
-
-            override fun getPointer(scope: AutofreeScope): CPointer<UShortVar> {
-                val result = scope.allocArray<UShortVar>(chars.size + 1)
-                nativeMemUtils.putCharArray(chars, result.pointed, chars.size)
-                result[chars.size] = 0u
-                return result
-            }
-        }
+        return WCString(CharArray(this.length, { i -> this.get(i)}))
```
PavelPunegov: Parentheses and return is not needed. `get() =` could be used here

gradle.properties
```diff
@@ -15,7 +15,7 @@
 #
 
 # A version of the Kotlin "toolchain" (compiler, runtime, stdlib etc) used by the build script.
-buildKotlinVersion=1.3.0-rc-116
+buildKotlinVersion=1.3.20
 buildKotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_130_CompilerAllPlugins),number:1.3.0-rc-116,tag:kotlin-native,pinned:true/artifacts/content/maven
```
PavelPunegov: Repo should be also updated to maven or `cache-redirector`

PR #2616, opened 2019-02-02
Ensure all @protocol at least conform to NSObject

### LINKED ISSUE #2610: Exported Obj-C protocols should require conforming to NSObject
`@protocol`s in the generated Obj-C header for frameworks currently do not require that objects that implement NSObject:
```
@protocol Example
@required
- (void)test;
@end
```
Instead, protocols should required that objects also implement the NSObject protocol:
```
@protocol Example <NSObject>
@required
- (void)test;
@end
```
By doing this, the compiler will warn you if you try to implement a protocol on a Swift object and pass it into a Kotlin Native library. Without such a warning, you'll get a runtime crash, when Kotlin code starts calling Obj-C selector-type methods on the Swift object that implements the protocol.

benasher44:
This fixes #2610.

I tested this by building the calculator sample project to ensure it still built and that the results I got were what I expected.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
@@ -126,6 +126,8 @@ object StubRenderer {
         val protocols = clazz.superProtocols
         if (protocols.isNotEmpty()) {
             protocols.joinTo(this, separator = ", ", prefix = " <", postfix = ">")
+        } else if (clazz is ObjCProtocol) {
+            append("<NSObject>")
         }
     }
 


SvyatoslavScherbina:
Conforming to `NSObject` protocol is not the equivalent to extending `NSObject` class. The latter is required by Kotlin/Native runtime.
The former may be confusing:
1) Kotlin protocols don't inherit `NSObject` protocol in sense of OOP.
2) Compiler error would force a developer to make his class conforming to `NSObject` protocol without giving any particular hint to achieve this by extending `NSObject` class.

benasher44:
They’re not equivalent, but making your class inherit from NSObject is the most practical way to make your class conform to NSObject in Swift (and I believe is what is suggested in the compiler-suggested fixit in Swift). In Obj-C, this would never be an issue: it’d be extraordinary for there to be a scenario where you consider NSObject the class and protocol separately because all objects inherit from NSObject (except for some very rare circumstances).

1. This might be true technically, but all classes that are exposed from Kotlin that conform to a protocol exposed from Kotlin must also conform to NSObject because they inherit from NSObject. Otherwise, Swift/Obj-C wouldn’t be able to talk to those objects.
2. Swift and Obj-C devs are very familiar with this issue. In Obj-C, you’ll almost never run into this (see earlier explanation). In Swift, this is a common issue; the common remedy is to make your class inherit from NSObject to get the NSObject protocol conformance.

The main issue here is that if you have a Swift class that you want to conform to a protocol that was exposed by Kotlin, it has to conform to NSObject the protocol as well (which in Swift means making your class inherit from NSObject, which I know seems odd but is very familiar if you work in a mixed codebase). If it doesn’t, Kotlin will not be able to message instances of that class, and you’ll get a runtime crash when you pass that Swift class that conforms to the protocol to Kotlin.

SvyatoslavScherbina:
> but making your class inherit from NSObject is the most practical way to make your class conform to NSObject in Swift (and I believe is what is suggested in the compiler-suggested fixit in Swift)

Unfortunately I don't see such a suggestion from Swift compiler.

> The main issue here is that if you have a Swift class that you want to conform to a protocol that was exposed by Kotlin, it has to conform to NSObject the protocol as well

Technically this is not correct: Kotlin/Native runtime doesn't require an object to conform to `NSObject` protocol. Currently the required behaviour is added to `NSObject` class:
https://github.com/JetBrains/kotlin-native/blob/67d6d47e5f1e3ec4657252545db310a8921be699/runtime/src/main/cpp/ObjCExport.mm#L407
So the real issue is that we don't add such a method to Swift classes, and I would prefer to fix that issue instead. I'd like to check whether we are able and have time to fix this, and then consider this PR as a backup path.

benasher44:
Hm I’m surprised that’s not the suggestion. In any case, it’s definitely the most common resolution. As stated earlier, it’s common practice when writing Swift in a mixed codebase.

If Kotlin relies on the Obj-C runtime, don’t objects that it takes in from Swift have to conform to NSObject to be able to particpicate in receiving Obj-C messages and other features that rely on the Obj-C runtimes? From a practical standpoint, that’s what’s you’re doing when you force a class to also conform to NSObject.

Because Swift doesn’t have a root class like NSObject (where you could add an extension like that), I don’t think what you want to do here will be possible, but I’m certainly open to other fixes.

benasher44:
I’ll post the crash trace here today. IIRC, the crash was about being unable to check if the class responded to a selector, which conforming to NSObject fixes. I’ll get back to you on the specifics soon though.

benasher44:
> The funny thing is that it is not necessary: Swift itself has good interoperability with Objective-C, so Swift class has some proper corresponding Objective-C class even if it doesn't extend `NSObject`. Instances of the Swift class are represented as instances of this Objective-C class when passed to Objective-C.
And additional methods can be added to this Objective-C class e.g. using `class_addMethod` at runtime.

As Swift evolves, this continues to be less-and-less accurate. As of Swift 4, you have to explicitly mark methods as @objc to expose them to Obj-C (also depends on whether or not you have the Swift 3 Obj-C method inference compiler setting enabled as well). The direction seems to be that interoperability will be good, but only when it's explicit.

From what I can tell from our crash, making a Swift class conform to a protocol declared by Kotlin (that doesn't require NSObject conformance) doesn't automatically expose the protocol methods to Swift.

SvyatoslavScherbina:
Well, that makes some sense.
On the other hand, Objective-C `id` type is mapped to Swift `AnyObject`, and it is still possible to pass any `AnyObject` to Objective-C, without conforming any Kotlin protocol. So the crash you've encountered is not specific to conforming to protocols.

> From what I can tell from our crash, making a Swift class conform to a protocol declared by Kotlin (that doesn't require NSObject conformance) doesn't automatically expose the protocol methods to Swift.

What do you mean? Kotlin protocol methods are visible to Swift by default.

benasher44:
> Well, that makes some sense.
On the other hand, Objective-C `id` type is mapped to Swift `AnyObject`, and it is still possible to pass any `AnyObject` to Objective-C, without conforming any Kotlin protocol. So the crash you've encountered is not specific to conforming to protocols.

As of Swift 4, they're exposed as `Any`, but yes generally it's save to cast to `AnyObject` and treat it like on.

They're visible to Swift, but the methods from the Swift objects that implement the protocol are not necessarily vislble to Kotlin/Obj-C.

benasher44:
In short, the `NSObject` extension may not always work for objects defined in Swift and passed to Kotlin, unless something compels the user to make their object inherit from NSObject. This PR should fix that (albeit implicitly because it's the easiest way to get NSObject protocol conformance for a Swift object).

SvyatoslavScherbina:
> In short, the `NSObject` extension may not always work for objects defined in Swift and passed to Kotlin, unless something compels the user to make their object inherit from NSObject.

Exactly! That's why I'd like to add required methods to Swift Objective-C root classes using other techniques, e.g. with `class_addMethod`.

> This PR should fix that (albeit implicitly because inheriting from NSObject is the easiest way to get NSObject protocol conformance for a Swift object).

How does this PR fix crash caused by passing a Swift object to Kotlin, if it doesn't inherit any Kotlin protocol?

benasher44:
> Exactly! That's why I'd like to add required methods to Swift Objective-C root classes using other techniques, e.g. with `class_addMethod`.

As far as I know, they don't have root classes.

> How does this PR fix crash caused by passing a Swift object to Kotlin, if it doesn't inherit any Kotlin protocol?

It doesn't fix it in all cases, but for the cases where you're passing a Swift object that conforms to a protocol defined in your Kotlin library, it will essentially force you to make your Swift object inherit from NSObject (because this PR forces Kotlin protocols to say that they require NSObject conformance).

SvyatoslavScherbina:
> As far as I know, they don't have root classes.

As far as I know, they do. When passed to Objective-C, Swift classes instances are represented as Objective-C classes instances, and these classes have Swift's own Objective-C root class declared in Swift runtime: https://github.com/apple/swift/blob/eb083ce84df313fadbae4c668dcf20660e09ab06/stdlib/public/runtime/SwiftObject.mm#L201

SvyatoslavScherbina:
Here is what I mean: https://github.com/JetBrains/kotlin-native/pull/2644

benasher44:
I see. Looks like they do under the hood then. Looking forward to the fix! Thanks!

olonho:
Shall we closes this one?

benasher44:
Done. Thanks!

PR #2636, opened 2019-02-07
Calculate percentage difference based on border values of intervals.

LepilkinaElena:


tools/benchmarksAnalyzer/src/tests/AnalyzerTests.kt
```diff
@@ -59,7 +59,7 @@ class AnalyzerTests {
         val inputs = createMeanVarianceBenchmarks()
 
         val percent = inputs.first.calcPercentageDiff(inputs.second)
-        val expectedMean = -10.0
+        val expectedMean = -9.99809998
```
olonho: Why?

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -40,7 +40,18 @@ data class MeanVarianceBenchmark(val meanBenchmark: BenchmarkResult, val varianc
                 other.varianceBenchmark.score >= 0 &&
                 other.meanBenchmark.score - other.varianceBenchmark.score != 0.0,
                 { "Mean and variance should be positive and not equal!" })
-        val mean = (meanBenchmark.score - other.meanBenchmark.score) / other.meanBenchmark.score
+        val exactMean = (meanBenchmark.score - other.meanBenchmark.score) / other.meanBenchmark.score
+        // Analyze intervals. Calculate difference between border points.
+        val (bigValue, smallValue) = if (meanBenchmark.score > other.meanBenchmark.score) Pair(this, other) else Pair(other, this)
+        val bigValueIntervalStart = bigValue.meanBenchmark.score - bigValue.varianceBenchmark.score
+        val smallValueINtervalEnd = smallValue.meanBenchmark.score + smallValue.varianceBenchmark.score
```
olonho: I*n*terval.

tools/benchmarksAnalyzer/src/tests/AnalyzerTests.kt
LepilkinaElena: Because I made change to use border points of intervals for calculating, not exact means. So Result of test is changed on variance value of intervals.

Issue #939, opened 2017-10-13
x86 32-bit targets


msink:
Support for 32-bit x86 targets - Windows and Linux - is the only missing part for native desktop applications,  almost everythig else is in place already.

olonho:
Why do you need 32-bit WIndows targets? Are there so many of them as targets for new development?

msink:
WinXP is still in use, and will be long time. See #787 
Even WinCE is in use on embedded devices like GPS trackers.

olonho:
Adding an additional platform imposes some overhead on us, but we're ready to consider external contribution of Windows/x86_32 support.  x86 Android and especially mobile Windows platforms are not something with significant market share, so we're not currently planning to support those.

msink:
OK, thanks. But let's leave this issue open :)

atsushieno:
Well, Android x86 target is real on emulators especially with HAXM enabled. It is something significant.

akbcode:
Windows x86 is still an important target. Many cheap Windows tablets and convertibles sold in emerging markets have only 1GB or 2GB RAM. These devices ship with 32-bit version of Windows.

vegansk:
My use case is that I'm searching for the language that allows me to create libraries from the same sources for the frontend (typescript), backend (scala), and for the legacy application (c++) for windows x86_32.

damianw:
32-bit support is also necessary to support WatchOS - both x86 for the simulator and 32-bit ARM for the actual Apple Watch. I have an iOS project that I would like to integrate a kotlin-native framework into, but currently cannot because the core infrastructure, which would use the framework, is used by the project's WatchOS app.

zaxebo1:
32 bit support is important even for "MS Windows Vista 32 bit Desktops" and "MS Windows 7 32 bit Desktops"

zaxebo1:
@vegansk you can see haxe language https://en.wikipedia.org/wiki/Haxe#Compiler

Dico200:
I've been doing some work with a compiled program on 32-bit x86, would love to start doing it in Kotlin.

ferbeb:
I really liked the idea to replace native code with Kotlin Native, but not supporting Windows x86 makes it useless to me. What does this have to do with legacy 32bit operating systems? I'm looking to write a dynamic library plugin for an existing x86 application on Windows 10 x64 , obviously x64 .dll won't work.

olonho:
[...]

PR #2640, opened 2019-02-07
mingw_x86 target

msink:
toolchain: [msys2-mingw-w64-i686-gcc-7.3.0-clang-llvm-lld-6.0.1.zip](https://yadi.sk/d/krBzyqeAqr2x8g)

### DIFF
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -358,6 +358,30 @@ dependencies.mingw_x64 = \
     msys2-mingw-w64-x86_64-gcc-7.3.0-clang-llvm-lld-6.0.1 \
     libffi-3.2.1-mingw-w64-x86-64
 
+# Windows i686, based on mingw-w64.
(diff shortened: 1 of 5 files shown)


olonho:
It will unlikely work out of the box, runtime is not ready for windows_x86 yet.

msink:
There are many votes for it in #939, including mine :)
So at least - should arch be named `X86`?

msink:
Great, it can succesfully create 32-bit .dll:
`kotlinc test.kt -target mingw_x86 -produce dynamic -o test`

And I really needed .dll, not .exe, so it already works for me.

msink:
Just ont problem - had to restart `gradlew bundle` several times - it fails randomly, probably because of low memory. Restarting helps.

olonho:
OK, we will try to get this merged eventually.

sbogolepov:
Thank you, merged with #2656.

PR #2644, opened 2019-02-08
Stop requiring to inherit NSObject when passing Swift object to Kotlin

SvyatoslavScherbina:


runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -399,6 +407,20 @@ static void initializeObjCExport() {
 
   added = class_addMethod(booleanClass, toKotlinSelector, (IMP)boxedBooleanToKotlinImp, toKotlinTypeEncoding);
   RuntimeAssert(added, "Unable to add 'toKotlin:' method to __NSCFBoolean class");
+
+  for (const char* swiftRootClassName : { "SwiftObject", "_TtCs12_SwiftObject" }) {
```
benasher44: Will this name mangling come out the same for every swift object?

benasher44: it migth also be possible that this is unstable across Swift versions 😬

SvyatoslavScherbina: First name is used in current Swift versions.
Second one is supposed to be used in upcoming Swift stable ABI: https://github.com/apple/swift/pull/13748/files#diff-97994a258ff1675c0281c247a4a82718R35
So I believe it is planned to remain stable.

PR #2645, opened 2019-02-11
Write Kotlin uncaught exception stacktrace to iOS crash log

SvyatoslavScherbina:


runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -229,8 +230,23 @@ void OnUnhandledException(KRef throwable) {
   }
 }
 
+#if KONAN_REPORT_BACKTRACE_TO_IOS_CRASH_LOG
+static bool terminating = false;
```
olonho: Why not just keep them in anonymous namespace instead of static?

SvyatoslavScherbina: Anonymous namespaces are too C++ :)
Using `static` is mostly a matter of taste. I like `static` because it makes visibility of the function obvious and doesn't require a developer to look around.

runtime/src/main/cpp/ObjCExceptions.cpp
```diff
@@ -0,0 +1,125 @@
...
+  *bufferPointer = '\0';
+}
+
+#if !defined(MACHSIZE)
+#error "Define MACHSIZE to 32 or 64"
+#endif
+
+#if MACHSIZE == 32
+
+typedef struct mach_header mach_header_host;
```
olonho: Technically, it's more like `_target` than `_host`.

runtime/src/main/cpp/ObjCExceptions.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <mach-o/loader.h>
+#include <CoreFoundation/CFRunLoop.h>
+#include "Natives.h"
+#include "ObjCExceptions.h"
+#include "Types.h"
+
+extern "C" OBJ_GETTER(Kotlin_Throwable_getStackTrace, KRef throwable);
+
+static void writeStackTraceToBuffer(KRef throwable, char* buffer, unsigned long bufferSize) {
+  if (bufferSize < 2) return;
```
olonho: Why `2`?

SvyatoslavScherbina: To make sure it doesn't overflow the buffer if no address is written.

runtime/src/main/cpp/ObjCExceptions.cpp
olonho: Maybe just 
```
#if MACHSIZE == 32
#define ARGH_SUFFIX(x) x
#else 
#define ARGH_SUFFIX(x) x##_64
#endif``

File: runtime/src/main/cpp/ObjCExceptions.cpp [hunk 4]
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      if (bufferEqualsString(segmentCommand->segname, "__DATA")) {
+        section_host* sections = reinterpret_cast<section_host*>(segmentCommand + 1);
+        for (uint32_t sectionIndex = 0; sectionIndex < segmentCommand->nsects; ++sectionIndex) {
+          section_host* section = &sections[sectionIndex];
+
+          if (bufferEqualsString(section->sectname, "__cf_except_bt") && bufferEqualsString(section->segname, "__DATA")) {
+            *size = section->size;
+            return reinterpret_cast<char*>(reinterpret_cast<uintptr_t>(header) + section->addr - textVmaddr);
```
olonho: Likely `return reinterpret_cast<char*>(header) + section->addr - textVmaddr` shall do.

runtime/src/main/cpp/ObjCExceptions.cpp
SvyatoslavScherbina: I would prefer to avoid using macros this heavily.

runtime/src/main/cpp/ObjCExceptions.cpp
SvyatoslavScherbina: What happens if `reinterpret_cast<char*>(header) + section->addr` overflows?

olonho: Here it's unlikely important, it usually matters when comparing result with some another pointer.

PR #2650, opened 2019-02-12
Fix typed array access for platforms not allowing unaligned access.

olonho:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -168,7 +168,8 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
                         // and general ABI requires 4-byte alignment on 64-bit long fields as mentioned in
                         // https://developer.apple.com/library/archive/documentation/Xcode/Conceptual/iPhoneOSABIReference/Articles/ARMv6FunctionCallingConventions.html#//apple_ref/doc/uid/TP40009021-SW1
                         // See https://github.com/ktorio/ktor/issues/941 for the context.
-                        "-DKONAN_NO_64BIT_ATOMIC=1")
+                        "-DKONAN_NO_64BIT_ATOMIC=1",
+                        "-DKONAN_NO_UNALIGNED_ACCESS=1")
```
msink: For ANDROID_ARM32 and LINUX_ARM32 too?

PR #2654, opened 2019-02-13
Rename usages of konan.home to org.jetbrains.kotlin.native.home

PavelPunegov:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanToolRunner.kt
```diff
@@ -68,7 +68,7 @@ internal abstract class KonanCliRunner(
     }
 
     override val additionalSystemProperties = mutableMapOf(
-            "konan.home" to project.konanHome,
+            "org.jetbrains.kotlin.native.home" to project.konanHome,
```
olonho: Shall konan.home be kept here?

PavelPunegov: No, this map is used to set system properties for the new java process that starts compiler. After the start it will search konanHome using this property in `Distribution.kt`

PR #2656, opened 2019-02-13
MinGW x86 target

sbogolepov:
Based on #2640


platformLibs/src/platform/mingw/posix.def
```diff
@@ -56,6 +57,7 @@ static short posix_htons(short x) {
 #define FARPROC     void*
 #define WINAPI      __stdcall
 #define CALLBACK    __stdcall
```
olonho: Do not think so: see http://www.c-jump.com/bcc/c261c/ASM/Procedures/P77_0070_pascal_stdcall.htm and https://reviews.llvm.org/D25204. Doesn't `__pascal` work?

PR #2659, opened 2019-02-14
Several benchmarks for C interop

LepilkinaElena:


performance/cinterop/src/nativeInterop/cinterop/struct.def
```diff
@@ -0,0 +1,105 @@
...
+    } else if (strstr(date, "Wednesday") != NULL) {
+        return Wednesday;
+    } else if (strstr(date, "Thursday") != NULL) {
+        return Thursday;
+    } else if (strstr(date, "Friday") != NULL) {
+        return Friday;
+    } else if (strstr(date, "Saturday") != NULL) {
+        return Saturday;
+    }
+    exit(1);
```
SvyatoslavScherbina: Why not use `abort()` instead?

performance/cinterop/src/nativeInterop/cinterop/struct.def
```diff
@@ -0,0 +1,105 @@
...
+    newElement.integer = coefficient * element.integer;
+    return newElement;
+}
+
+struct ElementS multiplyElementS(struct ElementS element, int coefficient) {
+    struct ElementS newElement;
+    newElement.integer = coefficient * element.integer;
+    newElement.floatValue = coefficient * element.floatValue;
+    int i = 0;
+    while (element.string[i] != NULL)  {
```
SvyatoslavScherbina: Using `NULL` as `char` value confuses.

performance/cinterop/src/nativeInterop/cinterop/types.def
```diff
@@ -0,0 +1,62 @@
...
+       if(charToFind == string[i])
+           ++frequency;
+    }
+    return frequency;
+}
+
+const char* findSuitableString(const char* stringsArray[], int arraySize, const char* stringToFind) {
+    for(int i = 0; i < arraySize; ++i) {
+        if (strcmp(stringsArray[i], stringToFind) == 0) {
+            return stringToFind;
```
SvyatoslavScherbina: When called from Kotlin, `stringToFind` lifetime ends after return, so Kotlin code gets dangling pointer in this case.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
```diff
@@ -0,0 +1,55 @@
...
+actual fun macrosBenchmark() {
+    memScoped {
+        val ints = new_list_int()
+        for (i in 1..benachmarkSize) {
+            list_push_front_int(ints, i)
+        }
+        val floats = new_list_float()
+        // Copy integer list to float one.
+        with (ints?.pointed) {
+            this?.let {
```
SvyatoslavScherbina: Can this combination of `with`/`let` be replaced by `ints?.pointed?.apply {` here?

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
```diff
@@ -0,0 +1,55 @@
...
+                            ?: error("Null elements in list are not expected!")
+                    )
+                    current = current?.pointed?._next
+                }
+            }
+        }
+        // Reverse list.
+        var previous: CPointer<list_elem_float>? = null
+        var current = floats?.pointed?._first
+        var next = current?.pointed?._next
```
SvyatoslavScherbina: I guess `next` can be moved to the loop body to make the code cleaner.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.macrosBenchmarks
+import kotlinx.cinterop.*
+
+const val benachmarkSize = 1000
+
+actual fun macrosBenchmark() {
```
SvyatoslavScherbina: Btw, this benchmark doesn't seem to bench macros, C macros don't have any runtime cost.
So what is this benchmark for actually?

performance/cinterop/gradle.properties
```diff
@@ -0,0 +1 @@
+org.jetbrains.kotlin.native.home=../../dist
```
olonho: Is it indeed needed?

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
```diff
@@ -0,0 +1,55 @@
...
+        for (i in 1..benachmarkSize) {
+            list_push_front_int(ints, i)
+        }
+        val floats = new_list_float()
+        // Copy integer list to float one.
+        with (ints?.pointed) {
+            this?.let {
+                var current = _first
+                while(current != null) {
+                    list_push_front_float(floats, current?.pointed?._data?.toFloat()
```
olonho: Better do `while<space>{`

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
...
+package org.jetbrains.structsBenchmarks
+import kotlinx.cinterop.*
+import kotlin.math.sqrt
+
+const val benchmarkSize = 10000
+
+actual fun structBenchmark() {
+    memScoped {
+        val containsFunction = staticCFunction<CPointer<ElementS>?, CPointer<ElementS>?, Int> { first, second ->
+            var result: Int;
```
SvyatoslavScherbina: Redundant `;`.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
```diff
@@ -0,0 +1,55 @@
...
+        var current = floats?.pointed?._first
+        var next = current?.pointed?._next
+        while(current != null) {
+            current?.pointed?._next = previous
+            previous = current
+            current = next
+            next = current?.pointed?._next
+        }
+        floats?.pointed?._first = previous
+        if (list_front_float(floats) != benachmarkSize.toFloat())
```
olonho: Comparison of general floats (not fitting precisely into IEEE representation) may not work  as expected.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
...
+            if (first == null || second == null) {
```
SvyatoslavScherbina: Please consider using `if` or `when` as expression here, i.e. `return if (...) ...`.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
...
+                result = 0
+            }
+            result
+        }
+        val elementsList = mutableListOf<ElementS>()
+        // Fill list.
+        for (i in 1..benchmarkSize) {
+            val element = alloc<ElementS>()
+            element.floatValue = i + sqrt(i.toDouble()).toFloat()
+            element.integer = i.toLong()
```
SvyatoslavScherbina: Note that `integer` has `long` type which isn't mapped to `kotlin.Long` on some platforms.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
...
+            itoa(i, element.string, 10)
+            element.contains = containsFunction
+
+            elementsList.add(element)
+        }
+        val summary = elementsList.map { multiplyElementS(it.readValue(), (0..10).random()) }
+                .reduce { acc, it -> sumElementSPtr(acc.ptr, it.ptr)!!.pointed.readValue() }
+        val intValue = summary.useContents { integer }
+        elementsList.last().contains!!(elementsList.last().ptr, elementsList.first().ptr)
```
SvyatoslavScherbina:
```
elementsList.last().apply { contains!!(this, elementsList.first().ptr) }
```

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
LepilkinaElena: Used macros produced function and structs. Macros itself has no runtime cost. Do you mean that I should remove it or rename?

performance/cinterop/gradle.properties
LepilkinaElena: Before it was needed, because it was used from subproject directory. I can test later again when this functionality will be fixed.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
LepilkinaElena: `inline fun <T> T.apply(block: T.() -> Unit): T`
But my function return Int.

SvyatoslavScherbina: Do you mean that using it with `T = Int` may cause redundant boxing?
`elementsList.last()` should at least be extracted to a variable.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/macrosBenchmark.kt
SvyatoslavScherbina: Rename or include into other benchmark.

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsBenchmark.kt
LepilkinaElena: I ment that contains has return value type Int, so `block: T.() -> Int` by Unit is expected

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/structsProducedByMacrosBenchmark.kt
```diff
@@ -0,0 +1,53 @@
...
+
+package org.jetbrains.structsProducedByMacrosBenchmarks
+import kotlinx.cinterop.*
+import kotlin.math.abs
+
+const val benachmarkSize = 1000
+
+actual fun macrosBenchmark() {
+    memScoped {
+        val ints = new_list_int()
```
SvyatoslavScherbina: Does memory taken by this list leak?

performance/cinterop/src/main/kotlin-native/org/jetbrains/cinteropBenchmarks/typesBenchmark.kt
```diff
@@ -0,0 +1,129 @@
...
+        // Generate random strings.
+        for (i in 1..benchmarkSize) {
+            charFrequency(randomString, randomChar.toByte())
+        }
+    }
+
+    actual fun stringToKotlinBenchmark() {
+        memScoped {
+            for (i in 1..benchmarkSize) {
+                val result = findSuitableString(strings.toCStringArray(this), benchmarkSize, "a")?.toKString()
```
SvyatoslavScherbina: Memory leaks here.

performance/cinterop/gradle.properties
LepilkinaElena: Yes, it's needed
```
 Classpath of the tool is empty: cinterop
  Probably the 'org.jetbrains.kotlin.native.home' project property contains an incorrect path.```

PR #2666, opened 2019-02-15
Add basic support for variable-length array parameters in cinterop

SvyatoslavScherbina:


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -532,6 +532,11 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
                 IncompleteArrayType(elemType)
             }
 
+            CXType_VariableArray -> {
+                val elemType = convertType(clang_getArrayElementType(type))
```
olonho: `elementType`?

SvyatoslavScherbina: Sure!

PR #2672, opened 2019-02-15
Rewrite interop tools command line options to have help messages without crashes

LepilkinaElena:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -174,28 +158,27 @@ private fun findFilesByGlobs(roots: List<Path>, globs: List<String>): Map<Path,
...
-    val defFile = arguments.def?.let { File(it) }
-    val manifestAddend = arguments.manifest?.let { File(it) }
+    val defFile = argParser.get<String>("def")?.let { File(it) }
+    val manifestAddend = argParser.get<String>("manifest")?.let { File(it) }
 
-    if (defFile == null && arguments.pkg == null) {
-        usage()
-        return null
+    if (defFile == null && argParser.get<String>("pkg") == null) {
+        argParser.printError("-def or -pkg should provided!")
```
SvyatoslavScherbina: Does it show help in this case as before?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
```diff
@@ -16,97 +16,53 @@
...
+                defaultValue = System.getProperty("user.dir")),
+        OptionDescriptor(ArgType.String(), "staticLibrary", description = "embed static library to the result",
+                isMultiple = true, delimiter = ","),
+        OptionDescriptor(ArgType.String(), "temporaryFilesDir", description = "save temporary files to the given directory")
+    )
+
+fun getCInteropArguments(): List<OptionDescriptor> {
+    val options = listOf(
+            OptionDescriptor(ArgType.String(), "import", description = "a semicolon separated list of headers, prepended with the package name",
+                    isMultiple = true, delimiter = ";"),
```
SvyatoslavScherbina: Using `;` delimiter here doesn't look correct to me.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
```diff
@@ -16,97 +16,53 @@
...
+            OptionDescriptor(ArgType.String(), "target", description = "native target to compile to"),
+            OptionDescriptor(ArgType.String(), "def", description = "the library definition file"),
+            OptionDescriptor(ArgType.String(), "header", "hd", "header file to produce kotlin bindings for",
```
SvyatoslavScherbina: Is it possible to keep the old short name for this option, `-h`? I would prefer to avoid altering command line interface.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
```diff
@@ -16,97 +16,53 @@
...
+                    isMultiple = true, delimiter = ","),
```
SvyatoslavScherbina: What's the motivation behind adding `,` delimiter to this and other options?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
LepilkinaElena: yes

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
LepilkinaElena: You're right, it's splitted deeper in code. Fixed.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
LepilkinaElena: I don't know. In my opinion, it's common practice to use -h for getting help message. People expect such behaviour. I changed flags everywhere. @olonho ?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
LepilkinaElena: This allows to save old behaviour when option can be splitted with some delimiter and get all values already splitted.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
SvyatoslavScherbina: `cinterop` CLI interface is unpopular but still public. Also there is at least `kotlin-multiplatform` Gradle plugin which is developed in other source repository.

LepilkinaElena: I understand, I don't know situation well, but getting error on asking help with common usage short name isn't very good too.
In MPP I changed to full name https://github.com/JetBrains/kotlin/pull/2117

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
SvyatoslavScherbina: Now I see that this appears to be a (weird) default behaviour of `CommonToolArguments` machinery.
I'm not sure this was the intended behaviour for `cinterop`, but I guess it is better to keep it anyway.

LepilkinaElena: It doesn't worked for interop tools. I can add them and resend to compiler.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -173,8 +174,36 @@ class ArgParser(optionsList: List<OptionDescriptor>, argsList: List<ArgDescripto
         if (!descriptor.isMultiple && !processedValues.getValue(descriptor.longName).isEmpty()) {
             printError("Option ${descriptor.longName} is used more than one time!")
         }
-        descriptor.type.check(value, descriptor.longName)
-        processedValues.getValue(descriptor.longName).add(value)
+        val savedValues = descriptor.delimiter?.let {
+            // Escape values on quotes.
```
SvyatoslavScherbina: What does this fix?

LepilkinaElena: This is fix for parameters in quotes with space. If parameter can be written with some delimiter, current parser should work properly with them after default compiler parser is fixed(issue discussed in slack)

SvyatoslavScherbina: As far as I can see, this code can't fix the issue discussed.

LepilkinaElena: This code process case when user can write cinterop -lopt "-f \"some name\" -another_flag". String inside quotes wouldn't be splitted.

LepilkinaElena: May be it isn't solution for full problem, but as I understood such case also can be and should be parsed properly.

SvyatoslavScherbina: > This code process case when user can write cinterop -lopt "-f "some name" -another_flag". String inside quotes wouldn't be splitted.

User is not supposed to pass such an option, because it doesn't have any effect. Linker options are passed when compiling the binary and handled with compiler arguments parser.

https://youtrack.jetbrains.com/issue/KT-29970
How does this patch fix anything mentioned in this issue?

olonho:
Please rebase it on master.

PR #2676, opened 2019-02-17
Use different C types for boolean types in C and Objective-C mode.

olonho:


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -469,13 +469,49 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
...
+                isSigned = true,
+                spelling = clang_getTypeSpelling(type).convertAndDispose()
+        )
+
+        CXTypeKind.CXType_Float, CXTypeKind.CXType_Double -> FloatingType(
+                size = type.getSize().toInt(),
+                spelling = clang_getTypeSpelling(type).convertAndDispose()
+        )
+
+        CXTypeKind.CXType_Bool -> boolType()
```
SvyatoslavScherbina: Isn't it `CBoolType` then?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/TypeUtils.kt
```diff
@@ -31,9 +31,10 @@ val StructDecl.isAnonymous: Boolean
  * TODO: use libclang to implement?
  */
 fun Type.getStringRepresentation(): String = when (this) {
-    is VoidType -> "void"
-    is CharType -> "char"
-    is BoolType -> "BOOL"
+    VoidType -> "void"
+    CharType -> "char"
+    CBoolType -> "bool"
```
SvyatoslavScherbina: Why not `_Bool`?

PR #2680, opened 2019-02-18
Removed a stale leftover commentary.

alexander-gorshenev:


alexander-gorshenev:
without fix
> $ du -sk  dist/klib/common/stdlib/linkdata
> 3052    dist/klib/common/stdlib/linkdata

with fix
> $ du -sk  dist/klib/common/stdlib/linkdata
> 884     dist/klib/common/stdlib/linkdata

PR #2682, opened 2019-02-18
Slack reporter and KonanTestSuite rewritten to Kotlin

PavelPunegov:


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KonanTestSuite.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    fun add(other: Statistics) {
+        passed += other.passed
+        failed += other.failed
+        error += other.error
+        skipped += other.skipped
+    }
+}
+
+open class KonanTestSuite(val name: String, val statistics: Statistics) {
```
olonho: abstract?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KonanTestSuite.kt
```diff
@@ -0,0 +1,139 @@
...
+
+        open fun skip(): TestResult {
+            statistics.skip()
+            return TestResult(TestStatus.SKIPPED)
+        }
+    }
+
+    open fun createTestCase(name: String) = KonanTestCase(name)
+
+    open fun start() {}
```
olonho: Maybe skip body and require subclass to implement?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KonanTestSuite.kt
vvlevchenko: No, there `KonanTestSuite` used to be default implementation for command line launches and so on. The single child is for TeamCity execution.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KonanTestSuite.kt
PavelPunegov: By default KonanTestSuite is used by the standalone cli-exectution, while subclasses are used for the TeamCity

vvlevchenko: I'd suggest to stay at this phase (naive and working groovy to kotlin conversion) to let me continue with refactoring and modification this part of code.

PR #2684, opened 2019-02-18
KT-29970: correctly handle commandline arguments with whitespace

jonnyzzz:
compilerOpts, linkerOpts, and other files may have arguments with spaces: paths, framework names, etc. The PR fixes the way we deal with such paths to make sure we do not split them blindly by whitespace. 

There are few more fixes required for the MPP plugin code in Kotlin main repository, see the issue for more details
https://youtrack.jetbrains.com/issue/KT-29970

### DIFF
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
@@ -18,7 +18,7 @@ package org.jetbrains.kotlin.native.interop.tool
 
 import org.jetbrains.kotlin.cli.common.arguments.*
 import org.jetbrains.kotlin.konan.util.ARGUMENT_NO_DELIMITER
-import org.jetbrains.kotlin.konan.util.parseCommandLineString
+import org.jetbrains.kotlin.konan.util.CommandLineOpts
 
 // TODO: unify camel and snake cases.
 // Possible solution is to accept both cases
@@ -69,10 +69,12 @@ class CInteropArguments : CommonInteropArguments() {
     @Argument(value = "-compilerOpts", shortName = "-copt", valueDescription = "<arg>", description = "additional compiler options", delimiter = " ")
     var compilerOpts: Array<String> = arrayOf()
 
-    @Argument(value = "-linkerOpts", shortName = "-lopt", valueDescription = "<arg>", description = "additional linker options", delimiter = ARGUMENT_NO_DELIMITER)
-    var linkerOptsInternal: Array<String> = arrayOf()
-
-    val linkerOpts: Array<String> by parseCommandLineString { linkerOptsInternal }
+    private val linkerOptsHolder = CommandLineOpts()
+    val linkerOpts: List<String> by linkerOptsHolder
+    @Argument(value = "-linkerOpt", valueDescription = "<arg>", description = "single additional linker option", delimiter = ARGUMENT_NO_DELIMITER)
+    var linkerOptInternal by linkerOptsHolder.singletonOpt
+    @Argument(value = "-linkerOpts", shortName = "-lopt", valueDescription = "<args>", description = "additional linker options", delimiter = ARGUMENT_NO_DELIMITER)
+    var linkerOptsInternal by linkerOptsHolder.multipleOpts
 
     @Argument(value = "-shims", description = "wrap bindings by a tracing layer") 
     var shims: Boolean = false
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
@@ -17,6 +17,8 @@
 package org.jetbrains.kotlin.native.interop.tool
 
 import org.jetbrains.kotlin.cli.common.arguments.*
+import org.jetbrains.kotlin.konan.util.ARGUMENT_NO_DELIMITER
+import org.jetbrains.kotlin.konan.util.parseCommandLineString
 
 // TODO: unify camel and snake cases.
 // Possible solution is to accept both cases
@@ -67,8 +69,10 @@ class CInteropArguments : CommonInteropArguments() {
     @Argument(value = "-compilerOpts", shortName = "-copt", valueDescription = "<arg>", description = "additional compiler options", delimiter = " ")
     var compilerOpts: Array<String> = arrayOf()
 
-    @Argument(value = "-linkerOpts", shortName = "-lopt", valueDescription = "<arg>", description = "additional linker options", delimiter = " ")
-    var linkerOpts: Array<String> = arrayOf()
+    @Argument(value = "-linkerOpts", shortName = "-lopt", valueDescription = "<arg>", description = "additional linker options", delimiter = ARGUMENT_NO_DELIMITER)
+    var linkerOptsInternal: Array<String> = arrayOf()
+
+    val linkerOpts: Array<String> by parseCommandLineString { linkerOptsInternal }
 
     @Argument(value = "-shims", description = "wrap bindings by a tracing layer") 
     var shims: Boolean = false
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
@@ -11,6 +11,8 @@ import org.jetbrains.kotlin.cli.common.arguments.Argument
 import org.jetbrains.kotlin.cli.common.messages.MessageCollector
 import org.jetbrains.kotlin.config.AnalysisFlag
 import org.jetbrains.kotlin.config.AnalysisFlags
+import org.jetbrains.kotlin.konan.util.ARGUMENT_NO_DELIMITER
+import org.jetbrains.kotlin.konan.util.parseCommandLineString
 
 class K2NativeCompilerArguments : CommonCompilerArguments() {
     // First go the options interesting to the general public.
@@ -60,8 +62,10 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-nopack", description = "Don't pack the library into a klib file")
     var nopack: Boolean = false
 
-    @Argument(value="-linker-options", deprecatedName = "-linkerOpts", valueDescription = "<arg>", description = "Pass arguments to linker", delimiter = " ")
-    var linkerArguments: Array<String>? = null
+    @Argument(value="-linker-options", deprecatedName = "-linkerOpts", valueDescription = "<arg>", description = "Pass arguments to linker", delimiter = ARGUMENT_NO_DELIMITER)
+    var linkerArgumentsInternal: Array<String> = arrayOf()
+
+    val linkerArguments: Array<String> by parseCommandLineString { linkerArgumentsInternal }
 
     @Argument(value = "-nostdlib", description = "Don't link with stdlib")
     var nostdlib: Boolean = false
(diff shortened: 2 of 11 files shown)


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
```diff
@@ -64,11 +66,15 @@ class CInteropArguments : CommonInteropArguments() {
     @Argument(value = HEADER_FILTER_ADDITIONAL_SEARCH_PREFIX, shortName = "-hfasp",  valueDescription = "<file>", description = "header file to produce kotlin bindings for") 
     var headerFilterPrefix: Array<String> = arrayOf()
 
-    @Argument(value = "-compilerOpts", shortName = "-copt", valueDescription = "<arg>", description = "additional compiler options", delimiter = " ")
-    var compilerOpts: Array<String> = arrayOf()
+    @Argument(value = "-compilerOpts", shortName = "-copt", valueDescription = "<arg>", description = "additional compiler options", delimiter = ARGUMENT_NO_DELIMITER)
```
SvyatoslavScherbina: This part interferes with #2672. On the other hand, it doesn't seem to fix anything from [KT-29970](
https://youtrack.jetbrains.com/issue/KT-29970).

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -60,8 +62,10 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-nopack", description = "Don't pack the library into a klib file")
     var nopack: Boolean = false
 
-    @Argument(value="-linker-options", deprecatedName = "-linkerOpts", valueDescription = "<arg>", description = "Pass arguments to linker", delimiter = " ")
-    var linkerArguments: Array<String>? = null
+    @Argument(value="-linker-options", deprecatedName = "-linkerOpts", valueDescription = "<arg>", description = "Pass arguments to linker", delimiter = ARGUMENT_NO_DELIMITER)
+    var linkerArgumentsInternal: Array<String> = arrayOf()
+
+    val linkerArguments: Array<String> by parseCommandLineString { linkerArgumentsInternal }
```
SvyatoslavScherbina: Handling quotes and escaping here makes sense, however quoting arguments in Gradle plugin to unquote them in the compiler doesn't seem to be proper solution to me.
Instead I suggest adding new command-line argument, `-linkerOpt`, which would allow passing next argument to the linker as-is.

This new option should be handled preserving mutual order with the old one, i.e.
```
-linkerOpt arg1 -linkerOpts "arg2 arg3" -linkerOpt "arg 4"
```
->
```
[arg1, arg2, arg3, arg 4]
```

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt
```diff
@@ -57,3 +60,69 @@ fun String.removeSuffixIfPresent(suffix: String) =
...
+    //com.intellij.openapi.util.text.StringUtilRt#splitHonorQuotes
+
+    val result = mutableListOf<String>()
+    val builder = StringBuilder(cmdString.length)
+
+    var inQuotes = false
+    for (i in 0 until cmdString.length) {
+        val c = cmdString[i]
+        if (c == ' ' && !inQuotes) {
+            if (builder.isNotEmpty()) {
```
SvyatoslavScherbina: This makes the parser ignore empty arguments, which doesn't look correct.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt
```diff
@@ -57,3 +60,69 @@ fun String.removeSuffixIfPresent(suffix: String) =
...
+                result.add(builder.toString())
+                builder.setLength(0)
+            }
+            continue
+        }
+        if ((c == '"' || c == '\'') && !(i > 0 && cmdString[i - 1] == '\\')) {
```
SvyatoslavScherbina: This parser fails to handle a lot of complicated cases, e.g. `' " abc "  '`.
I consider implementing escaping support harmful unless it has reliable widely recognized behaviour.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
jonnyzzz: It fixes any possible escaped arguments here. Do you mean there are no such arguments possible?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
jonnyzzz: What is the best place to write tests for that code? I was looking for the right module for it. Please recommend one to add tests there.

jonnyzzz: btw, the exact case would work. The `@Argument` annotation generate the following array for us:
```
[arg1, "arg2 arg3", "arg 4"]
```
the quotes around the second and the third values are removed by the commanline parser, when you start the process
the artificial `delimiter` value will not let it to split arguments
Next after processing the quotes we will have:
```
[arg1, arg2, arg2, arg, 4]
```
One may need double quotes to pass a `arg 4` as a single argument, e.g. `-linerOps "\"arg 4\""`

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt
jonnyzzz: let me write a test for that case

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt
jonnyzzz: In general, the way how arguments are parsed by a process depends on the process. It depends on the OS. You may have the same commandline processed differently by java, cmd, or bash.
I think the general idea to pass quoted commandline string is broken from the very beginning. 

Alternative could be to add yet anothe sub-command to pass exactly one argument, e.g. 
```-linkerOpt "abc" -linkerOpt `' " abc " '````
Both commands may blindly include the whole value as one array element. Simliarly we may do for the `.def` files, but it may require to patch the way we use `Properties` class in the implementation


What do you think?

**Comment by jonnyzzz** [comment]
@SvyatoslavScherbina what should I do to have TeamCity builds alive? Some builds are failing because the `tc-dist` task is not found

File: Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt (hunk 1 above)
Comment (SvyatoslavScherbina) [reply Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt]: I mean that this part doesn't seem to fix anything from [KT-29970](https://youtrack.jetbrains.com/issue/KT-29970). It fixes an issue which you haven't encountered and reported yet, that's why it can be fixed later to avoid merge conflicts now.

Comment (jonnyzzz) [reply Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt]: It is not totally true - https://youtrack.jetbrains.com/issue/KT-29970#focus=streamItem-27-3295557.0-0

File: backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt (hunk 2 above)
Comment (SvyatoslavScherbina) [reply backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt]: > The `@Argument` annotation generate the following array for us:

How can it generate this array if `-linkerOpt` is not supported yet?

> One may need double quotes to pass a `arg 4` as a single argument, e.g. `-linerOps "\"arg 4\""`

What's the motivation behind applying unreliable solution to pass data between Gradle plugin and compiler?

File: shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt (hunk 4 above)
Comment (SvyatoslavScherbina) [reply shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt]: > Alternative could be to add yet anothe sub-command to pass exactly one argument,

How is it different from the solution I proposed in https://github.com/JetBrains/kotlin-native/pull/2684#discussion_r257906960 ?

**Comment by SvyatoslavScherbina** [comment]
> what should I do to have TeamCity builds alive? Some builds are failing because the `tc-dist` task is not found

@jonnyzzz maybe rebase to more recent `master` branch commits?

File: Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt (hunk 1 above)
Comment (SvyatoslavScherbina) [reply Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt]: How does this affect you when using e.g. `A B C.framework`?

Comment (jonnyzzz) [reply Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt]: It does not

File: backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt (hunk 2 above)
Comment (jonnyzzz) [reply backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt]: Quotes in quotes are ok. 

Unreliable? - I suggest to consider a way of passing arguments between processes (Gradle, cinterop, kotlinc-native, klib metadata) in a way, where we do not need to encode/decode arguments many times. The approach is error-prone by definition. That is what looks _reliable_ to me. 

I suggested adding the new `-linkerOpt` argument to _avoid_ us from parsing longer lines of multiple arguments on and on again. It has nothing related to the current change around the `@Argument` annotation.

Comment (SvyatoslavScherbina) [reply backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt]: > I suggest to consider a way of passing arguments between processes (Gradle, cinterop, kotlinc-native, klib metadata)

This is *exactly* what I've suggested above, but you've mentioned some problems that don't happen with such a solution:

>  [arg1, arg2, arg2, arg, 4]

> One may need double quotes to pass a `arg 4` as a single argument, e.g. `-linerOps "\"arg 4\""`

Comment (jonnyzzz) [reply backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt]: 👍 sorry, I did not spot you used two different options in the very first comment: `-linkerOpt` and `-linkerOpts`. 
I will try addressing that

File: shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt [hunk 5]
```diff
@@ -59,3 +62,69 @@ fun String.removeSuffixIfPresent(suffix: String) =
...
+
+
+
+/// we use the fake delimiter to support the feature
+/// of the [@Argument] parser to merge several arguments
+/// into the same argument (it works only for [Array<String>]
+/// type, but we do not need the separator
+///
+/// the right fix is to update the [@Argument] to support paths separation
+const val ARGUMENT_NO_DELIMITER = "\\n\\t\\t\\n\\t\\t\\n\\ue000\\ue001\\ue002\\n\\t\\t\\t\\t\\n"
```
olonho: Nope, this is not the way I believe this feature be implemented. As Slava mentioned, please just add `linkerOpt` and `linkerOpts` arguments for CLI, and concatenate them + implement "\ " escaping for " "-separated options in .def files.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/util/Util.kt
olonho: At the moment there’s no explicit testing infrastructure here, so better just test on existing samples and your case.

jonnyzzz:
I applied the new approach. It works on my example. 
There is _another_ command line arguments parser included and added to some of the entry points (with no unit tests). 
I guess it makes sense to merge the linkerOpts effort and use the same parser for all executables

olonho:
Reworked in https://github.com/JetBrains/kotlin-native/pull/2805 and other related commits.

PR #2689, opened 2019-02-19
Support UTF-32 conversion.

olonho:


msink:
Maybe better as in C++: `u16str` and `u32str` always 16/32 bit, and `wcstr` depends on `sizeof(wchar_t)` ?

msink:
BTW - if I want to compile some K/N library as shared lib for Windows/Linux/Mac, and export all Kotlin strings on Linux and Mac as utf8 and on Windows as utf16 (no conversions) - is it possible somehow?

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeMem.kt
```diff
@@ -150,15 +150,45 @@ fun CPointer<UShortVar>.toKString(): String {
...
     while (index < length) {
-        bytes[index] = nativeBytes[index].toShort().toChar()
+        chars[index] = nativeBytes[index].toShort().toChar()
         ++index
     }
-    return String(bytes)
+    return String(chars)
 }
 
+fun CPointer<IntVar>.toKString(): String {
```
SvyatoslavScherbina: Shouldn't we mention UTF-32 explicitly in the name of this function?

olonho: How would you suggest to call it?

SvyatoslavScherbina: I don't know. Maybe `utf32ToKString`, `toKStringFromUTF32`, `decodeUtf32` or `decodeFromUtf32` etc.
The second variant appears to be more discoverable (considering existing `toKString` functions), but we may decide to cut discoverability down to be able to choose better name.

olonho:
> BTW - if I want to compile some K/N library as shared lib for Windows/Linux/Mac, and export all Kotlin strings on Linux and Mac as utf8 and on Windows as utf16 (no conversions) - is it possible somehow?

Internally strings are in UTF-16, so it's unlikely possible.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeMem.kt
olonho: toKStringFromUtf32 looks OKish to me.

PR #2695, opened 2019-02-20
Support static ObjC frameworks

ilmat192:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -134,6 +134,9 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-Xruntime", deprecatedName = "--runtime", valueDescription = "<path>", description = "Override standard 'runtime.bc' location")
     var runtimeFile: String? = null
 
+    @Argument(value = STATIC_FRAMEWORK_FLAG, description = "Create a framework with a static library instead of a dynamic one.")
```
SvyatoslavScherbina: The final dot seems to be redundant.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LinkStage.kt
```diff
@@ -15,6 +15,20 @@ typealias BitcodeFile = String
 typealias ObjectFile = String
 typealias ExecutableFile = String
 
+private fun determineLinkerOutput(context: Context): LinkerOutputKind {
+    val staticFramework = context.config.configuration.getBoolean(KonanConfigKeys.STATIC_FRAMEWORK)
```
SvyatoslavScherbina: Can be declared under `if`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LinkStage.kt
```diff
@@ -15,6 +15,20 @@ typealias BitcodeFile = String
...
+    return if (context.config.produce == CompilerOutputKind.FRAMEWORK) {
+        if (staticFramework) LinkerOutputKind.STATIC_LIBRARY else LinkerOutputKind.DYNAMIC_LIBRARY
```
SvyatoslavScherbina: Why not keep this branch within `when` below?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LinkStage.kt
```diff
@@ -15,6 +15,20 @@ typealias BitcodeFile = String
...
+    } else {
+        when (context.config.produce) {
```
olonho: Why not CompilerOutputKind.STATIC_FRAMEWORK and CompilerOutputKind.DYNAMIC_FRAMEWORK?

SvyatoslavScherbina: 1) Static framework is much like a hack.
2) Static framework is a bit closer to dynamic one than static library to dynamic library. For example, they use the same naming convention.
3) It is much easier to implement this way.

olonho: Not sure if I like it this way, as it's somewhat confusing and different from other compilation outputs. Maybe ask @alexander-gorshenev.

SvyatoslavScherbina: It is ok to make it different from other compilation outputs, see 2.

Static framework seems much like a flavour of the framework rather than a different output kind. Consider bitcode embedding as an example of other flavour. Do you suggest to make it a separate compilation output too?

Finally, I guess this way it would be easier for users to configure the build.

ilmat192: Also I don't think we need to provide a separate binary type for static frameworks in gradle DSL. It's better to implement this setting as a separate property in a framework configuration block (like `binaries.framework { staticFramework = true/false }`). So implementing this setting in the compiler as a flag instead of an output kind makes this approach in DSL more straightforward.

olonho: OK then, hopefully it will work for users fine.

befovy:
Hello, I read this merge request, but still don't known how to configure a static framework or library in `build.gradle.kts`
Can someone give an answer? Thanks very much.

SvyatoslavScherbina:
@befovy see `isStatic = true` in the documentation: https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#building-final-native-binaries

PR #2701, opened 2019-02-22
Create an empty inlining rule that suppresses a warning about performance

PavelPunegov:


olonho:
As agreed with Stas and Roman.

PR #2703, opened 2019-02-22
Compile only benchmarks: helloWorld and videoplayer

LepilkinaElena:


performance/gradle/benchmark.gradle
```diff
@@ -96,7 +96,7 @@ task konanJsonReport {
                                                   'compilerVersion': "${konanVersion}".toString(),
                                                   'flags': kotlin.targets.native.compilations.main.extraOpts.collect{ "\"$it\"" },
                                                   'benchmarks': benchContents,
-                                                  'compileTime': nativeCompileTime,
+                                                  'compileTime': [nativeCompileTime],
```
olonho: Why list? What else shall be here?

performance/videoplayer/build.gradle
```diff
@@ -0,0 +1,68 @@
+def dist = findProperty('org.jetbrains.kotlin.native.home') ?: 'dist'
+
+def linkerOpts = []
```
olonho: Hmm, this looks like not very awesome idea. Isn't it possible to use composite build with the build in `samples`?

performance/gradle/benchmark.gradle
LepilkinaElena: There can be several measurements of compile time to get more stable result.

performance/videoplayer/build.gradle
LepilkinaElena: Then we have no opportunity 

1) to measure clean way without gradle

2) measure several times

3) use default arguments without -opt, -debug

PR #2706, opened 2019-02-23
[build][test][environment] improvemnt of test reporting infrostructure

vvlevchenko:
1. reduced number of json file generation
2. Groovy's JsonSlurp replaced with Kotlin Json serialization/deserialization
3. improved naive groovy -> kotlin convertion (46d32c606)
4. some refactoring.


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/ExternalReportUtils.kt
```diff
@@ -0,0 +1,23 @@
+//@file:JvmName("ExternalTest")
```
olonho: Remove if no longer needed.

buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -593,8 +603,10 @@ class RunInteropKonanTest extends KonanTest {
 }
 
 class LinkKonanTest extends KonanTest {
```
olonho: Wouldn't be easier to have common superclasses for all tests enabled and disabled in dev runs, i.e. `DeveloperTest` and `BuildserverTest` where this property is overriden?

vvlevchenko: This hack was introduced long times ago and its elimination is what Pavel working on. I've added only comment, to avoid occasional removing this property, which could be caused by following Idea suggestions.

PR #2714, opened 2019-02-25
Rework deprecated warning in arg parser

LepilkinaElena:


tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,6 +17,21 @@
 
 package org.jetbrains.kliopt
 
+enum class DeprecatedState {
+    ALL {
+        override fun helpMessage(descriptor: Descriptor): String = "Warning: Option/argument -${descriptor.longName} is deprecated!"
```
SvyatoslavScherbina: It makes sense to mention both short and long variants here, e.g. `"Warning: Option/argument -do/-deprecatedOption is deprecated!"`

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,6 +17,21 @@
...
+    },
+    ONLY_SHORT {
+        override fun helpMessage(descriptor: Descriptor): String =
+                "Warning: Short form -${(descriptor as? OptionDescriptor)?.shortName} of option -${descriptor.longName} is deprecated!"
```
SvyatoslavScherbina: I believe that it would be more user-friendly to give a hint here, e.g. `"Warning: -h option is deprecated, use -header instead"`.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,6 +17,21 @@
...
+    },
+    ONLY_FULL {
+        override fun helpMessage(descriptor: Descriptor): String =
+                "Warning: Full form of option -${descriptor.longName} is deprecated!"
```
SvyatoslavScherbina: Same here.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -172,25 +187,30 @@ class ArgParser(optionsList: List<OptionDescriptor>, argsList: List<ArgDescripto
...
+            argDescriptors.getValue(name).deprecated?.let { println ("${it.helpMessage(argDescriptors.getValue(name))}") }
             processedValues.getValue(name).add(arg)
             return true
         }
         return false
     }
 
-    private fun saveAsOption(descriptor: OptionDescriptor, value: String, processedValues: Map<String, MutableList<String>>) {
+    private fun saveAsOption(descriptor: OptionDescriptor, value: String,
+                             processedValues: Map<String, MutableList<String>>, useShort: Boolean = false) {
```
SvyatoslavScherbina: `useShort` parameter name confuses. It is rather `shortNameUsed` or something like this.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -172,25 +187,30 @@ class ArgParser(optionsList: List<OptionDescriptor>, argsList: List<ArgDescripto
...
         if (!descriptor.isMultiple && !processedValues.getValue(descriptor.longName).isEmpty()) {
             printError("Option ${descriptor.longName} is used more than one time!")
         }
-        descriptor.deprecatedWarning?.let { if (processedValues.getValue(descriptor.longName).isEmpty()) println ("Warning: $it") }
+        descriptor.deprecated?.let {
+            if (processedValues.getValue(descriptor.longName).isEmpty())
```
SvyatoslavScherbina: What this check is supposed to prevent? Reporting the same warning twice? If so, then it probably has side effects.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,18 +17,23 @@
...
-    };
+sealed class DeprecatedState(val newName: String? = null) {
+    class AllForms(newName: String? = null) : DeprecatedState(newName) {
+        override fun helpMessage(descriptor: Descriptor) = "Warning: Option/argument -${descriptor.longName}" +
+                "${(descriptor as? OptionDescriptor)?.let { it.shortName?.let{ "-$it" } ?: ""} ?: ""} is deprecated!" +
+                "${newName?.let {" Use -$it instead."} ?: ""}"
+    }
+    class OnlyShort(newName: String? = null) : DeprecatedState(newName) {
+        override fun helpMessage(descriptor: Descriptor) =
+                "Warning: Short form -${(descriptor as? OptionDescriptor)?.shortName} of option " +
```
SvyatoslavScherbina: Does user need to know that this is *short form* of some option? I don't think so. The only piece of information useful here is what option deprecated and how should it be replaced.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,18 +17,23 @@
...
+                        "-${descriptor.longName} is deprecated!${newName?.let {" Use -$it instead."} ?: ""}"
```
SvyatoslavScherbina: Can we advice to use `-${descriptor.longName}` instead?

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -17,18 +17,23 @@
...
-    },
-    ONLY_SHORT {
-        override fun helpMessage(descriptor: Descriptor): String =
-                "Warning: Short form -${(descriptor as? OptionDescriptor)?.shortName} of option -${descriptor.longName} is deprecated!"
-    },
-    ONLY_FULL {
-        override fun helpMessage(descriptor: Descriptor): String =
-                "Warning: Full form of option -${descriptor.longName} is deprecated!"
-    };
+sealed class DeprecatedState(val newName: String? = null) {
```
SvyatoslavScherbina: The entire machinery looks much like a case of over-engineering to me.
Most of the features aren't used currently. Do we have any examples except for `-h`?

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
LepilkinaElena: It can be important, because if I read that option is deprecated I'd think that option is deprecated, not only it's short form.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
LepilkinaElena: There can be cases when user wants to change old short form to new short form.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
LepilkinaElena: It's common tool, it can be used not only in interop tools.

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
olonho: Indeed looks a bit too complicated for the problem solved. Maybe just introduce two separate options, one being deprecated, another is not, and only one is deprecated, and consumer takes concatenation of the two.

LepilkinaElena: I didn't understand approach you suggested.
Also I don't understand what is complicated here? Getting optional value in constructor?

PR #2719, opened 2019-02-26
Added documentation about performance measurement

LepilkinaElena:


HACKING.md
```diff
@@ -92,4 +92,58 @@ To update the blackbox compiler tests set TeamCity build number in `gradle.prope
...
+    
+ **konanRun** task can be run separately for one/several benchmark applications:
+ 
+    ../gradlew :cinterop:konanRun
+    
+ **konanRun** task has parameter `filter` which allows to run only some subset of benchmarks :
+ 
+    ../gradlew :cinterop:konanRun --filter=struct,macros
+    
+ There are also tasks for running benchmarks on JVM (pay attention, some benchmarks e.g. cinterop benchmarks can't be run on JVM)
```
olonho: BTW, why cannot we run interop benches on JVM? It mostly works.

LepilkinaElena: I used functions from kotlin/native stdlib. To try run them on JVM I need to write realization for JVM. But I don't think that it's very useful, because it can't be really compared.

PR #2729, opened 2019-02-28
Fix setting -shims flag in interop plugin

LepilkinaElena:


tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/IncrementalSpecification.groovy
```diff
@@ -208,7 +208,7 @@ class IncrementalSpecification extends BaseKonanSpecification {
         "linkerOpts"             | "'--help'"
         "includeDirs"            | "'src'"
         "includeDirs.allHeaders" | "'src'"
-        "extraOpts"              | "'-shims', 'false'"
+        "extraOpts"              | "'-shims', 'true'"
```
SvyatoslavScherbina: Why did this changed?

LepilkinaElena: -shims in cinterop command line is translated as option is on. Before false was ignored and -shims went to command line, so I just wanted to save same behaviour.

SvyatoslavScherbina: Would the test pass if the entire `extraOpts` line was removed?

LepilkinaElena: It couldn't be removed, because test's purpose to check rebuild of cinterop  if extrasOpts (and other parameters as separate cases) is changed.

PR #2730, opened 2019-02-28
Remove -shims. Warning for -linkerOpts for native flavor

LepilkinaElena:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -177,8 +177,11 @@ private fun processCLib(args: Array<String>, additionalArgs: Map<String, Any> =
 
     val def = DefFile(defFile, tool.substitutions)
 
+    if (flavorName == "native" && argParser.getOrigin("linkerOpts") == ArgParser.ValueOrigin.SET_BY_USER) {
+        warn("Don't use option -linkerOpts/-lopt for native flavor. Necessary libraries will be get from def file.")
```
SvyatoslavScherbina: `"-linkerOpts/-lopt option is not supported by cinterop. Please add linker options to .def file or binary compilation instead"`

PR #2731, opened 2019-02-28
Corner cases with failed code size measurement. Choice tools for buil…

LepilkinaElena:
…d depends on os


tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/Statistics.kt
```diff
@@ -83,7 +83,7 @@ data class MeanVarianceBenchmark(val meanBenchmark: BenchmarkResult, val varianc
 }
 
 fun geometricMean(values: List<Double>, totalNumber: Int? = null) =
```
olonho: `totalNumber = values.size` can be used in default argument.

PR #2734, opened 2019-02-28
Stdlib array copying and sorting from templates

ilya-g:
I've rebased two branches on master: `stdlib-sorting` and `stdlib-array-copying`. There was one merge conflict in `collections/Arrays.kt`.


olonho:
Have you checked if it affects performance in any way?

ilya-g:
@olonho No, I haven't. I don't expect it to affect performance, do you?

The tests in e.g. Linux Performance tests, if they somehow relate to this code, don't show significant execution time difference.

PR #2737, opened 2019-03-01
Add Platform detection support.

olonho:


runtime/src/main/kotlin/kotlin/native/Platform.kt
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+/**
+ * Operating system family.
+ */
+public enum class OsFamily {
+    OTHER,
```
qwwdfsad: Probably it should be called `UNKNOWN`

PR #2739, opened 2019-03-01
Added scripts to turn off/on services on linux machine to decrease no…

LepilkinaElena:
…ises in measurements


performance/scripts/linux_services.list
```diff
@@ -0,0 +1,12 @@
+cups.service
```
olonho: Would be better to embed this list into the script, if possible (just not repeating it twice).

LepilkinaElena: I thought that separate lists are better to have separate list and script that can turn on/off different sets of services. 
We also need to do similar on mac in future, although there are other commands, we can have separated lists for mac and linux. It's possible that we start use another distributive of linux and services can differ, for example.
But if you think it's better to have several lists in one script, I'll do this.

PR #2742, opened 2019-03-04
 Refactor ObjCExportHeaderGenerator to allow per declaration translation

SvyatoslavScherbina:
Also split ObjCExport header and code generation phases.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
@@ -8,6 +8,7 @@ import org.jetbrains.kotlin.backend.konan.descriptors.konanLibrary
 import org.jetbrains.kotlin.backend.konan.ir.KonanSymbols
 import org.jetbrains.kotlin.backend.konan.llvm.*
 import org.jetbrains.kotlin.backend.konan.lower.ExpectToActualDefaultValueCopier
+import org.jetbrains.kotlin.backend.konan.objcexport.ObjCExport
 import org.jetbrains.kotlin.backend.konan.serialization.*
 import org.jetbrains.kotlin.cli.common.messages.AnalyzerWithCompilerReport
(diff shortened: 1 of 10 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -275,6 +284,7 @@ internal val toplevelPhase = namedUnitPhase(
         name = "Compiler",
         description = "The whole compilation process",
         lower = frontendPhase then
+                objCExportPhase then
```
olonho: This is an interesting problem, if the pipeline to be built conditionally for each platform or include noop phases like this one. @alexander-gorshenev WDYT?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -47,6 +48,14 @@ internal val frontendPhase = konanUnitPhase(
...
 
+internal val objCExportPhase = konanUnitPhase(
+        op = {
+            objCExport = ObjCExport(this)
+        },
+        name = "ObjCExport",
+        description = "Objective-C header generation"
+)
+
 internal val psiToIrPhase = konanUnitPhase(
```
olonho: And I'd  like top level `object` here.

SvyatoslavScherbina: How is this related to my PR?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
SvyatoslavScherbina: Unfortunately, header generation phase is part of frontend. Luckily, this phase doesn't affect IR.
Btw, frontend is already different for different backends.

PR #2747, opened 2019-03-06
Support for properties conflicting with C keyword

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
```diff
@@ -331,7 +332,10 @@ internal class ObjCExportNamerImpl(
         assert(mapper.isObjCProperty(property))
 
         StringBuilder().apply {
-            append(property.name.asString())
+            val candidate = property.name.asString()
```
SvyatoslavScherbina: There is an extension point to prevent this kind of name conflicts: you can override `reserved` method in `propertyNames`.

PR #2765, opened 2019-03-14
Server and client to show summary performance information

LepilkinaElena:


tools/benchmarks/shared/src/main/kotlin/org/jetbrains/report/json/ConvertedFromJson.kt
```diff
@@ -36,6 +36,14 @@ interface ConvertedFromJson {
                 element.unquoted()
             else
                 error("Field '$name' in '$element' is expected to be a string. Please, check origin files.")
+
+    fun elementToStringOrNull(element: JsonElement, name:String): String? =
```
olonho: `name: String`

```when (element) {
       is Foo -> ..
       is Bar -> ..
```

tools/benchmarksAnalyzer/src/main/kotlin/main.kt
```diff
@@ -77,8 +77,64 @@ fun getFileContent(fileName: String, user: String? = null): String {
     }
 }
 
+fun getBenchmarkReport(fileName: String, user: String? = null): BenchmarksReport {
+    val benchsResults = getFileContent(fileName, user)
```
olonho: Why not just 
```
fun getBenchmarkReport(fileName: String, user: String? = null) =
     BenchmarksReport.create(JsonTreeParser.parse(getFileContent(fileName, user)))
```?

File: tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt [hunk 3]
```diff
@@ -133,6 +133,34 @@ class SummaryBenchmarksReport (val currentReport: BenchmarksReport,
         }
     }
 
+    enum class Metric(val suffix: Regex) {
+        EXECUTION_TIME(Regex("(?<!(\\.(codeSize|compileTime)))$")),
```
olonho: Why codeSize?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -0,0 +1,284 @@
...
+                buildNumber?.let {
+                    // Check if interesting build id is in cache.
+                    buildsList[it]?.let { buildsList.values }
+                } ?: buildsList.values
+            }
+
+    operator fun get(target: String, buildId: String? = null): Collection<String> {
+        val builds = getBuilds(target, buildId)
+
+        if (builds == null || builds.isEmpty()) {
```
olonho: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-empty.html

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
LepilkinaElena: Execution time has no suffix. So just negative lookbehind assertion to check that there is no suffix.

tools/performance-server/ui/autocomplete/jquery.autocomplete.js
```diff
@@ -0,0 +1,997 @@
...
+        select: function (i) {
+            var that = this;
+            that.hide();
+            that.onSelect(i);
+        },
+
+        moveUp: function () {
+            var that = this;
+
+            if (that.selectedIndex === -1) {
```
olonho: Not sure if `===` to be used here.

LepilkinaElena: This is just taken js library https://github.com/devbridge/jQuery-Autocomplete

tools/performance-server/ui/chartist/scss/chartist.scss
```diff
@@ -0,0 +1,241 @@
+@import "settings/chartist-settings";
```
olonho: Not sure if we shall commit third-party code like that this way.

LepilkinaElena: I had problems with auto installing bower and library as npm module, I had to install it calling install command by myself. I couldn't do this on kubernetes. 

I saved license file for this library.

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
olonho: As we control both sides, maybe makes sense to introduce more reliable reporting mechanism here?

tools/kliopt/org/jetbrains/kliopt/KliOpt.kt
```diff
@@ -120,7 +122,7 @@ class ArgDescriptor(
 
 // Arguments parser.
 class ArgParser(optionsList: List<OptionDescriptor>, argsList: List<ArgDescriptor> = listOf<ArgDescriptor>(),
-                useDefaultHelpShortName: Boolean = true) {
+                val actions: Map<String, Action> = mapOf<String, Action>(), useDefaultHelpShortName: Boolean = true) {
```
olonho: Does `val actions: Map<String, Action> = emptyMap()` work here?

tools/benchmarksAnalyzer/src/main/kotlin/org/jetbrains/analyzer/SummaryBenchmarksReport.kt
LepilkinaElena: Do you suggest  to create one extra property of Benchmark - metric type?

PR #2767, opened 2019-03-14
Stop forcing @ExportForCompiler functions IR deserialization

SvyatoslavScherbina:
Also fix forward declarations in IR deserializer


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanIrModuleDeserializer.kt
```diff
@@ -165,6 +168,18 @@ class KonanIrModuleDeserializer(
                 if (!deserializedTopLevels.contains(it)) reachableTopLevels.add(it) // Assuming forward declarations are always top levels.
             }
 
+            if (descriptor != null) {
+                val module = descriptor.module
+                if (module !in deserializedModuleProtoSymbolTables && !module.isForwardDeclarationModule) {
```
alexander-gorshenev: Does `module !in deserializedModuleProtoSymbolTables` supposed to mean "we haven't seen this module yet"?

SvyatoslavScherbina: Yes.

PR #2768, opened 2019-03-14
Refine LLVM bitcode imports detection

SvyatoslavScherbina:
Don't link native libraries if external function requires only library-stored bitcode.
Also make Objective-C interop stubs bitcode less harmful by excluding selectors.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
@@ -296,8 +296,13 @@ internal class Llvm(val context: Context, val llvmModule: LLVMModuleRef) {
         return LLVMAddFunction(llvmModule, "llvm.memcpy.p0i8.p0i8.i32", functionType)!!
     }
 
-    internal fun externalFunction(name: String, type: LLVMTypeRef, origin: CompiledKonanModuleOrigin): LLVMValueRef {
-        this.imports.add(origin)
+    internal fun externalFunction(
+            name: String,
+            type: LLVMTypeRef,
+            origin: CompiledKonanModuleOrigin,
+            independent: Boolean = false
+    ): LLVMValueRef {
+        this.imports.add(origin, onlyBitcode = independent)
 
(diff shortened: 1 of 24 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
```diff
@@ -350,16 +356,28 @@ internal class Llvm(val context: Context, val llvmModule: LLVMModuleRef) {
...
+            usedBitcode.add(library)
+            if (!onlyBitcode) {
+                usedNativeDependencies.add(library)
+            }
         }
 
-        override fun isImported(library: KonanLibrary): Boolean = library in usedLibraries
+        override fun bitcodeIsUsed(library: KonanLibrary) = library in usedBitcode
+
+        override fun nativeDependenciesAreUsed(library: KonanLibrary) = library in usedNativeDependencies
```
olonho: Great!

runtime/src/main/kotlin/kotlin/native/internal/Annotations.kt
```diff
@@ -104,4 +104,12 @@ internal annotation class PointsTo(vararg val onWhom: Int)
...
\ No newline at end of file
+internal annotation class TypedIntrinsic(val kind: String)
+
+/**
+ * Indicates that `@SymbolName external` function is implemented in library-stored bitcode
+ * and doesn't have native dependencies.
+ */
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.BINARY)
+annotation class Independent
```
olonho: Not sure if the name is too enlightening, what about smth like `LocallyDefined`? Also what confuses me, is that functions in bitcode usually call code in the native library, why can we ignore that? Or it's only applied to declarations without real external deps?

SvyatoslavScherbina: I agree that the name is confusing, but the semantics is too subtle. `LocallyDefined` is nice, but does it explain why Objective-C class member stubs are locally defined, while category member stubs aren't? Bodies of stubs of both kinds are indistinguishable.

> Or it's only applied to declarations without real external deps?

Exactly.

PR #2771, opened 2019-03-14
Refactor CommonBackendContext, remove confusing methods

udalov:
Please merge this only after https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2815 is accepted and merged into Kotlin master.

I've run `:backend.native:tests:run` via the composite build to verify that tests still work.


vvlevchenko:
will be applied together with #2781

PR #2774, opened 2019-03-15
Add documentation for primitive types conversion

qurbonzoda:
These docs are already added to big kotlin.

### DIFF
--- a/runtime/src/main/kotlin/kotlin/Primitives.kt
+++ b/runtime/src/main/kotlin/kotlin/Primitives.kt
@@ -191,18 +191,58 @@ public final class Byte private constructor() : Number(), Comparable<Byte> {
     public inline operator fun unaryMinus(): Int =
             -this.toInt()
 
+    /** Returns this value. */
     public inline override fun toByte(): Byte =
             this
+    /**
+     * Converts this [Byte] value to [Char].
+     *
+     * Returns the `Char` with the numeric value equal to this value sign-extended to 16 bits.
+     */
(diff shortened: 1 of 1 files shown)


runtime/src/main/kotlin/kotlin/Primitives.kt
```diff
@@ -191,18 +191,58 @@ public final class Byte private constructor() : Number(), Comparable<Byte> {
...
+     */
     @TypedIntrinsic(IntrinsicType.SIGN_EXTEND)
     external public override fun toChar(): Char
+    /**
+     * Converts this [Byte] value to [Short].
+     *
+     * The resulting `Short` value represents the same numerical value as this `Byte`.
+     *
+     * The least significant 8 bits of the resulting `Short` value are the same as the binary representation of this `Byte` value,
+     * whereas the most significant 8 bits are filled with the bit sign of this value.
```
olonho: This is definition of sign-extension, why we need to explicitly explain it.

ilya-g: We do not expect every reader to be aware of what sign extension/zero extension is.

olonho: This is inconsistent with `toChar` docs above then.

ilya-g: Ok, let's merge this in the current revision to avoid diverging the docs between platforms. We'll note this inconsistency about `toChar` and try to provide a better wording for its description in a separate PR.

PR #2776, opened 2019-03-15
FreezableAtomicReference prototype.

olonho:


olonho:
Merging, hopefully it will make sense for consumers.

PR #2787, opened 2019-03-19
Add ssh arguments for ssh executor.

PavelPunegov:


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -190,7 +191,7 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
                 upload(executable)
                 executable = "$remoteDir/${File(executable).name}"
                 execFile = executable
-                commandLine = arrayListOf("/usr/bin/ssh", remote) + commandLine
+                commandLine = arrayListOf("/usr/bin/ssh") + sshArgs + remote + commandLine
```
olonho: Maybe smth like 
```
val sshHome = System.getenv("SSH_HOME") ?:  "/usr/bin"
```
?

PavelPunegov: Yes, it may help in some cases like cross-testing on windows

PR #2789, opened 2019-03-20
Introduce infrastructure for testing `-Xcoverage` flag

sbogolepov:
Also added filter for compiler-generated functions.


backend.native/tests/build.gradle
```diff
@@ -3212,66 +3215,57 @@ task runKonanRegexTests(type: RunKonanTest) {
...
-            args "-o", "$dir/program.profdata"
-        }
-        // Should fail in case of corrupted __llvm_coverage.
-        execLlvm("llvm-cov") {
-            def program = buildExePath()
-            def suffix = target.family.exeSuffix
-            args "report"
-            args "$program.$suffix"
-            args "-instr-profile", "$dir/program.profdata"
+if (getTarget(project).name == "ios_x64" || getTarget(project).name == "macos_x64") {
```
PavelPunegov: Could be replaced with method `isAppleTarget(project)`

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/CoverageTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    override fun configure(closure: Closure<Any>): Task {
+        super.configure(closure)
+        if (!(project.property("useCustomDist") as Boolean)) {
+            setRootDependency("${target}CrossDist", "${target}CrossDistRuntime", "commonDistRuntime", "distCompiler")
+        }
+        dependsOn(project.tasks.getByName("run$binaryName"))
+        return this
+    }
+
+    private fun setRootDependency(vararg s: String) = s.forEach { dependsOn(project.rootProject.tasks.getByName(it)) }
```
PavelPunegov: Isn't this mean that all `crossdist`s will depend on the test task? The test task should depend on `crossDist` and build task

PR #2791, opened 2019-03-20
Rewrite tests infrastructure to Kotlin.

PavelPunegov:
Split build and run logic. Build tests with Konan gradle plugin.


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    var disabled: Boolean
+        get() = !enabled
+        @Optional set(value) { enabled = !value }
+
+    /**
+     * Test output directory. Used to store processed sources and binary artifacts.
+     */
+    lateinit var outputDirectory: String
```
ilmat192: Can we make `outputDirectory`, `testLogger` and `executable` abstract and get rid of `lateinits`? We still initialize these properties in constructors of subclasses.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+
+    @Suppress("UnstableApiUsage")
+    override fun configure(config: Closure<*>): Task {
+        super.configure(config)
+
+        // Set Gradle properties for the better navigation
+        group = LifecycleBasePlugin.VERIFICATION_GROUP
+        description = "Kotlin/Native test infrastructure task"
+
+        if (!::arguments.isInitialized) {
```
ilmat192: Can we just initilize `arguments` with an empty list and get rid of `lateinit` and this check?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+    @TaskAction
+    open fun run() = project.executeAndCheck(project.file(executable).toPath(), arguments)
+
+    // Converts to runner's pattern
+    private fun String.convertToPattern() = this.replace('/', '.').replace(".kt", "") + (".*")
+
+    internal fun ProcessOutput.print(prepend: String = "") {
+        if (project.verboseTest)
+            println(prepend + """
+                |stdout:$stdOut
```
ilmat192: May be insert a `\n` between `"stdout:"`, `"stderr:"` and the corresponding outputs?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+                |stderr:$stdErr
+                |exit code: $exitCode
+                """.trimMargin())
+    }
+}
+
+/**
+ * Create a test task of the given type. Supports configuration with Closure passed form build.gradle file.
+ */
+fun <T: KonanTest> Project.createTest(name: String, type: Class<T>, config: Closure<*>): T =
```
ilmat192: It may make sense to add an overloaded method taking an `Action<T>` as a last parameter. In Gradle Kotlin DSL, a lambda can be passed as such an `Action` so this method will be useful in *.kts build scripts.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+    lateinit var statistics: Statistics
+
+    @TaskAction
+    override fun run() = runProcess(
+            executor = project.executor::execute,
+            executable = executable,
+            args = arguments
+    ).run {
+        statistics = parse(stdOut)
+        print()
```
ilmat192: IMHO it's better to use `let { ... }` here. This `print()`  call uses implicit `this` as a receiver but at first sight it looks like a call of the top-level `kotlin.print()` function. Within the `let` block this call looks less confusing: `it.print()`

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+open class KonanLocalTest : KonanTest() {
+    init {
+        // local tests built into a single binary with the known name
+        val target = project.testTarget
+        outputDirectory = project.testOutputLocal
+        executable = "$outputDirectory/${target.name}/localTest.${target.family.exeSuffix}"
+        testLogger = Logger.SILENT
+    }
+
+    @Optional
```
ilmat192: Why do we mark these properties as `@Optional`? `@Optional` makes sense only in pair with other annotations like `@Input` or `@InputFiles` ([link](https://docs.gradle.org/current/userguide/more_about_tasks.html)). But as far as I understand we don't use up-to-date checks for these tasks and always run them so there is no need for such annotations.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+     * Should this test fail or not.
+     */
+    @Optional
+    var expectedFail = false
+
+    /**
+     * Used to validate output as a gold value.
+     */
+    @Optional
+    lateinit var goldValue: String
```
ilmat192: May be replace some of `lateinits` here and in other classes with just nullable properties? It looks like a null value is more suitable for expressing "there is no value for this property" than an uninitialized lateinit. So I suggest to keep only lateinits that _must_ be configured by a user (like `source`) and replace other ones with nullable properties.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+
+    @Optional
+    var enableKonanAssertions = true
+
+    private var _flags: MutableList<String> = mutableListOf()
+    /**
+     * Compiler flags used to build a test.
+     */
+    var flags: MutableList<String>
+        get() {
```
ilmat192: It's possible to use `field` here and get rid of `_flags`:
```kotlin
get() {
    if (enableKonanAssertions) field.add("-ea")
    return field
}

set(value) {
    field = value
}
```
Also is't it better to avoid mutation of the backing field? May be make this property just `List<String>` (not mutable) and replace this getter with something like: 

```kotlin
 get() = if (enableKonanAssertions) field + "-ea" else field
```
?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+/**
+ * This is another way to run the konanc compiler. It runs a konanc shell script.
+ *
+ * @note This task is not intended for regular testing as project.exec + a shell script isolate the jvm from IDEA.
+ * @see KonanLocalTest to be used as a regular task.
+ */
+open class KonanDriverTest : KonanStandaloneTest() {
+    override fun configure(config: Closure<*>): Task {
+        super.configure(config)
+        if (doBefore != null) doFirst(doBefore!!)
```
ilmat192: It's possible to use the `?.let`-idiom here to avoid `!!`:
```kotlin
doBefore?.let { doFirst(it) }
```

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+        super.run()
+    }
+
+    private fun konan() {
+        val dist = project.rootProject.file(project.findProperty("org.jetbrains.kotlin.native.home") ?:
+        project.findProperty("konan.home") ?: "dist")
+        val konancDriver = if (HostManager.hostIsMingw) "konanc.bat" else "konanc"
+        val konanc = File("${dist.canonicalPath}/bin/$konancDriver").absolutePath
+
+        File(executable).mkdirs()
```
ilmat192: Shouldn't it be `File(executable).parentFile.mkdirs()`?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -0,0 +1,436 @@
...
+
+        val args = mutableListOf("-output", executable).apply {
+            if (project.testTarget != HostManager.host) {
+                add("-target")
+                add(project.testTarget.visibleName)
+            }
+            addAll(getSources())
+            addAll(flags)
+            addAll(project.globalTestArgs)
+            if (enableKonanAssertions) add("-ea")
```
ilmat192: We've already taken "-ea" into account in the `flags` property.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/TestDirectives.kt
```diff
@@ -0,0 +1,64 @@
...
+    // Remove diagnostic parameters in external tests.
+    val srcText = srcFile.readText().replace(Regex("<!.*?!>(.*?)<!>")) { match -> match.groupValues[1] }
+
+    if (srcText.contains("// WITH_COROUTINES")) {
+        val coroutineHelpersFileName = "$outputDirectory/helpers.kt"
+        createFile(coroutineHelpersFileName, createTextForHelpers(true))
+        result.add(coroutineHelpersFileName)
+    }
+
+    val filePattern = Pattern.compile("(?m)// *FILE: *(.*)")
```
ilmat192: May be replace spaces in this regex with general whitespaces (`\s`)?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/TestDirectives.kt
```diff
@@ -0,0 +1,64 @@
...
+    val matcher = filePattern.matcher(srcText)
+
+    if (!matcher.find()) {
+        // There is only one file in the input
+        val filePath = "$outputDirectory/${srcFile.name}"
+        registerKtFile(result, filePath, srcText)
+    } else {
+        // There are several files
+        var processedChars = 0
+        while (true) {
```
ilmat192: It looks like we can replace this loop with a `do..while` one:
```kotlin
do {
    ...
    val nextFileExists = matcher.find()
    ...
} while(!nextFileExists)
```

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/Utils.kt
```diff
@@ -1,24 +1,115 @@
...
+    }
+}
+
+//region Task dependency.
+
+fun Project.setDistDependencyFor(taskName: String) {
+    project.setDistDependencyFor(project.tasks.getByName(taskName))
+}
+
+fun Project.setDistDependencyFor(t: Task) {
```
ilmat192: IMHO making this function an extension for `Task` makes its call more readable: `task.dependsOnDist()` or `task.setDistDependency()`.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/Utils.kt
```diff
@@ -1,24 +1,115 @@
...
+
+/**
+ * Sets the same dependencies for the receiver task from the given [task]
+ * and make [task] depend on receiver
+ */
+fun Task.setSameDependenciesAs(task: Task) {
+    val dependencies = task.dependsOn.toList() // save to the list, otherwise it will cause cyclic dependency.
+    this.dependsOn(dependencies)
+    // Run task should depend on compile task.
+    task.dependsOn(this)
```
ilmat192: This effect (making `task` depending on the receiver) doesn't look intuitive. May be remove it from this function and specify such a dependency manually at a caller side? Or at least rename the function to reflect this effect?

backend.native/tests/build.gradle
```diff
@@ -249,6 +256,84 @@ void dependsOnPlatformLibs(Task t) {
...
 
+/**
+ * Creates a task for a standalone test. Configures runner and adds building task.
+ */
+Task standaloneTest(String name, Closure configureClosure) {
+    return KotlinNativeTestKt.createTest(project, name, KonanStandaloneTest) { task ->
+        task.configure(configureClosure)
+        if (task.enabled) {
+            konanArtifacts {
+                program(name, targets: [target.name ?: 'host']) {
```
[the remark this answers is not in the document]
PavelPunegov: Fixed, it couldn't

backend.native/tests/build.gradle
```diff
@@ -3077,32 +3160,86 @@ kotlinNativeInterop {
...
+
+/**
+ * Creates a task for the interop test. Configures runner and adds library and test build tasks.
+ */
+Task interopTest(String name, Closure<KonanInteropTest> configureClosure) {
+    return KotlinNativeTestKt.createTest(project, name, KonanInteropTest) { task ->
+        task.configure(configureClosure)
+        if (task.enabled) {
+            konanArtifacts {
+                def conf = project.kotlinNativeInterop[task.interop]
```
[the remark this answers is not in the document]
PavelPunegov: Yes, it should have been done. I just was stuck with several issues in interop. I am going to fix it

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/TestDirectives.kt
PavelPunegov: It actually can be rewritten using stdlib functions for parsing. This is a straight forward conversion and I'm going to return to it later with rewriting external tests.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
PavelPunegov: Removed most of lateinits.

PR #2792, opened 2019-03-20
Change axises and delete action in API

LepilkinaElena:


tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -64,6 +64,22 @@ object LocalCache {
     fun buildExists(target: String, buildNumber: String) =
             buildsInfo[target][buildNumber]?.let { true } ?: false
 
+    fun delete(target: String, builds: Iterable<String>, bintrayUser: String, bintrayPassword: String) {
```
olonho: Why `delete` is needed at all?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -64,6 +64,22 @@ object LocalCache {
...
+        // Delete from bintray.
+        val buildsDescription = getBuildsInfoFromBintray(target).lines()
+        buildsDescription.filter {
+            val buildNumber = it.substringBefore(',')
+            buildNumber !in builds
+        }
+        // Upload new version of file.
+        val uploadUrl = "$uploadBintrayUrl/$bintrayPackage/latest/$target/$buildsFileName?publish=1&override=1"
+        sendUploadRequest(uploadUrl, buildsDescription.joinToString("\n"), bintrayUser, bintrayPassword)
```
olonho: Why it uploads smth?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -242,6 +261,12 @@ fun router() {
         response.sendStatus(200)
     })
 
+    router.get("/delete/:target", { request, response ->
+        val buildsToDelete: List<String> = request.query.builds.toString().split(",").map { it.trim() }
+        LocalCache.delete(request.params.target, buildsToDelete, request.query.user, request.query.key)
+        response.sendStatus(200)
```
olonho: Shalln't it only succeed if smth was indeed deleted?

tools/performance-server/src/main/kotlin/routes/route.kt
LepilkinaElena: We have to remove some builds after some time. We couldn't save all builds for very long period. Moreover, in case of some runs with errors.

tools/performance-server/src/main/kotlin/routes/route.kt
LepilkinaElena: It removes data from file with description of builds and should renew it on bintray.

PR #2800, opened 2019-03-21
Better encapsulation of deserializer.

alexander-gorshenev:


alexander-gorshenev:
This is a native counterpart of https://github.com/JetBrains/kotlin/pull/2202

PR #2813, opened 2019-03-25
Cheaper stack

olonho:


runtime/src/main/cpp/Memory.h
```diff
@@ -455,6 +466,8 @@ extern "C" {
     ObjHeader* result = name(__VA_ARGS__, OBJ_RESULT);  \
     return result;                                      \
   }
+#define HEAP_RETURN_SLOT(slot) \
```
SvyatoslavScherbina: Why not enforce return slot to be always stack-located instead?
I believe that this is
1) possible
2) simpler
3) more efficient

olonho: It's not just stack allocated, it must be in slots seen when walking the stack. Let's first stabilize this design and see its performance implications and the proceed with optimisations.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -616,38 +646,71 @@ inline uint32_t freeableSize(MemoryState* state) {
...
-    RuntimeAssert(container->objectCount() == 1,
-        "cycle collector shall only work with single object containers");
+    RuntimeAssert(!Atomic && !container->shareable(), "Cycle collector shalln't be used with shared objects yet");
+    RuntimeAssert(container->objectCount() == 1, "cycle collector shall only work with single object containers");
     // We do not use cycle collector for frozen objects, as we already detected
     // possible cycles during freezing.
     // Also do not use cycle collector for provable acyclic objects.
     int color = container->color();
     if (color != CONTAINER_TAG_GC_PURPLE && color != CONTAINER_TAG_GC_GREEN) {
+      RuntimeAssert(!container->shareable(), "Shareable cannot be cycle collected");
```
SvyatoslavScherbina: Isn't it the same as checked above?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -616,38 +646,71 @@ inline uint32_t freeableSize(MemoryState* state) {
...
+      GarbageCollect();
+    }
+  }
+  state->toRelease->push_back(container);
+}
+
 inline void initThreshold(MemoryState* state, uint32_t gcThreshold) {
   state->gcThreshold = gcThreshold;
-  state->toFree->reserve(gcThreshold);
+  state->toRelease->reserve(gcThreshold);
```
SvyatoslavScherbina: Why was `toFree->reserve(gcThreshold)` removed?

SvyatoslavScherbina: Do you suppose that it is unlikely to get filled up to threshold because `toRelease` will become full earlier?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1107,15 +1181,14 @@ void FreeContainer(ContainerHeader* container) {
 
 void ObjectContainer::Init(const TypeInfo* typeInfo) {
   RuntimeAssert(typeInfo->instanceSize_ >= 0, "Must be an object");
-  uint32_t alloc_size =
-      sizeof(ContainerHeader) + typeInfo->instanceSize_;
+  uint32_t alloc_size = sizeof(ContainerHeader) + typeInfo->instanceSize_;
   header_ = AllocContainer(alloc_size);
+  RuntimeCheck(header_ != nullptr, "Cannot alloc memory");
```
SvyatoslavScherbina: So the check below isn't required anymore?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1548,52 +1618,143 @@ void UpdateHeapRefIfNull(ObjHeader** location, const ObjHeader* object) {
...
-  MEMORY_LOG("Garbage collect\n")
+  GC_LOG(">>> GC: toFree %d toRelease %d\n", state->toFree->size(), state->toRelease->size())
 
 #if GC_ERGONOMICS
   auto gcStartTime = konan::getTimeMicros();
 #endif
 
   state->gcInProgress = true;
 
+  incrementStack(state);
```
SvyatoslavScherbina: Do I understand correctly that long-living stack references become cycle candidates during every GC?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1638,8 +1800,7 @@ void Kotlin_native_internal_GC_resume(KRef) {
   MemoryState* state = memoryState;
   if (state->gcSuspendCount > 0) {
     state->gcSuspendCount--;
-    if (state->toFree != nullptr &&
-        freeableSize(state) >= state->gcThreshold) {
+    if (state->toFree != nullptr && freeableSize(state) >= state->gcThreshold) {
```
SvyatoslavScherbina: Btw, I guess `gcSuspendCount == 0` should be checked here too.

runtime/src/main/cpp/Memory.cpp
olonho: Yes, it's no longer even close to that threshold.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1650,6 +1811,7 @@ void Kotlin_native_internal_GC_stop(KRef) {
 #if USE_GC
   if (memoryState->toFree != nullptr) {
     GarbageCollect();
+    // TODO: what shall we do with toRelease here? Just leak?
```
SvyatoslavScherbina: Btw, this GC.start/stop feature doesn't seem to work properly.
Consider e.g.
```
kotlin.native.internal.GC.stop()
Any().freeze()
```

runtime/src/main/cpp/Worker.cpp
```diff
@@ -373,34 +371,35 @@ void* workerRoutine(void* argument) {
...
-    } catch (ObjHolder& e) {
+        result = transfer(resultHolder.obj(), job.transferMode);
+      } catch (ObjHolder& e) {
         ok = false;
         if (worker->errorReporting())
             ReportUnhandledException(e.obj());
+      }
+      // Notify the future.
+      job.future->storeResultUnlocked(result, ok);
+      resultHolder.clear();
```
SvyatoslavScherbina: As far as I see, by this moment another thread may have got this object already, so clearing the holder here may cause data race.

Btw, can `transfer` have `ObjHolder` parameter instead and clear it by itself? This looks making sense to me.

runtime/src/main/cpp/Memory.cpp
olonho: Reworked.

runtime/src/main/cpp/Memory.cpp
olonho: Exactly. Any alternatives?

runtime/src/main/cpp/Memory.cpp
olonho: Yeah, more testing is needed, not sure if in this PR.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1759,7 +1761,17 @@ void GarbageCollect() {
...
   incrementStack(state);
   processDecrements(state);
+  size_t beforeDecrements = state->toRelease->size();
   decrementStack(state);
+  size_t afterDecrements = state->toRelease->size();
+  ssize_t stackReferences = afterDecrements - beforeDecrements;
+  if (stackReferences * 5 > state->gcThreshold) {
+     auto newThreshold = state->gcThreshold * 3 / 2 + 1;
+     if (newThreshold < kMaxErgonomicThreshold) {
+       state->gcThreshold = newThreshold;
```
SvyatoslavScherbina: Is space in `toRelease` not reserved intentionally?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1817,28 +1834,35 @@ void Kotlin_native_internal_GC_resume(KRef) {
...
+  if (memoryState->toRelease != nullptr) {
     GarbageCollect();
-    // TODO: what shall we do with toRelease here? Just leak?
+    konanDestructInstance(memoryState->toRelease);
     konanDestructInstance(memoryState->toFree);
     konanDestructInstance(memoryState->roots);
+    memoryState->toRelease = nullptr;
     memoryState->toFree = nullptr;
     memoryState->roots = nullptr;
+    state->gcSuspendCount = 0;
```
SvyatoslavScherbina: What if GC is stopped while suspended?

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: Not yet, but I believe we should eventually apply some efforts to find some.

runtime/src/main/cpp/Memory.cpp
olonho: Reworked here.

runtime/src/main/cpp/Memory.cpp
olonho: Hard to tell what is right, but moved zeroing above.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -635,20 +753,72 @@ inline void DecrementRC(ContainerHeader* container) {
...
+      // Also do not use cycle collector for provable acyclic objects.
+      int color = container->color();
+      if (color != CONTAINER_TAG_GC_PURPLE && color != CONTAINER_TAG_GC_GREEN) {
+        container->setColorAssertIfGreen(CONTAINER_TAG_GC_PURPLE);
+        if (!container->buffered()) {
+          container->setBuffered();
+          state->toFree->push_back(container);
+          if (state->gcSuspendCount == 0 && state->toFree->size() >= state->gcThreshold) {
+              GC_LOG("Calling GC from DecrementRC: %d\n", state->toRelease->size())
+              garbageCollect(state, false);
```
SvyatoslavScherbina: This `DecrementRC` variant is used only during GC. What's the purpose of invoking GC inside it?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -111,9 +111,10 @@ volatile int allocCount = 0;
 volatile int aliveMemoryStatesCount = 0;
 
 // Forward declarations.
+__attribute__((noinline))
```
SvyatoslavScherbina: `NO_INLINE`?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1216,8 +1206,9 @@ ContainerHeader* AllocContainer(MemoryState* state, size_t size) {
   // We recycle elements of finalizer queue for new allocations, to avoid trashing memory manager.
   ContainerHeader *container = state->finalizerQueue, *previous = nullptr;
   while (container != nullptr) {
-    // TODO: shall it be >= or ==?
-    if (container->hasContainerSize() && container->containerSize() == size) {
+    // TODO: shall it be == instead?
+    if (container->hasContainerSize() &&
+        container->containerSize() >= size && container->containerSize() <= 2 * size) {
```
SvyatoslavScherbina: `container->containerSize() <= 2 * size` is too permissive. Can it be more like `container->containerSize() <= size + 8`?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1808,14 +1799,12 @@ void UpdateStackRef(ObjHeader** location, const ObjHeader* object) {
...
   ObjHeader* old = *location;
-  if (old != object) {
-    if (object != nullptr) {
-      AddStackRef(object);
-    }
-    *const_cast<const ObjHeader**>(location) = object;
-    if (old != nullptr ) {
-      ReleaseStackRef(old);
-    }
+  if (object != nullptr) {
```
SvyatoslavScherbina: Does removing `old != object` check have significant positive effect?

runtime/src/main/cpp/Memory.h
olonho: Did that.

runtime/src/main/cpp/Memory.cpp
olonho: Just remove excessive branch, but can return.

runtime/src/main/cpp/Memory.h
SvyatoslavScherbina: Did that help?

runtime/src/main/cpp/Memory.h
```diff
@@ -561,27 +509,73 @@ void EnsureNeverFrozen(ObjHeader* obj);
...
+  // As they go in pair, sizeof(FrameOverlay) % sizeof(void*) == 0 is always held.
+  int32_t parameters;
+  int32_t count;
+};
+
 // Class holding reference to an object, holding object during C++ scope.
 class ObjHolder {
  public:
-   ObjHolder() : obj_(nullptr) {}
+   ObjHolder() : obj_(nullptr) {
```
SvyatoslavScherbina: So it is actually now more like `ObjStackHolder`, `StackRefHolder` or even `StackFrame`.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -62,10 +62,11 @@ enum {
 THREAD_LOCAL_VARIABLE KInt g_currentWorkerId = 0;
 
 KNativePtr transfer(ObjHolder* holder, KInt mode) {
+  holder->hold();
```
SvyatoslavScherbina: Why not just
```
void* result = CreateStablePointer(holder.obj())
holder.clear()
if (!ClearSubgraphReferences(holder->obj(), mode == CHECKED)) {
  ThrowWorkerInvalidState();
}
return result;
```

Technically almost the same, but avoids leaking abstractions and pairs with `AdoptStablePointer` semantically.

PR #2814, opened 2019-03-26
Support Gradle test patterns in the test runner (KT-30533)

ilmat192:


runtime/src/main/kotlin/kotlin/native/internal/test/TestStatistics.kt
```diff
@@ -26,26 +26,44 @@ internal class MutableTestStatistics: TestStatistics {
     override var totalSuites: Int = 0; private set
 
     override val failed: Int
-        get() = failedTests_.size
+        get() = _failedTests.size
```
olonho: Why _ before? Generally _ after gives better tooling hints.

ilmat192: According to the coding conventions: https://kotlinlang.org/docs/reference/coding-conventions.html#property-names.

snrostov:
Is there any issue about stacktraces to track it?

https://github.com/JetBrains/kotlin-native/blob/5a9aac55da52db0395c8f84c3c5f251b9df050be/runtime/src/main/kotlin/kotlin/native/internal/test/TeamCityLogger.kt#L54

runtime/src/main/kotlin/kotlin/native/internal/test/TestStatistics.kt
olonho: Hmm, maybe coding convention better be rethought here. Anyway, not in this PR.

PR #2818, opened 2019-03-27
Add quick workaround for https://bugs.swift.org/browse/SR-10177

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -226,6 +231,48 @@ internal class ObjCExport(val context: Context) {
...
+                "-emit-llvm",
+                "-c", "-o", bitcode.absolutePath
+        )
+
+        val result = Command(clangCommand).getResult(withErrors = true)
+
+        if (result.exitCode == 0) {
+            context.llvm.additionalProducedBitcodeFiles += bitcode.absolutePath
+        } else {
+            // Note: ignoring compile errors intentionally.
```
olonho: Somewhat not very consistent - if error is important, why we ignore it?

SvyatoslavScherbina: 1) Because the same error encountered when importing the framework is more debuggable, e.g. because it doesn't involve any temp files and occurs at the moment a user can expect it
2) To keep the old behaviour.

PR #2825, opened 2019-03-27
Update XCode to 10.2.

olonho:


README.md
```diff
@@ -27,7 +27,7 @@ Then, build the compiler and libraries:
 
 To build with experimental targets support compile with `-Porg.jetbrains.kotlin.native.experimentalTargets`.
 
-The build can take about an hour on a Macbook Pro.
+./gradlew dist -PsharedProjectPathThe build can take about an hour on a Macbook Pro.
```
SvyatoslavScherbina: ?

konan/konan.properties
```diff
@@ -25,6 +25,7 @@ dependencyProfiles = default alt
 airplaneMode = false
 # if true, ignores Xcode version check.
 ignoreXcodeVersionCheck = false
+useFixedXcodeVersion = 10.2
```
SvyatoslavScherbina: Doesn't this force a user to use Xcode 10.2?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/FrameworkTest.kt
```diff
@@ -85,9 +85,21 @@ open class FrameworkTest : DefaultTask() {
     }
 
     private fun runTest(testExecutable: Path) {
+        val target = project.testTarget()
+        val platform = project.platformManager().platform(target)
+        assert(platform.configurables is AppleConfigurables)
```
SvyatoslavScherbina: What's the motivation behind asserting the condition to be checked by `as` at next line?

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/FrameworkTest.kt
```diff
@@ -85,9 +85,21 @@ open class FrameworkTest : DefaultTask() {
...
+        val configs = platform.configurables as AppleConfigurables
+        val swiftPlatform = when (target) {
+            KonanTarget.IOS_X64 -> "iphoneossimulator"
```
SvyatoslavScherbina: `iphonesimulator`?

konan/konan.properties
olonho: Maybe filter that out in release bundle.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/FrameworkTest.kt
olonho: Not much, I guess.

konan/konan.properties
olonho: Not sure if I like approach taken here, let's discuss.

Issue #2803, opened 2019-03-22
compile as static library not work with android


vlike:
kotlinc-native -version
```
info: kotlinc-native 1.3.11 (JRE 1.8.0_131-b11)
Kotlin/Native: 1.0.3
```

kotlin code:
test.kt
```
fun wxtest(){
    println("===Hello test!")
}
```

compile command:
` kotlinc-native -produce static -target android_arm32 -o knlib ./test.kt`

CMakeLists.txt config in android studio:
```
target_link_libraries(
        ...
        ...
        ...

        ${MAIN_PATH}/cpp/other/libs/armeabi-v7a/libknlib.a

        z m
)
```

cpp code:
```
#include "knlib_api.h"

char *test2() {
    knlib_ExportedSymbols* lib = knlib_symbols();
    lib->kotlin.root.wxtest();
    return nullptr;
}
```

message:

```
Build command failed.
Error while executing process /Volumes/data/SDK/android-sdk-macosx/cmake/3.6.4111459/bin/cmake with arguments {--build /Volumes/data/test/app/.externalNativeBuild/cmake/debug/armeabi-v7a --target native-lib}
[1/2] Building CXX object CMakeFiles/native-lib.dir/src/main/cpp/demo/test2.cpp.o
[2/2] Linking CXX shared library ../../../../build/intermediates/cmake/debug/obj/armeabi-v7a/libnative-lib.so

[...]

PR #2828, opened 2019-03-28
Embed some linker options to static frameworks

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CompilerOutput.kt
```diff
@@ -121,3 +127,24 @@ private fun parseAndLinkBitcodeFile(llvmModule: LLVMModuleRef, path: String) {
...
 }
+
+private fun embedAppleLinkerOptionsToBitcode(llvm: Llvm, config: KonanConfig) {
+    fun findEmbeddableOptions(options: List<String>): List<List<String>> {
+        val result = mutableListOf<List<String>>()
+        val iterator = options.iterator()
+        loop@while (iterator.hasNext()) {
+            val option = iterator.next()
+            result += when {
+                option.startsWith("-l") -> listOf(option)
```
olonho: What about `-l foo`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmLinkOptions.kt
```diff
@@ -0,0 +1,11 @@
+package org.jetbrains.kotlin.backend.konan.llvm
+
+import llvm.LLVMAddNamedMetadataOperand
+import llvm.LLVMModuleRef
+
+fun embedLlvmLinkOptions(module: LLVMModuleRef, options: List<List<String>>) {
+    options.forEach {
```
olonho: Shalln't we assert that it only works on Apple targets?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CompilerOutput.kt
SvyatoslavScherbina: Not supported by the linker.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmLinkOptions.kt
SvyatoslavScherbina: This is kinda not true.

olonho: You mean, we could use it on other platforms? Like to fix https://github.com/JetBrains/kotlin-native/issues/2803

SvyatoslavScherbina: Maybe.
E.g. https://github.com/llvm-mirror/llvm/commit/e7676fec11b02e4b698b5ffc99e1901246a7bf66
Not sure whether it is supported by GNU ld.

PR #2846, opened 2019-04-03
Use compiler-generated C stubs to call Objective-C methods

SvyatoslavScherbina:
Also
*   Use compiler-generated C stubs when overriding all Objective-C methods
*   Support block types in C stubs generator
*   Generate trap before each unreachable instruction in debug mode

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
@@ -6,8 +6,10 @@ import org.jetbrains.kotlin.cli.common.messages.MessageCollector
 import org.jetbrains.kotlin.konan.exec.Command
 import org.jetbrains.kotlin.konan.file.*
 import org.jetbrains.kotlin.konan.target.ClangArgs
+import org.jetbrains.kotlin.konan.target.Family
+import org.jetbrains.kotlin.konan.target.KonanTarget
 
-class CStubsManager {
+class CStubsManager(private val target: KonanTarget) {
 
     fun getUniqueName(prefix: String) = "$prefix${counter++}"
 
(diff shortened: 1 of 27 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
```diff
@@ -18,13 +20,24 @@ class CStubsManager {
     fun compile(clang: ClangArgs, messageCollector: MessageCollector, verbose: Boolean): File? {
         if (stubs.isEmpty()) return null
 
-        val cSource = createTempFile("cstubs", ".c").deleteOnExit()
+        val compilerOptions = mutableListOf<String>()
```
olonho: Maybe add TODO about better file extension computation?

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -643,6 +643,12 @@ static ALWAYS_INLINE id Kotlin_ObjCExport_refToObjCImpl(ObjHeader* obj) {
   RETURN_RESULT_OF(Kotlin_ObjCExport_refFromObjC, obj);
 }
 
+extern "C" OBJ_GETTER(Kotlin_Interop_CreateObjCObjectHolder, id obj) {
+  RuntimeAssert(obj != nullptr, "");
```
olonho: Maybe more meaningful message.

backend.native/tests/interop/objc/smoke.kt
```diff
@@ -213,6 +214,16 @@ fun createAndAbandonWeakRef(obj: NSObject) {
     WeakReference(obj)
 }
 
+fun testExceptions() {
```
olonho: Do we have test for foreign exception?

backend.native/tests/interop/objc/smoke.m
```diff
@@ -76,3 +76,9 @@ -(void)dealloc {
   unexpectedDeallocation = YES;
 }
 @end;
+
+@implementation BlockCaller
++(void)callBlock:(void (^)(void))block {
+    block();
```
olonho: What we gonna do for `staticCFunction`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -626,7 +676,7 @@ internal class InteropLoweringPart1(val context: Context) : IrBuildingTransforme
 
             return builder.at(expression).run {
                 val classPtr = getRawPtr(expression.extensionReceiver!!)
-                callAllocAndInit(classPtr, initMethodInfo, arguments)
+                callAllocAndInit(classPtr, initMethodInfo, arguments, expression, callee as IrSimpleFunction)
```
olonho: Cannot this cast fail?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
SvyatoslavScherbina: This is exactly what I've meant in the TODO below.

backend.native/tests/interop/objc/smoke.kt
SvyatoslavScherbina: Not yet.

backend.native/tests/interop/objc/smoke.m
SvyatoslavScherbina: Could you clarify? Both blocks and C callbacks are allowed to throw Kotlin exceptions.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
SvyatoslavScherbina: No, because this code is executed only when `callee` is factory method, which is always a (simple) function.

backend.native/tests/interop/objc/smoke.m
olonho: I mean, maybe test both?

backend.native/tests/interop/objc/smoke.kt
olonho: Could we try to throw C++ exception in the test?

SvyatoslavScherbina: Sure, but then it would crash the entire program.

PR #2847, opened 2019-04-03
Local delegated property was missing from IR serialization.

alexander-gorshenev:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/LegacyDescriptorUtils.kt
```diff
@@ -309,3 +316,42 @@ fun createAnnotation(
...
+class WrappedVariableDescriptorWithAccessors
+    : VariableDescriptorWithAccessors, WrappedCallableDescriptor<IrLocalDelegatedProperty>(Annotations.EMPTY, SourceElement.NO_SOURCE) {
+    override val getter: VariableAccessorDescriptor?
+        get() = error("Don't ask wrapped descriptor for so many details.")
+    override val setter: VariableAccessorDescriptor?
+        get() = error("Don't ask wrapped descriptor for so many details.")
+    override val isDelegated: Boolean
+        get() = error("Don't ask wrapped descriptor for so many details.")
+
+    override fun getContainingDeclaration() = (owner.parent as IrFunction).descriptor
```
SvyatoslavScherbina: Not sure if this is correct. For example, local delegated property can be declared in anonymous initializer.

SvyatoslavScherbina: On the other hand, it doesn't seem to be used somewhere, is it?

alexander-gorshenev: I am not aware of such places.

PR #2850, opened 2019-04-05
Kpgalligan/20190315/generics

kpgalligan:
This is a draft post to start the discussion around generics output in Objective-C headers. There are a number of limitations involved, but this is also a much discussed feature, so we've made an attempt to implement them and see if the development community finds this useful. A longer explanation of the design and compromise decisions can be found here: https://www.kgalligan.com/kotlin-native-interop-generics/


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/CustomTypeMapper.kt
```diff
@@ -16,7 +16,7 @@ import org.jetbrains.kotlin.types.TypeUtils
 
 internal interface CustomTypeMapper {
     val mappedClassId: ClassId
-    fun mapType(mappedSuperType: KotlinType, translator: ObjCExportTranslatorImpl): ObjCNonNullReferenceType
+    fun mapType(mappedSuperType: KotlinType, translator: ObjCExportTranslatorImpl, typeParamProvider: TypeParamProvider): ObjCNonNullReferenceType
```
olonho: We do not shorted nouns in identifiers, `TypeParameterProvider` shall be used here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -48,6 +50,11 @@ internal class ObjCExportTranslatorImpl(
 
     private val kotlinAnyName = namer.kotlinAnyName
 
+    //Flag to enable generic output
```
olonho: // Comments shall be space-separated from double-slash.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -48,6 +50,11 @@ internal class ObjCExportTranslatorImpl(
...
+    private val writeGenerics:Boolean by lazy {
+        System.getProperty("framework.writeGenerics", "false").toBoolean()
```
olonho: Maybe instead use property in .def file or CLI argument? We usually do not use Java properties to configure tools behavior.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -58,10 +65,19 @@ internal class ObjCExportTranslatorImpl(
...
+        val referenceClassExists = generator?.referenceClassExists(descriptor)
+        return if(referenceClassExists != null && referenceClassExists){
+            translateName
+        } else {
+            generator?.requireClassOrInterface(descriptor)
 
-        return translateClassOrInterfaceName(descriptor).also {
-            generator?.referenceClass(it.objCName, descriptor)
+            return translateName.also {
+                val sb = StringBuilder(it.objCName)
```
olonho: `val builder = StringBuilder(it.objCName)`

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -58,10 +65,19 @@ internal class ObjCExportTranslatorImpl(
...
+                if(writeGenerics)
```
olonho: `if<space>(cond)`

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -599,7 +633,43 @@ internal class ObjCExportTranslatorImpl(
...
+
+    internal fun ClassDescriptor.genericParamDeclaration(paramProvider: TypeParamProvider, includeUpperBounds:Boolean = false): List<String> = declaredTypeParameters.map { typeParam ->
+        val param = "${typeParam.variance.objcDeclaration()}${typeParam.name.asString()}"
+
+        if(includeUpperBounds) {
+            val upperBounds = typeParam.upperBounds
+            val upperBoundTypes = upperBounds
+                    .filter { kt ->
+                        val classDescriptor = kt.getErasedTypeClass()
+                        !(classDescriptor == builtIns.any || classDescriptor.classId in mapper.hiddenTypes || classDescriptor.isInlined() || classDescriptor.defaultType.isObjCObjectType())
```
olonho: Not very readable condition.

backend.native/tests/framework/values_generics/values.kt
```diff
@@ -0,0 +1,344 @@
+/*
```
olonho: Maybe it makes sense to avoid copy-pasting `backend.native/tests/framework/values/values.kt` and instead do generic-specific tests.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
kpgalligan: I'm pulling this out. The upper bounds constraint on generics was one of the last things added and was disabled due to some complicated types. If reintroduced this logic will need to be resurrected in some form, but it's in commit history for reference.

backend.native/tests/framework/values_generics/values.kt
kpgalligan: Updated in latest

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
kpgalligan: Moved to CLI.

kpgalligan:
Updated with feedback. Much appreciated. Please let me know if you have more.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -973,3 +1018,50 @@ private fun ObjCExportNamer.ClassOrProtocolName.toNameAttributes(): List<String>
 
 private fun swiftNameAttribute(swiftName: String) = "swift_name(\"$swiftName\")"
 private fun objcRuntimeNameAttribute(name: String) = "objc_runtime_name(\"$name\")"
+
+interface TypeParameterProvider{
+    fun typeAvailable(kotlinType: KotlinType):Boolean
```
SvyatoslavScherbina: I would prefer this method to be more like `fun getName(typeParameter: TypeParameterDescriptor): String?`
And this interface could be named like `ObjCExportScope` etc.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -973,3 +1018,50 @@ private fun ObjCExportNamer.ClassOrProtocolName.toNameAttributes(): List<String>
...
+}
+
+internal class ClassParameterProvider(container:DeclarationDescriptor): TypeParameterProvider {
+    constructor(method: FunctionDescriptor) : this(method.containingDeclaration)
+    private val nameSet:Set<String>
+    init {
+        nameSet = mutableSetOf()
+        if(container is ClassDescriptor && !container.isInterface) {
+            container.declaredTypeParameters.forEach {
```
SvyatoslavScherbina: Does this work for inner class when outer class has its own type parameters?
If yes, does this work when there is a name clash between inner and outer type parameters?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -973,3 +1018,50 @@ private fun ObjCExportNamer.ClassOrProtocolName.toNameAttributes(): List<String>
...
+private fun Variance.objcDeclaration():String = when(this){
+    Variance.OUT_VARIANCE -> "__covariant "
+    Variance.IN_VARIANCE -> "__contravariant "
+    else -> ""
+}
+
+//To anybody reviewing this code, I don't know my way
+//around the Kotlin parsed type system too well. What
+//I'm looking for here is generic type name minus the '?'
+//I'm sure there's a much simpler way to get that
```
SvyatoslavScherbina: Type parameters are represented with `TypeParameterDescriptor` in Kotlin compiler frontend.
You can get one using `TypeUtils.getTypeParameterDescriptorOrNull(this)`.

backend.native/tests/framework/values_generics/values_generics.kt
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fun variContraType():GenVarIn<SomeData>{
+    val compileVariIn:GenVarIn<BaseData> = GenVarIn(SomeData(1890))
+    val compileVariInSD:GenVarIn<SomeData> = compileVariIn
+    return compileVariInSD
+}
+
+open class GenBase<T:Any>(val t:T)
+class GenEx<TT:Any, T:Any>(val myT:T, baseT:TT):GenBase<TT>(baseT)
```
SvyatoslavScherbina: Would the test pass if `TT` was named `T` instead?
I.e. `class GenEx<T:Any, S:Any>(val myT:S, baseT:T):GenBase<T>(baseT)`

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -45,16 +45,12 @@ internal class ObjCExportTranslatorImpl(
         val builtIns: KotlinBuiltIns,
         val mapper: ObjCExportMapper,
         val namer: ObjCExportNamer,
-        val warningCollector: ObjCExportWarningCollector
+        val warningCollector: ObjCExportWarningCollector,
+        val objcGenerics:Boolean
```
olonho: We usually space separate colon.

backend.native/tests/framework/values_generics/values_generics.kt
kpgalligan: I updated test with your example and it's OK.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
kpgalligan: Updated. Thanks.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
kpgalligan: It did not, but updated. The outer class type params aren't visible to the inner class, so they're all 'id' (Any? to swift), but they do work.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
kpgalligan: Updated

backend.native/tests/framework/values_generics/values_generics.kt
SvyatoslavScherbina: Btw, renaming `T` in `GenEx` to `TT` makes the resulting header invalid.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1019,49 +1021,36 @@ private fun ObjCExportNamer.ClassOrProtocolName.toNameAttributes(): List<String>
...
+        } else {
+            emptyList()
         }
     }
 
-    override fun typeAvailable(kotlinType: KotlinType): Boolean =
-            nameSet.contains(kotlinType.typeParameterName())
+    override fun getName(typeParameterDescriptor: TypeParameterDescriptor?): String? =
+            if(typeParameterDescriptor != null && types.contains(typeParameterDescriptor)){
+                typeParameterDescriptor.name.asString()
```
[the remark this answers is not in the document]
kpgalligan: Yeah, interesting. Xcode seems to be OK with `id`, which is a surprise, but it should be renamed. I want to think about that a bit to see if there are other problematic names.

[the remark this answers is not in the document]
SvyatoslavScherbina: > Xcode seems to be OK with id, which is a surprise

`id` is just a type name. It is not forbidden to reuse such names for anything else.
The problem here is that we are trying to refer `id` in `foo`, i.e. from the scope where it is overridden by type parameter.

> if there are other problematic names.

Sure!
Potentially type parameter name can override name of any type. E.g.
0) Some "built-in" types emitted (like `id`)
1) Objective-C types imported to Kotlin (e.g. `NSObject`) when exported to Objective-C header
2) Kotlin class names (not that serious, because Kotlin class names are prefixed)

backend.native/tests/framework/values_generics/values_generics.kt
kpgalligan: This is related to how the supertype is declared. I wasn't outputting the type arguments, which compiles but creates issues. I'm not sure the best way to get the supertype. I can get the super _class_, but without the type args. During compile, classes are mostly represented by `LazyClassDescriptor`, which has `computeSupertypes()`, which returns `Collection<KotlinType>`, and I can get what I need from there, but that's obviously not a good idea (it's not public). Before I go digging around more, is there an obvious way to get the equivalent with a `ClassDescriptor`?

SvyatoslavScherbina: Try `classDescriptor.typeConstructor.supertypes`.

kpgalligan: Thanks! I didn't see this reply till just now. Spent more time yesterday trying to navigate the parsing code. Assuming I can get this to work, should have an update today.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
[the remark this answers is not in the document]
kpgalligan: Went through a couple iterations on 2. I was hoping to delay writing names until the end as it looks like stubs delay rendering, but they can evaluate earlier, so that didn't work. Then I landed on the far easier method of checking type param names and objc class/protocol names, and appending underscores to either. So, in the following:

```kotlin
class GenClashNames<ValuesGenericsClashnameClass, ValuesGenericsClashnameProtocol, ValuesGenericsValues_genericsKt>() {
    fun foo(): Any = ClashnameClass("nnn")
}

data class ClashnameClass(val str: String)
interface ClashnameProtocol {
    val str: String
}
```

```objc
@interface ValuesGenericsGenClashNames<ValuesGenericsClashnameClass, 
```

The type param name is above, the class name written later is:

```objc
@interface ValuesGenericsClashnameClass_ : KotlinBase
```
This is rare, obviously, and swift names don't change. Let me know what you think. The alternatives seemed like a lot of work for little gain.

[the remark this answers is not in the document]
SvyatoslavScherbina: This is ok for now.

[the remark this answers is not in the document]
kpgalligan: After working with that for a while, it turns out that if you have a type parameter name that clashes with a class name, using the target class in the generic means the target class's name will eval before the generic, and nothing will clash. I've been trying to cause a test issue but have been unable to, so I rolled back some of what I said above.

The type parameter name will be mangled, but not the class name itself.

Added some tests to hopefully catch a failing issue if that code changes in the future.

```kotlin
class GenClashNames<ValuesGenericsClashnameClass, ValuesGenericsClashnameProtocol, ValuesGenericsClashnameParam, ValuesGenericsValues_genericsKt>() {
    fun foo() = ClashnameClass("nnn")

    fun bar(): ClashnameProtocol = object : ClashnameProtocol{
        override val str = "qqq"
    }

    fun baz(arg: ClashnameParam): Boolean {
        return arg.str == "meh"
    }
}

open class ClashnameClass(val str: String)
interface ClashnameProtocol {
    val str: String
}
data class ClashnameParam(val str: String)
```

produces

```objc
@interface ValuesGenericsGenClashNames<ValuesGenericsClashnameClass_, ValuesGenericsClashnameProtocol_, ValuesGenericsClashnameParam_, ValuesGenericsValues_genericsKt> : KotlinBase
- (instancetype)init __attribute__((swift_name("init()"))) __attribute__((objc_designated_initializer));
+ (instancetype)new __attribute__((availability(swift, unavailable, message="use object initializers instead")));
- (ValuesGenericsClashnameClass *)foo __attribute__((swift_name("foo()")));
- (id<ValuesGenericsClashnameProtocol>)bar __attribute__((swift_name("bar()")));
- (BOOL)bazArg:(ValuesGenericsClashnameParam *)arg __attribute__((swift_name("baz(arg:)")));
@end;

__attribute__((swift_name("ClashnameClass")))
@interface ValuesGenericsClashnameClass : KotlinBase
- (instancetype)initWithStr:(NSString *)str __attribute__((swift_name("init(str:)"))) __attribute__((objc_designated_initializer));
@property (readonly) NSString *str __attribute__((swift_name("str")));
@end;

__attribute__((swift_name("ClashnameProtocol")))
@protocol ValuesGenericsClashnameProtocol
@required
@property (readonly) NSString *str __attribute__((swift_name("str")));
@end;

__attribute__((objc_subclassing_restricted))
__attribute__((swift_name("ClashnameParam")))
@interface ValuesGenericsClashnameParam : KotlinBase
- (instancetype)initWithStr:(NSString *)str __attribute__((swift_name("init(str:)"))) __attribute__((objc_designated_initializer));
- (BOOL)isEqual:(id _Nullable)other __attribute__((swift_name("isEqual(:)")));
- (NSUInteger)hash __attribute__((swift_name("hash()")));
- (NSString *)description __attribute__((swift_name("description()")));
- (NSString *)component1 __attribute__((swift_name("component1()")));
- (ValuesGenericsClashnameParam *)doCopyStr:(NSString *)str __attribute__((swift_name("doCopy(str:)")));
@property (readonly) NSString *str __attribute__((swift_name("str")));
@end;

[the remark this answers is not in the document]
Comment (SvyatoslavScherbina) [reply backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt]: How have you implemented this type parameter name mangling logic? Which part of code is it located at, and which data does it use?

> using the target class in the generic means the target class's name will eval before the generic

Could you clarify?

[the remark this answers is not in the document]
Comment (kpgalligan) [reply backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt]: Still need to push. Wanted a more complete implementation before pushing and have been distracted with other concurrent work. Will push now, but am still poking around some.

I have implemented this. It's in `ObjCExportNamerImpl`. It does not extend from `Mapping`, as the concept of `nameToElements` didn't make a lot of sense, but that means there's some code duplication. To refactor possibly.

To clarify the statement, if the type param name is `ValuesGenericsClashnameClass`, that will clash with `ClashnameClass`. That means we want to rename the param to `ValuesGenericsClashnameClass_`, but that's only a problem if we're using that target class, `ClashnameClass`, in the generic parent class (`GenClashNames` in the example above). If we encountered the type param name first, then the class itself would need a name mangling, but because the generator code evaluates properties, methods, and parent classes first, we always resolve the target class's name before we need to assign that type param, so the target class's name itself never needs mangling because of the type param name clashing.

Summary, the type param may be mangled because is clashes with a class name, but not the other way around.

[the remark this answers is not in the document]
Comment (kpgalligan) [reply backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt]: Also, to clarify, "Wanted a more complete implementation". I'm just cleaning up tests and had planned to push last night, but wanted better test coverage and also want a full review that with the experimental flag off that nothing is changed, but the implementation is "complete" assuming I don't find anything with tests and review.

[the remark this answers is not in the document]
Comment (SvyatoslavScherbina) [reply backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt]: Ok, thank you.

> the generator code evaluates properties, methods, and parent classes first

Yes, this is the detail I missed.

> Summary, the type param may be mangled because is clashes with a class name, but not the other way around.

That's great!

File: backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt [hunk 14]
```diff
@@ -129,7 +150,30 @@ internal class ObjCExportTranslatorImpl(
...
+        fun superClassGenerics(genericExportScope: ObjCExportScope): List<ObjCNonNullReferenceType> {
+            val parentType = computeSuperClassType(descriptor)
+            return if(parentType != null) {
+
+                parentType.arguments.map { typeProjection ->
+                    val typeParamClass = typeProjection.type.getErasedTypeClass()
+
+                    if (!typeProjection.type.isTypeParameter() && mapper.shouldBeExposed(typeParamClass)) {
+                        if (typeParamClass.isInterface)
+                            referenceProtocol(typeParamClass)
```
SvyatoslavScherbina: What is this required for? How is it different from what `mapReferenceTypeIgnoringNullability` does?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -316,8 +378,8 @@ internal class ObjCExportTranslatorImpl(
     }
 
     private fun StubBuilder.translatePlainMembers(methods: List<FunctionDescriptor>, properties: List<PropertyDescriptor>) {
-        methods.forEach { +buildMethod(it, it) }
-        properties.forEach { +buildProperty(it, it) }
+        methods.forEach { +buildMethod(it, it, genericExportScope(it.containingDeclaration)) }
```
SvyatoslavScherbina: It would be more consistent to receive `ObjCExportScope` as an argument here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -909,7 +987,19 @@ abstract class ObjCExportHeaderGenerator internal constructor(
     }
 
     internal fun generateClass(descriptor: ClassDescriptor) {
+        //If type args reference this class in parent declaration, we have a special case forward class declaration
+        //needed. If upper bounds are added back to type parameters, this may need to be expanded.
+        fun selfReferencingClassType(descriptor: ClassDescriptor): Boolean {
+            val parentType = computeSuperClassType(descriptor)
+            return parentType != null && parentType.arguments.any { descriptor == it.type.getErasedTypeClass() }
```
SvyatoslavScherbina: It is not the only case.
Please try disabling this optimization instead:
https://github.com/JetBrains/kotlin-native/blob/3a0be1c3319a44fcf8a9d5764fb581a3e268ba5a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt#L1018
(i.e. always add forward declaration).
I find it a better option. I would enable it back after some refactoring if required.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -55,13 +61,28 @@ internal class ObjCExportTranslatorImpl(
...
+        }
+    }
+
     private fun referenceClass(descriptor: ClassDescriptor): ObjCExportNamer.ClassOrProtocolName {
         assert(mapper.shouldBeExposed(descriptor))
         assert(!descriptor.isInterface)
-        generator?.requireClassOrInterface(descriptor)
+        val translateName = translateClassOrInterfaceName(descriptor)
+        val referenceClassExists = generator?.referenceClassExists(descriptor)
+        return if(referenceClassExists != null && referenceClassExists){
```
japplin: missing spaces before and after ()

kpgalligan:
Extensions could in theory also have generics defined, but I hadn't put much thought into that and how they're translated currently. Those now explicitly ignore generics and can potentially be refactored. The forward declaration now writes all classes to avoid the self-reference issue, but only if generics flag enabled. In a code base of any reasonable size that forward declaration will get long, but I'm not sure that'll have much impact. My experience with j2objc says you can give the compiler some pretty huge Objective-C without much trouble, but refactoring that later would be good.

SvyatoslavScherbina:
> Extensions could in theory also have generics defined, but I hadn't put much thought into that and how they're translated currently.

It's ok. I'm not sure if proper full support is possible here, e.g.
```
class Foo<T : Bound>
fun <T : MoreSpecificBound> Foo<T>.bar()
``` 

doesn't seem to be representable in Objective-C.

> In a code base of any reasonable size that forward declaration will get long, but I'm not sure that'll have much impact.

I don't think it will have any impact.
Having forward declarations in one line shouldn't affect the compiler. And the size is limited by the amount of classes.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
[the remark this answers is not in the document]
SvyatoslavScherbina: > the generator code evaluates properties, methods, and parent classes first

Btw, what if type parameter is used in properties or methods?

```kotlin
class Foo<ValuesGenericsBar> {
    fun foo(p: ValuesGenericsBar) {}
    fun bar(p: Bar) {}
}

class Bar
```

->

```obj-c
__attribute__((objc_subclassing_restricted))
__attribute__((swift_name("Foo")))
@interface ValuesGenericsFoo<ValuesGenericsBar> : KotlinBase
// ...
- (void)fooP:(ValuesGenericsBar _Nullable)p __attribute__((swift_name("foo(p:)")));
- (void)barP:(ValuesGenericsBar *)p __attribute__((swift_name("bar(p:)")));
@end;
```

I'm still not sure if it makes much sense to fix this now though.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
```diff
@@ -424,13 +436,71 @@ internal class ObjCExportNamerImpl(
...
+
+        private fun validName(element: TypeParameterDescriptor, name: String): Boolean {
+            assert(element.containingDeclaration is ClassDescriptor)
+
+            val nameSet = classNameSet(element)
+            return !objCClassNames.nameExists(name) && !objCProtocolNames.nameExists(name) && name !in nameSet
+        }
+
+        private fun classNameSet(element: TypeParameterDescriptor): MutableSet<String> {
+            return typeParameterNameClassOverrides.getOrPut(element.containingDeclaration as ClassDescriptor) {
```
SvyatoslavScherbina: Doesn't seem correct for (yet unsupported) outer class type parameters.
```
class Foo<T> {
    inner class Bar<T>
}
```

backend.native/tests/framework/values_generics/values.swift
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func testGenericInterface() throws {
+    let a: NoGeneric = SomeGeneric<SomeData>(t: SomeData(num: 52))
+    try assertEquals(actual: (a.myVal() as! SomeData).num, expected: 52)
+}
+
+func testGenericInheritance() throws {
+    let ge = GenEx<SomeData, SomeOtherData>(myT:SomeOtherData(str:"Hello"), baseT:SomeData(num: 11))
+    try assertTrue(ge.t is SomeData, "base property not SomeData")
```
SvyatoslavScherbina: AFAIK, this checks only the dynamic type, not the static one.
So this assert doesn't ensure that `ge.t` is statically known to be `SomeData` and can be assigned to a a `SomeData`-typed variable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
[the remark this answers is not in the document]
kpgalligan: Well, I'm not happy that I missed that. My testing got caught up in the outer class definition. Anyway, the RenderedStub will render the name as `buildAsDeclaredOrInheritedMethods` creates a Set, which hash/equals results in rendering the underlying stub. To avoid this, we'd either need run a pass through the classes properties and methods before calling `collectMethodsOrProperties` (or some similar first pass, perhaps higher up), or mangle the class name.

It is obviously unlikely that names would clash, and if we're expecting significant refactoring before generics would be standard, if ever, then fixing later is OK, as long as it's addressed at some point.

Let me know if you want me to deal with it now. I'd be inclined to pass through class members in `StubBuilder.translateClassMembers` to flush out name issues.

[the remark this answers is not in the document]
SvyatoslavScherbina: I find it ok to stop at this point.
We have some plans to fully rework clash handling in generated header, and I believe that we'll handle then these clashes between type parameters and classes.

backend.native/tests/framework/values_generics/values.swift
kpgalligan: Yeah, that check doesn't really do what I want. I've replaced how that works (in next push) for all type checks. Rather than check type, if the type isn't correct Swift needs a conversion, so the fact that it compiles and runs should imply that the types are as expected.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
kpgalligan: Generics are definitely not supporting outer generics on nested or inner types at this point. Although the resulting syntax isn't great, it's possible to cast your way around it. Can be added, though. Will defer to you. Will add if you think this would be valuable.

backend.native/tests/framework/values_generics/values.swift
SvyatoslavScherbina: What do you mean?

How does [the test](https://github.com/JetBrains/kotlin-native/blob/3f38a28068cd5951ab0c21e18181d9d1d2ebc2d1/backend.native/tests/framework/values_generics/values.swift#L150) now ensure that `ge.t` is declared as `SomeData`?

kpgalligan: That line is old. The [last push](https://github.com/kpgalligan/kotlin-native/blob/3f38a28068cd5951ab0c21e18181d9d1d2ebc2d1/backend.native/tests/framework/values_generics/values.swift#L150) was referencing the value instead of "is SomeData", which should fail to compile if not `SomeData`. 

```
try assertEquals(actual: ge.t.num, expected: 11)
```

However, I have some more testing to figure out as the properties and methods of the base class being written in the child class means disabling the superclass generics didn't cause that test to fail.

```
__attribute__((objc_subclassing_restricted))
__attribute__((swift_name("GenEx")))
@interface MainGenEx<TT, T> : MainGenBase
- (instancetype)initWithMyT:(T)myT baseT:(TT)baseT __attribute__((swift_name("init(myT:baseT:)"))) __attribute__((objc_designated_initializer));
- (instancetype)initWithT:(id)t __attribute__((swift_name("init(t:)"))) __attribute__((objc_designated_initializer)) __attribute__((unavailable));
- (TT)goBase __attribute__((swift_name("goBase()")));
@property (readonly) T myT __attribute__((swift_name("myT")));
@property (readonly) TT t __attribute__((swift_name("t")));
@end;
```

Property 't' is still SomeData even though we don't tell MainGenBase what `T` is because `t` is also written to `MainGenEx`. If we specify `T` as `Any`, Swift will report true with `is SomeData` but requires a cast to `SomeData`.

In [this example](https://github.com/kpgalligan/kotlin-native/blob/3f38a28068cd5951ab0c21e18181d9d1d2ebc2d1/backend.native/tests/framework/values_generics/values.swift#L157), GenExAny is defined as:

```
class GenExAny<TT:Any, T:Any>(val myT:T, baseT:TT):GenBase<Any>(baseT)
```

That forces the cast to `SomeData`. If GenEx was also writing `t` as `Any/id`, it should fail to compile.

However, I went in and disabled the generic superclass write to see what would fail and while one thing did fail, it was less than I expected. Will be diving into that more and update with better testing.

SvyatoslavScherbina: `ge.t.num` doesn't fail because Swift seems to allow accessing any method or property on `id`.
That's why the proper way to test the type is to store the value into explicitly typed variable:
```
let geT: SomeData = ge.t
```

kpgalligan: Tests updated to specific types and assigned as above. I wasn't understanding why the parent class property was being written out, but I see that now. It [filters by string](https://github.com/JetBrains/kotlin-native/blob/ea11c377a61ea968c199fa6100a1bcbf7057f000/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt#L392).

```objc
__attribute__((swift_name("GenEx")))
@interface MainGenEx<TT, T> : MainGenBase
- (instancetype)initWithMyT:(T)myT baseT:(TT)baseT __attribute__((swift_name("init(myT:baseT:)"))) __attribute__((objc_designated_initializer));
- (instancetype)initWithT:(id)t __attribute__((swift_name("init(t:)"))) __attribute__((objc_designated_initializer)) __attribute__((unavailable));
- (TT)goBase __attribute__((swift_name("goBase()")));
@property (readonly) T myT __attribute__((swift_name("myT")));
@property (readonly) TT t __attribute__((swift_name("t")));
@end;
```

I'm not sure why redefining the property exists in current code, so I'm a little reluctant to try to be smarter about filtering for generics. The possible parameter name differences make redefining more likely, but it should still work as expected.

I added an example that won't redefine the property to the test just to make sure that works.

```kotlin
open class GenBase<T:Any>(val t:T)
class GenEx<TT:Any, T:Any>(val myT:T, baseT:TT):GenBase<TT>(baseT)
class GenEx2<T:Any, S:Any>(val myT:S, baseT:T):GenBase<T>(baseT)
```

Can write a smarter filter, but want to be clear on the cases it exists for.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -181,6 +181,9 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-Xcoverage-file", valueDescription = "<path>", description = "Save coverage information to the given file")
     var coverageFile: String? = null
 
+    @Argument(value = "-Xobjc-generics", description = "Write objc header with generics support")
```
SvyatoslavScherbina: Could description be more like `"Enable experimental generics support for framework header"`?
To be consistent with other options and clearly stating experimental status.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -599,7 +660,14 @@ internal class ObjCExportTranslatorImpl(
         return if (classDescriptor.isInterface) {
             ObjCProtocolType(referenceProtocol(classDescriptor).objCName)
         } else {
-            ObjCClassType(referenceClass(classDescriptor).objCName)
+            val typeArgs = if (objcGenerics) {
+                kotlinType.arguments.map { typeProjection ->
```
SvyatoslavScherbina: Btw, this doesn't match `ObjCClassExportScope`:
```kotlin
class Outer<T> {
    inner class Inner<S>
}

fun foo(obj: Outer<String>.Inner<Any>) {}
``` 

->

```objc
@interface HelloOuterInner<S> : KotlinBase
```
but
```objc
+ (void)fooObj:(HelloOuterInner<id, NSString *> *)obj __attribute__((swift_name("foo(obj:)")));
```

Supporting inner classes properly in `ObjCClassExportScope` would fix this.

backend.native/tests/framework/values_generics/values.swift
SvyatoslavScherbina: > It filters by string.

This machinery has to be revised later. Let's keep current simple filter for now.

SvyatoslavScherbina:
I believe we can merge this PR after getting serious issues fixed, ignoring remaining ones for now, to make this feature included into 1.3.40.

Issues to be fixed before merge:
https://github.com/JetBrains/kotlin-native/pull/2850#discussion_r281134028
https://github.com/JetBrains/kotlin-native/pull/2850#discussion_r281141583

cc @olonho

kpgalligan:
There is an issue that comes up with generics, Swift, and inner classes. The header defines the Swift class name with a '.' between the outer and inner classes, which can cause a crash. For example, from the tests, this is OK

```swift
let innerClass = GenOuter.GenInner<SomeData, SomeOtherData>(GenOuter<SomeOtherData>(a: SomeOtherData(str: "ggg")), c: SomeData(num: 66), aInner: SomeOtherData(str: "ttt"))
```

However, when you go to use that in Swift, it can actually crash the Swift compiler with a seg fault. Uncomment [these lines](https://github.com/JetBrains/kotlin-native/blob/897d5c03245fcb2ab35fedc74ba584408c91a100/backend.native/tests/framework/values_generics/values.swift#L196) in the test to see this.

I've done some local header testing, and manually removing the '.' from the Swift class name allows the Swift to compile and run. Turning off generics and keeping the '.' will also compile and run (if you remove the generics from the Swift code as well, obviously).

In summary, if '.' is a valid character in a class name, this looks like a Swift compiler bug. I think we have the following options: 

1. Leave it as is, although this is a frustrating bug if you happen to hit it
2. Not include the '.' in class names
3. Not include the '.' in class names, but only if generics are enabled (to avoid breaking current code)

SvyatoslavScherbina:
Generally '.' is a valid character in class name. See [Import as member (SE-0044)](https://github.com/apple/swift-evolution/blob/master/proposals/0044-import-as-member.md).
The document doesn't state that this feature works for classes too, however in fact it does, and this is used by system frameworks.

So this crash is likely a bug in Swift compiler.
Reproducer:
`1.h`:
```objc
#import <Foundation/Foundation.h>
@interface A : NSObject
@end;

NS_SWIFT_NAME(A.B)
@interface AB<T> : NSObject
@end;

void takeAB(AB<id>* ab);
```

`1.swift`:
```swift
let ab = A.B<AnyObject>()
takeAB(ab)
```

```bash
swiftc 1.swift -import-objc-header 1.h
```

Also it's not clear how importing as member should work if outer class has generics, because Swift itself seems to require specifying type arguments even when accessing static members (like nested classes):

```swift
class A<T> {
    class B<S> {}
}

let ab = A<Any>.B<Any>()
```

So let's avoid using '.' in class names if `-Xobjc-generics` is enabled and either inner or outer class has generics.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportNamer.kt
```diff
@@ -191,38 +192,51 @@ internal class ObjCExportNamerImpl(
...
-            val containingDeclaration = descriptor.containingDeclaration
-            if (containingDeclaration is ClassDescriptor) {
-                append(getClassOrProtocolSwiftName(containingDeclaration))
-
-                val importAsMember = when {
-                    descriptor.isInterface || containingDeclaration.isInterface -> {
-                        // Swift doesn't support neither nested nor outer protocols.
-                        false
-                    }
+            generateClassOrProtocolSwiftName(descriptor)
```
SvyatoslavScherbina: The original implementation was supposed to use mangled outer class name, but the new one mangles the concatenated name instead, which is not desirable.

kpgalligan:
Updated Swift generics naming issue fix. For a class like the following

```kotlin
class GenOuterDeep2() {
    inner class Before()
    inner class GenShallowOuterInner() {
        inner class GenShallowInner<T>() {
            inner class GenDeepInner()
        }
    }
    inner class After()
}```

The Swift names will be as follows

```swift
GenOuterDeep2()
GenOuterDeep2.Before(rec)
GenOuterDeep2.After(rec)
GenOuterDeep2GenShallowOuterInner(rec)
//etc
```

Dropping the '.' only happens if a deeper class has a defined generic, but otherwise naming remains as it was.

kpgalligan:
One more coming. Still an issue with nested.

PR #2852, opened 2019-04-08
ObjectiveCInterop benchmarks

LepilkinaElena:


performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.h
```diff
@@ -0,0 +1,27 @@
+#import <Foundation/Foundation.h>
+
+@protocol CustomNumber
+@required
+- (id<CustomNumber>)add: (id<CustomNumber>)other;
+- (id<CustomNumber>)sub: (id<CustomNumber>)other;
+@end
+
+@interface Complex : NSObject<CustomNumber> {
+    double re;
```
SvyatoslavScherbina: ivars are typically declared in `@implementation`.

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.m
```diff
@@ -0,0 +1,61 @@
+#import <stdio.h>
+#import "complexNumbers.h"
+
+@interface Complex ()
+- (void)zeroComplex;
```
SvyatoslavScherbina: Is this method used?

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.h
```diff
@@ -0,0 +1,27 @@
...
+    double im;
+    NSString *format;
+}
+
+@property (nonatomic, readonly) double re;
+@property (nonatomic, readonly) double im;
+@property (retain, nonatomic, readwrite) NSString *format;
+
+- (id)initWithRe: (double)re andIm: (double)im;
++ (Complex *)complexWithRe: (double)re andIm: (double)im;
```
SvyatoslavScherbina: AFAIK, name convention doesn't suppose using `and` here.

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.m
```diff
@@ -0,0 +1,61 @@
...
+    if (self = [super init]) {
+        re = _re;
+        im = _im;
+        format = @"re: %.1lf im: %.1lf";
+    }
+    return self;
+}
+
++ (Complex *)complexWithRe: (double)re andIm: (double)im {
+    return [[Complex alloc] initWithRe: re andIm: im];
```
SvyatoslavScherbina: Do you compile this file with ARC enabled?

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
```
SvyatoslavScherbina: Somewhat obsolete header. The new one is also much shorter.

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
```diff
@@ -0,0 +1,145 @@
...
+    actual fun generateNumbersSequence(): List<Any> {
+        val result = mutableListOf<Complex>()
+        for (i in 1..benchmarkSize) {
+            result.add(Complex(randomNumber(), randomNumber()))
+        }
+        return result
+    }
+
+    actual fun sumComplex() {
+        complexNumbersSequence.map { (it as Complex).add(it) }.reduce { acc, it -> (acc as Complex).add(it) }
```
SvyatoslavScherbina: Is cast required here?

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.m
```diff
@@ -0,0 +1,61 @@
...
+}
+- (Complex *)add: (Complex *)other {
```
SvyatoslavScherbina: Super method has different signature.

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.m
LepilkinaElena: Yes, as I understood, this is modern way of writting on ObjC and Xcode has such settings by default. Could it work with Kotlin properly? In ObjC I used `@autoreleasepool` in main to test.

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
LepilkinaElena: Yes, because I had to use Any type if I want to use this method also as benchmark.

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.m
SvyatoslavScherbina: > Yes

Then `retain` here shouldn't be required:
https://github.com/JetBrains/kotlin-native/pull/2852/files#diff-c4aac30e2bfe14c88545c761df9a7809R13

> Could it work with Kotlin properly?

Yes, it works with Kotlin properly.

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
SvyatoslavScherbina: Well, `Sequence<Complex>` is `Sequence<Any>` too, so you can use trivial wrapper as a benchmark.

performance/objcinterop/src/nativeInterop/cinterop/complexNumbers.h
```diff
@@ -0,0 +1,23 @@
...
+
+@interface Complex : NSObject<CustomNumber>
+
+@property (nonatomic, readonly) double re;
+@property (nonatomic, readonly) double im;
+@property (retain, nonatomic, readwrite) NSString *format;
+
+- (id)initWithRe: (double)re andIm: (double)im;
++ (Complex *)complexWithRe: (double)re im: (double)im;
+- (void)setFormat: (NSString *)newFormat;
```
SvyatoslavScherbina: This method is not required, it is generated automatically for `readwrite` property.

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
LepilkinaElena: I was suggested to use better solution `actual typealias`.

performance/objcinterop/src/main/kotlin-native/org/jetbrains/objCinteropBenchmarks/complexNumbers.kt
```diff
@@ -0,0 +1,136 @@
...
+    actual fun generateNumbersSequence(): List<Complex> {
+        val result = mutableListOf<Complex>()
+        for (i in 1..benchmarkSize) {
+            result.add(Complex(randomNumber(), randomNumber()))
+        }
+        return result
+    }
+
+    actual fun sumComplex() {
+        complexNumbersSequence.map { it.add(it) }.reduce { acc, it -> acc?.add(it) }
```
SvyatoslavScherbina: Using `?.` here may potentially hide an issue, and the benchmark then wouldn't measure Objective-C interop.
Adding nullability annotations to Objective-C header may help to avoid the necessity to handle `null` here.

PR #2855, opened 2019-04-09
Set declaration parent in SharedVariableManager.

sfs:
This is part of https://github.com/JetBrains/kotlin/pull/2242

As part of that PR I changed the place where the parent field is set in SharedVariablesManager. The parent has to be set based on the original declaration, because of possibly nested declarations. This is the same fix for kotlin-native.


max-kammerer:
Seems it could be merged independently of JetBrains/kotlin#2242

alexander-gorshenev:
why is it not just
```
- newDeclaration.parent = irFunction
+ newDeclaration.parent = declaration.parent
```

in SharedVariablesLowering.kt?
That way you don't need to triplicate it all over the backends.
And .parent assignment is nearby the insertion to the scope.

sfs:
There's no technical reason, I just didn't want declareSharedVariable to generate incomplete IR and wasn't aware that this is also the behavior in kotlin-native. I can change it in SharedVariablesLowering instead, if you prefer.

alexander-gorshenev:
>  I can change it in SharedVariablesLowering instead, if you prefer.
yes, please.

olonho:
So closing this one?

PR #2860, opened 2019-04-12
Eliminate most of incorrect SymbolTable usages

SvyatoslavScherbina:
Also do some cleanup.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -1012,9 +1050,9 @@ internal class CAdapterGenerator(
     private var functionIndex = 0
     fun nextFunctionIndex() = functionIndex++
 
-    internal val kGetTypeFuncType =
+    internal val kGetTypeFuncType get() =
```
homuroll: `lazy`?

SvyatoslavScherbina: Somewhat conflicts the idea of having `codegen` for a limited amount of time only.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -203,6 +203,8 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
 
     lateinit var objCExport: ObjCExport
 
+    lateinit var cAdapterGenerator: CAdapterGenerator
```
olonho: Hmm, this guy doesn't seem to belong to context.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -160,29 +162,26 @@ internal class ObjCExportCodeGenerator(
...
+                    val superClass = it.superClassNotAny ?: objCClassForAny
+
+                    dataGenerator.emitEmptyClass(it.binaryName, superClass.binaryName)
+                    // Note: it is generated only to be visible for linker.
+                    // Methods will be added at runtime.
+
+                    createTypeAdapter(it, superClass)
+                }
+
+                is ObjCProtocolForKotlinInterface -> createTypeAdapter(it, superClass = null)
```
olonho: Shalln't we have catching `else` here?

SvyatoslavScherbina: This `when` is used as an expression and thus forced to be exhaustive.

PR #2861, opened 2019-04-12
Update versions in samples and docker file

ilmat192:


samples/androidNativeActivity/gradle.properties
```diff
@@ -5,7 +5,7 @@ org.gradle.parallel=true
 org.gradle.workers.max=4
 
 # Pin Kotlin version:
-kotlin_version=1.3.30-eap-113
+kotlin_version=1.3.30
```
olonho: Maybe we shall add easily discoverable marker, like CHANGE_KOTLIN_VERSION_HERE in comment, so that it's easy not to miss all such cases?

ilmat192: I've added the `CHANGE_VERSION_WITH_RELEASE` marker and mentioned it in the release doc.

samples/gradle.properties
```diff
@@ -4,7 +4,8 @@ kotlin.code.style=official
 org.gradle.parallel=true
 
 # Pin Kotlin version:
-kotlin_version=1.3.30-eap-113
+# CHANGE_VERSION_WITH_RELEASE
+kotlin_version=1.3.30
```
olonho: Conflicts with https://github.com/JetBrains/kotlin-native/pull/2879

samples/androidNativeActivity/gradle.properties
ddolovov: Marker is better than nothing. :+1:

But still it requires editing multiple files. I think we could elaborate a better approach when only one file needs to be edited. Added this to https://jetbrains.quip.com/RO9VA7uCbqjF (point #4)

samples/gradle.properties
ddolovov: We need to use Gradle plugin version `1.3.40-dev-1480`. It has the necessary fixes with arguments, etc.

PR #2866, opened 2019-04-14
Support escaped arguments in argfiles.

mpetrov:
This is already fixed in `K2JVMCompiler` (KT-26122).

Without this change `K2Native` would read the argfile line
`'-Xdump-perf=/tmp/perf'` as a file name rather than a `-Xdump-prefs` flag.


mpetrov:
@kevin1e100 as FYI

PR #2876, opened 2019-04-16
Update compiler to 1.3.40-dev-1842

PavelPunegov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -209,6 +209,8 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
 
     override val configuration get() = config.configuration
 
+    override val internalPackageFqn: FqName = FqName.fromSegments(listOf("kotlin", "native", "internal"))
```
olonho: Why not in https://github.com/JetBrains/kotlin-native/blob/master/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/RuntimeNames.kt?

PR #2890, opened 2019-04-17
Support building tetris sample for MinGW32

ddolovov:


samples/tetris/build.gradle
```diff
@@ -104,12 +112,27 @@ afterEvaluate {
...
 // If host platform is Linux and RaspberryPi target is activated.
 private boolean isRaspberryPiBuild() {
     MPPTools.isLinux() && Boolean.parseBoolean(project.findProperty('tetris.raspberrypi.build') as String)
-}
\ No newline at end of file
+}
+
+private boolean isMingwX86Build() {
+    MPPTools.isWindows() && (
+            System.getenv("ProgramFiles(x86)") == null
```
msink: Why? Host must be 64-bit, so here cannot be null.

samples/tetris/build.gradle
```diff
@@ -60,10 +62,16 @@ kotlin {
     }
 }
 
-if (hostPreset == kotlin.presets.mingwX64) {
+if ((hostPreset == kotlin.presets.mingwX86) || (hostPreset == kotlin.presets.mingwX64)) {
     tasks.create('compileWindowsResources', Exec) {
+        def windresDir = null
+        if (hostPreset == kotlin.presets.mingwX86)
+            windresDir = "${MPPTools.kotlinNativeDataPath()}/dependencies/msys2-mingw-w64-i686-gcc-7.4.0-clang-llvm-6.0.1/bin"
```
olonho: Ugly :(. Maybe at least smth like `fun targetToolchainPath(hostPreset)`.

samples/tetris/build.gradle
ddolovov: 32-bit host does not have `ProgramFiles(x86)` environment variable. That's why 32-bit tetris will be built under 32-bit host. What's not correct here?

samples/tetris/build.gradle
ddolovov: Another _bad_ function for _MPPTools_?

ddolovov: I would suggest to add this to K/N Samples Wishlist but leave it as is for now.

samples/tetris/build.gradle
olonho: 32-bit Windows hosts cannot run Kotlin/Native compiler.

msink: Kotlin/Native does not support 32-bit hosts.

ddolovov: OK

PR #2894, opened 2019-04-17
Add option to use regexes for filtering and add property for passing compiler args when running benchmarks.

punzki:


LepilkinaElena:
It's good idea to describe subset with regexes, but I think it'd be better to add new flag --filter-regex for this. --filter flag is useful for use cases when you debug one benchmark. As far as we have benchmarks with dot in names it will be uncomfortable to add extra backslashes to escape wildcard characters of regex (It's not critical with dot, but there could be other problems in future)

punzki:
> It's good idea to describe subset with regexes, but I think it'd be better to add new flag --filter-regex for this. --filter flag is useful for use cases when you debug one benchmark. As far as we have benchmarks with dot in names it will be uncomfortable to add extra backslashes to escape wildcard characters of regex (It's not critical with dot, but there could be other problems in future)

I was thinking that this would not break any behavior, since a full regex match is required, therefore to run a single test, you still have to enter the full name. Also as you pointed out, we don't need to escape backslashes. Thoughts?

Also, are there TeamCity builds or any other continuous builds that run these benchmarks?

LepilkinaElena:
> Also as you pointed out, we don't need to escape dots with backslashes. 

It's true for now, but people should understand that this regex can match not only with dot. We have no limitations for benchmarks names, so it would better on my opinion separate flags to be sure that people understand what they want. If users don't need regexes they shouldn't think about their syntax and wildcards. It's more for future. @olonho what do you think?

> Also, are there TeamCity builds or any other continuous builds that run these benchmarks?

Yes, we have both TeamCity configurations and inner server for saving and showing results of benchmarks, but they are in inner network of JetBrains.

punzki:
@LepilkinaElena Not a problem, I can just create a separate flag for regex. Users can use either (or both).

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
```diff
@@ -15,14 +15,17 @@ import java.io.File
...
 
     var buildType = "RELEASE"
     var workingDir: Any = project.projectDir
     var outputFileName: String? = null
     @Input
-    @Option(option = "filter", description = "filter")
+    @Option(option = "filter", description = "Benchmarks to run (comma-separated)")
     var filter: String = ""
+    @Input
+    @Option(option = "filterRegex", description = "Benchmarks to run, as regexes (comma-separated)")
```
LepilkinaElena: May be "Benchmarks to run  described by regular expressions" is better?

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
```diff
@@ -45,12 +48,17 @@ open class RunKotlinNativeTask @Inject constructor(
         this.dependsOn += taskName
     }
 
+    private fun splitCommaSeparatedOption(arg: String, optionName: String) =
```
LepilkinaElena: As far as this function is exactly repeated may be better put it to Utils file. It can be even done as extension function for String.

performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -25,13 +25,18 @@ abstract class Launcher(val numWarmIterations: Int, val numberOfAttempts: Int, v
...
-        runningBenchmarks.forEach {
-            val benchmark = benchmarks[it]
-            benchmark ?: error("Benchmark $it wasn't found!")
+    fun launch(filters: Collection<String>? = null,
+               filterRegexes: Collection<String>? = null): List<BenchmarkResult> {
+        val regexes = filterRegexes?.map { it.toRegex() } ?: listOf()
+        val filterSet = filters?.toHashSet() ?: hashSetOf()
+        // Filter benchmarks using given filters, or run all benchmarks if none were given.
+        val runningBenchmarks = if (filterSet.isNotEmpty() || regexes.isNotEmpty()) {
+            benchmarks.filterKeys { bm -> bm in filterSet || regexes.any { r -> r.matches(bm) } }
```
LepilkinaElena: As far as I know we try escape such short forms like `bm`, `r`.

punzki: Spelled out "benchmark" and removed "r". In this case I need to give at least one of them a name. I can't use the default "it" name for both since both are used in the innermost lambda.

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
punzki: Moved to Utils file as extension function.

performance/buildSrc/src/main/kotlin/RunKotlinNativeTask.kt
punzki: Done.

LepilkinaElena:
Thank you!

PR #2896, opened 2019-04-18
 Support variadic Objective-C methods

### LINKED ISSUE #1834: Can't use NSString.initWithFormat
Hello!

I try to create formatted string in iOS platform module.
In Swift code should be:
```
NSString(format: "test %s", "test 2")
```
but in Kotlin/Native i can't use same init method. 
```
public convenience init(format: NSString, _ args: CVarArg...)
``` 
was converted to 
```
@kotlinx.cinterop.ObjCMethod public external fun platform.Foundation.NSString.initWithFormat(format: kotlin.String, arguments: platform.posix.va_list? /* = kotlinx.cinterop.CPointer<kotlinx.cinterop.ByteVar /* = kotlinx.cinterop.ByteVarOf<kotlin.Byte> */>? */): kotlin.String { /* compiled code */ }
```

and then call in kotlin was:
```
NSString().initWithFormat("test %s", "test 2")
```

what throw exception:
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** initialization method -initWithFormat:locale:arguments: cannot be sent to an abstract object of class __NSCFConstantString: Create a concrete instance!'
```

how create NSString with format in kotlin/native?

SvyatoslavScherbina:
Fix #1834

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
@@ -57,7 +57,8 @@ internal interface KotlinStubs {
 
 private class KotlinToCCallBuilder(
         val irBuilder: IrBuilderWithScope,
-        val stubs: KotlinStubs
+        val stubs: KotlinStubs,
+        val isObjCMethod: Boolean
 ) {
 
     val cBridgeName = stubs.getUniqueCName("knbridge")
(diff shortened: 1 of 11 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -173,23 +174,98 @@ private fun KotlinToCCallBuilder.addArguments(arguments: List<IrExpression?>, ca
...
+         *
+         *   val secondTmp = *arrayOf(...)
+         *   val firstTmp = ...
+         *   foo(firstTmp, secondTmp)
+         *
+         *
+         **/
+
+        val variable = argument.symbol.owner
+        if (variable is IrVariable && variable.origin == IrDeclarationOrigin.IR_TEMPORARY_VARIABLE && !variable.isVar) {
```
olonho: :(

LouisCAD: What made you sad?

PR #2904, opened 2019-04-22
Added benchmarks for swift reverse interop

LepilkinaElena:


performance/swiftinterop/src/main/kotlin/org/jetbrains/launcher/SwiftLauncher.kt
```diff
@@ -0,0 +1,32 @@
+/*
```
olonho: Better use shorter copyright with correct year.

performance/swiftinterop/src/main/kotlin/org/jetbrains/launcher/SwiftLauncher.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.kliopt.*
+
+class SwiftLauncher(numWarmIterations: Int, numberOfAttempts: Int, prefix: String): Launcher(numWarmIterations, numberOfAttempts, prefix) {
+    override val benchmarks = BenchmarksCollection(
+            mutableMapOf(
+            )
+    )
+}
+
+/*fun main(args: Array<String>) {
```
olonho: Please no dead commented-out code.

performance/swiftinterop/swiftSrc/main.swift
```diff
@@ -0,0 +1,42 @@
...
+let args = KotlinArray(size: Int32(CommandLine.arguments.count), init: {index in
+    CommandLine.arguments[Int(truncating: index)]
+})
+runner.runBenchmarks(args: args, run: { (parser: ArgParser) -> [BenchmarkResult] in
+    var swiftBenchmarks = SwiftInteropBenchmarks()
+    var swiftLauncher = SwiftLauncher(numWarmIterations: parser.get(name: "warmup") as! Int32,
+        numberOfAttempts: parser.get(name: "repeat") as! Int32, prefix: parser.get(name: "prefix") as! String)
+    swiftLauncher.benchmarks["createMultigraphOfInt"] =  swiftBenchmarks.createMultigraphOfInt
+    swiftLauncher.benchmarks["fillCityMap"] =  swiftBenchmarks.fillCityMap
+    swiftLauncher.benchmarks["searchRoutesInSwiftMultigraph"] =  swiftBenchmarks.searchRoutesInSwiftMultigraph
```
olonho: Extra space after ` = `.

performance/swiftinterop/swiftSrc/main.swift
```diff
@@ -0,0 +1,42 @@
...
+    swiftLauncher.benchmarks["searchTravelRoutes"] =  swiftBenchmarks.searchTravelRoutes
+    swiftLauncher.benchmarks["availableTransportOnMap"] =  swiftBenchmarks.availableTransportOnMap
+    swiftLauncher.benchmarks["allPlacesMapedByInterests"] =  swiftBenchmarks.allPlacesMapedByInterests
+    swiftLauncher.benchmarks["getAllPlacesWithStraightRoutesTo"] =  swiftBenchmarks.getAllPlacesWithStraightRoutesTo
+    swiftLauncher.benchmarks["goToAllAvailablePlaces"] =  swiftBenchmarks.goToAllAvailablePlaces
+    swiftLauncher.benchmarks["removeVertexAndEdgesSwiftMultigraph"] =  swiftBenchmarks.removeVertexAndEdgesSwiftMultigraph
+    swiftLauncher.benchmarks["removePlacesAndRoutes"] =  swiftBenchmarks.removePlacesAndRoutes
+    return swiftLauncher.launch(benchmarksToRun: parser.getAll(name: "filter"))
```
olonho: Two return statements?

LepilkinaElena: These are return from different closures. First  from run  argument, second from collect argument. As I understood return statement should exist in closure in Swift.

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -0,0 +1,165 @@
...
+
+interface Cost {
+    operator fun plus(other: Cost): Cost
+    operator fun minus(other: Cost): Cost
+    override operator fun equals(other: Any?): Boolean
+    operator fun compareTo(other: Cost): Int
+}
+
+data class Edge<T>(val id: UInt, val from: T, val to: T, val cost: Cost) {
+    override operator fun equals(other: Any?): Boolean {
```
SvyatoslavScherbina: `hashCode` should be overridden too.

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -0,0 +1,165 @@
...
+
+class Multigraph<T>() {
+    private var edges = mutableMapOf<T, MutableList<Edge<T>>>()
+    private var idCounter = 0u
+
+    val allVertexes: Set<T>
+        get() {
+            val outerVertexes = edges.keys
+            val innerVertexes = edges.map { (key, values) ->
+                values.map { it.to }.filter { it in outerVertexes }
```
SvyatoslavScherbina: What is this code supposed to do?
As far as I see, `innerVertexes` is a subset of `outerVertexes`, so `outerVertexes.union(innerVertexes) == outerVertexes`.

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -0,0 +1,165 @@
...
+        edges.forEach { (_, value) ->
+            value.forEach {
+                if (it.id == id)
+                    return it
+            }
+        }
+        throw EdgeAbsenceMultigraphException("Edge with id $id wasn't found.")
+    }
+
+    fun copy(other: Multigraph<T>): Multigraph<T> {
```
SvyatoslavScherbina: `other` is unused?

performance/swiftinterop/swiftSrc/benchmarks.swift
```diff
@@ -0,0 +1,170 @@
...
+        while (!edges.isEmpty) {
+            multigraphCopy.removeEdge(id: UInt32(edges.randomElement()!))
+            let vertexes = multigraphCopy.allVertexes
+            multigraphCopy.removeVertex(vertex: vertexes.randomElement()!)
+            edges = multigraphCopy.allEdges
+        }
+    }
+
+    func removePlacesAndRoutes() {
+        var cityMapCopy = cityMap.copy() as! CityMap
```
SvyatoslavScherbina: What do you expect to get by calling `cityMap.copy`?

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -0,0 +1,165 @@
...
+            val edge = getEdgeById(id)
+            edges[edge.from]?.remove(edge)
+        } catch (exception: EdgeAbsenceMultigraphException) {
+            println("WARNING: no edge with id $id was found.")
+        }
+    }
+
+    fun removeVertex(vertex: T) {
+        edges.remove(vertex)
+        edges.map { (key, values) ->
```
SvyatoslavScherbina: What is this code supposed to do?

performance/swiftinterop/swiftSrc/benchmarks.swift
LepilkinaElena: Deep copy of object

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -0,0 +1,165 @@
+/*
```
SvyatoslavScherbina: Kotlin code does a lot of heavy computations for simple operations. Not sure if a benchmark using these operations is suitable for measuring Swift -> Kotlin interop.

performance/swiftinterop/swiftSrc/benchmarks.swift
SvyatoslavScherbina: Not sure if `cityMap.copy()` creates deep copy.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,134 @@
...
+    override operator fun minus(other: Cost) =
+            if (other is RouteCost) {
+                RouteCost(if (moneyCost > other.moneyCost) moneyCost - other.moneyCost else 0.0,
+                        if (timeCost > other.timeCost) timeCost - other.timeCost else 0.0,
+                        interests.subtract(other.interests), transport.subtract(other.transport))
+            } else {
+                error("Expected type is RouteCost")
+            }
+
+    override operator fun compareTo(other: Cost) =
```
SvyatoslavScherbina: This operation is not symmetric.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,134 @@
...
+}
+
+class Place(val geoCoordinateX: Double, val geoCoordinateY: Double, val name: String, val interestCategory: Interest) {
+    companion object {
+        var counter = 0u
+    }
+    val id: UInt
+    init {
+        id = counter
+        counter++
```
SvyatoslavScherbina: Is this code used? It doesn't look like it works properly.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    override operator fun equals(other: Any?): Boolean {
+        return if (other is Place) geoCoordinateX == other.geoCoordinateX && geoCoordinateY == other.geoCoordinateY &&
```
SvyatoslavScherbina: How would this be different from default `equals` implementation if `Place` was a `data class`?

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,134 @@
...
+    companion object {
+        var counter = 0u
+    }
+    val id: UInt
+    init {
+        id = counter
+        counter++
+    }
+
+    override operator fun equals(other: Any?): Boolean {
```
SvyatoslavScherbina: `hashCode`.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,134 @@
...
+        return if (other is Place) geoCoordinateX == other.geoCoordinateX && geoCoordinateY == other.geoCoordinateY &&
+                name == other.name && interestCategory == other.interestCategory
+        else false
+    }
+
+    fun compareTo(other: Place): Int {
```
SvyatoslavScherbina: Not symmetric.

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
LepilkinaElena: I mostly tried to repeat and measure some scenarios that could be in real applications with swift-kotlin interop, when common part is written on kotlin and swift just used function to get some results to show in UI. I can add very small functions to try measure mostly interop layer. But I amn't sure if such microbenchmarks can be stable enough. Do you suggest just add or replace all?

performance/swiftinterop/swiftSrc/benchmarks.swift
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+import Foundation
+import CityMap
+
+class SimpleCost: Cost {
+    var value: Int
```
SvyatoslavScherbina: `var` -> `let`

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
LepilkinaElena: Could you explain? Do you suppose that there would be wrong ids? I checked ids are right.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
LepilkinaElena: Place isn't data class, it has auto-generated id field.

SvyatoslavScherbina: > it has auto-generated id field.

How does this prevent `Place` from being a `data class`?

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
SvyatoslavScherbina: In Kotlin/Native you can't mutate fields in singletons.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
LepilkinaElena: Sorry, I was sure that all fields should be in constructor.

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
SvyatoslavScherbina: > I mostly tried to repeat and measure some scenarios that could be in real applications with swift-kotlin interop, when common part is written on kotlin

Makes sense.

> and swift just used function to get some results to show in UI.

This doesn't seem the only case to me.

> But I amn't sure if such microbenchmarks can be stable enough

Calling small Kotlin functions from Swift doesn't imply making a microbenchmark. You can have some non-trivial logic in Swift instead.

LepilkinaElena: I added 2 simple benchmarks, we already have similar small benchmarks in Ring, so seems that there shouldn't be problems.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,138 @@
...
+                when {
+                    this == other -> 0
+                    moneyCost < other.moneyCost -> -1
+                    moneyCost > other.moneyCost -> 1
+                    timeCost < other.timeCost -> -1
+                    timeCost > other.timeCost -> 1
+                    interests.size < other.interests.size -> -1
+                    interests.size > other.interests.size -> 1
+                    transport.size < other.transport.size -> -1
+                    else -> 1
```
SvyatoslavScherbina: Still not symmetric here.

performance/swiftinterop/src/main/kotlin/org/jetbrains/model/CityMap.kt
```diff
@@ -0,0 +1,138 @@
...
+        return id.toInt()
+    }
+
+    fun compareTo(other: Place): Int {
+        return when {
+            this == other -> 0
+            geoCoordinateX < other.geoCoordinateX -> -1
+            geoCoordinateX > other.geoCoordinateX -> 1
+            geoCoordinateY < other.geoCoordinateY -> -1
+            else -> 1
```
SvyatoslavScherbina: Still not symmetric here.
Consider using functions from https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.comparisons/index.html.
E.g. you can create comparator once (storing into top-level variable).

performance/swiftinterop/src/main/kotlin/org/jetbrains/multigraph/Multigraph.kt
```diff
@@ -16,6 +16,9 @@ data class Edge<T>(val id: UInt, val from: T, val to: T, val cost: Cost) {
     override operator fun equals(other: Any?): Boolean {
         return if (other is Edge<*>) (from == other.from && to == other.to && cost == other.cost) else false
     }
+
+    override fun hashCode(): Int =
```
SvyatoslavScherbina: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html

> If two objects are equal according to the `equals()` method, then calling the `hashCode` method on each of the two objects must produce the same integer result.

PR #2912, opened 2019-04-24
Improve stack slots object references handling.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2359,12 +2288,8 @@ OBJ_GETTER(ReadHeapRefLocked, ObjHeader** location, int32_t* spinlock) {
   MEMORY_LOG("ReadHeapRefLocked: %p\n", location)
   lock(spinlock);
   ObjHeader* value = *location;
-  // We do not use UpdateRef() here to avoid having ReleaseRef() on return slot under the lock.
-  if (value != nullptr) {
-    AddStackRef(value);
-  }
   unlock(spinlock);
-  updateReturnRefAdded(OBJ_RESULT, value);
+  UpdateReturnRef(OBJ_RESULT, value);
```
SvyatoslavScherbina:
```kotlin
import platform.posix.*

@SharedImmutable
val ref = AtomicReference<Any?>(Any().freeze())
    
fun main() {
    val worker = Worker.start()
    val future = worker.execute(TransferMode.SAFE, { null }) {
        val value = ref.value
        sleep(2)
        println(value)
    }
    sleep(1)
    ref.value = null

    future.consume {}
    worker.requestTermination().consume {}
}
```

backend.native/tests/runtime/workers/worker10.kt
```diff
@@ -78,4 +78,24 @@ val topSharedData = Data(43)
 
     worker.requestTermination().result
     println("OK")
+}
+
+val ref = AtomicReference<Any?>(Any().freeze())
+
+val semaphore = AtomicInt(0)
+
+@Test fun runTest2() {
```
SvyatoslavScherbina: This test doesn't reproduce [the original bug](https://github.com/JetBrains/kotlin-native/pull/2912#discussion_r278066070) on [cb7e81a7](https://github.com/JetBrains/kotlin-native/commit/cb7e81a7e20f9efcd81cf4a94660e488da933324).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2359,13 +2272,20 @@ OBJ_GETTER(ReadHeapRefLocked, ObjHeader** location, int32_t* spinlock) {
...
   ObjHeader* value = *location;
-  // We do not use UpdateRef() here to avoid having ReleaseRef() on return slot under the lock.
-  if (value != nullptr) {
-    AddStackRef(value);
-  }
   unlock(spinlock);
-  updateReturnRefAdded(OBJ_RESULT, value);
-  return value;
+  if (value != nullptr)
+    rememberNewContainer(value->container());
```
SvyatoslavScherbina: It's too late to do this after `unlock`.
Another thread may have changed `*location` and performed GC by this moment.
`IncrementRC` must be done under lock.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1554,6 +1489,16 @@ void garbageCollect(MemoryState* state, bool force) {
   GC_LOG("<<< GC: toFree %d toRelease %d\n", state->toFree->size(), state->toRelease->size())
 }
 
+void rememberNewContainer(ContainerHeader* container) {
```
SvyatoslavScherbina: This operation should also be applied to every reference read from the "heap" that could be released by other thread, e.g.

* References from `StableRef`
* References from `KRefSharedHolder`
* Values of `MutableData.buffer`

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1476,6 +1409,8 @@ void processDecrements(MemoryState* state) {
      toRelease->pop_back();
      if (isMarkedAsRemoved(container))
        continue;
+     if (container->shareable())
+       container = realShareableContainer(container);
```
SvyatoslavScherbina: Why not enqueue `ObjHeader*` instead?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1554,6 +1489,16 @@ void garbageCollect(MemoryState* state, bool force) {
   GC_LOG("<<< GC: toFree %d toRelease %d\n", state->toFree->size(), state->toRelease->size())
 }
 
+void rememberNewContainer(ContainerHeader* container) {
+  if (container == nullptr) return;
+  // Instances can be allocated before actual runtime init - be prepared for that.
+  if (memoryState != nullptr) {
+    IncrementRC</* Atomic = */ false>(container);
```
SvyatoslavScherbina: > `/* Atomic = */ false`

Doesn't seem correct when reading from `AtomicReference`.

backend.native/tests/runtime/workers/worker10.kt
olonho: Hopefully now it does.

runtime/src/main/cpp/Memory.cpp
olonho: Not a big deal, I guess. We generally try to use containers in MM algorithms.

runtime/src/main/cpp/Memory.cpp
olonho: Yeah, noticed that as well.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: + `Konan_WeakReferenceCounter_get`

Not sure if I found all such references.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -555,6 +556,15 @@ inline void unlock(KInt* spinlock) {
 
 } // namespace
 
+ObjHeader* KRefSharedHolder::ref() const {
+  verifyRefOwner();
+#if USE_GC
+  if (obj_ != nullptr)
+    rememberNewContainer(obj_->container());
```
SvyatoslavScherbina: There are usages of `KRefSharedHolder::ref()` that don't require `rememberNewContainer`.
Most of them now becomes suboptimal, but few of the usages are now incorrect (e.g. `KotlinBase._tryRetain`).

So we might have two variants of this function, e.g. `ref` and `localRef`, the second one is not supposed to outlive the holder and thus could be used when the reference doesn't escape.

olonho: Seems that this change makes more harm then good. Let's think over if this behavior is indeed important.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -559,9 +559,10 @@ inline void unlock(KInt* spinlock) {
 ObjHeader* KRefSharedHolder::ref() const {
   verifyRefOwner();
 #if USE_GC
-  if (obj_ != nullptr)
+  if (obj_ != nullptr && owner_ != memoryState)
```
SvyatoslavScherbina: This is wrong. Frozen object may be released by other thread.

backend.native/tests/runtime/workers/worker10.kt
```diff
@@ -118,4 +119,28 @@ val semaphore = AtomicInt(0)
...
+    assertEquals(expected, weak.get())
+}
+
+val stableHolder = StableRef.create(("hello" to "world").freeze())
+
+@Test fun runTest4() {
+    val worker = Worker.start()
+    semaphore.value = 0
+    val future = worker.execute(TransferMode.SAFE, { WeakReference(stableHolder.get()) }) {
+        ensureWeakIs(it, "hello" to "world")
```
SvyatoslavScherbina: This test doesn't reproduce the original issue.
If this worker task is modified to be like that in previous tests, then the issue gets reproduced.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: As we've discussed earlier, this should be reverted.

PR #2915, opened 2019-04-24
Unify compiler and cinterop flags names

LepilkinaElena:

PR #2921, opened 2019-04-26
Update kotlin compiler

LepilkinaElena:


LepilkinaElena:
I'm waiting until this version is built6 now there are some problems with agents, but locally there is no problems with merge (locally build and tests passed). I need this version to merge https://github.com/JetBrains/kotlin-native/pull/2915

PR #2935, opened 2019-04-30
Update sdk version for ios targets

PavelPunegov:


konan/konan.properties
```diff
@@ -83,7 +83,7 @@ linkerNoDebugFlags.ios_arm32 = -S
 linkerDynamicFlags.ios_arm32 = -dylib
 llvmLtoNooptFlags.ios_arm32 = -O1
 llvmLtoDynamicFlags.ios_arm32 = -relocation-model=pic
-linkerKonanFlags.ios_arm32 = -lSystem -lc++ -lobjc -framework Foundation -sdk_version 11.2
+linkerKonanFlags.ios_arm32 = -lSystem -lc++ -lobjc -framework Foundation -sdk_version 12.1
```
SvyatoslavScherbina: Maybe update to 12.2? AFAIK, it corresponds to our sysroots.

benasher44: FWIW, arm32 devices can't run iOS 12. Max major iOS version for arm32 devices is iOS 10

benasher44: Actually this doesn't matter. This isn't deployment target. Ignore me.

PR #2936, opened 2019-04-30
Fix #2931

### LINKED ISSUE #2931: objc_release crash on enums
I'm not 100% sure this is just for enums, but it looks like a memory management race condition in objc interop.

Create an enum class

```kotlin
enum class Direction {
    NORTH, SOUTH, WEST, EAST, NORTHWEST, NORTHEAST
}
```

From Swift, loop through that a lot from more than one thread.

```swift
@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
    var window: UIWindow?
    
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {

        for i in 0..<2 {
            AppDelegate.loadInBackground()
        }
        usleep(10000000)
        
        return true
    }
    
    static func loadInBackground()  {
        DispatchQueue.global(qos: .background).async {
            for i in 0..<100 {
                let et : Direction?
                switch i % 7 {
                case 0:
                    et = Direction.east
                case 1:
                    et = Direction.north
                case 2:
                    et = Direction.northeast
                case 3:
                    et = Direction.northwest
                case 4:
                    et = Direction.south
                case 5:
                    et = Direction.west
                default:
                    et = nil
                }
                if i % 1000 == 0 {
                    print("entity: \(et)")
                }
            }
            print("This is run on the background queue")
        }
    }
//AppDelegate etc
}
```

On multiple threads that will trigger a crash in objc_release.

[Xcode screenshot](https://drive.google.com/file/d/1ndWDOegGracAdUiYVVLZfMBwauB31Q2R/view?usp=sharing) for more context

I've tried this on 1.3.31. The version above has 2 threads and 100 loops, which should in theory make it unlikely to happen, but it's been happening every time. If repro isn't working, try higher loop count (assuming the problem is a memory/arc race).

[Slightly different form](https://drive.google.com/file/d/1Y3CoTQCO7RuSXuh4VfIcNptCPeRIe9Gy/view?usp=sharing). Happens occasionally.

## Update with some additional info

This is very easy to trigger and doesn't really need a loop.

```swift
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
        for i in 0..<2 {
            AppDelegate.loadInBackground(tcount: i)
        }
        usleep(2000000)
        
        return true
    }
    
    static func loadInBackground(tcount: Int)  {
        DispatchQueue.global(qos: .background).async {
            let et = Direction.east
        }
    }
```

Adding a delay in the top loop prevents the issue, which suggests a race condition, but it seems unlikely that DispatchQueue would be so precise in the example above as to cause the crash every time.

Anyway, this avoids crashing. Smaller usleep delay values eventually start crashing again.

```swift
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
        for i in 0..<2 {
            AppDelegate.loadInBackground(tcount: i)
            usleep(20000)
        }
        usleep(2000000)
        
        return true
    }
```

Some other things I've tried.

```kotlin
fun makeDir():Direction{
    initRuntimeIfNeeded()
    return Direction.EAST
}
```

Also crashes.

Putting the call in [an autoreleasepool](https://drive.google.com/file/d/1ODsAx-qhUUNAKhKNphQ10ID9Z1KX6S3_/view?usp=sharing) moves the place where it crashes, but still does.

Finally, just as a test I did the following which was no help.

```kotlin
enum class Direction {
    @SharedImmutable
    NORTH,
    @SharedImmutable
    SOUTH,
    @SharedImmutable
    WEST,
    @SharedImmutable
    EAST,
    @SharedImmutable
    NORTHWEST,
    @SharedImmutable
    NORTHEAST
}
```

I've also tried this on 1.3.21 and it seems to be the same.

SvyatoslavScherbina:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -1125,15 +1125,23 @@ inline size_t containerSize(const ContainerHeader* container) {
...
+#endif
+
+  MetaObjHeader* meta = konanConstructInstance<MetaObjHeader>();
   meta->typeInfo_ = typeInfo;
 #if KONAN_NO_THREADS
   *location = reinterpret_cast<TypeInfo*>(meta);
 #else
   TypeInfo* old = __sync_val_compare_and_swap(location, typeInfo, reinterpret_cast<TypeInfo*>(meta));
-  if (old->typeInfo_ != old) {
+  if (old != typeInfo) {
```
olonho: In my tests just this line is sufficient for race to go away, code above just duplicates the check in `has_meta_object()`. My test was was Swift program from my comment in #2931 with 100 background jobs enclosed in while loop and with another commit from this PR applied.

olonho:
Can we have a test for this issue?

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: In my tests just this line is not sufficient for race to go away.

`test.kt`:
```kotlin
enum class Direction {
    NORTH, SOUTH, WEST, EAST, NORTHWEST, NORTHEAST
}

fun toString(obj: Any) = obj.toString()
```

`test.swift`:
```swift
import Test

func loadInBackground()  {
    DispatchQueue.global(qos: .background).async {
        TestKt.toString(obj: Direction.east)
    }
}

for i in 0..<2 {
    loadInBackground()
}

usleep(100000)
```

> code above just duplicates the check in `has_meta_object()`

Code above doesn't just duplicate the check in `has_meta_object()`. It prevents from data race when `typeInfo` is different from that read in `has_meta_object()`:

1.  Thread 1 loads original type info and ensures that `has_meta_object()` is `false`. Then it allocates `meta1`.
2.  Thread 2 successfully checks, allocates and stores meta object `meta2`.
3.  Thread 1 loads `meta2` as `typeInfo` and makes `meta1` pointing `meta2` instead of original type info. Then it makes successful compare-and-swap, and now the object points to `meta1` which points to `meta2`.

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -112,21 +112,13 @@ inline static OBJ_GETTER(AllocInstanceWithAssociatedObject, const TypeInfo* type
...
-    void* old = __sync_val_compare_and_swap(&obj->meta_object()->associatedObject_, nullptr, (void*)associatedObject);
-    if (old == nullptr) {
-      return associatedObject;
-    } else {
-      Kotlin_ObjCExport_releaseAssociatedObject((void*)associatedObject);
-      return (id)old;
-    }
-  }
+static inline id AtomicCompareAndSwapAssociatedObject(ObjHeader* obj, id expectedValue, id newValue) {
+  return (id)__sync_val_compare_and_swap(&obj->meta_object()->associatedObject_, (void*)expectedValue, (void*)newValue);
```
olonho: Maybe `reinterpret_cast<id>` here?

SvyatoslavScherbina: After taking a fresh look, I consider turning this into 
```
id* location = reinterpret_cast<id*>(&obj->meta_object()->associatedObject_);
return __sync_val_compare_and_swap(location, expectedValue, newValue);
```
WDYT?

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -264,11 +261,21 @@ - (instancetype)copyWithZone:(NSZone *)zone {
         freeWhenDone:NO] autorelease];
   } else {
     // TODO: consider making NSString subclass to avoid copying here.
-    NSString* result = [[NSString alloc] initWithBytes:utf16Chars
+    NSString* candidate = [[NSString alloc] initWithBytes:utf16Chars
```
olonho: Unrelated to this change, but https://developer.apple.com/documentation/foundation/nsstring/1413830-initwithbytesnocopy?language=objc may be better choice here, lifetime of Kotlin and Obj-C objects likely match, so we can use single storage.

SvyatoslavScherbina: > lifetime of Kotlin and Obj-C objects likely match,

Unfortunately, this is not correct: default `NSString` implementation can't prevent Kotlin object from removing. Consider the case like
```kotlin
// Kotlin
fun toString(x: Int) = x.toString()
```

```swift
// Swift
let str = TestKt.toString(x: 42)
// Kotlin string doesn't have any references to it anymore.
```

 We need custom `NSString` implementation holding Kotlin object reference like `KotlinBase` does. `TODO` above refers exactly to this approach.

PR #2946, opened 2019-05-06
Custom exception hook in reverse C interop

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -744,6 +745,7 @@ private fun KotlinStubs.mapType(
     type.isLong() -> TrivialValuePassing(irBuiltIns.longType, CTypes.longLong)
     type.isFloat() -> TrivialValuePassing(irBuiltIns.floatType, CTypes.float)
     type.isDouble() -> TrivialValuePassing(irBuiltIns.doubleType, CTypes.double)
+    type.isStringClassType() && typeLocation.shallUseCStrings -> CStringValuePassing(symbols)
```
SvyatoslavScherbina: It was intentionally designed to forbid this case.

backend.native/tests/produce_dynamic/simple/hello.kt
```diff
@@ -96,3 +97,15 @@ fun useInlineClasses(ic1: IC1, ic2: IC2, ic3: IC3) {
     assert(ic2.value == "bar")
     assert(ic3.value is Base)
 }
+
+fun setCErrorHandler(callback: CPointer<CFunction<(String) -> Unit>>?) {
```
SvyatoslavScherbina: `CPointer<CFunction<(String) -> Unit>>` is not supposed to work this way.
Use `CPointer<ByteVar>` instead of `String`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -671,6 +669,9 @@ private sealed class TypeLocation(val element: IrElement) {
     class BlockReturnValue(val blockLocation: TypeLocation) : TypeLocation(blockLocation.element)
 }
 
+private val TypeLocation.shallUseCStrings: Boolean
+    get() = this is TypeLocation.FunctionArgument || this is TypeLocation.FunctionPointerParameter
```
SvyatoslavScherbina: This is wrong, see `interop_objc_smoke` test.
Also this will likely break support for C function types with `NSString*`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
olonho: Let's discuss offline.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -455,6 +456,8 @@ private class ExportedElement(val kind: ElementKind,
                     "result", cfunction[0].second, Direction.KOTLIN_TO_C, builder)
             builder.append("  return $result;\n")
         }
+        builder.append("   } catch (KExceptionObjHolder& e) { TerminateWithUnhandledException(e.obj()); } \n")
```
SvyatoslavScherbina: How does this work if nobody throws `KExceptionObjHolder`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -838,8 +841,8 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
         output("typedef float              ${prefix}_KFloat;")
         output("typedef double             ${prefix}_KDouble;")
         output("typedef void*              ${prefix}_KNativePtr;")
-        output("struct ${prefix}_KType;")
-        output("typedef struct ${prefix}_KType ${prefix}_KType;")
+        output("struct KType;")
```
SvyatoslavScherbina: Why so?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
olonho: Maybe due to type deduplication. Anyway, made it consistent.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
olonho: Just because same reason as with `KObjHeader*` - it's an opaque reference type.

SvyatoslavScherbina: `KObjHeader*` seems to be private for generated `.cpp` file, but `KType` is user-visible. C compiler indeed can handle multiple such `KType` declarations, but these don't represent the same type, so I'm not sure if it's ok to merge them.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
SvyatoslavScherbina: Actually the test (`produce_dynamic`) passes without this try-catch. And I can't reproduce original issue on macOS neither with nor without this try-catch.
Is the issue Windows-specific? Were you able to reproduce the issue and ensure that it is fixed by this patch?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
olonho: Yeah, may make sense, as indeed this way it's easier to mix types from two potentially different runtimes, if more than one library is used.

PR #2947, opened 2019-05-06
Add initial experimental support for @AssociatedObjectKey

SvyatoslavScherbina:


runtime/src/main/cpp/TypeInfo.h
```diff
@@ -114,6 +114,9 @@ struct TypeInfo {
     WritableTypeInfo* writableInfo_;
 #endif
 
+    // Null-terminated array of pairs (TypeInfo*, OBJ_GETTER0(getAssociatedObjectInstance)).
+    void** associatedObjects;
```
olonho: Why not type properly?

SvyatoslavScherbina: Because checking for null terminator becomes somewhat dirty in this case.
We get forced either to including extra pointer or making last element in this array half-defined. Do you find this ok?

olonho: Other similar structures in this type have explicit count next to it. We do it this way, or just zero TypeInfo in the terminator. I probably like the first one more.

SvyatoslavScherbina: The first approach applied the same way as for other data would make every TypeInfo larger. We can use additional indirection here, but using zero terminator looks much simpler to me.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/RuntimeNames.kt
```diff
@@ -11,4 +11,5 @@ object RuntimeNames {
     val independent = FqName("kotlin.native.internal.Independent")
     val filterExceptions = FqName("kotlin.native.internal.FilterExceptions")
     val kotlinNativeInternalPackageName = FqName.fromSegments(listOf("kotlin", "native", "internal"))
+    val associatedObjectKey = FqName("kotlin.reflect.AssociatedObjectKey")
```
olonho: Maybe some other package, like `kotlin.annotations`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -6,19 +6,17 @@
 package org.jetbrains.kotlin.backend.konan.llvm
 
 import llvm.*
+import org.jetbrains.kotlin.backend.konan.*
 import org.jetbrains.kotlin.backend.konan.Context
-import org.jetbrains.kotlin.backend.konan.computePrimitiveBinaryTypeOrNull
 import org.jetbrains.kotlin.backend.konan.descriptors.*
 import org.jetbrains.kotlin.backend.konan.ir.*
+import org.jetbrains.kotlin.backend.konan.ir.isAnonymousObject
```
olonho: Are those two indeed used?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/RuntimeNames.kt
SvyatoslavScherbina: I like the idea of having both ends of this machinery (`@AssociatedObjectKey` and `getAssociatedObject()`) in the same package. Not sure if this makes much sense though.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
SvyatoslavScherbina: These imports are required because of ambiguity appeared after adding `import org.jetbrains.kotlin.ir.util.*`.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -0,0 +1,25 @@
...
+@Experimental(level = Experimental.Level.ERROR)
+annotation class ExperimentalAssociatedObjects
+
+@ExperimentalAssociatedObjects
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+annotation class AssociatedObjectKey
+
+@ExperimentalAssociatedObjects
+fun KClass<*>.getAssociatedObject(key: KClass<*>): Any? =
```
elizarov: That is somewhat un-Kotlin API (non typesafe & verbose). Public API should look something like

```
inline fun <reified T> KClass<*>.getAssociatedObject(): T
```

But even that way it does not look "native" with respect to the rest of reflection API (the rest of reflection consists mostly of vals). I wonder if @udalov can suggest something better.

SvyatoslavScherbina: > `inline fun <reified T> KClass<*>.getAssociatedObject(): T`

Have you omitted `key` parameter intentionally?

elizarov: Writing `getAssiatedObject(Key::class)` looks less convenient to me than `getAssociatedObject<Key>()` (that's what refied type are parameters for)

SvyatoslavScherbina: Ok. Then this function can't return `T`, because `T` is the annotation class then.
Maybe we should have `@AssociatedObject(KClass<*> key, KClass<*> value)` then instead.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -17,9 +17,5 @@ annotation class ExperimentalAssociatedObjects
 annotation class AssociatedObjectKey
 
 @ExperimentalAssociatedObjects
-fun KClass<*>.getAssociatedObject(key: KClass<*>): Any? =
-        if (this is KClassImpl<*> && key is KClassImpl<*>) {
-            this.getAssociatedObjectImpl(key)
-        } else {
-            null
-        }
+inline fun <reified T> KClass<*>.findAssociatedObject(): Any? =
```
elizarov: Needs docs! (here and for other public declarations)

ilya-g: And also explicit `public` visibility for declarations.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Experimental(level = Experimental.Level.ERROR)
+@Retention(value = AnnotationRetention.BINARY)
+public annotation class ExperimentalAssociatedObjects
+
+/**
+ * Makes the annotated annotation class an associated object key.
+ *
+ * An associated object key annotation should have single [KClass] parameter.
+ * When applied to a class with reference to a singleton as an argument, it binds
```
udalov: Maybe consider using the term "object declaration" instead of "singleton" because the latter might give impression that enum entry is also possible, which is not

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -0,0 +1,46 @@
...
+public annotation class AssociatedObjectKey
+
+/**
+ * If [T] is an @[AssociatedObjectKey]-annotated annotation class and [this] class is annotatated with @[T] (`S::class`),
+ * returns the instance of singleton `S`.
+ *
+ * Otherwise returns `null`.
+ */
+@ExperimentalAssociatedObjects
+public inline fun <reified T> KClass<*>.findAssociatedObject(): Any? =
```
udalov: `T : Annotation`

SvyatoslavScherbina: Done.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
SvyatoslavScherbina: Makes sense. Somewhat reworded.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -0,0 +1,46 @@
...
+ * [general compatibility guarantees](https://kotlinlang.org/docs/reference/evolution/components-stability.html) given for the standard library:
+ * the behavior of such API may be changed or the API may be removed completely in any further release.
+ *
+ * > Beware using the annotated API especially if you're developing a library, since your library might become binary incompatible
+ * with the future versions of the standard library.
+ *
+ * Any usage of a declaration annotated with `@ExperimentalAssociatedObjects` must be accepted either by
+ * annotating that usage with the [UseExperimental] annotation, e.g. `@UseExperimental(ExperimentalAssociatedObjects::class)`,
+ * or by using the compiler argument `-Xuse-experimental=kotlin.reflect.ExperimentalAssociatedObjects`.
+ */
```
ilya-g: Better to copy-paste docs from another experimental annotation, e.g. `ExperimentalTypeInference`.

SvyatoslavScherbina: > Better to copy-paste docs from another experimental annotation

Exactly! That's why I've copied this from [`ExperimentalStdlibApi`](https://github.com/JetBrains/kotlin/blob/43bf962d790a532981020b729d94ba3da87a316f/libraries/stdlib/src/kotlin/annotations/ExperimentalStdlibApi.kt#L6).

> e.g. `ExperimentalTypeInference`.

Do you find it better source?

ilya-g: I mean not from `ExperimentalStdlibApi` but from another one, for example from `ExperimentalTypeInference`

SvyatoslavScherbina: Done.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
```diff
@@ -0,0 +1,41 @@
...
+ * When applied to a class with reference to an object declaration as an argument, it binds
+ * the object to the class, making this binding discoverable at runtime using [findAssociatedObject].
+ */
+@ExperimentalAssociatedObjects
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+public annotation class AssociatedObjectKey
+
+/**
+ * If [T] is an @[AssociatedObjectKey]-annotated annotation class and [this] class is annotatated with @[T] (`S::class`),
```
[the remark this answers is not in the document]
ilya-g: Another wording suggestion, change type parameter of key to `K` and then
```
If this class is annotated with an annotation of type [K], which is itself an associated object key (see [AssociatedObjectKey] annotation), 
returns an instance of the class specified by that annotation or `null` otherwise.
```

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
SvyatoslavScherbina: > I don't think that pursuing reified type parameter brings much value.

However it brings some of it, and it is more consistent with related reflection APIs.
I prefer using reified parameter here.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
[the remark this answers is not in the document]
SvyatoslavScherbina: Kinda unreadable:
> an annotation of type [K]

> (see [AssociatedObjectKey] annotation)

> an instance of the class specified by that annotation

which one?

Also I consider my version more self-contained and not relying on `AssociatedObjectKey` documentation that much.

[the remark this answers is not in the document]
ilya-g: I was having hard time deciphering what "@[T] (`S::class`)" means :)
That's the problem inherent to documentation of meta-annotations.

Is this better?
> an instance of the class specified by that `K` annotation

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/RuntimeNames.kt
ilya-g: `kotlin.annotation` package is for general annotation facility support types. This annotation is specialized for reflection use cases, so its `kotlin.reflect` package placement looks logical.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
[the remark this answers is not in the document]
SvyatoslavScherbina: > I was having hard time deciphering what "@[T] (S::class)" means :)

It is rendered as `@T (S::class)`. A space character may be confusing but is required for kdoc parser to recognize `[T]` properly. I can avoid using link here to make it `T(S::class)` which is supposed to be pretty straightforward and understandable.

> Is this better?

Still has a couple of issues:

* Relies on  `AssociatedObjectKey` documentation too much.
* "an instance of the class" is too low-level. Kotlin object declarations don't seem to be referred as classes in the official documentation, so "instance of the class" is somewhat unclear from the language point of view.

runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt
ilya-g: Ok, we can start from the reified overload, and expose the non-reified one if there will be request for that.

backend.native/tests/codegen/associatedObjects/associatedObjects1.kt
```diff
@@ -0,0 +1,77 @@
...
+@Associated2(Baz::class)
+class Foo
+
+object Bar
+object Baz
+
+@Test
+@UseExperimental(ExperimentalAssociatedObjects::class)
+fun testGlobalOptimizations1() {
+    val i1 = I1ImplHolder::class.findAssociatedObject<Associated1>()!! as I1
```
olonho: Is `findAssociatedObject` right? Maybe `getAssociatedObject`?

SvyatoslavScherbina: This variant was selected on design meeting.

olonho: Shalln't this method take second type parameter of object type and be reified on that?

SvyatoslavScherbina: See [above](https://github.com/JetBrains/kotlin-native/pull/2947#discussion_r287829580).

PR #2952, opened 2019-05-07
Implement -Xframework-import compiler flag

SvyatoslavScherbina:
To be used to add imports to framework header to workaround missing imports issues.

See #2951


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -119,6 +119,13 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var exportedLibraries: Array<String>? = null
 
+    @Argument(
+            value = "-Xframework-import",
```
olonho: Maybe `-Xframework-import-headers`?

PR #2964, opened 2019-05-12
Small usability tweaks.

olonho:


README.md
```diff
@@ -14,6 +14,7 @@ Prerequisites:
 *   install JDK for your platform, instead of JRE. The build requires ```tools.jar```, which is not included in JRE;
 *   on macOS install Xcode 10.2
 *   on Fedora 26+ ```yum install ncurses-compat-libs``` may be needed
+*   on recent Ubuntu ```apt install libncurses5` is needed
```
sbogolepov: Maybe generalize the requirement? `libncurses5` is required on all Linux distros, not just Ubuntu or Fedora.

olonho: Guess we'd better link it into LLVM on the next compiler update, or compile without this dep.

PR #2966, opened 2019-05-13
[kotlin compiler][update] 1.3.50-dev-67

vvlevchenko:


vvlevchenko:
> Please do not merge the changes of this PR related to stdlib functions, otherwise they will cause conflict with PR #2973

I'm glad to follow your request, but I afraid it's impossible, changes to stdlib is a part of update, without them master will be red until #2973 won't be applied. there're two option merge both PRs or merge on master on by one.

PR #2967, opened 2019-05-13
Support nullable string return type in reverse C interop.

olonho:


runtime/src/main/cpp/KString.cpp
```diff
@@ -712,6 +712,7 @@ OBJ_GETTER(CreateStringFromUtf8, const char* utf8, uint32_t lengthBytes) {
 }
 
 char* CreateCStringFromString(KConstRef kref) {
+  if (kref == nullptr) return nullptr;
```
SvyatoslavScherbina: Does conversion from null `const char*` pointer to `String?` work properly? Do we have a test for this case?

olonho: You mean in direct interop?

SvyatoslavScherbina: No, I mean reverse interop when `String?` is used as parameter type.

PR #2968, opened 2019-05-13
Map returned Unit to void in more cases when producing framework

SvyatoslavScherbina:
Including function-typed parameters and return types


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt
```diff
@@ -138,10 +138,25 @@ private fun ObjCExportMapper.bridgeType(kotlinType: KotlinType): TypeBridge = ko
...
+            if (kotlinType.isFunctionType) {
+                bridgeFunctionType(kotlinType)
+            } else {
+                ReferenceBridge
+            }
         }
 )
 
+private fun ObjCExportMapper.bridgeFunctionType(kotlinType: KotlinType): TypeBridge {
+    val numberOfParameters = kotlinType.arguments.size - 1
```
olonho: Better comment why -1, I guess.

SvyatoslavScherbina: Done.

PR #2969, opened 2019-05-13
Fix #2830, #2959

### LINKED ISSUE #2830: Cast internal class object to public interface fails in Swift
When  I declare in Kotlin :
```
public interface InterA {}
internal class ClassA: InterA {}
```
And expose in Swift an `InterA` backed by a `ClassA` instance,
Some cast operations fail: 
```
    private func getObject<T>(_ value: InterA) -> T? {
        let castedValue = value as? T // nil
        let test1 = value is T // false
        let test2 = value is InterA // true
        let test3 = T.self == InterA.self // true
        return castedValue
    }

    let inter: InterA? = getObject(commonManager.api.a) // commonManager.api.a is a ClassA
```
It is not possible to cast `value` in `T`, even if `T` is `InterA`. If I change getObject to take `value` as `Any`, even `test2` fails.

Everything works well if ClassA is public.

The use case behind that issue is to be able tu use a DI framework such as Swinject.

### LINKED ISSUE #2959: Precondition failed: NSArray element failed to match the Swift Array Element type
Hi!

I have a interface, inheriting from another interface like this:

```
interface Book : BaseBook {
    val name: String
}

interface BaseBook {
    val id: Long
}
```
Function for loading books (iOS and Android):
```
Utils.sdk()?.booksController().getBooks(successCallback:{ [weak self] books in
             print(books[0].name) <--------Error occurs here
                return KotlinUnit()
            }, errorCallback: { [weak self] error in
                return KotlinUnit()
        })
```
Error:
```
Precondition failed: NSArray element failed to match the Swift Array Element type
Expected FrameworkBook but found kobjcc0: file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1001.2.63.12/swift/stdlib/public/core/ArrayBuffer.swift, line 346
2019-05-10 14:44:01.547868+0200 SRSDK Demo[9918:12973242] Precondition failed: NSArray element failed to match the Swift Array Element type
Expected FrameworkBook but found kobjcc0: file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1001.2.63.12/swift/stdlib/public/core/ArrayBuffer.swift, line 346
```
If I change Book to data class, everything works as expected. Is there something wrong with the way I am handling interfaces and objects or might there be a bug in K/N?

P.S. If I debug the iOS app, I can see the objects in array (the size of the array is correct too), but instead of objects and their properties I get:
```
[0] = (kobjcc0 *) 0x6000020f0180
    KotlinBase
        NSObject
            isa = (Class) kobjcc0
```

SvyatoslavScherbina:
Fix #2830
Fix #2959

### DIFF
--- a/runtime/src/main/cpp/ObjCExport.mm
+++ b/runtime/src/main/cpp/ObjCExport.mm
@@ -326,18 +326,21 @@ static const ObjCTypeAdapter* getTypeAdapter(const TypeInfo* typeInfo) {
   return typeInfo->writableInfo_->objCExport.typeAdapter;
 }
 
-static Protocol* getProtocolForInterface(const TypeInfo* interfaceInfo) {
+static void addProtocolForAdapter(Class clazz, const ObjCTypeAdapter* protocolAdapter) {
+  Protocol* protocol = objc_getProtocol(protocolAdapter->objCName);
+  if (protocol != nullptr) {
+    class_addProtocol(clazz, protocol);
+    class_addProtocol(object_getClass(clazz), protocol);
+  } else {
+    // TODO: construct the protocol in compiler instead, because this case can't be handled easily.
+  }
+}
+
+static void addProtocolForInterface(Class clazz, const TypeInfo* interfaceInfo) {
   const ObjCTypeAdapter* protocolAdapter = getTypeAdapter(interfaceInfo);
   if (protocolAdapter != nullptr) {
-    Protocol* protocol = objc_getProtocol(protocolAdapter->objCName);
-    if (protocol != nullptr) {
-      return protocol;
-    } else {
-      // TODO: construct the protocol in compiler instead, because this case can't be handled easily.
-    }
+    addProtocolForAdapter(clazz, protocolAdapter);
   }
-
-  return nullptr;
 }
 
 static const TypeInfo* getOrCreateTypeInfo(Class clazz);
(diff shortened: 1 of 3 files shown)


runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -326,18 +326,21 @@ - (instancetype)copyWithZone:(NSZone *)zone {
...
 }
 
-static Protocol* getProtocolForInterface(const TypeInfo* interfaceInfo) {
+static void addProtocolForAdapter(Class clazz, const ObjCTypeAdapter* protocolAdapter) {
+  Protocol* protocol = objc_getProtocol(protocolAdapter->objCName);
+  if (protocol != nullptr) {
+    class_addProtocol(clazz, protocol);
+    class_addProtocol(object_getClass(clazz), protocol);
+  } else {
+    // TODO: construct the protocol in compiler instead, because this case can't be handled easily.
```
olonho: Shalln't we assert/inform user here?

SvyatoslavScherbina: I don't think so.
This can happen if the protocol is not present in any binary. It's possible only if the protocol isn't referenced at compile time. But operations like type checks do reference the protocol, so it's not easy to observe this case.
Btw, it shouldn't happen while we have #2818 applied, but I would prefer not to rely upon it.

PR #2971, opened 2019-05-14
Worker.executeAfter() API for scheduling jobs on worker.

olonho:


runtime/src/main/cpp/Worker.cpp
```diff
@@ -157,32 +211,79 @@ class Worker {
...
+      return 0;
+    } else {
+      return job.executeAfter.whenExecute - now;
+    }
+  }
+
+  void waitForQueueLocked() {
+    while (queue_.size() == 0) {
+      KLong closestToRun = checkDelayedLocked();
+      if (closestToRun >= 0) {
```
SvyatoslavScherbina: Maybe avoid waiting if `closestToRun == 0`?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -234,16 +335,45 @@ class State {
...
+      auto it = workers_.find(id);
+      if (it == workers_.end()) return false;
+      worker = it->second;
+
+    }
+    Job job;
+    job.kind = JOB_EXECUTE_AFTER;
+    job.executeAfter.operation = CreateStablePointer(operation);
+    if (afterMicroseconds == 0) {
+      worker->putJob(job, false);
```
SvyatoslavScherbina: Shouldn't the entire function body be executed under `lock_`? What happens if someone stops the worker before this line?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -285,7 +415,7 @@ class State {
     struct timeval tv;
     struct timespec ts;
     gettimeofday(&tv, nullptr);
-    KLong nsDelta = millis * 1000000LL;
+    KLong nsDelta = millis * 10000000LL;
```
SvyatoslavScherbina: `10000000LL`? Seven zeroes?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -374,27 +504,46 @@ void* workerRoutine(void* argument) {
     ObjHolder resultHolder;
     while (true) {
       Job job = worker->getJob();
-      if (job.function == nullptr) {
+      if (job.kind == JOB_TERMINATE) {
+        if (job.terminationRequest.waitDelayed) {
+          if (worker->waitDelayed()) {
+            worker->putJob(job, false);
```
SvyatoslavScherbina: Is it possible that another job have been added by this moment? This would make worker execute this job despite it is already terminating.

runtime/src/main/cpp/Worker.cpp
olonho: Typo.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -144,10 +184,24 @@ class Worker {
...
+    // Cleanup jobs in the queue.
     for (auto job : queue_) {
-      DisposeStablePointer(job.argument);
-      job.future->cancelUnlocked();
+      switch (job.kind) {
+        case JOB_REGULAR:
+          DisposeStablePointer(job.regularJob.argument);
+          job.regularJob.future->cancelUnlocked();
+          break;
+        case JOB_EXECUTE_AFTER: {
```
SvyatoslavScherbina: Shouldn't we handle `delayed_` too?

runtime/src/main/cpp/Worker.cpp
olonho: Yeah, that's intended behavior. It people want forced termination - it is available.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -374,27 +504,46 @@ void* workerRoutine(void* argument) {
...
+      if (job.kind == JOB_EXECUTE_AFTER) {
+         ObjHolder operationHolder, dummyHolder;
+         KRef obj = DerefStablePointer(job.executeAfter.operation, operationHolder.slot());
+         try {
+           WorkerLaunchpad(obj, dummyHolder.slot());
+         } catch (ExceptionObjHolder& e) {
+           if (worker->errorReporting())
+             ReportUnhandledException(e.obj());
+         }
+         DisposeStablePointer(job.executeAfter.operation);
```
SvyatoslavScherbina: Why not perform this immediately after `DerefStablePointer`?

runtime/src/main/cpp/Worker.cpp
SvyatoslavScherbina: So whether a job added after `requestTermination()` is executed depends on presence of delayed jobs? Is this behaviour intended? Looks a bit confusing.

olonho: Well, I do not believe it's too problematic. when we start waiting for delayed jobs it's kind of natural that other requests are processed as well. Guess we can fix it if it will indeed become problematic.

runtime/src/main/cpp/Worker.cpp
olonho: Not sure if it's too important. It is important in other places where ownership is transferred, but here we kind of own (or share) this lambda anyway.

romanandreyvich:
This pull request has been merged to master on 14 may but still not available (in Kotlin 1.3.40). Do you have an idea, when it will be included in a release version?

benasher44:
Thanks for this! Was able to delete a bunch of code focused on keeping track of and cleaning up `Future`s to prevent leaks: https://github.com/Autodesk/coroutineworker/pull/21/files#diff-a1e65a339a700664f0635d6a4517042eR74

PR #2972, opened 2019-05-14
Draft commit for enabling GUI client lldb debugging

kpgalligan:
konan_lldb.py was primarily designed for cli but not really optimized for GUI. We did a version to run in Xcode, but it drifted a lot from the "stock" version. This commit has a version that is much closer to "stock", and should run as before on the command line, but if run by Xcode or AppCode, will optimize a bit for GUI needs. Comments to follow about decisions and where messy hacks should probably be cleaned up.

### DIFF
--- a/llvmDebugInfoC/src/scripts/konan_lldb.py
+++ b/llvmDebugInfoC/src/scripts/konan_lldb.py
@@ -23,12 +23,30 @@
 
 import lldb
 import struct
+import os
+
+def is_appcode():
+    return 'AppCode' in os.getenv('XPC_SERVICE_NAME', '')
+
+def is_xcode():
+    return 'Xcode' in os.getenv('CLIENT_TYPE', '')
 
 NULL = 'null'
+
+CLIENT_APPCODE = is_appcode()
+CLIENT_XCODE = is_xcode()
+VISUAL_DEBUGGER = CLIENT_APPCODE or CLIENT_XCODE
 
 def log(msg):
     if False:
         print(msg)
+
+def exelog(stmt):
+    if False:
+        f = open(os.getenv('HOME', '') + "/lldbexelog.txt", "a")
+        f.write(stmt)
+        f.write("\n")
+        f.close()
 
 def lldb_val_to_ptr(lldb_val):
     addr = lldb_val.GetValueAsUnsigned()
@@ -37,20 +55,16 @@ def lldb_val_to_ptr(lldb_val):
 
 def evaluate(expr):
     result = lldb.debugger.GetSelectedTarget().EvaluateExpression(expr, lldb.SBExpressionOptions())
-    log("{} => {}".format(expr, result))
+    evallog = "{} => {}".format(expr, result)
+    log(evallog)
+    exelog(evallog)
     return result
 
 def is_instance_of(addr, typeinfo):
     return evaluate("(bool)IsInstance({}, {})".format(addr, typeinfo)).GetValue() == "true"
 
-
-def is_string(value):
-    return is_instance_of(lldb_val_to_ptr(value), "theStringTypeInfo")
-
-
-def is_array(value):
-    return evaluate("(int)Konan_DebugIsArray({})".format(lldb_val_to_ptr(value))).unsigned == 1
-
+def is_string_or_array(value):
+    return evaluate("(bool)IsInstance({0}, theStringTypeInfo) ? 1 : ((int)Konan_DebugIsArray({0}) ? 2 : 0)".format(lldb_val_to_ptr(value))).unsigned
 
 def check_type_info(value):
     """This method checks self-referencing of pointer of first member of TypeInfo including case when object has an
@@ -63,9 +77,6 @@ def check_type_info(value):
 
 def type_info(value):
     return evaluate("*(void **)((uintptr_t)(*(void**){0:#x}) & ~0x3)".format(value.unsigned)).unsigned
-
-#
-# Some kind of forward declaration.
 
 
 __FACTORY = {}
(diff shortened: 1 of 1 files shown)


llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -23,34 +23,48 @@
 
 import lldb
 import struct
+import os
+
+def is_appcode():
+    return 'AppCode' in os.getenv('XPC_SERVICE_NAME', '')
```
kpgalligan: Hacky, but seems to work. 'XPC_SERVICE_NAME' has 'AppCode' in the string. Should probably be something more formal.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -23,34 +23,48 @@
...
 
+CLIENT_APPCODE = is_appcode()
+CLIENT_XCODE = is_xcode()
+VISUAL_DEBUGGER = CLIENT_APPCODE or CLIENT_XCODE
+
 def log(msg):
     if False:
         print(msg)
 
+def exelog(stmt):
```
kpgalligan: Separate log for executed statements. Cutting down on these is the primary optimization technique, and there may be many, so put in a file is easier to manage.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -23,34 +23,48 @@
...
 
-
-def is_string(value):
-    return is_instance_of(lldb_val_to_ptr(value), "theStringTypeInfo")
-
-
-def is_array(value):
-    return evaluate("(int)Konan_DebugIsArray({})".format(lldb_val_to_ptr(value))).unsigned == 1
-
+def is_string_or_array(value):
```
kpgalligan: Reduced evaluate calls. There are still several calls per-instance, and reducing them will improve performance.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -92,19 +103,20 @@ def kotlin_object_type_summary(lldb_val, internal_dict):
...
-    return __FACTORY['string'](lldb_val) if is_string(lldb_val) else __FACTORY['array'](lldb_val) if is_array(
-        lldb_val) else __FACTORY['object'](lldb_val)
-
+    soa = is_string_or_array(lldb_val)
+    return __FACTORY['string'](lldb_val) if soa == 1 else __FACTORY['array'](lldb_val) if soa == 2 \
+        else __FACTORY['object'](lldb_val)
 
 class KonanHelperProvider(lldb.SBSyntheticValueProvider):
-    def __init__(self, valobj):
+    def __init__(self, valobj, amString):
```
kpgalligan: Hacky. String should probably be its own class. The purpose here is to avoid the extra 'Konan_DebugGetFieldCount' call.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -231,7 +244,15 @@ def offset(self):
 
 class KonanObjectSyntheticProvider(KonanHelperProvider):
     def __init__(self, valobj, tip):
-        super(KonanObjectSyntheticProvider, self).__init__(valobj)
+        # Save an extra call into the process
+        if tip in SYNTHETIC_OBJECT_LAYOUT_CACHE:
```
kpgalligan: if tip in SYNTHETIC_OBJECT_LAYOUT_CACHE, we don't need to call 'Konan_DebugGetFieldCount'

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -241,7 +262,10 @@ def __init__(self, valobj, tip):
         else:
             log("TIP: {:#x} HIT".format(tip))
         self._children = SYNTHETIC_OBJECT_LAYOUT_CACHE[tip]
-        self._values = [self._read_value(index) for index in range(self._children_count)]
+        if VISUAL_DEBUGGER:
```
kpgalligan: Lazy loading values. This *may* not be very useful, but was ported from earlier version.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -265,16 +289,24 @@ def __none(iterable, f):
...
+            result = self._read_value(index)
+            self._values[index] = result
+        return result
+
 
     # TODO: fix cyclic structures stringification.
     def to_string(self):
-        return dict([(self._children[i].name(), self._deref_or_obj_summary(i)) for i in range(self._children_count)])
-
+        if VISUAL_DEBUGGER:
```
kpgalligan: This is the primary "optimization". Visual debugger really doesn't need the summary string. Can get rid of TODO comment.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -297,10 +339,17 @@ def get_child_index(self, name):
...
-        return self._values[index]
+        result = self._values[index]
+        if result is None:
+            result = self._read_value(index)
+            self._values[index] = result
+        return result
 
     def to_string(self):
-        return [self._deref_or_obj_summary(i) for i in range(self._children_count)]
+        if VISUAL_DEBUGGER:
```
kpgalligan: Same as object summary. This is much worse for arrays.

llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -284,10 +316,20 @@ def __init__(self, valobj):
...
-        self._values = [self._read_value(i) for i in range(self._children_count)]
+        if VISUAL_DEBUGGER:
+            self._values = [None for index in range(self.cap_children_count())]
+        else:
+            self._values = [self._read_value(i) for i in range(self.cap_children_count())]
+
+    def cap_children_count(self):
+        #AppCode limits array len automatically
+        if CLIENT_XCODE:
+            return min(self._children_count, 20)
```
kpgalligan: Xcode will just show you all values in an array, so for example if you have 100k, you'll have problems. AppCode shows in increments of 50, which actually works well without modification.

olonho:
Any updates on this change? Kevin, do you need a review?

tylerwilson:
I tested this on my local machine with AppCode, and it made a *huge* difference in speed. From unusable to usable. Still some jumping around issues, but as mentioned in the Slack channel likely due to debug info not being as precise as needed. 👍

vvlevchenko:
thank for cooperation. Change was bit enhanced and rethought and finally applied in #3561

PR #2973, opened 2019-05-14
String <-> CharArray, ByteArray conversions

ilya-g:
This is another take on PR #2950 with the following differences:
- Kotlin compiler is updated to 1.3.50-dev-44
- correct experimental stdlib api opt-in in tests
- shorter copyright headers
- bound checks in CPP implementations are restored because we need to examine their usages in Kotlin/Native codebase more thoroughly.


ilya-g:
cc @vvlevchenko @qurbonzoda

PR #2981, opened 2019-05-16
Add more targets in echoServer sample

ddolovov:


samples/echoServer/build.gradle
```diff
@@ -17,3 +17,28 @@ kotlin {
 MPPTools.createRunTask(project, 'runProgram', kotlin.targets.echoServer) {
     args 3000
 }
+
+private def determinePreset() {
```
olonho: We shall build for both host and few cross targets, like it's done in `samples/androidNativeActivity`

ddolovov: ok. I will rework.

PR #2984, opened 2019-05-17
[build] split configuration for compiler and stdlib

vvlevchenko:


build.gradle
```diff
@@ -44,7 +44,7 @@ ext {
 
     kotlinCommonStdlibModule="org.jetbrains.kotlin:kotlin-stdlib-common:${kotlinVersion}:sources"
     kotlinCompilerModule="org.jetbrains.kotlin:kotlin-compiler:${kotlinVersion}"
-    kotlinStdLibModule="org.jetbrains.kotlin:kotlin-stdlib:${kotlinVersion}"
+    kotlinStdLibModule="org.jetbrains.kotlin:kotlin-stdlib:${kotlinStdlibVersion}"
     kotlinTestCommonModule="org.jetbrains.kotlin:kotlin-test-common:${kotlinVersion}:sources"
```
ilya-g: Please take `kotlin-test-annotations-common`, `kotlin-test-common` by `kotlinStdlibVersion` too

build.gradle
```diff
@@ -44,7 +44,7 @@ ext {
 
     kotlinCommonStdlibModule="org.jetbrains.kotlin:kotlin-stdlib-common:${kotlinVersion}:sources"
     kotlinCompilerModule="org.jetbrains.kotlin:kotlin-compiler:${kotlinVersion}"
-    kotlinStdLibModule="org.jetbrains.kotlin:kotlin-stdlib:${kotlinVersion}"
```
ilya-g: I suppose it should be `kotlinCommonStdlibModule` _instead of_ `kotlinStdLibModule`. The former is used to build native runtime and the latter is used as a dependency for compiler.

build.gradle
```diff
@@ -42,11 +42,11 @@ ext {
     platformManager = new PlatformManager(DistributionKt.buildDistribution(projectDir.absolutePath),
             ext.experimentalEnabled)
 
-    kotlinCommonStdlibModule="org.jetbrains.kotlin:kotlin-stdlib-common:${kotlinVersion}:sources"
     kotlinCompilerModule="org.jetbrains.kotlin:kotlin-compiler:${kotlinVersion}"
-    kotlinStdLibModule="org.jetbrains.kotlin:kotlin-stdlib:${kotlinVersion}"
-    kotlinTestCommonModule="org.jetbrains.kotlin:kotlin-test-common:${kotlinVersion}:sources"
-    kotlinTestAnnotationsCommonModule="org.jetbrains.kotlin:kotlin-test-annotations-common:${kotlinVersion}:sources"
+    kotlinStdLibModule="org.jetbrains.kotlin:kotlin-stdlib:${kotlinStdlibVersion}"
```
ilya-g: If `kotlinStdlibModule` is a dependency only for compiler, it's better to take the same version of it as the compiler, i.e. `kotlinVersion`.

backend.native/tests/build.gradle
```diff
@@ -114,36 +120,38 @@ def update_external_tests() {
...
+            exclude 'stdlib/**'
         }
     }
 
     delete(externalStdlibTestsDir)
     externalStdlibTestsDir.mkdirs()
+    def stdlibTestsFiles = configurations.update_stdlib_tests.asFileTree
     copy {
-        files.each {
+        compilerTestsFiles.each {
```
ilya-g: stdlibTestsFiles?

gradle.properties
```diff
@@ -18,9 +18,12 @@
...
 remoteRoot=konan_tests
-kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-259,branch:default:true,pinned:true/artifacts/content/maven
-kotlinVersion=1.3.50-dev-259
-testKotlinVersion=1.3.50-dev-259
+kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-267,branch:default:true,pinned:true/artifacts/content/maven
+kotlinVersion=1.3.50-dev-267
+kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-280,branch:default:true,pinned:false/artifacts/content/maven
+kotlinStdlibVersion=1.3.50-dev-280
+testKotlinCompilerVersion=1.3.50-dev-267
+testKotlinStdlibVersion=1.3.50-dev-280
```
ilya-g: I believe after this change we'll always be able to use the same version of tests as the version of common stdlib, so we can join `kotlinStdlibVersion` and `testKotlinStdlibVersoin`. But if you have considerations why it still might be needed, then keep it separate.

PR #2990, opened 2019-05-20
Implement ObjCExportLazy

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportLazy.kt
```diff
@@ -0,0 +1,403 @@
+package org.jetbrains.kotlin.backend.konan.objcexport
```
olonho: Missing copyright.

PR #3004, opened 2019-05-22
Objective c interop benchmarks migration

LepilkinaElena:
Old one https://github.com/JetBrains/kotlin-native/pull/2852


performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -19,12 +19,17 @@ package org.jetbrains.benchmarksLauncher
...
 import org.jetbrains.report.BenchmarkResult
 import org.jetbrains.kliopt.*
+import kotlin.reflect.KFunction0
 
 abstract class Launcher(val numWarmIterations: Int, val numberOfAttempts: Int, val prefix: String = "") {
     class Results(val mean: Double, val variance: Double)
 
     abstract val benchmarks: BenchmarksCollection
 
+    fun add(name: String, benchmark: KFunction0<Any>) {
```
SvyatoslavScherbina: Please use function type `() -> Any?` here.
Passing Swift function references as `KFunction0<Any>` is incorrect but not detected at runtime due to lack of runtime type checks in Kotlin/Native.

LepilkinaElena: It works such way. Or could there be other side effects?

SvyatoslavScherbina: > It works such way.

Yes, because Kotlin/Native lacks required runtime checks here.

> Or could there be other side effects?

Yes. Calling any of `KFunction` APIs on such an object would fail in an unpredictable way.

PR #3007, opened 2019-05-23
API for toFrozen operation.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,56 @@ void FreezeSubgraph(ObjHeader* root) {
...
+    AddHeapRef(object);
+    return;
+  }
+  auto substitution = translationMap->find(object);
+  if (substitution != translationMap->end()) {
+    *location = substitution->second;
+    AddHeapRef(*location);
+  } else {
+    ObjHolder result;
+    auto* clone = AllocInstance(object->type_info(), result.slot());
```
SvyatoslavScherbina: IIRC, we are considering making this part opt-in, i.e. enabled only for certain classes.
Do you plan to change this later thus breaking the behaviour if we decide in favor of opt-in?

olonho: Yeah, we just haven't decided how to distinguish such types, so idea here is to try how fitting the solution is.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,56 @@ void FreezeSubgraph(ObjHeader* root) {
...
+    SetHeapRef(location, clone);
+    deepFrozenCopyTo(clone, object, translationMap);
+  }
+}
+
+void deepFrozenCopyTo(ObjHeader* to, ObjHeader* from, ObjHeaderMap* translationMap) {
+   (*translationMap)[from] = to;
+   ::memcpy(to, from, objectSize(from));
```
olonho: This behavior has a bug, if GC or exception is thrown in the middle of copying, as half-copied object is incorrect GC wise, need to come up with the better mechanism here.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,56 @@ void FreezeSubgraph(ObjHeader* root) {
...
+       changeToCopyIfNeeded(ArrayAddressOfElementAt(array, index), translationMap);
+     }
+   }
+}
+
+OBJ_GETTER(ToFrozenForm, ObjHeader* object) {
+  RuntimeAssert(object != nullptr && !PermanentOrFrozen(object), "Must be mutable object");
+  ObjHolder result;
+  ObjHeaderMap translationMap;
+  auto* clone = AllocInstance(object->type_info(), result.slot());
```
olonho: We may have an Array here.

backend.native/tests/runtime/workers/freeze1.kt
```diff
@@ -11,6 +11,15 @@ import kotlin.native.concurrent.*
 
 data class Node(var previous: Node?, var data: Int)
 
+fun makeAcyclic(count: Int): Node {
```
olonho: Use that in test.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,96 @@ void FreezeSubgraph(ObjHeader* root) {
 #endif
 }
 
+void deepFrozenCopyTo(ObjHeader* to, ObjHeader* from, ObjHeaderMap* translationMapm, ObjHeader** firstBlocker);
```
homuroll: `translationMap**m**` - a typo?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,96 @@ void FreezeSubgraph(ObjHeader* root) {
...
+
+bool cannotClone(ObjHeader* original) {
+  if (original == nullptr)
+    return false;
+  if ((original->type_info()->flags_ & TF_NON_CLONEABLE) != 0)
+    return true;
+  if (original->has_meta_object() && ((original->meta_object()->flags_ & MF_NEVER_FROZEN) != 0))
+    return true;
+  return false;
```
SvyatoslavScherbina: It's also not correct to clone Kotlin object with non-trivial Objective-C counterpart (e.g. for instance of Objective-C subclass of Kotlin class).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,96 @@ void FreezeSubgraph(ObjHeader* root) {
...
 
+void deepFrozenCopyTo(ObjHeader* to, ObjHeader* from, ObjHeaderMap* translationMapm, ObjHeader** firstBlocker);
+
+bool cannotClone(ObjHeader* original) {
+  if (original == nullptr)
+    return false;
+  if ((original->type_info()->flags_ & TF_NON_CLONEABLE) != 0)
+    return true;
+  if (original->has_meta_object() && ((original->meta_object()->flags_ & MF_NEVER_FROZEN) != 0))
+    return true;
```
SvyatoslavScherbina: Not sure if this behaviour is desirable. We can provide `ensureNevenCopied` and `MF_NEVER_COPIED` instead.

runtime/src/main/cpp/TypeInfo.h
```diff
@@ -54,7 +54,8 @@ enum Konan_RuntimeType {
 enum Konan_TypeFlags {
   TF_IMMUTABLE = 1 << 0,
   TF_ACYCLIC   = 1 << 1,
-  TF_INTERFACE = 1 << 2
+  TF_INTERFACE = 1 << 2,
+  TF_NON_CLONEABLE = 1 << 3
```
SvyatoslavScherbina: Why not `TF_CLONEABLE`?
`TF_NON_CLONEABLE` looks somewhat confusing, especially with opt-in approach.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,96 @@ void FreezeSubgraph(ObjHeader* root) {
...
+    ::memcpy(clone->array() + 1, original->array() + 1, objectSize(original) - sizeof(ArrayHeader));
+  } else {
+    clone = AllocInstance(original->type_info(), OBJ_RESULT);
+    // Copy content only.
+    ::memcpy(clone + 1, original + 1, objectSize(original) - sizeof(ObjHeader));
+  }
+  RETURN_OBJ(clone);
+}
+
+void changeToCopyIfNeeded(ObjHeader** location, ObjHeaderMap* translationMap, ObjHeader** firstBlocker) {
```
SvyatoslavScherbina: The name doesn't reflect the semantics with respect to all the complexities. Maybe `patchRawCopiedReference`?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2252,6 +2253,96 @@ void FreezeSubgraph(ObjHeader* root) {
...
+  if (cannotClone(object)) {
+    firstBlocker = object;
+  } else {
+    memoryState->gcSuspendCount++;
+    clone = makeShallowClone(object, result.slot());
+    deepFrozenCopyTo(clone, object, &translationMap, &firstBlocker);
+    memoryState->gcSuspendCount--;
+  }
+  if (firstBlocker != nullptr) {
+    ThrowFreezingException(object, firstBlocker);
```
SvyatoslavScherbina: This exception would be a bit confusing. It should report "unable to copy" instead of "freezing has failed".

runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt
```diff
@@ -42,12 +42,25 @@ public fun <T> T.freeze(): T {
...
+/**
+ * Transforms an object to the frozen form by creating frozen version of the object, by either
+ * deep copy of mutable objects and reusing frozen references. Note, that consequent calls
+ * of this function may return different instances.
+ *
+ * @return the frozen object form
+ * @see ensureNeverFrozen
+ */
+public inline fun <reified T> T.toFrozen(): T =
+        @Suppress("UNCHECKED_CAST", "NON_PUBLIC_CALL_FROM_PUBLIC_INLINE")
```
SvyatoslavScherbina: `@PublishedApi` can be added to `toFrozenInternal` to avoid unreliable suppression of compiler error.

runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt
```diff
@@ -42,12 +42,25 @@ public fun <T> T.freeze(): T {
 public val Any?.isFrozen
     get() = isFrozenInternal(this)
 
+
+/**
+ * Transforms an object to the frozen form by creating frozen version of the object, by either
```
SvyatoslavScherbina: Is "either" required here?

runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt
```diff
@@ -42,12 +42,25 @@ public fun <T> T.freeze(): T {
...
+ * deep copy of mutable objects and reusing frozen references. Note, that consequent calls
+ * of this function may return different instances.
+ *
+ * @return the frozen object form
+ * @see ensureNeverFrozen
+ */
+public inline fun <reified T> T.toFrozen(): T =
```
SvyatoslavScherbina: Maybe add a note to `freeze` documentation about its side effects with reference to `toFrozen` as a safer alternative?

runtime/src/main/cpp/Memory.cpp
olonho: Not sure if it's dramatically better.

runtime/src/main/cpp/Memory.cpp
olonho: Hard to tell.

runtime/src/main/kotlin/kotlin/Annotations.kt
```diff
@@ -148,4 +148,14 @@ public annotation class DslMarker
...
-public annotation class PublishedApi
\ No newline at end of file
+public annotation class PublishedApi
+
+/**
+ * When applied to class marks the fact, that its instances could be copied by value with [toFrozen()]
+ * function by cloning object instance, and so when used in models could be safely cloned.
+ * Data classes are assumed to be shared by value by default.
+ */
+// TODO: how to properly mark as experimental?
```
SvyatoslavScherbina: > how to properly mark as experimental?

Like this:

https://github.com/JetBrains/kotlin-native/blob/0f6859f29c43fc0f55430fbd6014d760400e066b/runtime/src/main/kotlin/kotlin/reflect/AssociatedObjects.kt#L19

runtime/src/main/kotlin/kotlin/Array.kt
```diff
@@ -17,6 +17,7 @@ import kotlin.native.internal.PointsTo
  * for more information on arrays.
  */
 @ExportTypeInfo("theArrayTypeInfo")
+@ShareByValue
```
SvyatoslavScherbina: Array classes don't have custom `equals`/`hashCode`, so this doesn't seem correct according to our discussion.

runtime/src/main/kotlin/kotlin/collections/HashMap.kt
```diff
@@ -7,6 +7,7 @@ package kotlin.collections
 
 import kotlin.native.concurrent.isFrozen
 
+@ShareByValue
```
SvyatoslavScherbina: This class doesn't handle the case when a copied key's hash code is different from the original one's.

runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt
```diff
@@ -42,12 +46,37 @@ public fun <T> T.freeze(): T {
 public val Any?.isFrozen
     get() = isFrozenInternal(this)
 
+
+/**
+ * Transforms an object to the frozen form by creating frozen version of the object, by
+ * deep copy of mutable objects and reusing frozen references. Note, that consequent calls
+ * of this function may return different instances. To make a clone class must be eitehr data class or
```
SvyatoslavScherbina: either

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -80,6 +80,14 @@ internal fun ThrowIncorrectDereferenceException() {
             "Trying to access top level value not marked as @ThreadLocal or @SharedImmutable from non-main thread")
 }
 
+@ExportForCppRuntime
+internal fun ThrowCannotCloneException(original: Any, blocker: Any): Nothing {
+    val originalName = original::class.qualifiedName ?: original::class.simpleName ?: "<object>"
+    val blockerName = blocker::class.qualifiedName ?: blocker::class.simpleName ?: "<object>"
+    throw IllegalArgumentException(
+            "Error trying to clone $original ($originalName) because $blocker ($blockerName) cannot be cloned")
```
SvyatoslavScherbina: The operation applied to `original` is not `clone`, it is (currently) `toFrozen`.
I believe the exception message should reflect that.

runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt
```diff
@@ -42,12 +46,37 @@ public fun <T> T.freeze(): T {
...
+ * explicitly marked by [kotlin.ShareByValue] annotation.
+ *
+ * @return the frozen object form
+ * @see ensureNeverFrozen
+ * @see ensureNeverCloned
+ * @see kotlin.ShareByValue
+ */
+public inline fun <reified T> T.toFrozen(): T =
```
Takhion: Have you considered naming this something like `copyFrozen()`, since users are used to the `copy()` function in data classes? I think it also conveys better how this is different from `freeze()`.

Takhion: > ```text
> To make a clone class must be either data class or explicitly marked by [kotlin.ShareByValue] annotation.
> ```

What if this was generated _only_ for the types that support it?
Either directly on the types (like the `copy()` function in data classes), or by having the types implement a special internal marker interface? The second option would actually be ideal, since it would allow targeting any cloneable type!

SvyatoslavScherbina:
Abandoned.

PR #3016, opened 2019-05-26
Perform runtime patching of signal handlers.

olonho:


Interop/Indexer/build.gradle
```diff
@@ -76,7 +76,7 @@ model {
         clangstubs(NativeLibrarySpec) {
             sources {
                 c.source.srcDir 'prebuilt/nativeInteropStubs/c'
-		        cpp.source.srcDir 'prebuilt/nativeInteropStubs/cpp'
+                cpp.source.srcDir 'prebuilt/nativeInteropStubs/cpp'
```
SvyatoslavScherbina: `signalChaining.cpp` is not a prebuilt stub, so it should be located in other directory.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+#if defined(__linux__)
+#include <link.h>
+#endif
+
+typedef void (*toggleCrashRecovery_t)(int);
+typedef int (*sigaction_t)(int, const struct sigaction *, struct sigaction *);
+
+static sigaction_t realSigaction = 0;
+
+static int mySigaction(int sig, const struct sigaction *act, struct sigaction * oact) {
```
SvyatoslavScherbina: Why not use `libjsig`?

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+  return realSigaction(sig, act, oact);
+}
+
+__attribute__((constructor))
+static void initSignalChaining() {
+  toggleCrashRecovery_t clang_sym = 0;
+  void** base = 0;
+  Dl_info info;
+
+  clang_sym = (toggleCrashRecovery_t)dlsym(RTLD_DEFAULT, "clang_toggleCrashRecovery");
```
SvyatoslavScherbina: IIRC, this code is linked with `libclang` and can use its headers, so `clang_toggleCrashRecovery` can be accessed directly.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+#endif
+
+typedef void (*toggleCrashRecovery_t)(int);
+typedef int (*sigaction_t)(int, const struct sigaction *, struct sigaction *);
+
+static sigaction_t realSigaction = 0;
+
+static int mySigaction(int sig, const struct sigaction *act, struct sigaction * oact) {
+  if (sig == SIGILL || sig == SIGFPE|| sig ==  SIGSEGV || sig == SIGBUS || sig == SIGUSR1|| sig == SIGUSR2) return 0;
+  return realSigaction(sig, act, oact);
```
SvyatoslavScherbina: What's the purpose of allowing libclang to handle other signals?
Seems unreliable because JVM can eventually rely on other signals too.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+  // And then patch GOT.
+#if defined(__linux__)
+  {
+    // On Linux we have to be a bit tricky, as there's unmapped gap between code and GOT.
+     struct link_map* linkmap = 0;
+     if (dladdr1((void*)clang_sym, &info, (void**)&linkmap, RTLD_DL_LINKMAP) == 0) return;
+     base = (void**)linkmap->l_ld;
+  }
+#endif
+  realSigaction = (sigaction_t)dlsym(RTLD_DEFAULT, "sigaction");
```
SvyatoslavScherbina: Why not access it directly?

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+  for (int index = 0, patched = 0; patched < 1; index++) {
+    void* value = base[index];
+    if (value == realSigaction) {
+        base[index] = (void*)mySigaction;
+        patched++;
+    }
+    if (value == mySigaction) {
```
SvyatoslavScherbina: Is this required?

libclangext/src/main/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
+#if defined(__linux__) || defined(__APPLE__)
```
SvyatoslavScherbina: Can we avoid adding the same file twice?

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
```diff
@@ -0,0 +1,58 @@
...
+        patched++;
+    }
+  }
+}
+#endif // defined(__linux__) || defined(__APPLE__)
```
SvyatoslavScherbina: It makes sense to add some kind of runtime validator for this trick.
E.g. after processing C library check that signal handlers for `SIGSEGV` and others are not from `libclang` or `libclangstubs`. This will help us detect if this tricks stops working properly after libclang update.

libclangext/src/main/cpp/signalChaining.cpp
SvyatoslavScherbina: Btw, how does this help?
As far as I see, both copies should get linked into the same `libclangstubs` binary.

Interop/Indexer/build.gradle
olonho: Which one?

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: It makes no sense. If libjsig is not preloaded, JVM doesn't know about it. Also this way it's much simpler and has no path assumptions.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: Yeah,, but it doesn't change much.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: Mostly for termination handlers. Remember, race comes from the high frequency signal used for safepointing, generally clang chains OKish. Which signals you have in mind.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: No particular reason.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: Only to make patching idempotent.

libclangext/src/main/cpp/signalChaining.cpp
olonho: Yeah, if we'll decide where this code goes. Suggestions?

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: Not sure where to put such a check. If we reach custom sigaction handler - trick works. Otherwise, we can only get control in ctor/dtor and JNI methods. 
Suggestions?

Interop/Indexer/build.gradle
SvyatoslavScherbina: E.g. `src/nativeInteropStubs/cpp`

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
SvyatoslavScherbina: > Which signals you have in mind.

Any other signal JVM may find useful to handle someday.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
SvyatoslavScherbina: It can make this code simpler.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
SvyatoslavScherbina: > E.g. after processing C library

I mean after cinterop completes processing C library headers.
E.g. to `processCLib` epilogue.

libclangext/src/main/cpp/signalChaining.cpp
SvyatoslavScherbina: `Interop/Indexer`.
Having one copy is worth keeping it in somewhat unintuitive location.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: This part doesn't change for decade and the half, and if it will happen - we'll notice :).

Interop/Indexer/build.gradle
olonho: BTW, seems `disable-abi-checks.cpp` is also not generated.

Interop/Indexer/prebuilt/nativeInteropStubs/cpp/signalChaining.cpp
olonho: Well, this way we guarantee that symbol found by the default lookup is in dynamic library we want it to be. Also I kind of like this code being pretty universal wrt symbols it patches.

libclangext/src/main/cpp/signalChaining.cpp
olonho: Ah, OK, maybe I was overly cautious here.

SvyatoslavScherbina: Btw, we still may have to patch both `libclang` and `libclangstubs` (containing `clangext`).

PR #3020, opened 2019-05-27
Avoid recursive runtime init bug.

olonho:


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -81,14 +82,15 @@ RuntimeState* initRuntime() {
   if (!result) return nullptr;
   result->memoryState = InitMemory();
   bool firstRuntime = atomicAdd(&aliveRuntimesCount, 1) == 1;
+  runtimeInitInProgress = true;
```
SvyatoslavScherbina: Although this indeed fixes the bug, this code now looks much like requiring some refactoring: not having `::runtimeState` initialized during legitimate Kotlin code execution is somewhat confusing.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -116,7 +118,7 @@ void AppendToInitializersTail(InitNode *next) {
 }
 
 void Kotlin_initRuntimeIfNeeded() {
-  if (runtimeState == nullptr) {
+  if (runtimeState == nullptr && !runtimeInitInProgress) {
```
SvyatoslavScherbina: It makes sense to add a test for this.

PR #3021, opened 2019-05-27
Rewrite samples using new API

LepilkinaElena:


samples/curl/build.gradle
```diff
@@ -12,14 +12,16 @@ repositories {
...
-                // Add lib path to `libcurl` and its dependencies:
-                compilations.main.linkerOpts "-L${MPPTools.mingwPath()}/lib"
+    targetFromPreset(hostPreset, 'curl') {
+        binaries {
+            executable() {
+                entryPoint = 'sample.curl.main'
+                if (presets.mingwX64 == hostPreset) {
+                    // Add lib path to `libcurl` and its dependencies:
+                    linkerOpts += ["-L${MPPTools.mingwPath()}/lib".toString()]
+                    runTask.environment('PATH': "${MPPTools.mingwPath()}/bin".toString())
```
olonho: Why toString on String?

LepilkinaElena: This is Groovy String, but not Java String, so Kotlin has problems6 because there is no such type.

PR #3028, opened 2019-05-28
Actuals for MonoClock, DurationUnit and Duration formatting

ilya-g:
The corresponding common API review: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3003

### DIFF
--- a/runtime/src/main/cpp/ToString.cpp
+++ b/runtime/src/main/cpp/ToString.cpp
@@ -108,4 +108,16 @@ OBJ_GETTER(Kotlin_Long_toStringRadix, KLong value, KInt radix) {
   RETURN_RESULT_OF(Kotlin_toStringRadix<KLong>, value, radix)
 }
 
+OBJ_GETTER(Kotlin_DurationValue_formatToExactDecimals, KDouble value, KInt decimals) {
+  char cstring[32];
+  konan::snprintf(cstring, sizeof(cstring), "%.*f", decimals, value);
+  RETURN_RESULT_OF(CreateStringFromCString, cstring)
+}
+
+OBJ_GETTER(Kotlin_DurationValue_formatScientificImpl, KDouble value) {
+  char cstring[16];
+  konan::snprintf(cstring, sizeof(cstring), "%.2e", value);
+  RETURN_RESULT_OF(CreateStringFromCString, cstring)
+}
+
 } // extern "C"
--- a/runtime/src/main/kotlin/kotlin/time/MonoClock.kt
+++ b/runtime/src/main/kotlin/kotlin/time/MonoClock.kt
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package kotlin.time
+
+import kotlin.system.*
+
+@SinceKotlin("1.3")
+@ExperimentalTime
+public actual object MonoClock : AbstractLongClock(unit = DurationUnit.NANOSECONDS), Clock { // TODO: interface should not be required here
+    override fun read(): Long = getTimeNanos()
+    override fun toString(): String = "Clock(nanoTime)"
+}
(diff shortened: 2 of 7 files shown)


runtime/src/main/kotlin/kotlin/time/MonoClock.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+package kotlin.time
+
+import kotlin.system.*
+
+public actual object MonoClock : LongReadingClock(unit = DurationUnit.NANOSECONDS), Clock { // TODO: interface should not be required here
```
olonho: Indeed, API looks unnecessary object oriented.

runtime/src/main/kotlin/kotlin/time/formatToDecimals.kt
```diff
@@ -0,0 +1,21 @@
...
+
+internal actual fun formatUpToDecimals(value: Double, decimals: Int): String {
+    return formatToExactDecimals(value, decimals).trimEnd('0')
+}
+
+@SymbolName("Kotlin_Double_formatScientificImpl")
+internal external fun formatScientificImpl(value: Double): String
+
+internal actual fun formatScientific(value: Double): String {
+    return formatScientificImpl(value).replace("e+0", "e").replace("e+", "e")
```
olonho: Oh, looks ugly.

runtime/src/main/cpp/ToString.cpp
```diff
@@ -108,4 +108,16 @@ OBJ_GETTER(Kotlin_Long_toStringRadix, KLong value, KInt radix) {
   RETURN_RESULT_OF(Kotlin_toStringRadix<KLong>, value, radix)
 }
 
+OBJ_GETTER(Kotlin_Double_formatToExactDecimals, KDouble value, KInt decimals) {
+  char cstring[16];
```
olonho: Is 16 indeed always enough?

ilya-g: Yes, for large values we use the scientific format.

runtime/src/main/kotlin/kotlin/time/MonoClock.kt
ilya-g: Interface should not be required, because it's already implemented, but still has to be mentioned to make MPP expect-actual matcher happy.

runtime/src/main/kotlin/kotlin/time/formatToDecimals.kt
ilya-g: The easiest way to make tests pass with the given API ¯\\\_(ツ)\_/¯

olonho: Probably it shall be somehow rethought, maybe by not using Double here at all.

runtime/src/main/kotlin/kotlin/time/MonoClock.kt
olonho: I somewhat dislike using unneeded `object` on fast path (i.e. timestamps), where we pay init check every access. Could you please redesign it to be a top level function?

runtime/src/main/cpp/ToString.cpp
olonho: https://stackoverflow.com/questions/9999221/double-precision-decimal-places so considering zero-terminator and dot it seems to be insufficient.

ilya-g: This function is only called for values < 1e7 and decimals <= 3

ilya-g: I've renamed it to avoid impression of a general purpose number formatting function.

ilya-g: Actually it's called on numbers of any magnitude and with any non-negative decimals. I'll think what would be better to do here.

ilya-g: I've increased the buffer size to 32 chars: this should be enough with regard to the coming limitations on the function usage:
- decimals will be in range 0..12
- value will be less than 1e+14

runtime/src/main/kotlin/kotlin/time/formatToDecimals.kt
ilya-g: Now we only remove `0` after `e+`, but leave plus.

runtime/src/main/kotlin/kotlin/time/MonoClock.kt
ilya-g: `ClockMark` is now an abstract class instead of an interface, so it should perform better on repeated measuring.

PR #3036, opened 2019-05-30
Add kotlinx.cinterop.getOriginalKotlinClass(objCClass/objCProtocol)

SvyatoslavScherbina:
Enables getting KClass for Kotlin classes produced to Objective-C header.

Also add other minor improvements.

### DIFF
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
@@ -490,6 +490,11 @@ abstract class ObjCContainerStub(stubGenerator: StubGenerator,
             supers.add(classifier.type)
         }
 
+        if (!isMeta && container.isProtocolClass()) {
+            // TODO: map Protocol type to ObjCProtocol instead.
+            supers.add(KotlinTypes.objCProtocol.type)
+        }
+
         val keywords = when (container) {
             is ObjCClass -> "open class"
             is ObjCProtocol -> "interface"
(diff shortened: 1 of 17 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ObjCInterop.kt
```diff
@@ -61,6 +63,9 @@ fun IrClass.isExternalObjCClass(): Boolean = this.isObjCClass() &&
             it.descriptor.annotations.hasAnnotation(externalObjCClassFqName)
         }
 
+fun ClassDescriptor.isObjCForwardDeclaration(): Boolean =
+        this.findPackage().fqName.startsWith(objcnamesForwardDeclarationsPackageName)
```
olonho: Maybe check that it starts with `objcnames.`.

SvyatoslavScherbina: This is exactly what is done here.

SvyatoslavScherbina: `FqName.startsWith()` takes dots into account properly.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -690,3 +695,8 @@ private fun genProtocolGetter(
 
     return functionName
 }
+
+private fun ObjCClassOrProtocol.isProtocolClass(): Boolean = when (this) {
+    is ObjCClass -> (name == "Protocol" || binaryName == "Protocol")
+    is ObjCProtocol -> false
```
olonho: Interestingly, so what does it actually tests for?

SvyatoslavScherbina: That this is `Protocol` class (not the same as protocol).
`@protocol(Foo)` has `Protocol*` type.

PR #3038, opened 2019-05-30
Enforce Xcode version when building dist

SvyatoslavScherbina:


SvyatoslavScherbina:
> Why do you believe it's needed?

To ensure that distribution is built with proper Xcode version.

SvyatoslavScherbina:
I believe it is critically important to build the distribution with reference Xcode version, because we test mostly with this version and the distribution may actually differ in an unpredictable way when built with other Xcode.

This check helps to enforce this requirement, without the check it's possible that the requirement is not met when building the release bundle and this issue remains unnoticed.

We had the equivalent check before, but [it was disabled to avoid unwanted dependencies](https://github.com/JetBrains/kotlin-native-shared/pull/6).

Also if we decide to disable [the internal server availability heuristic](https://github.com/JetBrains/kotlin-native-shared/blob/0458542c4880a5589d090c0eca56386527e854dc/src/main/kotlin/org/jetbrains/kotlin/konan/util/DependencyProcessor.kt#L296), then having this check will become even more desirable since the new possibility of having build misconfigured will appear.

SvyatoslavScherbina:
Appears to be implemented in https://github.com/JetBrains/kotlin-native/pull/3543

PR #3045, opened 2019-06-03
Avoid using internal server DNS-based heuristic by default

SvyatoslavScherbina:
As it appears to be slow in some configurations


SvyatoslavScherbina:
Still kinda dangerous without #3038.

PR #3048, opened 2019-06-04
GitChurn expand

artdfel:
For now, the gitchurn sample does not measure file changes frequency but provide info about every contributor amount of file changes inside every directory he touched. Also, some lib method calls from the old version were separated to the GitRevwalk class.

### DIFF
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
@@ -66,17 +66,10 @@ private fun calculateChurn(workDir: String, limit: Int) {
         println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
         var changedFilesSum = 0
         listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
-            if(modification.filePath.contains("/")){
-                print("Dir:  ")
-            }else{
-                print("File: ")
-            }
-            println(modification.filePath)
-            println("      $counter")
-            println()
+            print(" ${modification.filePath}:  $counter \n")
             changedFilesSum += counter
         }
-        println("Made $changedFilesSum modifications in total")
+        println("\n Made $changedFilesSum modifications in total")
         println("________________________")
     }
 
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
@@ -65,8 +65,14 @@ private fun calculateChurn(workDir: String, limit: Int) {
     modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
         println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
         var changedFilesSum = 0
+        var flag = 1
         listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
-            print(" ${modification.filePath}:  $counter \n")
+            print(" ${modification.filePath}:  $counter \t")
+            if (flag > 3){
+                print("\n")
+                flag = 0
+            }
+            flag++
             changedFilesSum += counter
         }
         println("\n Made $changedFilesSum modifications in total")
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
@@ -65,15 +65,12 @@ private fun calculateChurn(workDir: String, limit: Int) {
     modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
         println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
         var changedFilesSum = 0
-        var flag = 1
-        listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
-            print(" ${modification.filePath}:  $counter \t")
-            if (flag > 3){
-                print("\n")
-                flag = 0
-            }
-            flag++
-            changedFilesSum += counter
+        listOfAuthorsChanges.sortedByDescending { it.second }.chunked( 4).forEach { chunk ->
+                chunk.forEach{ (modification, counter) ->
+                    print(" ${modification.filePath}:  $counter \t")
+                    changedFilesSum += counter
+                }
+            print("\n")
         }
         println("\n Made $changedFilesSum modifications in total")
         println("________________________")
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
@@ -35,21 +35,25 @@ fun main(args: Array<String>) {
 private fun calculateChurn(workDir: String, limit: Int) {
     println("Opening…")
     val repository = git.repository(workDir)
-    val map = mutableMapOf<String, Int>()
+    data class WhoAndWhere(val authorEmail: String = "", val filePath: String = "")
+    val modificationsByAuthor = mutableMapOf<WhoAndWhere, Int>()
     var count = 0
-    val commits = repository.commits()
-    val limited = commits.take(limit)
+    val commits = repository.commits().take(limit)
     println("Calculating…")
-    limited.forEach { commit ->
+    commits.forEach { commit ->
         if (count % 100 == 0)
-            println("Commit #$count [${commit.time.format()}]: ${commit.summary}")
+            println("Commit #$count [${commit.time.format()}] by ${commit.author.name!!.toKString()}: ${commit.summary}")
 
         commit.parents.forEach { parent ->
             val diff = commit.tree.diff(parent.tree)
             diff.deltas().forEach { delta ->
-                val path = delta.newPath
-                val n = map[path] ?: 0
-                map.put(path, n + 1)
+                val path = if (delta.newPath == delta.newPath.substringBefore('/'))
+                        delta.newPath
+                    else
+                        delta.newPath.substringBefore('/') + "/"
+                val key = WhoAndWhere(commit.author.email!!.toKString().toLowerCase(), path)
+                val n = modificationsByAuthor[key] ?: 0
+                modificationsByAuthor[key] = n + 1
             }
             diff.close()
             parent.close()
@@ -57,11 +61,23 @@ private fun calculateChurn(workDir: String, limit: Int) {
         commit.close()
         count++
     }
-    println("Report:")
-    map.toList().sortedByDescending { it.second }.take(10).forEach {
-        println("File: ${it.first}")
-        println("      ${it.second}")
-        println()
+    println("Named Report:")
+    modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
+        println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
+        var changedFilesSum = 0
+        listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
+            if(modification.filePath.contains("/")){
+                print("Dir:  ")
+            }else{
+                print("File: ")
+            }
+            println(modification.filePath)
+            println("      $counter")
+            println()
+            changedFilesSum += counter
+        }
+        println("Made $changedFilesSum modifications in total")
+        println("________________________")
     }
 
     repository.close()
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
@@ -31,33 +31,17 @@ class GitRepository(val location: String) {
             val name = array[index]!!.toKString()
             val remotePtr = allocPointerTo<git_remote>()
             git_remote_lookup(remotePtr.ptr, handle, name).errorCheck()
-            list.add(GitRemote(this@GitRepository, remotePtr.value!!))
+            list.add(GitRemote(remotePtr.value!!))
         }
         list
     }
 
-    fun commits(): Sequence<GitCommit> = memScoped {
-        val walkPtr = allocPointerTo<git_revwalk>()
-        git_revwalk_new(walkPtr.ptr, handle).errorCheck()
-        val walk = walkPtr.value
-        git_revwalk_sorting(walk, GIT_SORT_TOPOLOGICAL or GIT_SORT_TIME)
-        git_revwalk_push_head(walk).errorCheck()
-        generateSequence<GitCommit> {
-            memScoped {
-                val oid = alloc<git_oid>()
-                val result = git_revwalk_next(oid.ptr, walk)
 
-                when (result) {
-                    0 -> {
-                        val commitPtr = allocPointerTo<git_commit>()
-                        git_commit_lookup(commitPtr.ptr, handle, oid.ptr).errorCheck()
-                        val commit = commitPtr.value!!
-                        GitCommit(this@GitRepository, commit)
-                    }
-                    GIT_ITEROVER -> null
-                    else -> throw Exception("Unexpected result code $result")
-                }
-            }
+    fun commits() = memScoped {
+        val walk = GitRevwalk(handle, sort = GIT_SORT_TOPOLOGICAL or GIT_SORT_TIME)
+        walk.repush()
+        generateSequence<GitCommit> {
+            walk.nextCommit()
         }
     }
 }
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package sample.gitchurn
+
+
+import kotlinx.cinterop.*
+import libgit2.*
+
+/**
+ * This class provides revwalk functionality by wrapping imported C functions to be the one entity
+ */
+class GitRevwalk (private val repository_handle: CPointer<git_repository>, val sort: git_sort_t) {
+    private val arena = Arena()
+    private val handle = memScoped {
+        val loc = allocPointerTo<git_revwalk>()
+        git_revwalk_new(loc.ptr, repository_handle).errorCheck()
+        loc.value!!
+    }
+
+    fun close() {
+        git_revwalk_free(handle)
+        arena.clear()
+    }
+
+    fun repush(oidToStartFrom: String = "") { // oidToStartFrom determines, whether revwalk will start from the HEAD reference, or some exact commit.
+        git_revwalk_reset(handle)
+        git_revwalk_sorting(walk = handle, sort_mode = sort)
+        when(oidToStartFrom){
+            "" -> git_revwalk_push_head(handle)
+            else -> memScoped {
+                val oid = alloc<git_oid>()
+                git_oid_fromstr(oid.ptr, oidToStartFrom)
+                git_revwalk_push(handle, oid.ptr)
+            }
+        }
+    }
+
+    fun sortMode(i:UInt){
+        git_revwalk_sorting(walk = handle, sort_mode = i)
+    }
+
+    fun nextCommit() : GitCommit? {
+        val commitPtr = arena.allocPointerTo<git_commit>()// here i can take iterated oids pointing to different commits, sorted by ^
+        val o = arena.alloc<git_oid>()
+        return when(val n = git_revwalk_next(o.ptr, handle)){
+            0 -> {
+                git_commit_lookup(commitPtr.ptr, repository_handle, o.ptr).errorCheck()
+                val commit = commitPtr.value!!
+                GitCommit(repository_handle, commit)
+            }
+            GIT_ITEROVER -> null
+            else -> throw Exception("Unexpected result code $n")
+        }
+    }
+}
--- a/samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
+++ b/samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
@@ -10,7 +10,7 @@ import kotlinx.cinterop.*
 import libgit2.*
 
 /**
- * This class provides revwalk functionality by wrapping imported C functions to be the one entity
+ * This class provides revwalk functionality by wrapping imported C functions to be the one entity.
  */
 class GitRevwalk (private val repository_handle: CPointer<git_repository>, val sort: git_sort_t) {
     private val arena = Arena()
@@ -43,7 +43,8 @@ class GitRevwalk (private val repository_handle: CPointer<git_repository>, val s
     }
 
     fun nextCommit() : GitCommit? {
-        val commitPtr = arena.allocPointerTo<git_commit>()// here i can take iterated oids pointing to different commits, sorted by ^
+        //Here i can take iterated oids pointing to different commits, sorted by asc.
+        val commitPtr = arena.allocPointerTo<git_commit>()
         val o = arena.alloc<git_oid>()
         return when(val n = git_revwalk_next(o.ptr, handle)){
             0 -> {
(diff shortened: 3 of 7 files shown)


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanToolRunner.kt
```diff
@@ -72,7 +73,20 @@ internal abstract class KonanCliRunner(
             "java.library.path" to "${project.konanHome}/konan/nativelib"
     )
 
-    override val environment = mutableMapOf("LIBCLANG_DISABLE_CRASH_RECOVERY" to "1")
+    override val environment = mutableMapOf("LIBCLANG_DISABLE_CRASH_RECOVERY" to "1").apply {
```
SvyatoslavScherbina: I suppose this file was committed by mistake, and so should be reverted.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,11 +35,14 @@ fun main(args: Array<String>) {
 private fun calculateChurn(workDir: String, limit: Int) {
     println("Opening…")
     val repository = git.repository(workDir)
-    val map = mutableMapOf<String, Int>()
+    //val map = mutableMapOf<String, Int>()
```
SvyatoslavScherbina: Commented code is not supposed to be kept in the repository.
Old version can always be retrieved from Git history instead.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,11 +35,14 @@ fun main(args: Array<String>) {
...
+    val namedMap = mutableMapOf<Pair<String, String>, Int>()
```
SvyatoslavScherbina: The code becomes unreadable at this point.
To figure out `namedMap` meaning one have to refer to the code populating this map below.
When it is possible, `namedMap` should describe its purpose clearly by itself.

To achieve that, it makes sense to 
1.  Declare a `data class` instead of using `Pair`
2. Rename `namedMap` to something more meaningful.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,11 +35,14 @@ fun main(args: Array<String>) {
...
     var count = 0
-    val commits = repository.commits()
-    val limited = commits.take(limit)
-    println("Calculating…")
+    //val commits = repository.commits()
+    val myCommits = repository.myCommits()
+    //val limited = commits.take(limit)
+    val myLimited = myCommits.take(limit)
```
SvyatoslavScherbina: Names like `mySomething` should be avoided. "My" doesn't help to understand the purpose of the variable here.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -57,11 +60,54 @@ private fun calculateChurn(workDir: String, limit: Int) {
...
+
+    /*println("Report:")
     map.toList().sortedByDescending { it.second }.take(10).forEach {
         println("File: ${it.first}")
         println("      ${it.second}")
         println()
+    }*/
+    println("Named Report:")
+
+    namedMap.toList().filter { it.second > 20 }.groupBy { it.first.first }.values.forEach {
```
SvyatoslavScherbina: Another reason for using `data class` becomes obvious here:  `it.first.author` is more understandable than `it.first.first`.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -57,11 +60,54 @@ private fun calculateChurn(workDir: String, limit: Int) {
...
+        println("Author: ${it[0].first.first}")
+        var changedFilesSum = 0
+        it.sortedByDescending { it.second }.forEach {
+
+    //namedMap.toList().sortedByDescending { it.second }.takeWhile { it.second > 20 }.forEach {
+        //println("Author: ${it.first.first}")
+            println("File: ${it.first.second}")
+            println("      ${it.second}")
+            println()
+            changedFilesSum+=it.second
```
SvyatoslavScherbina: Spaces missing. Should be `changedFilesSum += it.second`.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
+package sample.gitchurn
+
+
+import kotlinx.cinterop.*
+import libgit2.*
+import platform.posix.alloca
```
SvyatoslavScherbina: Redundant import.

samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
```diff
@@ -60,4 +60,19 @@ class GitRepository(val location: String) {
             }
         }
     }
+
+    fun myCommits() = memScoped {
+        val walk = GitRevwalk(handle, sort = GIT_SORT_TOPOLOGICAL or GIT_SORT_TIME)
+        walk.repush()
```
SvyatoslavScherbina: The code below is ill-formatted.
Fortunately, Kotlin defines some kind of standard formatting convention: https://kotlinlang.org/docs/reference/coding-conventions.html#formatting

You can use Code -> Reformat Code feature in IntelliJ IDEA to make your code better conform the conventions.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * this class provides revwalk functionality without need of memory management etc
```
SvyatoslavScherbina: The comment should be a sentence, i.e. start with capital letter and end with a dot.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+        git_revwalk_new(loc.ptr, repository_handle).errorCheck()
+        loc.value!!
+    }
+
+    fun close() {
+        git_revwalk_free(handle)
+        arena.clear()
+    }
+
+    fun repush(pushMode: String = "") { // I want to use pushMode as string,
```
SvyatoslavScherbina: `pushMode: String` is kinda unclear.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+                                        // that provides different regex or references for pushing
+        git_revwalk_reset(handle)
+        git_revwalk_sorting(walk = handle, sort_mode = sort)
+        when(pushMode){
+            "" -> git_revwalk_push_head(handle)
+            else -> {
+                val oid: CPointer<git_oid>? = null
```
SvyatoslavScherbina: Not sure if this code is correct. Have you checked that it works as expected?

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+                git_oid_fromstr(oid, pushMode)
+                git_revwalk_push(handle, oid)}
+        }
+    }
+
+    @ExperimentalUnsignedTypes
```
SvyatoslavScherbina: This annotation is not required. IDEA incorrectly reports a warning here.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+    fun sortMode(i:UInt){
+        git_revwalk_sorting(walk = handle, sort_mode = i)
+    }
+
+    val oid: CPointer<git_oid>? // here i can take iterated oids pointing to different commits, sorted by ^
+        get() {
+            val o = arena.alloc<git_oid>()
+            val n = git_revwalk_next(o.ptr, handle)
+            when(n){
+                0 -> return  o.ptr
```
SvyatoslavScherbina: `when` can be used as an expression, so it is possible and desirable to write this code like
```kotlin
return when(n) {
    0 -> o.ptr
    GIT_ITEROVER -> null
    else -> throw Exception("Unexpected result code $n")
}
```

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,52 @@
...
+                git_revwalk_push(handle, oid)}
+        }
+    }
+
+    @ExperimentalUnsignedTypes
+    fun sortMode(i:UInt){
+        git_revwalk_sorting(walk = handle, sort_mode = i)
+    }
+
+    val oid: CPointer<git_oid>? // here i can take iterated oids pointing to different commits, sorted by ^
```
SvyatoslavScherbina: Do I understand correctly that this property returns (logically) different values on each access?
If so, then please use a function here, because property should not behave like this.

samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
```diff
@@ -60,4 +60,20 @@ class GitRepository(val location: String) {
             }
         }
     }
+
+    fun myCommits() = memScoped {
```
SvyatoslavScherbina: How is it different from `commits`?

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
     println("Opening…")
     val repository = git.repository(workDir)
-    val map = mutableMapOf<String, Int>()
+    data class WhoAndWhere(val authorEmail: String = "", val filePath: String = "")
+    val modificationsByAuthor = mutableMapOf<WhoAndWhere, Int>()
     var count = 0
-    val commits = repository.commits()
-    val limited = commits.take(limit)
+    val commits = repository.myCommits()
+    val commitsLimit = commits.take(limit)
```
SvyatoslavScherbina: `commitsLimit` looks misnamed. Also `commits` is used only once, so in Kotlin this code should be written more like
```
val commits = repository.myCommits().take(limit)
```

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
-            println("Commit #$count [${commit.time.format()}]: ${commit.summary}")
+            println("Commit #$count [${commit.time.format()}] by ${commit.author.name!!.toKString()}: ${commit.summary}")
 
         commit.parents.forEach { parent ->
             val diff = commit.tree.diff(parent.tree)
             diff.deltas().forEach { delta ->
-                val path = delta.newPath
-                val n = map[path] ?: 0
-                map.put(path, n + 1)
+                val path = if (delta.newPath == delta.newPath.substringBefore('/'))
```
SvyatoslavScherbina: Do I understand correctly that this code computes the path to the directory containing the file?
Somewhat unreadable.
Does it handle files located at the root directory properly?

samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
artdfel: My version uses new GitRevwalk class instance to work with libgit2 revwalk objects. In old `commits` it was called directly. I see, that it should be renamed, but I don't know how to name it properly.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
+                    delta.newPath
+                else
+                    delta.newPath.substringBefore('/') + "/"
+                val n = modificationsByAuthor[WhoAndWhere(parent.author.email!!.toKString().toLowerCase(), path)] ?: 0
+                modificationsByAuthor[WhoAndWhere(parent.author.email!!.toKString().toLowerCase(), path)] = n + 1
```
SvyatoslavScherbina: `WhoAndWhere(...)` is copy-pasted.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
             val diff = commit.tree.diff(parent.tree)
             diff.deltas().forEach { delta ->
-                val path = delta.newPath
-                val n = map[path] ?: 0
-                map.put(path, n + 1)
+                val path = if (delta.newPath == delta.newPath.substringBefore('/'))
+                    delta.newPath
+                else
+                    delta.newPath.substringBefore('/') + "/"
+                val n = modificationsByAuthor[WhoAndWhere(parent.author.email!!.toKString().toLowerCase(), path)] ?: 0
```
SvyatoslavScherbina: > `parent.author.email`

Is it correct to check author of `commit`'s parent instead of that of the `commit` itself?

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
artdfel: Not exactly, this code idea was to find out, which of the root directories this exact file lies in. It was a move towards "where this person was working" analysis, that differs from original "what files were touched what number of times". And yes, it handles files from the root, in the report they can be spotted by missing '/' in the end.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
         commit.close()
         count++
     }
-    println("Report:")
-    map.toList().sortedByDescending { it.second }.take(10).forEach {
-        println("File: ${it.first}")
-        println("      ${it.second}")
-        println()
+    println("Named Report:")
+    modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach {
```
SvyatoslavScherbina: Lack of named variables and properties (other than almost meaningless `it`, `.first` and `.second`) makes this code complicated.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,44 @@ fun main(args: Array<String>) {
...
+        println("Author: ${it[0].first.authorEmail}")
+        var changedFilesSum = 0
+        it.sortedByDescending { it.second }.forEach(fun(it: Pair<WhoAndWhere, Int>) {
```
SvyatoslavScherbina: What's the motivation behind using `fun(...) {}` syntax instead of plain lambda?

samples/gitchurn/src/gitChurnMain/kotlin/GitRepository.kt
SvyatoslavScherbina: So both `commits` and `myCommits` are supposed to produce the same result?
Why have both then?

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
SvyatoslavScherbina: > which of the root directories this exact file lies in

Are there multiple root directories?

> And yes, it handles files from the root, in the report they can be spotted by missing '/' in the end.

Does this code produce file paths for files from root directory and containing directory paths for others?

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
artdfel: IDEA told me, that there was two different `it`, one from the outer cycle, and one from the inner one. To make it more straightforward, it recommended to declare my `it` this way. Now I see, that I need just to rewrite this part with more named variables.

SvyatoslavScherbina: Btw, you can specify lambda parameter to have name other than `it`, e.g.
```Kotlin
{ parameter ->
}
```
You can also achieve that by using Alt+Enter on `it`.

On the other hand, using `fun(it: Pair<WhoAndWhere, Int>) {` doesn't seem to avoid shadowing one `it` by another.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,53 @@
...
+                git_revwalk_push(handle, oid.ptr)
+            }
+        }
+    }
+
+    fun sortMode(i:UInt){
+        git_revwalk_sorting(walk = handle, sort_mode = i)
+    }
+
+    fun oid() : CPointer<git_oid>? { // here i can take iterated oids pointing to different commits, sorted by ^
```
SvyatoslavScherbina: This method name doesn't help to know that this method returns next id.
Typically such a method should be named with `next` word, like in `Iterator`.

SvyatoslavScherbina: Also to make the entire class more high-level it makes sense to return `GitCommit` instead of `git_oid` from this method.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import kotlinx.cinterop.*
+import libgit2.*
+
+/**
+ * This class provides revwalk functionality by wrapping imported C functions to be the one entity
+ */
+class GitRevwalk (val repository_handle: CPointer<git_repository>, val sort: git_sort_t) {
+    val arena = Arena()
+    val handle: CPointer<git_revwalk> = memScoped {
```
olonho: No need for explicit typing here.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,53 @@
...
+        git_revwalk_free(handle)
+        arena.clear()
+    }
+
+    fun repush(oidToStartFrom: String = "") { // oidToStartFrom determines, whether revwalk will start from the HEAD reference, or some exact commit.
+        git_revwalk_reset(handle)
+        git_revwalk_sorting(walk = handle, sort_mode = sort)
+        when(oidToStartFrom){
+            "" -> git_revwalk_push_head(handle)
+            else -> memScoped {
```
olonho: Looks like typical pattern, why not factor out.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
artdfel: I meant all directories, that are inside of the root one.
> Does this code produce file paths for files from root directory and containing directory paths for others?
Yes, but only highest-level part of the directory path, if I use this term the right way.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
artdfel: I don't know, how to deal with groupBy. To use it I got to make List, and then my data comes to be paired, which leads to all this `.first-.second` thing. Is there something I misunderstand, or it can't be solved this way?

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
artdfel: This means I should not do this in the "else" case, or what?

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
SvyatoslavScherbina: Now I see. Thanks for the explanation.
I believe it should be commented with a couple of examples, e.g. 
```
// foo/bar/baz.txt -> foo/
// foo.txt -> foo.txt
```

Also `delta.newPath == delta.newPath.substringBefore('/')` looks equivalent to `!delta.newPath.contains('/')`. If this is true, then simpler variant is preferable.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
SvyatoslavScherbina: Consider using [destructuring declarations in lambdas](https://kotlinlang.org/docs/reference/multi-declarations.html#destructuring-in-lambdas-since-11).

Try giving proper names to entities, e.g. to `forEach {` lambda parameters. If you find it extremely difficult, then this must be a sign that your data is too complicated and should be simplified.

samples/gitchurn/src/gitChurnMain/kotlin/GitRevwalk.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    }
+
+    fun sortMode(i:UInt){
+        git_revwalk_sorting(walk = handle, sort_mode = i)
+    }
+
+    fun nextCommit() : GitCommit? {
+        val commitPtr = arena.allocPointerTo<git_commit>()// here i can take iterated oids pointing to different commits, sorted by ^
```
olonho: Put comment before line, start with capital letter and terminate with dot.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,49 @@ fun main(args: Array<String>) {
...
-        println("      ${it.second}")
-        println()
+    println("Named Report:")
+    modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
+        println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
+        var changedFilesSum = 0
+        listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
+            if(modification.filePath.contains("/")){
+                print("Dir:  ")
+            }else{
```
olonho: `} else {`

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,49 @@ fun main(args: Array<String>) {
...
-    map.toList().sortedByDescending { it.second }.take(10).forEach {
-        println("File: ${it.first}")
-        println("      ${it.second}")
-        println()
+    println("Named Report:")
+    modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
+        println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
+        var changedFilesSum = 0
+        listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
+            if(modification.filePath.contains("/")){
```
olonho: `if (`

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,49 @@ fun main(args: Array<String>) {
...
+                print("Dir:  ")
+            }else{
+                print("File: ")
+            }
+            println(modification.filePath)
+            println("      $counter")
+            println()
```
olonho: Unneeded extra linefeed.

samples/gitchurn/src/gitChurnMain/kotlin/GitChurn.kt
```diff
@@ -35,33 +35,49 @@ fun main(args: Array<String>) {
...
-    println("Report:")
-    map.toList().sortedByDescending { it.second }.take(10).forEach {
-        println("File: ${it.first}")
-        println("      ${it.second}")
-        println()
+    println("Named Report:")
+    modificationsByAuthor.toList().groupBy { it.first.authorEmail }.values.forEach { listOfAuthorsChanges ->
+        println("Author: ${listOfAuthorsChanges[0].first.authorEmail}")
+        var changedFilesSum = 0
+        listOfAuthorsChanges.sortedByDescending { it.second }.forEach { (modification, counter) ->
```
olonho: Produce more condensed output.
Like
```
Author: nikolay.igotti@jetbrains.com
experiments/: 765<tab> backend.native/: 201 <tab>runtime/: 147
```

artdfel: I have to disagree. Currently, there are 47 files and directories in the compiler repo. This amount of labels looks uninformative when I try to output it as a one-liner.

SvyatoslavScherbina:
Abandoned.

PR #3060, opened 2019-06-11
Add benchmark plugins

ilmat192:


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -0,0 +1,260 @@
...
+
+    private fun Project.determinePreset(): KotlinNativeTargetPreset =
+        defaultHostPreset(this).also { preset ->
+            logger.quiet("$project has been configured for ${preset.name} platform.")
+        } as KotlinNativeTargetPreset
+
+    private fun Project.configureSourceSets(kotlinVersion: String) {
+        with(kotlin.sourceSets) {
+            commonMain.dependencies {
+                implementation("org.jetbrains.kotlin:kotlin-stdlib-common:$kotlinStdlibVersion")
```
LepilkinaElena: Can there be situations when stdlib version isn't equal kotlin version? We use it from kotlin and kotlin-native? If we will separate it, then I should add it to report, because difference in stdlib can influence on results too.

LepilkinaElena:
One more small note `jvmRun` have no group `Benchmarking tasks`, so in list it looks a bit strange, I can latter add group and description for it too, to have consistent information about benchmark tasks.

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -0,0 +1,260 @@
...
+                val properties: Map<String, Any> = commonBenchmarkProperties + mapOf(
+                    "type" to "jvm",
+                    "compilerVersion" to kotlinVersion,
+                    "benchmarks" to benchContents,
+                    "compileTime" to listOf(jvmCompileTime),
+                    "codeSize" to getCodeSizeBenchmark(applicationName, jarPath.absolutePath)
+                )
+
+                val output = createJsonReport(properties)
+                buildDir.resolve(project.property("jvmJson") as String).writeText(output)
```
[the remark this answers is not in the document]
ilmat192: Fixed

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/benchmark/CompileBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+    
+    private fun Project.configureKonanRun(
+        benchmarkExtension: CompileBenchmarkExtension
+    ): Unit = with(benchmarkExtension) {
+        // Aggregate task.
+        val konanRun = tasks.create("konanRun") { task ->
+            task.dependsOn("configureBuild")
+
+            task.group = BenchmarkingPlugin.BENCHMARKING_GROUP
+            task.description = "Runs the compile time benchmark for Kotlin/Native."
```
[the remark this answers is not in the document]
ilmat192: Fixed

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/benchmark/CompileBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                }
+            }
+        }
+
+        // Report task.
+        tasks.create("konanJsonReport").apply {
+
+            group = BenchmarkingPlugin.BENCHMARKING_GROUP
+            description = "Builds the compile time benchmarking report for Kotlin/Native."
```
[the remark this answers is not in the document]
ilmat192: Fixed

buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
ilmat192: Usually Kotlin compiler and stdlib versions are equal but in general case they can be changed independently (see the root `gradle.properties` - we have separate properties for these versions there). Kotlin stdlib version may influence our performance because we use the same common part in our stdlib.

cc @vvlevchenko

ilmat192:
Fixed `jvmRun` group too.

performance/videoplayer/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+    )
+    PlatformInfo.isLinux() -> listOf("-compiler-option", "-I/usr/include/SDL2")
+    PlatformInfo.isWindows() -> listOf("-compiler-option", "-I$mingwPath/include/SDL2")
+    else -> error("Unsupported platform")
+}
+
+compileBenchmark {
+    applicationName = "Videoplayer"
+    repeatNumber = 10
+    buildSteps = mapOf(
```
PavelPunegov: I'd rather made `buildSteps` a DSL that gets `steps`, each of that could be a simple data class with named args `name` and `cmd` to look like:
```
buildSteps {
     step { 
            name="runCinterop"
            cmd="listOf()
     }
}
```

ilmat192: Fixed.

PR #3068, opened 2019-06-13
Handle zero-terminated strings in deprecated utf8 to String conversions

qurbonzoda:


runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -21,12 +21,20 @@ public fun ByteArray.stringFromUtf8() : String {
 @Deprecated("Use decodeToString instead", ReplaceWith("decodeToString(start, start + size)"))
 public fun ByteArray.stringFromUtf8(start: Int = 0, size: Int = this.size) : String {
     checkBoundsIndexes(start, start + size, this.size)
-    return stringFromUtf8Impl(start, size)
+    return stringFromUtf8Impl(start, realSize(this, size))
```
qurbonzoda: Real size is determined after bounds check. It's the original behaviour.

ilya-g:
Are there any tests that expect this behavior?

runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -21,12 +21,20 @@ public fun ByteArray.stringFromUtf8() : String {
 @Deprecated("Use decodeToString instead", ReplaceWith("decodeToString(start, start + size)"))
```
ilya-g: We'd also like to remove the deprecation replacement for now, until we design the replacement function.

qurbonzoda:
@ilya-g If you are asking about zero-terminated strings: `./gradlew interop_zlib`
If you are about checking bounds before determining real size: No tests

ilya-g:
> @ilya-g If you are asking about zero-terminated strings: `./gradlew interop_zlib`

I do not see where it fails on the build server before this change. @olonho Could you clarify?

runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -9,45 +9,46 @@ package kotlin.native
...
  */
-@Deprecated("Use decodeToString instead", ReplaceWith("decodeToString(start, start + size, throwOnInvalidSequence = true)"))
 public fun ByteArray.stringFromUtf8OrThrow(start: Int = 0, size: Int = this.size) : String {
     checkBoundsIndexes(start, start + size, this.size)
     try {
-        return stringFromUtf8OrThrowImpl(start, size)
+        return stringFromUtf8OrThrowImpl(start, realSize(this, size))
     } catch (e: CharacterCodingException) {
-        @Suppress("DEPRECATION")
         throw IllegalCharacterConversionException()
```
qurbonzoda: Now that `ByteArray.stringFromUtf8OrThrow()` is not deprecated, `IllegalCharacterConversionException` is also not deprecated. Because throwing deprecated exception from non-deprecated function looks odd.

runtime/src/main/kotlin/kotlin/native/Text.kt
qurbonzoda: +1

runtime/src/main/kotlin/kotlin/native/Text.kt
olonho: > > @ilya-g If you are asking about zero-terminated strings: `./gradlew interop_zlib`
> 
> I do not see where it fails on the build server before this change. @olonho Could you clarify?

It doesn't fail as it needs platform libraries, and is not being run by default.

PR #3073, opened 2019-06-17
Add a migration guide from kotlin-platform-native to the MPP plugin

ilmat192:


msink:
Maybe add hint how to configure all targets:

```
targets.all {
    if (this is KotlinNativeTarget) {
        ....
    }
}
```

ilmat192:
Added, thank you.

msink:
In your variant smartcast does not work.

ilmat192:
No, it's ok:
image

msink:
Well, sorry, I messed it: `targets.withType(KotlinNativeTarget::class.java)` works, `targets.withType<KotlinNativeTarget>()` does not.

GRADLE_PLUGIN.md
```diff
@@ -1,10 +1,268 @@
...
-for more information.
+### Applying the multiplatform plugin
+
+To apply the `kotlin-multiplatform` plugin, just add the following snippet into your build script:
+
+<div class="sample" markdown="1" theme="idea" mode="groovy">
+
+```groovy
+plugins {
+    id 'org.jetbrains.kotlin.multiplatform' version '1.3.40'
```
h0tk3y: If kotlinlang.org is the only place where this documentation is presented, then you may use `{{ site.data.releases.latest.version }}` Instead of a particular Kotlin version.

h0tk3y: Also consider using the syntax that is valid in both Groovy and Kotlin where it is feasible. This ensures that copying a sample into the build script works in more cases. Here, it would be `id("org.jetbrains.kotlin.multiplatform") version "..."`

GRADLE_PLUGIN.md
```diff
@@ -1,10 +1,268 @@
...
+    // You also can configure all native targets in a single block.
+    targets.withType(KotlinNativeTarget) {
+        // Native target configuration.
+    }
+}
+```
+
+</div>
+
+Each target includes two __compilations__: `main` and `test` compiling product and test sources respectively. A compilation is an abstraction
```
h0tk3y: ~~`product`~~ → `production`?

GRADLE_PLUGIN.md
```diff
@@ -1,10 +1,268 @@
...
+```
+
+</div> 
+
+See more about native binaries declaration at the [corresponding page](https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#building-final-native-binaries).
+
+
+### Publishing
+
+Both `kotlin-platform-native` and `kotlin-multiplatform` plugins automatically set up artifact publication when the `maven-publish` plugin is applied.
```
h0tk3y: A link to [Publishing a Multiplatform Library](https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#publishing-a-multiplatform-library) may be useful here.

jonnyzzz:
Please do not forget to merge documentation into `1.3.0-fixes` branch and also update the `_nav.yml` file if you'd like a new page to be added to the TOC

GRADLE_PLUGIN.md
```diff
@@ -1,10 +1,268 @@
...
+        dependencies {
+            implementation 'org.sample.test:testlibrary:1.0'
+        }
+    }
+}
+```
+
+</div>
+
+Note that a dependency declared for `commonMain` or `commonTest` source set must be published using the `kotlin-multiplatform` plugin.
```
h0tk3y: > a dependency <...> must be published <...>
To me, the phrasing is a bit confusing as it may imply that the dependency declaration in this project somehow limits publication options for this project. Consider rephrasing to something like this:
> a module referenced by a dependency <...> must be published

GRADLE_PLUGIN.md
ilmat192: > If kotlinlang.org is the only place where this documentation is presented, then you may use {{ site.data.releases.latest.version }} Instead of a particular Kotlin version.

Sometimes we use GitHub links to this documentation, so I think it's better to leave explicit version here.

> Also consider using the syntax that is valid in both Groovy and Kotlin where it is feasible. 

Fixed.

PR #3078, opened 2019-06-17
Implement primitive to run through worker's queue

olonho:


runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,6 +103,18 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
         executeAfterInternal(id, operation, afterMicroseconds)
     }
 
+
+    /**
+     * Process pending job(s) on the queue of this worker, returns `true` is something was processed
```
SvyatoslavScherbina: > returns `true` is something

typo.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,6 +103,18 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * and `false` otherwise. Note that jobs scheduled with [executeAfter] using non-zero timeout are
```
SvyatoslavScherbina: > Note that jobs scheduled with [executeAfter] using non-zero timeout are not processed this way.

Is this behaviour tested?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,6 +103,18 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * not processed this way. If termination request arrives while processing the queue via this API,
+     * worker is marked as terminated and will exit once the current request is done with.
```
SvyatoslavScherbina: > If termination request arrives while processing the queue via this API, worker is marked as terminated and will exit once the current request is done with.

Is this behaviour tested?

backend.native/tests/runtime/workers/worker11.kt
```diff
@@ -40,9 +39,49 @@ fun initJobs(count: Int) = Array<Job?>(count) { i -> Job(i, i * 2, i)}
...
+      }
+  }
+  val futures = Array(workers.size) { workerIndex ->
+      workers[workerIndex].execute(TransferMode.SAFE, {
+          workerIndex
+      }) {
+          index ->
+          assertEquals(0, counters[index].value)
+          // Process following request.
+          while (!Worker.current!!.processQueue()) {}
```
SvyatoslavScherbina: This test doesn't actually ensure that `processQueue` returns `false` if queue is empty.
The test will likely fail if `processQueue` returns `true` in this case, but it makes sense to avoid relying on non-deterministic effects.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
SvyatoslavScherbina: > Process pending job(s)

Can it really process multiple jobs?

backend.native/tests/runtime/workers/worker11.kt
olonho: It's not what it tests, so for that there shall be a different test.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Yes, if inside a job there's call to processQueue.

SvyatoslavScherbina: Then this would be the effect of the job, not that of `processQueue` directly.
This side effect is worth mentioning explicitly, leaving here singular "job".

runtime/src/main/cpp/Worker.cpp
```diff
@@ -494,42 +518,41 @@ void Future::cancelUnlocked() {
...
+  ObjHolder resultHolder;
+  if (terminated_) return JOB_TERMINATE;
+  Job job = getJob(blocking);
+  switch (job.kind) {
+    case JOB_NONE: {
+      break;
+    }
+    case JOB_TERMINATE: {
+      if (job.terminationRequest.waitDelayed) {
+        if (waitDelayed()) {
```
SvyatoslavScherbina: Can this block if `blocking == false`?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Do not think so.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -494,42 +518,41 @@ void Future::cancelUnlocked() {
...
-      if (job.kind == JOB_TERMINATE) {
-        if (job.terminationRequest.waitDelayed) {
-          if (worker->waitDelayed()) {
-            worker->putJob(job, false);
-            continue;
-          }
+JobKind Worker::processQueueElement(bool blocking) {
+  ObjHolder argumentHolder;
+  ObjHolder resultHolder;
+  if (terminated_) return JOB_TERMINATE;
```
SvyatoslavScherbina: So `Worker.processQueue()` would return `true` in this case?

olonho: Yes, maybe contract better be improved, like return false on termination requests and here.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Seems not yet, not sure how to better test it.

backend.native/tests/runtime/workers/worker11.kt
```diff
@@ -40,9 +39,56 @@ fun initJobs(count: Int) = Array<Job?>(count) { i -> Job(i, i * 2, i)}
...
+  }
+  futures2.forEach { it.result }
+  futures.forEach { it.result }
+  workers.forEach {
+      it.requestTermination().result
+  }
+
+  // Ensure terminated workers are no longer there.
+  workers.forEach {
+    assertFailsWith<IllegalStateException> { it.execute(TransferMode.SAFE, { Unit }) { println("ERROR") } }
```
SvyatoslavScherbina: Btw, is the implementation of this behaviour deterministic?

https://github.com/JetBrains/kotlin-native/blob/7811831674563f02b21409b9fa06b3dab7dd73f3/runtime/src/main/cpp/Worker.cpp#L540
```c++
job.terminationRequest.future->storeResultUnlocked(nullptr, true);
theState()->removeWorkerUnlocked(id());
break;
```

I.e. the future is notified before the worker is removed, so it is not guaranteed that `it.execute` throws an exception here.

Also what happens if the job is added after termination request was sent but before it gets completed?

runtime/src/main/cpp/Worker.cpp
SvyatoslavScherbina: Has this been fixed?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Not really yet.

runtime/src/main/cpp/Worker.cpp
olonho: Yeah 
```
JobKind kind = worker->processQueueElement(false);
return kind != JOB_NONE && kind != JOB_TERMINATE;
```

backend.native/tests/runtime/workers/worker11.kt
olonho: Swapped removing and notification statements. 

If job is added after termination request is sent, it depends how adding time correlates with the termination handler and if handler shall wait. Generally, if we do processing of pending requests -  we process whatever is available in the queue at the moment of processing. If we'll terminate unconditionally - queue content is silently ignored.

backend.native/tests/runtime/workers/worker11.kt
```diff
@@ -91,4 +91,27 @@ val counters = Array(COUNT) { AtomicInt(0) }.freeze()
...
-}
\ No newline at end of file
+}
+
+@Test fun runTest2() {
+    val workers = Array(COUNT) { Worker.start() }
+    val futures = Array(workers.size) { workerIndex ->
+        workers[workerIndex].execute(TransferMode.SAFE, { null }) {
+            // Here we processed termination request.
+            assertEquals(false, Worker.current!!.processQueue())
```
SvyatoslavScherbina: `processQueue` is non-blocking. Maybe wait until termination request sent?

olonho: Here we either see termination request or see nothing, in both cases `false` to be returned. So your suggestion will make test more deterministic, but seems it's correct even in its current form.

SvyatoslavScherbina: Exactly.

PR #3079, opened 2019-06-18
 Add KType.arguments and support 'typeOf' intrinsic

SvyatoslavScherbina:
Fix KT-29917 and KT-28625.

Also perform minor refactoring.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
@@ -9,12 +9,16 @@ import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.konan.Context
+import org.jetbrains.kotlin.backend.konan.error
 import org.jetbrains.kotlin.backend.konan.ir.typeWithoutArguments
 import org.jetbrains.kotlin.backend.konan.reportCompilationError
+import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.util.irCall
+import org.jetbrains.kotlin.ir.util.isTypeOfIntrinsic
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 
 /**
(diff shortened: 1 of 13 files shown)


runtime/src/main/kotlin/kotlin/native/internal/KTypeImpl.kt
```diff
@@ -21,6 +21,40 @@ internal class KTypeImpl(
     override fun hashCode(): Int {
         return (classifier?.hashCode() ?: 0) * 31 * 31 + this.arguments.hashCode() * 31 + if (isMarkedNullable) 1 else 0
     }
+
+    override fun toString(): String {
+        val classifierString = when (classifier) {
+            is KClass<*> -> classifier.qualifiedName ?: classifier.simpleName
```
olonho: Does it handle `object` properly?

SvyatoslavScherbina: What do you mean by `object` and how do you expect it to be handled?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
```diff
@@ -26,23 +30,32 @@ internal class PostInlineLowering(val context: Context) : FileLoweringPass {
...
-                        expression.startOffset, expression.endOffset)
+                val builder = createIrBuilder(expression)
 
-                return builder.irKClass(context, expression.symbol)
+                val symbol = expression.symbol
+                return if (symbol is IrClassSymbol) {
+                    builder.irKClass(context, symbol)
+                } else {
+                    // E.g. for `T::class` in a body of an inline function itself.
+                    builder.irCall(context.ir.symbols.ThrowNullPointerException.owner)
```
olonho: NPE?

SvyatoslavScherbina: Exactly.
Please note that this change doesn't actually affect current behaviour: it only moves handling this case here from `irKClass`.

runtime/src/main/kotlin/kotlin/native/internal/KTypeImpl.kt
olonho:
```diff --git a/backend.native/tests/codegen/ktype/ktype1.kt b/backend.native/tests/codegen/ktype/ktype1.kt
index a3d9fe3d4..aacb45ca6 100644
--- a/backend.native/tests/codegen/ktype/ktype1.kt
+++ b/backend.native/tests/codegen/ktype/ktype1.kt
@@ -17,17 +17,23 @@ class D
 fun <T> kTypeForCWithTypeParameter() = kType<C<T>>()
 
 class Outer<T> {
+    companion object Friend
     inner class Inner<S>
 }
 
+object Object
+
 private val pkg = "codegen.ktype.ktype1"
 
 @Test
+@UseExperimental(ExperimentalStdlibApi::class)
 fun test1() {
     assertEquals("$pkg.C<kotlin.Int?>", kType<C<Int?>>().toString())
     assertEquals("$pkg.C<$pkg.C<kotlin.Any>>", kType<C<C<Any>>>().toString())
 
     assertEquals("$pkg.C<kotlin.Any?>", kTypeForCWithTypeParameter<D>().toString())
+    assertEquals("$pkg.Object", kType<Object>().toString())
+    assertEquals("$pkg.Outer.Friend", kType<Outer.Friend>().toString())
 
     val innerKType = kType<Outer<D>.Inner<String>>()
     assertEquals(Outer.Inner::class, innerKType.classifier)
```

SvyatoslavScherbina: Works for me.

olonho: That's what I meant by `object` and it better be tested. Can we have an anonymous object construct somehow be passed as type parameter?

SvyatoslavScherbina: Yes we can, it doesn't have string representation (`toString` returns `"(not denotable)"`).

PR #3080, opened 2019-06-18
Fix helper class for coroutines

PavelPunegov:


buildSrc/plugins/src/main/kotlin/org/jetbrains/kotlin/coroutineTestUtil.kt
```diff
@@ -101,6 +101,12 @@ fun createTextForHelpers(isReleaseCoroutines: Boolean): String {
             |    var finished = false
             |
             |    var proceed: () -> Unit = {}
+            |    
+            |    fun reset() {
```
olonho: Better be packed with tests.

PavelPunegov: It is going to be a next step. This PR is just a quick fix

PR #3081, opened 2019-06-18
[build][restrict] flags turning build bundle operation to build restricted bundle

vvlevchenko:


build.gradle
```diff
@@ -441,6 +441,44 @@ task bundle(type: (isWindows()) ? Zip : Tar) {
     }
 }
 
+task bundleRestricted(type: Tar) {
+    dependsOn('crossDistPlatformLibs')
+    dependsOn('crossDist')
+    dependsOn('distSources')
+    def simpleOsName = HostManager.simpleOsName()
+    baseName = "kotlin-native-$simpleOsName-$konanVersionFull-restricted"
```
ilmat192: I think it's better to have `kotlin-native-restricted-$simpleOsName-$konanVersionFull` instead. It's easier to support from the Gradle plugin side and IMHO more intuitive.

build.gradle
```diff
@@ -441,6 +441,44 @@ task bundle(type: (isWindows()) ? Zip : Tar) {
     }
 }
 
+task bundleRestricted(type: Tar) {
+    dependsOn('crossDistPlatformLibs')
```
ilmat192: This task and the `bundle` task have a lot of common settings. I'm not sure but probably it worth moving this common part to a single place.  E.g. 
```
task bundle (...) {
    // Bundle-specific configuration
}

task bundleRestricted(...) {
    // Restricted bundle-specific configuration.
}

configure([bundle, bundleRestricted]) {
    // Common configuration.
}
```

In this case we get rid of settings duplication but the bundle configuration becomes divided into two parts. So I'm not sure if it is better or not. What do you think?

build.gradle
```diff
@@ -441,6 +441,44 @@ task bundle(type: (isWindows()) ? Zip : Tar) {
...
+    dependsOn('crossDist')
+    dependsOn('distSources')
+    def simpleOsName = HostManager.simpleOsName()
+    baseName = "kotlin-native-$simpleOsName-$konanVersionFull-restricted"
+    from("$project.rootDir/dist") {
+        include '**'
+        exclude 'dependencies'
+        exclude 'klib/testLibrary'
+        exclude 'dist/klib/platform/macos_x64'
```
ilmat192: Shouldn't these paths be relative to `dist` like the path above? I mean `klib/platform/macos_x64` instead of `dist/klib/platform/macos_x64`.

build.gradle
vvlevchenko: Sounds great in scenario of task duplication :). I've just forgotten about this fabulous `configure` 😄

PR #3082, opened 2019-06-19
Avoid reparsing C headers in cinterop whenever possible

SvyatoslavScherbina:
Thus optimize it.


SvyatoslavScherbina:
> What are perf numbers?

As we've discussed before, `cinterop` stubs generation (i.e. excluding stubs compilation) becomes ~25% faster on macos_x64 + ios_arm64 platform libs.

PR #3104, opened 2019-06-24
Build: Replace buildSrc with a separate included build

ilmat192:
Gradle doesn't allow us to substitute dependencies of buildSrc
with artifacts provided by a composite build. But our buildSrc
depends on kotlin-native-utils provided by the Big Kotlin repo.
Thus we cannot make changes in kotlin-native-utils and use them
in buildSrc during a development process.

This patch fixes the issue by transforming buildSrc into a separate
included build.


gradle.properties
```diff
@@ -27,7 +27,4 @@ konanVersion=1.3.50
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.workers.max=4
 #kotlinProjectPath=/Users/jetbrains/IdeaProjects/kotlin
-#sharedProjectPath=/Users/jetbrains/IdeaProjects/kotlin-native-shared
 #kotlinProjectPath=/Users/jetbrains/kotlin-native/kotlin
```
PavelPunegov: Please, remove this one too

build-tools/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+    implementation("io.ktor:ktor-client-core:1.2.1")
+    implementation("io.ktor:ktor-client-cio:1.2.1")
+
+    api("org.jetbrains.kotlin:kotlin-native-utils:$kotlinVersion")
+
+    // Located in <repo root>/shared and always provided by the composite build.
+    api("org.jetbrains.kotlin:kotlin-native-shared:$konanVersion")
+}
+
+sourceSets["main"].withConvention(KotlinSourceSet::class) {
```
PavelPunegov: Why this sourceSet in't defined like this? 
```
sourceSets { kotlin {  srcDir () } } 
```
What does `withConvention` do?

build-tools/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+    // We explicitly configure versions of plugins in settings.gradle.kts.
+    // due to https://github.com/gradle/gradle/issues/1697.
+    id("kotlin")
+    id("kotlinx-serialization")
+    groovy
+    `java-gradle-plugin`
+}
+
+val rootProperties = Properties().apply {
+    rootDir.resolve("../gradle.properties").reader().use(::load)
```
PavelPunegov: Is this correct? Isn't the `rootDir` points to the whole project's root?
If so then `"../gradle.properties"` seems to be incorrect

gradle.properties
ilmat192: Fixed.

build-tools/build.gradle.kts
ilmat192: The `kotlin` source directory set is added as a dynamic property using the corresponding Gradle machinery. But there is no extension getter allowing us just write something like this

```
sourceSets["main"].kotlin // kotlin property isn't available
```

`withConvention` is a way to access this property [recommended by the documentation](https://kotlinlang.org/docs/reference/using-gradle.html#targeting-the-jvm).

build-tools/build.gradle.kts
ilmat192: No, `rootDir` points to the `build-tools` directory because `build-tools` is a separate build included into the main one thus it have its own root directory.

PR #3106, opened 2019-06-24
Preparations for relaxed mode work.

olonho:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -177,6 +177,18 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
...
 
+                put(MEMORY_MODEL, when (arguments.memoryModel) {
+                    "relaxed" -> {
+                        configuration.report(STRONG_WARNING, "Relaxed memory model is not yet functional")
+                        MemoryModel.RELAXED
+                    }
+                    "strict" -> MemoryModel.STRICT
+                    else -> {
+                        configuration.report(ERROR, "Unsupported memory model ${arguments.memoryModel}")
+                        return
```
SvyatoslavScherbina: It is more safe to provide some value here (e.g. `MemoryModel.STRICT`) instead of returning

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -307,12 +307,18 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
     }
 
     private fun updateReturnRef(value: LLVMValueRef, address: LLVMValueRef) {
-        store(value, address)
+        if (context.memoryModel == MemoryModel.STRICT)
+            store(value, address)
+        else
+            call(context.llvm.updateReturnRefFunction, listOf(address, value))
```
SvyatoslavScherbina: Why is this required?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
olonho: Maybe, can include in followup patches.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Just not rely on C++/Kotlin inliner, which not always works.

PR #3110, opened 2019-06-25
Stub IR

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -65,195 +64,7 @@ private fun ObjCMethod.getFirstKotlinParameterNameCandidate(forConstructorOrFact
...
-                KotlinTypes.any.makeNullable(),
-                isVararg = true,
-                annotations = emptyList()
-        )
-    }
-
-    return result
-}
-
-class ObjCMethodStub(private val stubGenerator: StubGenerator,
```
SvyatoslavScherbina: It makes a lot of sense to keep reworked generator pieces in its places.
This would help to keep searching through git simple and to avoid placing all code to the single file.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package org.jetbrains.kotlin.native.interop.gen
+
+import org.jetbrains.kotlin.native.interop.indexer.*
+
+interface StubIrElement {
+    fun accept(visitor: StubIrVisitor)
```
SvyatoslavScherbina: Isn't this method implementable through the second one?
In this case I don't find enough motivation to have both declared and implemented in each class.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+ * For example, can be used for comments in textual representation.
+ */
+class StubContainerMeta(
+        val textAtStart: String = "",
+        val textAtEnd: String = ""
+)
+
+/**
+ * Trivial container of IR elements that can be used
+ * for module or ObjC category representation.
```
SvyatoslavScherbina: > that can be used for module or ObjC category representation.

Somewhat incomplete and redundant.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+
+    override fun <T, R> accept(visitorAux: StubIrVisitorAux<T, R>, data: T): R {
+        return visitorAux.visitSimpleStubContainer(this, data)
+    }
+}
+
+val StubContainer.children: List<StubIrElement>
+    get() {
+        val classes: List<StubIrElement> = classes
+        val funcs: List<StubIrElement> = functions
```
SvyatoslavScherbina: What are these variables required for?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+        val classifier: Classifier,
+        val typeArguments: List<TypeArgumentStub> = emptyList(),
+        override val nullable: Boolean = false
+) : StubType()
+
+
+/**
+ * Fallback variant for all cases where we cannot refer to specific [KotlinType] or [Classifier].
+ */
+class SymbolicStubType(
```
SvyatoslavScherbina: Note that this type is not convertible to metadata.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+
+interface StubElementWithOrigin : StubIrElement {
+    val origin: StubOrigin
+}
+
+interface AnnotationHolder {
+    val annotations: List<AnnotationStub>
+}
+
+sealed class AnnotationStub {
```
SvyatoslavScherbina: Why not make annotation representation more low-level, e.g. class fq name + arguments?
I expect both output formats requiring annotations in this form.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+
+class PropertyStub(
+        val name: String,
+        val type: StubType,
+        val kind: Kind,
+        val modality: MemberStubModality = MemberStubModality.NONE,
+        val receiverType: StubType? = null,
+        override val annotations: List<AnnotationStub> = emptyList()
+) : StubIrElement, AnnotationHolder {
+    sealed class Kind {
```
SvyatoslavScherbina: Why not have `val getter: PropertyAccessor.Getter` and `val setter: PropertyAccessor.Setter?` instead?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+    sealed class Qualifier {
+        class VAL(val overrides: Boolean) : Qualifier()
+        class VAR(val overrides: Boolean) : Qualifier()
+        object NONE : Qualifier()
+    }
+}
+
+class SuperClassInit(
+        val type: StubType,
+        val parameters: List<ValueStub> = listOf()
```
SvyatoslavScherbina: Arguments.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+        val type: StubType
+)
+
+class FunctionParameterStub(
+        val name: String,
+        val type: StubType,
+        override val annotations: List<AnnotationStub> = emptyList(),
+        val isVararg: Boolean = false,
+        val origin: StubOrigin = StubOrigin.None
+) : ValueStub(), AnnotationHolder
```
SvyatoslavScherbina: A parameter itself is not a value.
Maybe declare `class GetFunctionParameter(val parameter: FunctionParameterStub) : ValueStub()` instead?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,460 @@
...
+}
+
+sealed class PropertyAccessor : FunctionalStub {
+
+    sealed class Getter : PropertyAccessor() {
+        class SimpleGetter(
+                override val annotations: List<AnnotationStub> = emptyList(),
+                val constant: ConstantStub? = null
+        ) : Getter() {
+            override val parameters: List<FunctionParameterStub> = emptyList()
```
SvyatoslavScherbina: Can this be pushed up to `Getter` class?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -0,0 +1,273 @@
...
+ */
+class StubIrBridgeBuilder(
+        private val context: StubIrContext,
+        private val builderResult: StubIrBuilderResult) {
+
+    private val globalAddressExpressions = mutableMapOf<String, KotlinExpression>()
+
+    private fun getGlobalAddressExpression(cGlobalName: String, accessor: PropertyAccessor) =
+            globalAddressExpressions.getOrPut(cGlobalName) {
+                simpleBridgeGenerator.kotlinToNative(
```
SvyatoslavScherbina: It is somewhat incorrect to register only the first requesting accessor as `nativeBacked` for this stub. `globalAddressExpressions` should probably have both function parameters in its keys.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
```diff
@@ -0,0 +1,408 @@
...
+    override fun getKotlinClassForPointed(structDecl: StructDecl): Classifier {
+        val classifier = declarationMapper.getKotlinClassForPointed(structDecl)
+        return classifier
+    }
+}
+
+data class StubIrBuilderResult(
+        val stubs: SimpleStubContainer,
+        val declarationMapper: DeclarationMapper,
+        val namesToBeDeclared: List<String>,
```
SvyatoslavScherbina: Isn't this list computable from `stubs`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -0,0 +1,273 @@
...
+        ) { nativeValues ->
+            "${origin.function.name}(${nativeValues.joinToString()})"
+        }
+        bodyGenerator.returnResult(result)
+        functionBridgeBodies[function] = bodyGenerator.build()
+    }
+
+    private fun generateProtocolGetter(protocolGetterName: String, protocol: ObjCProtocol) {
+        val builder = NativeCodeBuilder(simpleBridgeGenerator.topLevelNativeScope)
+        val nativeBacked = object : NativeBacked {}
```
SvyatoslavScherbina: Было бы неплохо сюда принести `TODO`, которая была приписана к этой строчке: https://github.com/JetBrains/kotlin-native/pull/3110/files#diff-d16ba15be14e492e6988b703670530dcL515

SvyatoslavScherbina: It would be great to have the same `TODO` here as before:
https://github.com/JetBrains/kotlin-native/pull/3110/files#diff-d16ba15be14e492e6988b703670530dcL515

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+ * that can be found in the LICENSE file.
+ */
+package org.jetbrains.kotlin.native.interop.gen
+
+import org.jetbrains.kotlin.native.interop.indexer.ObjCProtocol
+
+private val StubOrigin.ObjCMethod.isOptional: Boolean
+        get() = container is ObjCProtocol && method.isOptional
+
+fun FunctionStub.isObjCMethodOptional(): Boolean = this.origin is StubOrigin.ObjCMethod &&
```
SvyatoslavScherbina: `isOptionalObjCMethod()`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrDriver.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.util.*
+
+class StubIrContext(
+        val configuration: InteropConfiguration,
+        val nativeIndex: NativeIndex,
+        val imports: Imports,
+        val platform: KotlinPlatform,
+        val libName: String
+) {
+    val libraryForCStubs = configuration.library.copy(
```
SvyatoslavScherbina: Be careful with this code when merging with master.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,669 @@
...
+                            out("")
+                            it.accept(this, element)
+                        }
+                    }
+                }
+            }
+        }
+
+        override fun visitTypealias(element: TypealiasStub, owner: StubContainer?) {
+            out("typealias ${renderClassifierDeclaration(element.alias.classifier)} = ${renderStubType(element.aliasee)}")
```
SvyatoslavScherbina: > `renderClassifierDeclaration(element.alias.classifier)`

Somewhat inconsistent with `element.alias` having type arguments.
Either make `element.alias` classifier or render the full type properly here.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: Because inferred type of concatenation was `List<Any>`. Made it simpler.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: There is no usages of `FunctionParameterStub` as `ValueStub` left so it can be removed.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: Since the list of annotation types is limited this approach is simpler from development point of view. 
It's easier to find usages of specific annotation type, easier to render (e.g. one can avoid rendering default parameter values) and it's less error-prone to typos and forgotten renamings that are quite common in low-level approach.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: Yeah, `SymbolicStubType` is a hacky approach. It'll be more clear how to do it right when I'll start to work on metadata-based "backend".

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrElementBuilders.kt
```diff
@@ -0,0 +1,912 @@
...
+
+        val baseTypeMirror = context.mirror(enumDef.baseType)
+        val baseType = WrapperStubType(baseTypeMirror.argType)
+
+        val clazz = (context.mirror(EnumType(enumDef)) as TypeMirror.ByValue).valueType.classifier
+
+        val enumVariants = enumDef.constants.map {
+            val literal = context.tryCreateIntegralStub(enumDef.baseType, it.value)
+                    ?: error("Cannot create enum value ${it.value} of type ${enumDef.baseType}")
+            EnumEntryStub(it.name.asSimpleName(), literal)
```
SvyatoslavScherbina: All `.asSimpleName()` should be moved to text emitter to avoid
1.  applying it twice
2.  applying it when emitting metadata

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: In this case it is harder to distinguish `const val x: Int = 5` and `val x: Int get() = 5`. If we add field `val isConst: Boolean` to `PropertyAccessor` it would be possible to get something like `const var ...`

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -7,14 +7,30 @@ package org.jetbrains.kotlin.native.interop.gen
...
+/**
+ * Compute which names will be declared by [StubContainer] in the given [pkgName]
+ */
+fun StubContainer.computeNamesToBeDeclared(pkgName: String): List<String> {
+
+    fun List<Classifier>.topLevelNamesInPackage() =
+            filter { it.pkg == pkgName }.map { it.topLevelName }
+
+    val classNames = classes.filterIsInstance<ClassStub.Simple>().map { it.classifier }.topLevelNamesInPackage()
+    val enumNames = classes.filterIsInstance<ClassStub.Enum>().map { it.classifier }.topLevelNamesInPackage()
```
SvyatoslavScherbina: This code would be more reliable if it was using `when` over sealed class instead of multiple `filterIsInstance`.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -7,14 +7,30 @@ package org.jetbrains.kotlin.native.interop.gen
...
+        false
+    }
+
+/**
+ * Compute which names will be declared by [StubContainer] in the given [pkgName]
+ */
+fun StubContainer.computeNamesToBeDeclared(pkgName: String): List<String> {
+
+    fun List<Classifier>.topLevelNamesInPackage() =
+            filter { it.pkg == pkgName }.map { it.topLevelName }
```
SvyatoslavScherbina: > `it.pkg == pkgName`

Can this be `false`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
SvyatoslavScherbina: The same argument is applicable to `ConstructorParamStub`.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrVisitor.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package org.jetbrains.kotlin.native.interop.gen
+
+interface StubIrVisitorAux<T, R> {
```
SvyatoslavScherbina: So now it can be renamed to `StubIrVisitor`.
And `aux` parameters could be more like `data` (like in backend IR visitors).

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+                            out("${modality}val $header ${renderGetter(kind.getter)}")
+                        } else {
+                            out("${modality}val $header")
+                            indent {
+                                out(renderGetter(kind.getter))
+                            }
+                        }
+                    }
+                    is PropertyStub.Kind.Var -> {
+                        val variableKind = if (nativeBridges.isSupported(kind.setter)) "var" else "val"
```
SvyatoslavScherbina: `nativeBridges.isSupported(kind.setter)` should be extracted to a variable.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+                                out(renderSetter(kind.setter))
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        // Try to use the provided name. If failed, mangle it with underscore and try again:
+        private tailrec fun getDeclarationName(scope: KotlinScope, name: String): String =
```
SvyatoslavScherbina: The name `getDeclarationName` is somewhat ambiguous in this new larger context.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+                .groupingBy { it.constant.value }
+                .reduce { _, accumulator, element ->
+                    if (element.isMoreCanonicalThan(accumulator)) {
+                        element
+                    } else {
+                        accumulator
+                    }
+                }
+
+        val (canonicalConstants, aliasConstants) = enum.entries.partition { canonicalsByValue[it.constant.value] == it }
```
SvyatoslavScherbina: This logic is not a part of text emitter, it should be applied during stub building.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrElementBuilders.kt
SvyatoslavScherbina: To ensure that 1 is not the case, consider testing with `asSimpleName()` adding back ticks unconditionally.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+
+    private fun renderFunctionReceiver(receiver: ReceiverParameterStub): String {
+        return renderStubType(receiver.type)
+    }
+
+    private fun renderFunctionParameter(parameter: FunctionParameterStub): String {
+        val annotations = if (parameter.annotations.isEmpty())
+            ""
+        else
+            parameter.annotations.joinToString { renderAnnotation(it) } + " "
```
SvyatoslavScherbina: Default separator used in `joinToString` is not applicable here.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -0,0 +1,419 @@
...
+        val type: StubType,
+        override val annotations: List<AnnotationStub> = emptyList(),
+        val isVararg: Boolean = false,
+        val origin: StubOrigin = StubOrigin.None
+) : AnnotationHolder
+
+enum class MemberStubModality {
+    OVERRIDE,
+    OPEN,
+    NONE,
```
SvyatoslavScherbina: There is no such thing as `NONE` modality.
In source code some declarations may have default modality, but in metadata modality is always explicit.
I believe it makes sense to make this modality explicit too, but render it with respect to container's default modality.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+            is ClassStub.Companion -> "companion object"
+            is ClassStub.Enum -> renderClassifierDeclaration(classStub.classifier)
+        }
+        val constructorParams = when (classStub) {
+            is ClassStub.Simple -> renderConstructorParams(classStub.constructorParams)
+            is ClassStub.Companion -> ""
+            is ClassStub.Enum -> renderConstructorParams(classStub.constructorParams)
+        }
+
+        val inheritance = run {
```
SvyatoslavScherbina: The code in this block is rather complicated.
I believe it can be made much simpler by extracting list of supers first (including super class and interfaces), maybe as mutable list to avoid sophisticated concatenations.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+    }
+
+    private fun renderSuperInit(superClassInit: SuperClassInit): String {
+        val parameters = superClassInit.arguments.joinToString(prefix = "(", postfix = ")") { renderValueUsage(it) }
+        return "${renderStubType(superClassInit.type)}$parameters"
+    }
+
+    private fun renderStubType(stubType: StubType): String = when (stubType) {
+        is WrapperStubType -> stubType.kotlinType.render(kotlinFile)
+        is ClassifierStubType -> kotlinFile.reference(stubType.classifier) + renderTypeArguments(stubType.typeArguments)
```
SvyatoslavScherbina: This doesn't apply nullability.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+        AnnotationStub.ObjC.Consumed ->
+            "@CCall.Consumed"
+        is AnnotationStub.ObjC.Constructor ->
+            "@ObjCConstructor(${annotationStub.selector.quoteAsKotlinLiteral()}, ${annotationStub.designated})"
+        is AnnotationStub.ObjC.ExternalClass -> {
+            val protocolGetter = annotationStub.protocolGetter.quoteAsKotlinLiteral()
+            val binaryName = annotationStub.binaryName.quoteAsKotlinLiteral()
+            "@ExternalObjCClass" + when {
+                annotationStub.protocolGetter.isEmpty() && annotationStub.binaryName.isEmpty() -> ""
+                annotationStub.protocolGetter.isEmpty() -> "($binaryName)"
```
SvyatoslavScherbina: Is this correct?

sbogolepov: It looks correct to me. What do you mean?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,670 @@
...
+        AnnotationStub.CCall.CString ->
+            "@CCall.CString"
+        AnnotationStub.CCall.WCString ->
+            "@CCall.WCString"
+        is AnnotationStub.CCall.Symbol ->
+            "@CCall(${annotationStub.symbolName.quoteAsKotlinLiteral()})"
+        is AnnotationStub.CStruct ->
+            "@CStruct(${annotationStub.struct.quoteAsKotlinLiteral()})"
+        is AnnotationStub.CNaturalStruct ->
+            "@CNaturalStruct(${annotationStub.struct.quoteAsKotlinLiteral()})"
```
SvyatoslavScherbina: Doesn't look correct.
Please compare stubs generated for Kotlin/JVM too (e.g. libclang and LLVM bindings).

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
sbogolepov: Yeah, this check can be avoided.

SvyatoslavScherbina: It makes sense to turn it into assertion.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
SvyatoslavScherbina: If you supply only one argument to `@ExternalObjCClass` annotation, it will be bound to its first parameter, `protocolGetter`.

sbogolepov: Shame on me.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,657 @@
...
+                accessor.constant != null -> renderValueUsage(accessor.constant)
+                else -> error("Bridge body for getter was not generated")
+            }
+        }
+
+        is PropertyAccessor.Getter.ArrayMemberAt -> "arrayMemberAt(${accessor.offset})"
+
+        is PropertyAccessor.Getter.MemberAt -> {
+            if (accessor.typeArguments.isEmpty()) {
+                "memberAt(${accessor.offset})"
```
SvyatoslavScherbina: The presence of `.value` shouldn't be deduced from emptiness of type arguments.
Additional flag better be added for this.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -286,6 +250,14 @@ private fun Type.isStret(target: KonanTarget): Boolean {
     }
 }
 
+private fun deprecatedInit(className: String, initParameterNames: List<String>, factory: Boolean): AnnotationStub {
+    val replacement = if (factory) "$className.create" else className
+    val replacementKind = if (factory) "factory method" else "constructor"
+    val replaceWith = "$replacement(${initParameterNames.joinToString { it }})"
```
SvyatoslavScherbina: Doesn't this require `it.asSimpleName()`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -0,0 +1,654 @@
...
+        val annotations = if (parameter.annotations.isEmpty())
+            ""
+        else
+            parameter.annotations.joinToString(separator = " ") { renderAnnotation(it) } + " "
+        val vararg = if (parameter.isVararg) "vararg " else ""
+        return "$annotations$vararg${parameter.name.asSimpleName()}: ${renderStubType(parameter.type)}"
+    }
+
+    private fun renderMemberModality(modality: MemberStubModality, container: StubContainer?): String =
+            if (container?.let { it.defaultMemberModality == modality } == true) {
```
SvyatoslavScherbina: Is it equivalent to `container?.defaultMemberModality == modality`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -446,139 +423,150 @@ abstract class ObjCContainerStub(stubGenerator: StubGenerator,
...
             val baseClass = container.baseClass
             val baseClassifier = if (baseClass != null) {
-                stubGenerator.declarationMapper.getKotlinClassFor(baseClass, isMeta)
+                context.getKotlinClassFor(baseClass, isMeta)
             } else {
                 if (isMeta) KotlinTypes.objCObjectBaseMeta else KotlinTypes.objCObjectBase
             }
-
-            supers.add(baseClassifier.type)
+            interfaces += WrapperStubType(baseClassifier.type)
```
SvyatoslavScherbina: Not an interface actually.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -446,139 +423,150 @@ abstract class ObjCContainerStub(stubGenerator: StubGenerator,
...
+        context,
         container,
-        metaContainerStub = object : ObjCContainerStub(stubGenerator, container, metaContainerStub = null) {}
-) {
-    override fun generate(context: StubGenerationContext) =
-            metaContainerStub!!.generate(context) + "" + super.generate(context)
+        metaContainerStub = object : ObjCContainerStubBuilder(context, container, metaContainerStub = null) {
+
+            override fun build(): List<StubIrElement> {
+                val (properties, methods) = buildBody()
```
SvyatoslavScherbina: This `buildBody` + `ClassStub.Simple()` combination looks nearly the same here and in next two classes.
Isn't it possible to extract the common logic?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
```diff
@@ -0,0 +1,352 @@
...
+
+class BridgeGenerationComponentsBuilder(
+        val getterToBridgeInfo: MutableMap<PropertyAccessor.Getter, BridgeGenerationComponents.GlobalGetterBridgeInfo> = mutableMapOf(),
+        val setterToBridgeInfo: MutableMap<PropertyAccessor.Setter, BridgeGenerationComponents.GlobalSetterBridgeInfo> = mutableMapOf(),
+        val enumToTypeMirror: MutableMap<ClassStub.Enum, TypeMirror> = mutableMapOf(),
+        val wCStringParameters: MutableSet<FunctionParameterStub> = mutableSetOf(),
+        val cStringParameters: MutableSet<FunctionParameterStub> = mutableSetOf()
+) {
+    fun build(): BridgeGenerationComponents = object : BridgeGenerationComponents {
+        override val getterToBridgeInfo = this@BridgeGenerationComponentsBuilder.getterToBridgeInfo
```
SvyatoslavScherbina: The common convention for builders and `build` methods is to copy mutable data.
E.g. by adding `.toMap()` here etc.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
```diff
@@ -0,0 +1,352 @@
...
+class StubIrBuilder(
+        private val context: StubIrContext,
+        private val verbose: Boolean = false
+) {
+
+    private val configuration = context.configuration
+    private val nativeIndex: NativeIndex = context.nativeIndex
+
+    private val classes = mutableListOf<ClassStub>()
+    private val functions = mutableListOf<FunctionalStub>()
```
SvyatoslavScherbina: Is `FunctionalStub` really required here instead of `FunctionStub`?

sbogolepov: Nope, it's a leftover from a refactoring.

PR #3111, opened 2019-06-25
Fix import packages in external tests

PavelPunegov:
Replace all occurrences of the package name in full qualified names in the line.

### DIFF
--- a/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
+++ b/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
@@ -443,9 +443,12 @@ class RunExternalTestGroup extends OldKonanTest {
                     }
                     if (line.contains("$pkg.") && ! (line =~ packagePattern || line =~ importRegex)
                             && ! vars.contains(pkg)) {
-                        def idx = line.indexOf("$pkg")
-                        if (! (idx > 0 && Character.isJavaIdentifierPart(line.charAt(idx - 1))) ) {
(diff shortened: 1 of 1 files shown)


buildSrc/plugins/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -443,9 +443,11 @@ class RunExternalTestGroup extends OldKonanTest {
                     }
                     if (line.contains("$pkg.") && ! (line =~ packagePattern || line =~ importRegex)
                             && ! vars.contains(pkg)) {
-                        def idx = line.indexOf("$pkg")
-                        if (! (idx > 0 && Character.isJavaIdentifierPart(line.charAt(idx - 1))) ) {
-                            line = line.substring(0, idx) + "$sourceName.$pkg" + line.substring(idx + pkg.length())
+                        def idx = 0
+                        while ((idx = line.indexOf("$pkg", idx)) > 0) {
```
SvyatoslavScherbina: Isn't this loop infinite if `line` contains `pkg`?

PavelPunegov: Yes, it is. Index increment is missing. Will fix it.

SvyatoslavScherbina: Still may happen to be infinite, if `sourceName.length() + 1 >= pkg.length()`.

PR #3116, opened 2019-06-27
New zero-terminated utf8 to String conversion API

qurbonzoda:


runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -9,52 +9,103 @@ package kotlin.native
...
-@SymbolName("Kotlin_ByteArray_stringFromUtf8")
-internal external fun ByteArray.stringFromUtf8Impl(start: Int, size: Int) : String
+/**
+ * Decodes a string from the bytes in UTF-8 encoding in this array.
+ * Bytes following the first occurrence of `0` byte, if it occurs, are not decoded.
+ *
+ * Malformed byte sequences are replaced by the replacement char `\uFFFD`.
+ */
+@SinceKotlin("1.3")
+public fun ByteArray.decodeZeroTerminatedToString() : String {
```
qurbonzoda: This new API should be discussed

runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -9,52 +9,103 @@ package kotlin.native
 /**
  * Converts an UTF-8 array into a [String]. Replaces invalid input sequences with a default character.
  */
+@Deprecated("Use decodeToString or decodeZeroTerminatedToString instead")
```
ilya-g: The message with two alternatives is fine, but I believe we should provide `ReplaceWith` to `decodeZeroTerminatedToString` as a behavior preserving option. Then one could erase "ZeroTerminated" if `decodeToString` is enough.

qurbonzoda: `ReplaceWith(...)` is not provided here to push user to choose the right option mindfully.

runtime/src/main/cpp/KString.cpp
```diff
@@ -49,11 +49,8 @@ OBJ_GETTER(utf8ToUtf16Impl, const char* rawString, const char* end, uint32_t cha
 }
 
 template<utf16to8 conversion>
-OBJ_GETTER(utf16ToUtf8Impl, KString thiz, KInt start, KInt size) {
+OBJ_GETTER(unsafeUtf16ToUtf8Impl, KString thiz, KInt start, KInt size) {
```
qurbonzoda: Methods from which bounds checking is removed are renamed to `unsafe*`.

backend.native/tests/runtime/text/utf8.kt
```diff
@@ -353,13 +434,53 @@ fun test8to16CustomBorders() {
...
+    checkZeroTerminatedUtf8to16Throws("\uFFFD", intArrayOf(-16, -97, -104, 0, '1'.toInt()))
+    checkZeroTerminatedUtf8to16Throws("\uFFFD1", intArrayOf(-16, -97, -104, '1'.toInt(), 0, -16, -97, -104))
+}
+
+fun testZeroTerminated8To16CustomBorders() {
+    val array = intArrayOf('a'.toInt(), 'a'.toInt(), 'a'.toInt(), 0, 'b'.toInt(), 'b'.toInt(), 'b'.toInt(), 0)
+    checkValidZeroTerminatedUtf8to16("aaa", array, 0, 5)
+    checkValidZeroTerminatedUtf8to16("a", array, 2, 2)
+    checkValidZeroTerminatedUtf8to16("", array, 3, 2)
+    checkValidZeroTerminatedUtf8to16("bb", array, 4, 2)
```
qurbonzoda: Here the terminating zero is searched starting from the given `start` index. Do we want to persist this behaviour in the new API?

runtime/src/main/cpp/Natives.h
```diff
@@ -137,7 +137,7 @@ KInt Kotlin_String_compareTo(KString thiz, KString other);
 KInt Kotlin_String_compareToIgnoreCase(KString thiz, KConstRef other);
 KChar Kotlin_String_get(KString thiz, KInt index);
 OBJ_GETTER(Kotlin_String_fromUtf8Array, KConstRef array, KInt start, KInt size);
```
qurbonzoda: `fromUtf8Array` seems to be an old api and this line should be removed.

backend.native/tests/interop/platform_zlib.kt
```diff
@@ -21,7 +21,7 @@ fun main(args: Array<String>) = memScoped {
         }.ptr
 
         if (inflateInit2(z, -15) == Z_OK && inflate(z, Z_FINISH) == Z_STREAM_END && inflateEnd(z) == Z_OK)
-            println(buffer.stringFromUtf8())
+            println(buffer.decodeZeroTerminatedToString())
```
olonho: `decodeZeroTerminatedToString` looks somewhat ugly. Nobody care if it is decoding or copying, just `fromTerminatedUtf8` looks sufficient.

backend.native/tests/runtime/text/utf8.kt
```diff
@@ -70,51 +139,115 @@ fun <T: Any> checkThrows(e: KClass<T>, string: String, action: () -> Unit) {
...
 }
 
-fun checkUtf16to8Throws(string: String, start: Int, size: Int) =
-        checkThrows(IllegalCharacterConversionException::class, string) {
-            string.toUtf8OrThrow(start, size)
-        }
 
-fun checkUtf8to16Throws(string: String, array: IntArray)
-        = checkThrows(IllegalCharacterConversionException::class, string) {
+// utils for checking malformed UTF-8 -> UTF-16 conversion
```
olonho: // We start comments with capital letter and end with dot and avoid abbreviations.

backend.native/tests/runtime/text/utf8.kt
```diff
@@ -70,51 +139,115 @@ fun <T: Any> checkThrows(e: KClass<T>, string: String, action: () -> Unit) {
...
 // region UTF-16 -> UTF-8 conversion
 fun test16to8() {
-    // Test manual conversion with replacement.
     // Valid strings.
-    checkUtf16to8Replacing("Hello", intArrayOf('H'.toInt(), 'e'.toInt(), 'l'.toInt(), 'l'.toInt(), 'o'.toInt()))
-    checkUtf16to8Replacing("Привет", intArrayOf(-48, -97, -47, -128, -48, -72, -48, -78, -48, -75, -47, -126))
-    checkUtf16to8Replacing("\uD800\uDC00", intArrayOf(-16, -112, -128, -128))
-    checkUtf16to8Replacing("", intArrayOf())
+    checkValidUtf16to8("Hello", intArrayOf('H'.toInt(), 'e'.toInt(), 'l'.toInt(), 'l'.toInt(), 'o'.toInt()))
+    checkValidUtf16to8("Привет", intArrayOf(-48, -97, -47, -128, -48, -72, -48, -78, -48, -75, -47, -126))
```
olonho: Why not proper uintArrayOf() with hex unsigned values?

backend.native/tests/runtime/text/utf8.kt
```diff
@@ -247,21 +352,22 @@ fun testPrint() {
 
     // https://github.com/JetBrains/kotlin-native/issues/1091
     val array = byteArrayOf(0xF0.toByte(), 0x9F.toByte(), 0x98.toByte(), 0xA5.toByte())
-    val badStr = array.stringFromUtf8()
-    assertEquals(2, badStr.length)
-    println(badStr)
+    convertUtf8to16(array) { badStr ->
```
olonho: badStr -> badString

runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -9,52 +9,103 @@ package kotlin.native
...
+ * @param startIndex the beginning (inclusive) of the subrange to decode, 0 by default.
+ * @param endIndex the end (exclusive) of the subrange to decode, size of this array by default.
+ * @param throwOnInvalidSequence specifies whether to throw an exception on malformed byte sequence or replace it by the replacement char `\uFFFD`.
+ *
+ * @throws IndexOutOfBoundsException if [startIndex] is less than zero or [endIndex] is greater than the size of this array.
+ * @throws IllegalArgumentException if [startIndex] is greater than [endIndex].
+ * @throws CharacterCodingException if the byte array contains malformed UTF-8 byte sequence and [throwOnInvalidSequence] is true.
+ */
+@SinceKotlin("1.3")
+public fun ByteArray.decodeZeroTerminatedToString(
```
olonho: One of unfortunate things is that we use ByteArray here. Technically, anything, which can get a byte by index (i.e. interop pointers) shall be usable for such an API. @SvyatoslavScherbina

backend.native/tests/runtime/text/utf8.kt
ilya-g: Yes, it seems logical: when we're working with a subrange of an array, we do not look to the contents out of this subrange.

runtime/src/main/kotlin/kotlin/native/Text.kt
```diff
@@ -71,11 +122,11 @@ public fun String.toUtf8() : ByteArray {
...
 public fun String.toUtf8(start: Int = 0, size: Int = this.length) : ByteArray {
     checkBoundsIndexes(start, start + size, this.length)
-    return toUtf8Impl(start, size)
+    return unsafeToUtf8Impl(start, size)
 }
 
-@SymbolName("Kotlin_String_toUtf8")
-internal external fun String.toUtf8Impl(start: Int, size: Int) : ByteArray
+@SymbolName("Kotlin_String_unsafeToUtf8")
+internal external fun String.unsafeToUtf8Impl(start: Int, size: Int) : ByteArray
```
ilya-g: Since we're renaming it anyway, I propose to name it `unsafeStringToUtf8Impl` to be consistent with `unsafeStringFromUtf8Impl`

runtime/src/main/kotlin/kotlin/native/Text.kt
ilya-g: I think it's ok to provide an automatic replacement to the new function with the same behavior, if the migration to the other alternative is easy enough.

backend.native/tests/interop/platform_zlib.kt
ilya-g: We want this name to follow the convention _"somethingToString"_. 
- "ToString" shows the target type one gets in the result
- "something" can be either a verb, like `decode`, or a description of source content, like `hex`,  `utf8` or `terminatedUtf8`

So another alternative could be `terminatedUtf8ToString()`. Though this would be at odds with our original goal not to have "utf8" in the name of `decodeToString` to allow providing an overload with the same name in kotlinx-io taking another encoding as an argument. 

Can we count on never needing encoding other than utf-8 in this use case?

backend.native/tests/runtime/text/utf8.kt
ilya-g: or even `ubyteArrayOf`

runtime/src/main/cpp/Natives.h
ilya-g: What is the purpose of this header declaration? It doesn't seem to have a definition.

runtime/src/main/kotlin/kotlin/text/Strings.kt
```diff
@@ -206,7 +206,7 @@ public actual fun CharSequence.repeat(n: Int): String {
         else -> {
             when (length) {
                 0 -> ""
-                1 -> this[0].let { char -> fromCharArray(CharArray(n) { char }, 0, n) }
+                1 -> this[0].let { char -> unsafeFromCharArray(CharArray(n) { char }, 0, n) }
```
ilya-g: We can use the public API `CharArray.concatToString()` (the one for entire array) in such cases. It does the same under the hood.

runtime/src/main/kotlin/kotlin/native/Text.kt
ilya-g: The points to discuss:
- Naming
- Placement: `kotlin.text` or `kotlin.native`

backend.native/tests/runtime/text/utf8.kt
ilya-g: It looks like a comment for designating a group of functions, so it might follow the format:
`// ---------------------------- comment ----------------------------`

runtime/src/main/kotlin/kotlin/native/Text.kt
olonho: Or kotlinx.cinterop. It is actually seems mostly used there.

backend.native/tests/interop/platform_zlib.kt
olonho: We must have similar convertors from UTF-16 and UTF-32. https://github.com/JetBrains/kotlin-native/blob/bff177a0728ac26f0839f9e52dddd49c88a18591/Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt#L509 have those extensions for C pointers. Maybe toKStringFromUtf[8|16|32] should be OK.

ilya-g: We discussed and liked this idea.

Given that there's `CPointer<ByteVar>.toKString()` in `kotlinx.cinterop`, we can provide the overload `ByteArray.toKString()` beside.

runtime/src/main/kotlin/kotlin/native/Text.kt
qurbonzoda: Moved to kotlinx.cinterop

runtime/src/main/kotlin/kotlin/native/Text.kt
qurbonzoda: Now two alternatives, `toKString` and `decodeToString`, are mentioned in message.
`ReplaceWith` holds `toKString`.

qurbonzoda: +1

runtime/src/main/cpp/Natives.h
qurbonzoda: This file have already been refactored in master, unnecessary declarations removed.

backend.native/tests/interop/platform_zlib.kt
qurbonzoda: done

backend.native/tests/runtime/text/utf8.kt
qurbonzoda: done

runtime/src/main/kotlin/kotlin/native/Text.kt
qurbonzoda: Renamed to `unsafeStringFromUtf8

runtime/src/main/kotlin/kotlin/text/Strings.kt
qurbonzoda: +1

backend.native/tests/runtime/text/utf8.kt
qurbonzoda: Tests here were not modified. I just refactored the utils to run a test for multiple methods, 
e.g. `toUtf8()`, `encodeToByteArray`, `toUtf8OrThrow` and `encodeToByteArray(throwOnInvalidSequence = true)` must all return the same UFT-8 bytes for a valid string.

Converting all these values to hex form seems not to be time-worthy.

backend.native/tests/runtime/text/utf8.kt
qurbonzoda: +1

runtime/src/main/kotlin/kotlin/native/Text.kt
qurbonzoda: Named `toKString()`

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt
```diff
@@ -554,6 +554,60 @@ public fun CPointer<IntVar>.toKStringFromUtf32(): String {
...
+ *
+ * @param startIndex the beginning (inclusive) of the subrange to decode, 0 by default.
+ * @param endIndex the end (exclusive) of the subrange to decode, size of this array by default.
+ * @param throwOnInvalidSequence specifies whether to throw an exception on malformed byte sequence or replace it by the replacement char `\uFFFD`.
+ *
+ * @throws IndexOutOfBoundsException if [startIndex] is less than zero or [endIndex] is greater than the size of this array.
+ * @throws IllegalArgumentException if [startIndex] is greater than [endIndex].
+ * @throws CharacterCodingException if the byte array contains malformed UTF-8 byte sequence and [throwOnInvalidSequence] is true.
+ */
+@UseExperimental(ExperimentalStdlibApi::class)
```
qurbonzoda: It is because `decodeToString` is still experimental on JVM.

I couldn't get why cinterop part of K/N is compilable on JVM.

runtime/src/main/kotlin/kotlin/text/Strings.kt
```diff
@@ -221,7 +221,7 @@ public actual fun CharSequence.repeat(n: Int): String {
 /**
  * Converts the characters in the specified array to a string.
  */
-public actual fun String(chars: CharArray): String = fromCharArray(chars, 0, chars.size)
+public actual fun String(chars: CharArray): String = unsafeStringFromCharArray(chars, 0, chars.size)
```
ilya-g: `concatToString()` ?

qurbonzoda: For some reason I thought it was a partial conversion, hence left `unsafeStringFromCharArray` to avoid bounds checking.

qurbonzoda: done!

Issue #2562, opened 2019-01-18
How to determine build type in kotlin-multiplatform project


dfpalomar:
Hello. I'm struggling to find a solution for a problem that seems simple.

I’m working on a multiplaform project, iOS and JVM (I’m not targeting Android directly) and depending on the build type (debug or release) I want to configure the logging level (i.e. to print only errors in release). Since there is no a _BuildConfig_ class available (like the one generated by the Android gradle plugin), how can I know from `commonMain` the build type?

msink:
#952

SvyatoslavScherbina:
Hello.
You can configure debug and release binaries to include alternative source sets.
See the example here: https://github.com/ilmat192/kotlin-native-gradle-samples/tree/master/debug-logger
Debug builds include [commonDebug](https://github.com/ilmat192/kotlin-native-gradle-samples/tree/master/debug-logger/src/commonDebug/kotlin) source set, release builds include [commonRelease](https://github.com/ilmat192/kotlin-native-gradle-samples/tree/master/debug-logger/src/commonRelease/kotlin) instead.

steve-the-edwards:
Seems like a lot of gradle work to just carry separate source sets.  Can we define a constant similar (a la BuildConfig.DEBUG) that is derived from the project build property in the build?

SvyatoslavScherbina:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native/-platform/is-debug-binary.html

SvyatoslavScherbina:
It should. Please report an issue if you discover that it doesn't.

PR #3129, opened 2019-06-28
Very preliminary relaxed mode draft.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -273,6 +279,14 @@ constexpr const char* MemoryStatistic::indexToName[];
 
 #endif  // COLLECT_STATISTIC
 
+inline bool isPermanentOrFrozen(ContainerHeader* container) {
```
SvyatoslavScherbina: Is it different from `isPermanentOrFrozen` declared in Memory.h?

runtime/src/main/cpp/MemoryPrivate.hpp
```diff
@@ -21,8 +21,12 @@
 
 extern "C" {
 
+MODEL_VARIANTS(void, ReleaseHeapRef, const ObjHeader* object);
+
 void AddRefFromAssociatedObject(const ObjHeader* object) RUNTIME_NOTHROW;
-void ReleaseRefFromAssociatedObject(const ObjHeader* object) RUNTIME_NOTHROW;
+void ReleaseRefFromAssociatedObject(const ObjHeader* object);
```
SvyatoslavScherbina: Is `RUNTIME_NOTHROW` removed intentionally?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1852,8 +1819,8 @@ OBJ_GETTER(readHeapRefLocked, ObjHeader** location, int32_t* spinlock) {
   if (container != nullptr)
     incrementRC<true>(container);
   unlock(spinlock);
-  if (container != nullptr)
-    enqueueDecrementRC</* CanCollect = */ true>(container);
+  if (value != nullptr)
+    ReleaseHeapRef(value);
```
SvyatoslavScherbina: Can this remove object before return? (like it was in `swapHeapRefLocked`).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1825,12 +1792,12 @@ OBJ_GETTER(swapHeapRefLocked,
     shallRelease = oldValue != nullptr;
   }
   unlock(spinlock);
-  if (shallRelease) {
-    releaseHeapRef(oldValue);
-  }
   // No need to rememberNewContainer(), as oldValue is already
   // present on this worker.
```
SvyatoslavScherbina: Btw, is this correct? Doesn't seem so when `oldValue != expectedValue`.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2393,38 +2482,55 @@ OBJ_GETTER(InitSharedInstanceStrict,
...
   zeroHeapRef(location);
 }
 
-void ZeroStackRef(ObjHeader** location) {
-  zeroStackRef(location);
+void ZeroStackRefStrict(ObjHeader** location) {
+  zeroStackRef<true>(location);
+}
+void ZeroStackRefRelaxed(ObjHeader** location) {
+  zeroStackRef<true>(location);
```
SvyatoslavScherbina: Is `true` correct here?

runtime/src/main/kotlin/kotlin/native/Platform.kt
```diff
@@ -58,6 +66,19 @@ public object Platform {
...
+    /**
+     * Memory model binary was compiled with.
+     */
+    public val memoryModel: MemoryModel
+        get() = MemoryModel.values()[Platform_getMemoryModel()]
+
+    /**
+     * If binary was compiled in debug mode.
+     */
+    public val isDebugBinary: Boolean
```
SvyatoslavScherbina: Maybe related: #2562

runtime/src/main/cpp/Memory.cpp
olonho: Comment was indeed wrong.

SvyatoslavScherbina: Comment is still wrong, and the behaviour is wrong too.
The following test crashes:
```kotlin
import kotlin.test.*

val atomicRef = AtomicReference<Any?>(Any().freeze())
val semaphore = AtomicInt(0)

@Test fun runTest() {
    semaphore.value = 0
    val worker = Worker.start()
    val future = worker.execute(TransferMode.SAFE, { null }) {
        val value = atomicRef.compareAndSwap(null, null)
        semaphore.increment()
        while (semaphore.value != 2) {}
        println(value.toString() != "")
    }
    while (semaphore.value != 1) {}
    atomicRef.value = null
    semaphore.increment()
    future.result
    worker.requestTermination().result
}
```
(based on `worker10`).
Generally `atomicRef.compareAndSwap(Any().freeze(), null)` is the same as `atomicRef.value`, and if the latter requires `rememberNewContainer`, then the former requires this too.

olonho: Thanks for bringing my attention to this problem. Fixed.

backend.native/tests/runtime/workers/worker10.kt
```diff
@@ -163,4 +163,22 @@ val stableHolder2 = StableRef.create(("hello" to "world").freeze())
     semaphore.increment()
     future.result
     worker.requestTermination().result
+}
+
+@Test fun runTest6() {
```
SvyatoslavScherbina: This test doesn't fail even without the fix applied.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1794,10 +1794,13 @@ OBJ_GETTER(swapHeapRefLocked,
...
   UpdateReturnRef(OBJ_RESULT, oldValue);
-  // No need to rememberNewContainer(), as if `oldValue` is not null - it is explicitly released
-  // anyway, and thus can not escape GC.
   if (shallRelease) {
+    // No need to rememberNewContainer() on this path, as if `oldValue` is not null - it is explicitly released
+    // anyway, and thus can not escape GC.
     ReleaseHeapRef(oldValue);
+  } else {
+    if (IsStrictMemoryModel && oldValue != expectedValue)
+      rememberNewContainer(oldValue->container());
```
SvyatoslavScherbina: Isn't it too late to perform this after unlock?

backend.native/tests/runtime/workers/worker10.kt
olonho: Yeah, better use fresh atomic ref here.

omiwrench:
Could we get some documentation on how to enable this relaxed mode?

PR #3136, opened 2019-07-01
Fix incorrect package replacement

PavelPunegov:
should check that current occurrence  is not a part of identifier. Otherwise it can incorrectly replace one package with another one with name being substring of the first one. ("another"/"other" packages)

### DIFF
--- a/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
+++ b/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
@@ -444,11 +444,10 @@ class RunExternalTestGroup extends OldKonanTest {
                     if (line.contains("$pkg.") && ! (line =~ packagePattern || line =~ importRegex)
                             && ! vars.contains(pkg)) {
                         def idx = 0
-                        while ((idx = line.indexOf(pkg, idx)) >= 0) {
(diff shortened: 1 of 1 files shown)


build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -444,11 +444,10 @@ class RunExternalTestGroup extends OldKonanTest {
                     if (line.contains("$pkg.") && ! (line =~ packagePattern || line =~ importRegex)
                             && ! vars.contains(pkg)) {
                         def idx = 0
-                        while ((idx = line.indexOf(pkg, idx)) >= 0) {
-                            if (!Character.isJavaIdentifierPart(line.charAt(idx - 1))) {
-                                line = line.substring(0, idx) + "$sourceName.$pkg" + line.substring(idx + pkg.length())
-                                idx += sourceName.length() + pkg.length() + 1
-                            }
+                        while ((idx = line.indexOf(pkg, idx)) >= 0 &&
+                                !Character.isJavaIdentifierPart(line.charAt(idx - 1))) {
```
SvyatoslavScherbina: Is it correct to skip the remaining occurrences when this becomes false?

PR #3138, opened 2019-07-01
[kotlin compiler][update] 1.3.50-dev-1637

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -243,6 +243,7 @@ internal val allLoweringsPhase = namedIrModulePhase(
                                 autoboxPhase then
                                 returnsInsertionPhase
                 ) then
+                patchDeclarationParents2Phase then //TODO: this was added for compiler's [1.3.50-dev-1637] update and should be removed ASAP.
```
alexander-gorshenev: Have @homuroll take a look at this.

PR #3141, opened 2019-07-02
Add more mac platform libs

ilmat192:


platformLibs/src/platform/osx/builtin.def
```diff
@@ -1,5 +1,6 @@
+depends = posix
```
olonho: How come?

ilmat192: Automatically :) The `set_depends` utility groups all special declarations added by Clang into each translation unit into a separate virtual header. Our current convention assumes that this virtual header is owned by the `posix` library and all other libraries (e.g. `darwin` and all new ones) import it. Thus this dependency appears.

olonho: I think dep shall be reverted and posix depend on this one, not vice versa.

ilmat192: It makes sense but I'm not sure if such dependency change can affect users or not. @SvyatoslavScherbina WDYT?

SvyatoslavScherbina: I think that this dependency is mostly technical, and it is somewhat easier to keep it in current state.

olonho: OK, let's go the current way.

PR #3142, opened 2019-07-02
Fixed bug with empty block returning Nothing + test

homuroll:


backend.native/tests/codegen/branching/empty_nothing_block.kt
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package codegen.branching.empty_nothing_block
+
+import kotlin.test.*
+
+@Test fun runTest() {
+    if (true) else true
```
SvyatoslavScherbina: IR produced for this code doesn't seem correct, so there is not much sense in supporting it in code generator.

olonho: ``` if (false) ;``` gives the same problematic code.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1724,8 +1724,11 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
                 }
             }
 
-            assert(value.type.isUnit())
-            return codegen.theUnitInstanceRef.llvm
+            return when {
+                value.type.isUnit() -> codegen.theUnitInstanceRef.llvm
+                value.type.isNothing() -> codegen.kNothingFakeValue
```
olonho: Maybe we shall use codegen.theUnitInstanceRef.llvm for now, and ask PS2IR to provide better type info.

homuroll:
Created issue to clear things up: https://youtrack.jetbrains.com/issue/KT-32581

SvyatoslavScherbina:
The issue is marked as fixed now. Shouldn't we close this PR?

PR #3144, opened 2019-07-02
Actuals for bit extension functions for signed ints

### LINKED ISSUE #2949: LLVM intrinsics for bit operations
The following bit operations are going to be introduced into the standard library ([KT-12749](https://youtrack.jetbrains.com/issue/KT-12749)):
- `countOneBits`
- `countLeadingZeroBits`, `countTrailingZeroBits`
- `takeHighestOneBit`, `takeLowestOneBit`

It seems to be possible to implement these operations with LLVM instructions as following:
- `countOneBits`: [`llvm.ctpop`](https://llvm.org/docs/LangRef.html#llvm-ctpop-intrinsic)
- `countLeadingZeroBits`: [`llvm.ctlz`](https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic) with `is_zero_undef` parameter equal to 0.
- `countTrailingZeroBits`: [`llvm.cttz`](https://llvm.org/docs/LangRef.html#llvm-cttz-intrinsic) with `is_zero_undef` parameter equal to 0.
- `takeHighestOneBit`, `takeLowestOneBit` can be implemented with `ctlz` and `cttz`, though since they require special handling of zero input, they can call these instructions with `is_zero_undef != 0`, e.g.

    ```
    fun Int.takeHighestOneBit(): Int = 
        if (this == 0) 0 else 1.shl(Int.SIZE_BITS - 1 - llvm.ctlz(this, 1))
    ```

ilya-g:
Closes #2949


gradle.properties
```diff
@@ -20,8 +20,8 @@ buildKotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds
 remoteRoot=konan_tests
 kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-1741,branch:default:any,pinned:true/artifacts/content/maven
 kotlinVersion=1.3.50-dev-1741
-kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-1741,branch:default:any,pinned:true/artifacts/content/maven
-kotlinStdlibVersion=1.3.50-dev-1741
+kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_dev_Compiler),number:1.3.50-dev-1429,branch:default:any,pinned:any/artifacts/content/maven
+kotlinStdlibVersion=1.3.50-dev-1429
```
ilya-g: I'll remove these changes when merging to kotlin-mirror/master

runtime/src/main/kotlin/kotlin/Numbers.kt
```diff
@@ -100,4 +100,180 @@ public actual inline fun Float.Companion.fromBits(bits: Int): Float = kotlin.fro
...
+        if (this == 0) Int.SIZE_BITS else countTrailingZeroBits(this)
+
+/**
+ * Returns a number having a single bit set in the position of the most significant set bit of this [Int] number,
+ * or zero, if this number is zero.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public actual fun Int.takeHighestOneBit(): Int =
+        if (this == 0) 0 else 1.shl(Int.SIZE_BITS - 1 - countLeadingZeroBits(this))
```
olonho: I'd use 32 here, as access to companion object looks like an overkill here.

olonho: As we do not have constant propagation, and those functions intended to be on perf sensitive paths.

runtime/src/main/kotlin/kotlin/Numbers.kt
```diff
@@ -100,4 +100,180 @@ public actual inline fun Float.Companion.fromBits(bits: Int): Float = kotlin.fro
...
+ * Rotating the number right by a negative bit count is the same as rotating it left by the negated bit count:
+ * `number.rotateRight(-n) == number.rotateLeft(n)`
+ *
+ * Rotating by a multiple of [Int.SIZE_BITS] (32) returns the same number, or more generally
+ * `number.rotateRight(n) == number.rotateRight(n % 32)`
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public actual fun Int.rotateRight(bitCount: Int): Int =
+        shl(Int.SIZE_BITS - bitCount) or ushr(bitCount)
```
olonho: How does it handle incorrect values?

runtime/src/main/kotlin/kotlin/Numbers.kt
```diff
@@ -100,4 +100,180 @@ public actual inline fun Float.Companion.fromBits(bits: Int): Float = kotlin.fro
...
+@SymbolName("Kotlin_Long_countTrailingZeroBits")
+private external fun countTrailingZeroBits(value: Long): Int
+
+/**
+ * Counts the number of consecutive least significant bits that are zero in the binary representation of this [Long] number.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public actual fun Long.countTrailingZeroBits(): Int =
+        if (this == 0L) Long.SIZE_BITS else countTrailingZeroBits(this)
```
olonho: Same as with Int versions.

runtime/src/main/kotlin/kotlin/Numbers.kt
ilya-g: Ok, I'll inline these constants manually, but leave a TODO to replace them back when constant propagation is ready. Btw, is there an issue about it to follow?

runtime/src/main/kotlin/kotlin/Numbers.kt
ilya-g: All `bitCount` values are correct. Shift operations take five least significant bits of bitCount argument, so for example `x.shl(-1) == x.shl(31)`

PR #3145, opened 2019-07-03
Don't explicitly resolve 'kotlinCommonSources' configuration

ddolovov:
in ':backend.native:unzipStdlibSources' task. This breaks composite build with Kotlin.


ilya-g:
Let's merge already, and in kotlin-mirror/master too

PR #3147, opened 2019-07-03
Rewriting infrastructure to get right measurement of GC work

LepilkinaElena:


performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -19,21 +19,38 @@ package org.jetbrains.benchmarksLauncher
...
+    fun add(name: String, benchmark: BenchmarkEntry) {
+        benchmarks[name] = benchmark
+    }
+
+    fun runBenchmark(benchmarkInstance: Any?, benchmark: BenchmarkEntry, repeatNumber: Int): Long {
+        var i = repeatNumber
+        return if (benchmark is InstanceBenchmarkEntry) {
+            cleanup()
+            measureNanoTime {
+                while (i-- > 0) benchmark.lambda(benchmarkInstance!!)
```
olonho: Maybe move `if (benchmark is InstanceBenchmarkEntry)` there?

LepilkinaElena: I made this  to avoid mostly all unconnected actions to make measurement as accurate as possible

runtime/src/main/cpp/Memory.cpp
```diff
@@ -70,14 +70,14 @@ static_assert(sizeof(ContainerHeader) % kObjectAlignment == 0, "sizeof(Container
...
-constexpr size_t kGcThreshold = 16 * 1024;
+constexpr size_t kGcThreshold = 8 * 1024;
 #if GC_ERGONOMICS
 // Ergonomic thresholds.
 // If GC to computations time ratio is above that value,
 // increase GC threshold by 1.5 times.
 constexpr double kGcToComputeRatioThreshold = 0.5;
 // Never exceed this value when increasing GC threshold.
-constexpr size_t kMaxErgonomicThreshold = 4 * 1024;
+constexpr size_t kMaxErgonomicThreshold = 16 * 1024;
```
olonho: And how it changes perf?

performance/cinterop/src/main/kotlin/main.kt
```diff
@@ -21,29 +21,26 @@ import org.jetbrains.structsBenchmarks.*
...
-                    "stringToC" to stringBenchmark::stringToCBenchmark,
-                    "stringToKotlin" to stringBenchmark::stringToKotlinBenchmark,
-                    "intMatrix" to intMatrixBenchmark::intMatrixBenchmark,
-                    "int" to intBenchmark::intBenchmark,
-                    "boxedInt" to boxedIntBenchmark::boxedIntBenchmark
+                    "macros" to FunctionBenchmarkEntry(::macrosBenchmark),
+                    "struct" to FunctionBenchmarkEntry(::structBenchmark),
+                    "union" to FunctionBenchmarkEntry(::unionBenchmark),
+                    "enum" to FunctionBenchmarkEntry(::enumBenchmark),
+                    "stringToC" to InstanceBenchmarkEntry.create(::StringBenchmark, { stringToCBenchmark() }),
```
olonho: FunctionBenchmarkEntry/InstanceBenchmarkEntry doesn't look too descriptive.
Maybe just  BenchmarkEntry with two different constructors? Or BenchmarkEntryWithInit?

runtime/src/main/cpp/Memory.cpp
LepilkinaElena: 8 and 16 show better results on benchmarks I run, I'll see result of full set on TC.

performance/cinterop/src/main/kotlin/main.kt
LepilkinaElena: I have different lambdas types and I need to separate them, may be then AbstractBenchmarkEntry for interface and BenchmarkEntry and BenchmarkEntryWithInit for children?

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -140,7 +141,6 @@ open class BenchmarkingPlugin: Plugin<Project> {
                     runTask!!.args(
                         "-w", nativeWarmup,
                         "-r", attempts,
-                        "-o", buildDir.resolve(nativeBenchResults).absolutePath,
```
ilmat192: With this patch we don't use this task to run benchmarks. But we still configure args here and then read these configured args in the `runKonan` task. It looks like a redundant step and I think it's better to configure `runKonan` directly.

We can even hide the default run task by assigning it to a `null` task group. Also it's possible to disable execution of this task using the corresponding property:
```
runTask.enabled = false
```

PR #3153, opened 2019-07-05
Generate C wrapper instead of bridge for all functions.

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -936,7 +936,9 @@ private class StructValuePassing(private val kotlinClass: IrClass, override val
         bridgeCallBuilder.prepare += kotlinPointed
 
         val cPointer = passThroughBridge(irGet(kotlinPointed), kotlinPointedType, CTypes.pointer(cType))
-        cBridgeBodyLines += "*${cPointer.name} = $expression;"
+        val kniStructResult = "kniStructResult"
+        cBridgeBodyLines += "${cType.render(kniStructResult)} = $expression;"
+        cBridgeBodyLines += "memcpy(${cPointer.name}, &$kniStructResult, sizeof($kniStructResult));"
```
SvyatoslavScherbina: What is this required for?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
...
+/**
+ * Some functions doesn't have an address (e.g. macros-based or builtins).
+ * To solve this problem we generate a wrapper function.
+ */
+private fun generateCCalleeWrapper(cCallSymbolName: String, origin: StubOrigin.Function): CCalleeWrapper =
+        if (origin.function.isVararg) {
+            CCalleeWrapper(origin.function.name, emptyList())
+        } else {
+            val function = origin.function
+            val wrapperName = "${cCallSymbolName}_wrapper"
```
SvyatoslavScherbina: May clash, better be more unique.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
...
+            val returnType = function.returnType.getStringRepresentation()
+            val parameters = function.parameters.mapIndexed { index, parameter ->
+                val parameterName = parameter.name
+                val name = if (parameterName == null || parameterName.isEmpty()) {
+                    "p$index"
+                } else {
+                    parameterName
```
SvyatoslavScherbina: Not sure if this is necessary.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
...
+                }
+                name to parameter.type.getStringRepresentation()
+            }
+
+            val callExpression = "${function.name}(${parameters.joinToString { it.first }});"
+            val wrapperBody = if (function.returnType is VoidType) {
```
SvyatoslavScherbina: Does this work for functions with return type being typedef to void?
We definitely need a test for this.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
...
+                callExpression
+            } else {
+                "return $callExpression"
+            }
+
+            val lines = listOf(
+                    "$returnType $wrapperName(${parameters.joinToString { "${it.second} ${it.first}" }}) {",
```
SvyatoslavScherbina: Does it make sense to add some inlining hint to this function?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
         val excludedStubs: Set<StubIrElement>
 )
 
+private data class CCalleeWrapper(val name: String, val lines: List<String>)
+
+/**
+ * Some functions doesn't have an address (e.g. macros-based or builtins).
```
SvyatoslavScherbina: > doesn't

don't

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -16,6 +19,44 @@ class BridgeBuilderResult(
...
+ * To solve this problem we generate a wrapper function.
```
olonho: Can we generate wrappers only for those functions?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -103,12 +144,17 @@ class StubIrBridgeBuilder(
...
+            val (wrapperName, wrapperLines) = generateCCalleeWrapper(cCallSymbolName, origin)
             simpleBridgeGenerator.insertNativeBridge(
                     function,
                     emptyList(),
-                    listOf("extern const void* $cCallSymbolName __asm(${cCallSymbolName.quoteAsKotlinLiteral()});",
-                            "extern const void* $cCallSymbolName = &${origin.function.name};")
+                    listOf(
+                        *wrapperLines.toTypedArray(),
+                        "extern const void* $cCallSymbolName __asm(${cCallSymbolName.quoteAsKotlinLiteral()});",
+                        "extern const void* $cCallSymbolName = &$wrapperName;"
```
olonho: Why extern here? You do provide value.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
olonho: +1, tailcall seems to be enough (and more efficient).

sbogolepov: It is required for structs with `const` fields. It was done like this [before].(https://github.com/JetBrains/kotlin-native/commit/471771e8b8c7e6e63b72df16538bc034af433c56).

SvyatoslavScherbina: CBridgeGen shouldn't get structs with const fields. All structs it manipulates are generated by cinterop.

Have you fixed any particular reproducible issue with this patch?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
sbogolepov: It's hard to distinguish functions that require wrapping from those that doesn't.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
sbogolepov: Yeah, definitely.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
SvyatoslavScherbina: https://github.com/JetBrains/kotlin-native/blob/master/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StructRendering.kt

sbogolepov: Yes, i've added these lines because of [this](https://github.com/JetBrains/kotlin-native/blob/master/backend.native/tests/interop/basics/types.kt#L6) failing test.

SvyatoslavScherbina: It turns out that the logic I referred above contains a bug:
https://github.com/JetBrains/kotlin-native/blob/712c056a547b0da1ea7dfca38e3b452d75e0bff7/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StructRendering.kt#L67
This actually can place `const` modifier.

This is not supposed to happen (e.g. to simplify CBridgeGen), so the proper way to fix this is to adjust cinterop (most likely fix `type.spelling` or don't rely on it in struct rendering).

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
SvyatoslavScherbina: The calls to wrapper are actually indirect (through the generated const global). I hope that optimizer is smart enough to make it direct.
Also we can try to avoid using globals in some cases. Globals are mostly required because generally there is no reliable way to refer the function from stubs directly (e.g. it may be C `static`). But we can reliably refer the wrapper when it is present. This may although complicate the semantics of `CCall` (make it hold the name of global for variadic functions and the name of the function for the rest).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
SvyatoslavScherbina: > the proper way to fix this is to adjust cinterop (most likely fix type.spelling or don't rely on it in struct rendering).

Is there any progress on this?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StructRendering.kt
```diff
@@ -74,7 +75,7 @@ private fun tryRenderVar(type: Type, name: String): String? = when (type) {
     is Typedef -> tryRenderVar(type.def.aliased, name)
     is ObjCPointer -> "void* $name"
     else -> null
-}
+}?.let { if (shouldOmitConstQualifier) it.substringAfterLast("const ") else it }
```
SvyatoslavScherbina: Why not fix `type.spelling` for `IntegerType` and `FloatingType` instead?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StructRendering.kt
```diff
@@ -62,7 +63,7 @@ private fun tryRenderUnion(def: StructDef): String? =
 
         }
 
-private fun tryRenderVar(type: Type, name: String): String? = when (type) {
+private fun tryRenderVar(type: Type, name: String, shouldOmitConstQualifier: Boolean = false): String? = when (type) {
```
SvyatoslavScherbina: Omitting `const` should be default behaviour.
Moreover, I don't know any usage of this function that requires `const` preserved.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBridgeBuilder.kt
```diff
@@ -104,14 +115,49 @@ class StubIrBridgeBuilder(
...
+        /**
+         * Some functions don't have an address (e.g. macros-based or builtins).
+         * To solve this problem we generate a wrapper function.
+         */
+        private fun generateCCalleeWrapper(cCallSymbolName: String, origin: StubOrigin.Function): CCalleeWrapper =
+                if (origin.function.isVararg) {
+                    CCalleeWrapper(origin.function.name, emptyList())
+                } else {
+                    val function = origin.function
+                    val wrapperName = generateFunctionWrapperName(cCallSymbolName)
```
SvyatoslavScherbina: Maybe use original function name instead of `cCallSymbolName` here?

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/NativeIndex.kt
```diff
@@ -265,10 +265,16 @@ object CBoolType : BoolType()
 
 object ObjCBoolType : BoolType()
 
-data class IntegerType(val size: Int, val isSigned: Boolean, val spelling: String) : PrimitiveType
+class IntegerType(val size: Int, val isSigned: Boolean, spelling: String) : PrimitiveType {
```
SvyatoslavScherbina: Not sure if removing `data` (including custom `hashCode`/`equals`/`toString`) is desirable here.
Can we fix the spelling on call site of these constructors, i.e. in `convertUnqualifiedPrimitiveType`?
Does `clang_getCanonicalType` remove `const` modifier from type?

sbogolepov: > Not sure if removing data (including custom hashCode/equals/toString) is desirable here.
Yup, it broke `interop_objc_smoke` test :)

sbogolepov: > Can we fix the spelling on call site of these constructors, i.e. in convertUnqualifiedPrimitiveType?

Yes.
> Does clang_getCanonicalType remove const modifier from type?

No. There is no C API for dropping `const` qualifier. It is possible to write a wrapper around  `Qualifiers::removeConst` but i don't think it's any better than string manipulation in this case.

PR #3154, opened 2019-07-05
Support -Xg0 compiler flag

SvyatoslavScherbina:
Enables generation of partial debug information
(e.g. useful for crash reports symbolication).

Supposed to avoid affecting '-opt' optimization quality.

### DIFF
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
@@ -126,6 +126,9 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var exportedLibraries: Array<String>? = null
 
+    @Argument(value = "-Xlight-debug", description = "Add light debug information")
+    var lightDebug: Boolean = false
+
     @Argument(
             value = "-Xframework-import-header",
             valueDescription = "<header>",
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
@@ -126,15 +126,15 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var exportedLibraries: Array<String>? = null
 
-    @Argument(value = "-Xlight-debug", description = "Add light debug information")
-    var lightDebug: Boolean = false
-
     @Argument(
             value = "-Xframework-import-header",
             valueDescription = "<header>",
             description = "Add additional header import to framework header"
     )
     var frameworkImportHeaders: Array<String>? = null
+
+    @Argument(value = "-Xg0", description = "Add light debug information")
+    var lightDebug: Boolean = false
 
     @Argument(value = "-Xprint-bitcode", deprecatedName = "--print_bitcode", description = "Print llvm bitcode")
     var printBitCode: Boolean = false
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
@@ -429,7 +429,7 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
 
     fun shouldContainDebugInfo() = config.debug
     fun shouldContainLocationDebugInfo() = shouldContainDebugInfo() || config.lightDebug
-    fun shouldContainSomeDebugInfo() = shouldContainDebugInfo() || shouldContainLocationDebugInfo()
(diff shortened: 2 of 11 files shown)


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -126,6 +126,9 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var exportedLibraries: Array<String>? = null
 
+    @Argument(value = "-Xlight-debug", description = "Add light debug information")
```
olonho: Not very descriptive name, maybe `-g0` or `-Xrelease-backtrace`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -436,6 +436,8 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
     fun shouldProfilePhases() = config.phaseConfig.needProfiling
 
     fun shouldContainDebugInfo() = config.debug
+    fun shouldContainLocationDebugInfo() = shouldContainDebugInfo() || config.lightDebug
+    fun shouldContainSomeDebugInfo() = shouldContainDebugInfo() || shouldContainLocationDebugInfo()
```
olonho: Not very clear, maybe shouldContainSomeDebugInfo -> shouldContainBasicDebugInfo?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -192,7 +196,17 @@ open class MacOSBasedLinker(targetProperties: AppleConfigurables)
...
             +linkerArgs
             +rpath(dynamic)
-        }) + if (debug) listOf(dsymUtilCommand(executable, outputDsymBundle)) else emptyList()
+        }
+
+        // TODO: revise debug information handling.
+        if (debug) {
+            result += dsymUtilCommand(executable, outputDsymBundle)
+            if (optimize) {
+                result += Command(strip, "-S", executable)
```
olonho: Neat! So how much binary grows actually?

SvyatoslavScherbina: ~0
(after strip)

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
sbogolepov: I think it's better to enable this behaviour by default and add option to omit debug info instead.

benasher44: I agree, especially since later build stages can run `strip` to remove all of it.

SvyatoslavScherbina: > Not very descriptive name, maybe -g0 or -Xrelease-backtrace?

It's a draft, the name is not chosen yet.

> I think it's better to enable this behaviour by default and add option to omit debug info instead.

1. I believe it makes sense to release it first as experimental (and thus optional) to ensure it doesn't break anything.
2. Consider the effect of `-Xlight-debug` on non-Apple platforms.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
SvyatoslavScherbina: The logic behind this name is: "if some debug info is emitted, then we must generate a DI header and properly finalize it".

olonho: Then maybe `shouldContainAnyDebugInfo`?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
benasher44: Even on non-Apple platforms, debug info can be stripped back out (even automatically at a later stage: generating the k/n binary), no? Seems like risk of making this opt-in is greater: every library has to opt-in to ensure complete debug info in production.

SvyatoslavScherbina: > Even on non-Apple platforms, debug info can be stripped back out (even automatically at a later stage: generating the k/n binary)

I mean current implementation. Consider the effect of enabling this option on non-Apple platforms using current implementation.

> every library has to opt-in to ensure complete debug info in production

Note that there is no debug info in libraries currently, it is added at "a later stage: generating the k/n binary".

benasher44: I see. I think I understand now. 👍

PR #3156, opened 2019-07-07
Fix missing include

apolyakov:
In case of COLLECT_STATISTIC option being equal to 1
compilation fails because of missing '#include <algorithm>'


olonho:
Which platform do you see it on?

apolyakov:
Ubuntu 18.04.2 LTS
~~~~
uname -a
Linux asus-k551 5.1.16-050116-generic #201907031232 SMP Wed Jul 3 12:35:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
~~~~

PR #3157, opened 2019-07-08
[Objective-C reverse interop] Fix @ObjCMethod annotation handling.

sbogolepov:


homuroll:
Already fixed in https://github.com/JetBrains/kotlin-native/pull/3158

PR #3162, opened 2019-07-08
Better GC heuristics and ergonomics control.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -1419,7 +1431,7 @@ void garbageCollect(MemoryState* state, bool force) {
   decrementStack(state);
   size_t afterDecrements = state->toRelease->size();
   ssize_t stackReferences = afterDecrements - beforeDecrements;
-  if (stackReferences * 5 > state->gcThreshold) {
+  if (state->gcErgonomics && stackReferences * 5 > state->gcThreshold) {
```
LepilkinaElena: Compilation error in case if GC_ERGONOMICS is 0, isn't it?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1917,17 +1945,39 @@ void startGC() {
...
+}
+
+KLong getGCThresholdAllocations() {
+  GC_LOG("getGCThresholdAllocation\n")
+  return memoryState->allocSinceLastGcThreshold;
+}
+
+void setTuneGCThreshold(KBoolean value) {
+  GC_LOG("setTuneGCThreshold %d\n", value)
+  memoryState->gcErgonomics = value;
```
LepilkinaElena: Compilation error in case if GC_ERGONOMICS is 0

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1917,17 +1945,39 @@ void startGC() {
...
+}
+
+KBoolean getTuneGCThreshold() {
+  GC_LOG("getTuneGCThreshold %d\n")
+  return memoryState->gcErgonomics;
```
LepilkinaElena: Same here

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -63,9 +63,36 @@ object GC {
         get() = getThreshold()
         set(value) = setThreshold(value)
 
+    /**
+     * GC allocation threshold, controlling how many bytes allocated since last
+     * collection will trigger new GC.
+     */
+    var thresholdAllocations: Long
+        get() = getThresholdAllocations()
+        set(value) = setThresholdAllocations(value)
```
LepilkinaElena: Not used now as I understood, value isn't changed. Is it for future tuning?

olonho: It's user API for GC tuning.

PR #3164, opened 2019-07-09
[Inliner] Fixed bug with incorrect eval order

homuroll:
We need to extract to temp locals all bound arguments
of function reference parameters of an inline call to
preserve the correct evaluation order

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
@@ -223,6 +223,7 @@ internal class FunctionInlining(val context: Context) : IrElementTransformerVoid
                     return super.visitCall(expression)
 
                 if (functionArgument is IrFunctionReference) {
+                    functionArgument.transformChildrenVoid(this)
                     val function = functionArgument.symbol.owner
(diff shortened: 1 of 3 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionInlining.kt
```diff
@@ -241,10 +242,10 @@ internal class FunctionInlining(val context: Context) : IrElementTransformerVoid
                     }.apply {
                         functionParameters.forEach {
                             val argument =
-                                    if (!unboundArgsSet.contains(it))
-                                        boundFunctionParametersMap[it]!!
-                                    else
+                                    if (unboundArgsSet.contains(it))
                                         valueParameters[unboundIndex++].second
+                                    else
+                                        boundFunctionParametersMap[it]!!
```
anton-bannykh: This change doesn't alter the code behaviour, does it?

homuroll: True, it's just more readable IMO.

anton-bannykh: ok

PR #3166, opened 2019-07-09
Update gradle: 5.5

ilmat192:


extracted/konan.metadata/build.gradle
```diff
@@ -59,7 +59,7 @@ protobuf {
 compileJava {
     dependsOn('renamePackage')
     doFirst {
-        delete 'build/generated'
+        delete 'build/generated/source/proto/main/java'
```
olonho: Why?

ilmat192: For some reason Gradle's `compileJava` task looks for sources in `generated/sources/annotationProcessor` and fails if this directory doesn't exist. So I've decided to delete only sources generated by protobug here.

PR #3170, opened 2019-07-10
StringBuilder allocates more during resize

ScottPierce:
The current algorithm for `StringBuilder`, when it needs a new Array, it increases the allocation only by 50%. This PR changes that allocation to double in size + 2, similar to the JVM algorithm.


RUSshy:
bit shift is faster 

``(array.size shl 1) + 2``

sbogolepov:
Such things are optimized by compiler.

RUSshy:
why give compiler more work when it can be fixed by user, it should help compile time no ?

PR #3176, opened 2019-07-11
Devirtualization enhancement

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt
```diff
@@ -456,6 +461,93 @@ internal object Devirtualization {
...
+                var multiNodesCount = 0
+                val multiNodesRepresentatives = IntArray(nodesCount)
+                for (i in order.size - 1 downTo 0) {
+                    val v = order[i]
+                    if (visited[v]) continue
+                    multiNodesRepresentatives[multiNodesCount++] = v
+                    val start = index
+                    paint(nodes[v])
+                    val end = index
+                    for (w in start until end) {
```
olonho: w is not the best name.

olonho:
Other that that - looks good.

PR #3177, opened 2019-07-11
Add lldb as dependency

kishmakov:
Dependencies bundles are versioned with internal Teamcity build numbers.

### DIFF
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -61,7 +61,7 @@ osVersionMin.macos_x64 = 10.11
 entrySelector.macos_x64 = -alias _Konan_main _main
 dependencies.macos_x64 = \
     libffi-3.2.1-3-darwin-macos \
-    lldb-171-macos.tar.gz \
+    lldb-171-macos \
     clang-llvm-6.0.1-darwin-macos
 
 target-sysroot-9-macos_x64.default = \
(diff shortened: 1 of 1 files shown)


msink:
mingw_x86 debugging not supported?

konan/konan.properties
```diff
@@ -61,6 +61,7 @@ osVersionMin.macos_x64 = 10.11
 entrySelector.macos_x64 = -alias _Konan_main _main
 dependencies.macos_x64 = \
     libffi-3.2.1-3-darwin-macos \
+    lldb-171-macos \
```
olonho: Can you have consistent versioning?

kishmakov: Bundles build queues are not synchronized currently. It is possible to hide this difference with artificial version numbers. I see two cons for this approach: 1) additional indirection layer which we would maintain 2) possible conflict with versioning for the builds in case we (as JB) would decide to publish them externally.

kishmakov:
Mike, this is for IDE experience, therefore targets platforms where developers works. Currently we are not planning to support 32bit Windows.

msink:
Sure, but 32-bit lldb works just fine on 64-bit Windows.

olonho:
We are not planning to support 32-bit hosts.

msink:
Of course, but as I said, 64-bit Windows host is fully compatible with 32-bit Windows programs, you can compile, run and debug them.
So it really is not cross-compilation.

PR #3188, opened 2019-07-12
Document order of array elements initialization (KT-32353)

qurbonzoda:


ilya-g:
(Not urgent) Please cherry pick this to v1.3.50-fixes

PR #3197, opened 2019-07-18
Callable references on suspend functions

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
```diff
@@ -254,9 +270,13 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
                 }
 
                 body = context.createIrBuilder(symbol, startOffset, endOffset).irBlockBody {
-                    if (!isKFunction)
-                        +irDelegatingConstructorCall(irBuiltIns.anyClass.owner.constructors.single())
-                    else +irDelegatingConstructorCall(kFunctionImplConstructorSymbol.owner).apply {
+                    val superConstructor = when {
+                        isKFunction -> kFunctionImplConstructorSymbol.owner
+                        isKSuspendFunction -> kSuspendFunctionImplConstructorSymbol.owner
+                        else -> irBuiltIns.anyClass.owner.constructors.single()
```
olonho: What could be on this path?

backend.native/tests/codegen/coroutines/functionReference_eqeq_name.kt
```diff
@@ -0,0 +1,29 @@
...
+}
+
+@Test fun runTest() {
+    val ref1 = ::foo
+    val rec = Foo(42)
+    val ref2 = rec::bar
+    val ref3 = ::foo
+    val ref4 = Foo(42)::bar
+    val ref5 = rec::bar
+    println(ref1.name)
```
olonho: Why not use asserts here?

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+        if (other !is KSuspendFunctionImpl<*>) return false
+        return fqName == other.fqName && bound == other.bound && receiver == other.receiver
+    }
+
+    override fun hashCode(): Int {
+        return (fqName.hashCode() * 31 + if (bound) 1 else 0) * 31 + receiver.hashCode()
+    }
+
+    override fun toString(): String {
+        return ("suspend function " + name) + " (Kotlin reflection is not available)"
```
olonho: String interpolation?

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+@FixmeReflection
+internal abstract class KSuspendFunctionImpl<out R>(
+        override val name: String, val fqName: String,
+        val bound: Boolean, val receiver: Any?,
+        override val returnType: KType
+): KSuspendFunction<R> {
+
+    override fun equals(other: Any?): Boolean {
+        if (other !is KSuspendFunctionImpl<*>) return false
+        return fqName == other.fqName && bound == other.bound && receiver == other.receiver
```
olonho: Can they differ by number or type of parameters?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
homuroll: A lambda.

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
homuroll: Hmm, they might actually be different by number of parameters. This was copied from `KFunctionImpl`, so there is a bug there as well.

backend.native/tests/codegen/coroutines/functionReference_eqeq_name.kt
homuroll: Dont see much of a difference. It's a test.

olonho: We generally try to use test infrastructure in tests.

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
homuroll: As it turns out, `fqName` here is our internal `symbolName`, so everything works fine. Still, adding `arity` instead of `bound` won't hurt and will bring us closer to Kotlin/JVM.

homuroll: But may be do it in a separate PR? It's not connected with the purpose of this PR.

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunctions.kt
```diff
@@ -0,0 +1,59 @@
...
+
+interface KSuspendFunction19<in P1, in P2, in P3, in P4, in P5, in P6, in P7, in P8, in P9, in P10, in P11, in P12, in P13, in P14, in P15, in P16, in P17, in P18, in P19, out R> : SuspendFunction19<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13, P14, P15, P16, P17, P18, P19, R>, KFunction<R>
+
+interface KSuspendFunction20<in P1, in P2, in P3, in P4, in P5, in P6, in P7, in P8, in P9, in P10, in P11, in P12, in P13, in P14, in P15, in P16, in P17, in P18, in P19, in P20, out R> : SuspendFunction20<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13, P14, P15, P16, P17, P18, P19, P20, R>, KFunction<R>
+
+interface KSuspendFunction21<in P1, in P2, in P3, in P4, in P5, in P6, in P7, in P8, in P9, in P10, in P11, in P12, in P13, in P14, in P15, in P16, in P17, in P18, in P19, in P20, in P21, out R> : SuspendFunction21<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13, P14, P15, P16, P17, P18, P19, P20, P21, R>, KFunction<R>
+
+// Since KSuspendFunctionN inherits Function{N+1} and we only have 0..22 range, comment this for now.
```
SvyatoslavScherbina: > Since KSuspendFunctionN inherits Function{N+1}

Does it?

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunction.kt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.reflect
+
+interface KSuspendFunction<out R> : KFunction<R>
```
SvyatoslavScherbina: What is this required for?

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    override fun hashCode(): Int {
+        return (fqName.hashCode() * 31 + if (bound) 1 else 0) * 31 + receiver.hashCode()
+    }
+
+    override fun toString(): String {
+        return "suspend function $name (Kotlin reflection is not available)"
```
SvyatoslavScherbina: > (Kotlin reflection is not available)

Why add this?

backend.native/tests/codegen/coroutines/functionReference_eqeq_name.kt
```diff
@@ -0,0 +1,29 @@
...
+    val ref2 = rec::bar
+    val ref3 = ::foo
+    val ref4 = Foo(42)::bar
+    val ref5 = rec::bar
+    assertEquals("foo", ref1.name)
+    assertEquals("bar", ref2.name)
+    assertFalse(ref1 == ref2)
+    assertTrue(ref1 == ref3)
+    assertFalse(ref2 == ref4)
+    assertTrue(ref2 == ref5)
```
SvyatoslavScherbina: As far as I see, test for `Foo::bar` is missing.

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunctions.kt
homuroll: You can cast `KSuspendFunction1<Int, Int>` to `Function2<Int, Continuation<Int>, Any?>`.

runtime/src/main/kotlin/kotlin/native/internal/KSuspendFunctionImpl.kt
homuroll: This was copied from `KFunctionImpl`. May be it now can be deleted.

backend.native/tests/codegen/coroutines/functionReference_eqeq_name.kt
homuroll: What do you mean?

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunction.kt
homuroll: It's not needed indeed.

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunctions.kt
SvyatoslavScherbina: But does `KSuspendFunctionN` actually inherit `Function{N+1}`?
I don't see any reason to disable `KSuspendFunction22`.

backend.native/tests/codegen/coroutines/functionReference_eqeq_name.kt
SvyatoslavScherbina: I mean that `Foo::bar` is a proper callable reference too, and it is different from e.g. `Foo(42)::bar`. It makes some sense to check this too.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
```diff
@@ -183,48 +183,64 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
...
+                    val continuationType = continuationClassSymbol.typeWith(referencedFunction.returnType)
+                    superTypes += functionClass.typeWith(functionParameterTypes + continuationType + irBuiltIns.anyNType)
+                    suspendFunctionClass = symbols.kSuspendFunctions[numberOfParameters].owner
+                    superTypes += suspendFunctionClass.typeWith(functionParameterTypes + referencedFunction.returnType)
+                }
+                else -> {
+                    superTypes += irBuiltIns.anyType
+                    functionClass = symbols.functions[numberOfParameters].owner
+                    superTypes += functionClass.typeWith(functionParameterTypes + referencedFunction.returnType)
+                    // Anonymous functions and lambdas never inherit SuspendFunctionX.
```
SvyatoslavScherbina: Why so?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
```diff
@@ -288,7 +308,9 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
                     superFunction.name,
                     Visibilities.PRIVATE,
                     Modality.FINAL,
-                    referencedFunction.returnType,
+                    if (referencedFunction.isSuspend && !superFunction.isSuspend)
```
SvyatoslavScherbina: When is it true?

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunctions.kt
homuroll: Nol it does not. But FE still tries to reference `Function23` if we use `KSuspendFunction22`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
homuroll: Suspend lambdas have type `SuspendLambdaX`, not `KSuspendLambdaX`. So only lambdas that can be at this point are not suspend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
homuroll: Hmm, looks like never. Might be an artefact of some previous experiments.

runtime/src/main/kotlin/kotlin/reflect/KSuspendFunctions.kt
SvyatoslavScherbina: As far as I see, it's not exactly FE.
The errors occurs during descriptor serialization due to this: https://github.com/JetBrains/kotlin/blob/ad1b795a69925e6ee8c79cbbd05cce8cd27a6704/compiler/serialization/src/org/jetbrains/kotlin/serialization/DescriptorSerializer.kt#L555
(when serializing super types for `KSuspendFunction22`).

Moreover, using plain suspend function type with 22 parameters leads to similar problem.

So I suppose it is better to at least describe this properly here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/CallableReferenceLowering.kt
SvyatoslavScherbina: AFAIR, the code with `suspendFunctionIrClass ` was intended to make this work:
```
import kotlin.coroutines.*

fun foo(block: (Continuation<Unit>) -> Any?) {
    block as (suspend () -> Unit)
}

fun main() {
    foo {}
}
```

Isn't the test broken now?

homuroll: You're right, I for some reason only supported that for KFunction.

olonho: Wouldn't say ` block as (suspend () -> Unit)` must work, although it is nice to have.

PR #3198, opened 2019-07-18
Support nullable types in reverse C interop.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -63,6 +63,13 @@ private operator fun String.times(count: Int): String {
     return builder.toString()
 }
 
+private val KotlinType.shortName
```
SvyatoslavScherbina: This is somewhat wrong. Consider the behaviour for types with arguments or nullability.
`predefinedTypes` should probably be `predefinedClasses` (or even `predefinedValueClasses`).

olonho: Not sure if I agree here. This is pretty much list of types we want default factories for. They do not have to be classes, just happen to be so for primitive types.

SvyatoslavScherbina: `shortName` and `createNullableName` properties are still wrong: don't work for arbitrary types. There are different options available here, for example to rename these (e.g. `shortNameForPredefinedType` and `createNullableNameForPredefinedType`).

olonho: PTAL

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -181,3 +181,40 @@ internal fun <T> listOfInternal(vararg elements: T): List<T> {
         result.add(elements[i])
     return result
 }
+
+
+@PublishedApi
```
SvyatoslavScherbina: Btw, have you seen https://github.com/JetBrains/kotlin-native/blob/9a54689033c8ac1f9eb679066c0418f10454de68/runtime/src/main/kotlin/kotlin/native/internal/Boxing.kt#L35 ?

olonho: Funny, will switch to those.

PR #3207, opened 2019-07-23
Document potentially cyclic cases in ownership transfer and add test.

olonho:


runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -76,6 +76,9 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
      * explicitly stored in object produced by [producer]. Scheduled job is being executed by the worker,
      * and result of such a execution is being disconnected from worker's object graph. Whoever will consume
      * the future, can use result of worker's computations.
+     * Note, that some technically disjoint subgraphs may lead to `kotlin.IllegalStateException`
+     * so `kotlin.native.internal.GC.collect()` could be called in the end of `producer` and `job`
+     * if garbage cyclic structures refer to to value being transferred.
```
SvyatoslavScherbina: Note that not only cyclic structures can lead to this error.
Since 1.3.40 any garbage referencing transferred graph does because of deferred reference counting:

```
import kotlin.test.*

// Fails since 1.3.40.
@Test
fun test1() {
    DetachedObjectGraph {
        Pair(Any(), Any()).first
    }
}

// Fails at least since 1.3.0.
@Test
fun test2() {
    DetachedObjectGraph {
        mapOf(Any() to Any()).values.first()
    }
}
```

PavelPunegov: "to to"

PR #3212, opened 2019-07-24
Remove unused metadata extension fields

alexander-gorshenev:


olonho:
Doesn't it require metadata version bump?

PR #3215, opened 2019-07-25
CLI library rework

LepilkinaElena:


endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/ExperimentalCli.kt
```diff
@@ -15,7 +15,7 @@ import kotlin.annotation.AnnotationTarget.*
  *
  * Any usage of a declaration annotated with `@ExperimentalCli` must be accepted either by
  * annotating that usage with the [UseExperimental] annotation, e.g. `@UseExperimental(ExperimentalCli::class)`,
- * or by using the compiler argument `-Xuse-experimental=org.jetbrains.kliopt.ExperimentalCli`.
+ * or by using the compiler argument `-Xuse-experimental=org.jetbrains.kotlinx.cli.ExperimentalCli`.
```
ilya-g: Incorrect FQN
```suggestion
 * or by using the compiler argument `-Xuse-experimental=kotlinx.cli.ExperimentalCli`.
```

ilya-g:
I don't quite get what are endorsed libraries in kotlin/native and what migration path is supposed for kotlinx.cli when it becomes a standalone MPP library?

olonho:
Guess idea (per talk with @elizarov) is that library is endorsed with K/N distro, and afterwards we use it more of our projects (including the compiler) and then make it endorsed with the Kotlin distro as an MPP library.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/ExperimentalCli.kt
LepilkinaElena: Yes, missed in one more place. Fixed.

ilya-g:
I'm still concerned about a mess that could happen for a user who depends on the current native version of [kotlinx.cli](https://github.com/Kotlin/kotlinx.cli) and gets completely different kotlinx.cli as an endorsed one. Thus, we need to discuss the consequences of making it endorsed a bit further.

msink:
`performance/numerical/*` too.

LepilkinaElena:
@msink Yes, forgot after rebase on master, thank you

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
```diff
@@ -22,6 +22,7 @@ import kotlinx.cli.*
 
 const val HEADER_FILTER_ADDITIONAL_SEARCH_PREFIX = "headerFilterAdditionalSearchPrefix"
 const val NODEFAULTLIBS = "nodefaultlibs"
+const val NOENDORSEDLIBS = "noendorsedlibs"
```
ilya-g: In Kotlin/JVM the similar compiler options are called `-no-stdlib` and `-no-reflect`. Makes sense to keep consistency here and consider calling it `-no-endorsedlibs` or `-no-endorsed-libs`.

LepilkinaElena: I named in same way as we already have flag for platform and standard libs. If change style, then `-nodefaultlibs` and  `-nostdlib` also should be renamed. We should deprecated old forms and add new ones.
@olonho what do you think?

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/ArgParser.kt
```diff
@@ -0,0 +1,467 @@
...
+    private val options = mutableMapOf<String, ParsingValue<*, *>>()
+    /**
+     * Map of arguments: key - full name of argument, value - pair of descriptor and parsed values.
+     */
+    private val arguments = mutableMapOf<String, ParsingValue<*, *>>()
+
+    /**
+     * Map with declared options.
+     */
+    private val declaredOptions = mutableMapOf<Int, OptionImpl<*, *>>()
```
LepilkinaElena: Need such maps, because modifiers can change option several times, before it's named.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/ArgumentValues.kt
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Single argument value.
+ *
+ * @property descriptor descriptor of option/argument.
+ */
+internal class ArgumentSingleValue<T: Any>(descriptor: Descriptor<T, T>): AbstractArgumentSingleValue<T>(descriptor),
+        ArgumentValueDelegate<T> {
+    override val value: T
```
LepilkinaElena: Created field to provide opportunity to get value if I don't use delegated properties.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,208 @@
...
+     * Needed to close set of variable [cliElement].
+     */
+    override val value: ArgumentValueDelegate<TResult>
+        get() = cliElement
+}
+
+/**
+ * Argument instance interface.
+ */
+interface Argument<T : Any, TResult>: CLIEntity<T, TResult>
```
LepilkinaElena: Separate `Argument` and `Option` interfaces to save opportunity provide different modifiers now and in future.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Argument with single nullable value.
+ */
+interface SingleNullableArgument<T : Any> : AbstractSingleArgument<T, T?>
+/**
+ * Argument with multiple values.
+ */
+interface MultipleArgument<T : Any, TResult : Collection<T>> : Argument<T, TResult>
+
+internal abstract class ArgumentImpl<T : Any, TResult>(owner: ArgParser): CLIEntityImpl<T, TResult>(owner), Argument<T, TResult> {
```
LepilkinaElena: Made all implementations internal, only interfaces are in public API.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,208 @@
...
+
+internal class SingleNullableArgumentImpl<T : Any>(descriptor: ArgDescriptor<T, T>, owner: ArgParser):
+        AbstractSingleArgumentImpl<T, T?>(owner),
+        SingleNullableArgument<T> {
+    init {
+        checkDescriptor(descriptor)
+        cliElement = ArgumentSingleNullableValue(descriptor)
+    }
+
+    override operator fun provideDelegate(thisRef: Any?, prop: KProperty<*>): ArgumentValueDelegate<T?> {
```
LepilkinaElena: So many similar methods only because different type of generic values.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Descriptors.kt
```diff
@@ -0,0 +1,152 @@
...
+ * @property shortName option short name.
+ * @property description text description of option.
+ * @property defaultValue default value for option.
+ * @property required if option is required or not. If it's required and not provided in command line and have no default value, error will be generated.
+ * @property multiple if option can be repeated several times in command line with different values. All values are stored.
+ * @property delimiter delimiter that separate option provided as one string to several values.
+ * @property deprecatedWarning text message with information in case if option is deprecated.
+ */
+internal class OptionDescriptor<T : Any, TResult>(
+        val optionFullFormPrefix: String,
```
LepilkinaElena: Have to add prefix form as property for options, because of making modifiers as open part of API, so `OptionDescriptor` can't be inner class any more.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/CommandLine.kt
olonho: Yes, let's make them consistent.

olonho: But this option could be called -no-endorsed-libs already.

cmd/run_konan
```diff
@@ -89,13 +89,13 @@ SHARED_JAR="${KONAN_HOME}/konan/lib/shared.jar"
 EXTRACTED_METADATA_JAR="${KONAN_HOME}/konan/lib/konan.metadata.jar"
 EXTRACTED_SERIALIZER_JAR="${KONAN_HOME}/konan/lib/konan.serializer.jar"
 KLIB_JAR="${KONAN_HOME}/konan/lib/klib.jar"
-KLIOPT_JAR="${KONAN_HOME}/konan/lib/kliopt-jvm.jar"
+KOTLINX_CLI_JAR="${KONAN_HOME}/konan/lib/kotlinx.cli-jvm.jar"
```
olonho: Could we eventually repack  all them into single jar?

LepilkinaElena: I think it should be done a bit later, when we start to change mechanism of using this library.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,178 @@
...
+    val value: ArgumentValueDelegate<TResult>
+        get() = cliElement
+
+    /**
+     * Origin of argument value.
+     */
+    val valueOrigin: ArgParser.ValueOrigin
+        get() = (value as ParsingValue<*, *>).valueOrigin
+
+    abstract operator fun provideDelegate(thisRef: Any?, prop: KProperty<*>): ArgumentValueDelegate<TResult>
```
ilya-g: The implementation looks the same in every inheritor. Can be lifted here.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -6,31 +6,14 @@ package kotlinx.cli
...
-internal data class CLIEntityWrapper(var entity: CLIEntityImpl<*, *>? = null)
+internal data class CLIEntityWrapper(var entity: CLIEntity<*, *>? = null)
 
 /**
  * Command line entity.
  *
  * @property owner parser which owns current entity.
  */
-internal abstract class CLIEntityImpl<T : Any, TResult>(val owner: CLIEntityWrapper): CLIEntity<T, TResult> {
+abstract class CLIEntity<T : Any, TResult> internal constructor(val owner: CLIEntityWrapper) {
```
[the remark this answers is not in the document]
LepilkinaElena: Fixed

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -175,16 +175,15 @@ private fun processCLib(args: Array<String>, additionalArgs: Map<String, Any> =
...
+            (cinteropArguments.linkerOptions.valueOrigin == ArgParser.ValueOrigin.SET_BY_USER) ||
+            (cinteropArguments.linkerOption.valueOrigin == ArgParser.ValueOrigin.SET_BY_USER)
     if (flavorName == "native" && isLinkerOptsSetByUser) {
-        warn("-linker-option(s)/-linkerOpts/-lopt option is not supported by cinterop. Please add linker options to .def file or binary compilation instead.")
+        warn("-linker-option(s)/-linkerOpts option is not supported by cinterop. Please add linker options to .def file or binary compilation instead.")
     }
 
-    val additionalLinkerOpts = cinteropArguments.linkerOpts.toTypedArray() + cinteropArguments.linkerOption.toTypedArray() +
-            cinteropArguments.linkerOptions.toTypedArray() + cinteropArguments.lopt.toTypedArray()
+    val additionalLinkerOpts = cinteropArguments.linkerOpts.value.value.toTypedArray() + cinteropArguments.linkerOption.value.value.toTypedArray() +
```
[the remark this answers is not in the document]
LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,178 @@
...
+    owner.entity = newArgument
+    return newArgument
+}
+
+/**
+ * Set default value for argument.
+ *
+ * @param value default value.
+ */
+fun <T: Any, TResult: Collection<T>> MultipleArgument<T>.default(value: TResult): MultipleArgument<T> {
```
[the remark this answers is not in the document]
LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,173 @@
...
+     */
+    val delegate: ArgumentValueDelegate<TResult>
+        get() = cliElement
+
+    /**
+     * Value of entity.
+     */
+    var value: TResult
+        get() = cliElement.value
+        set(value) = delegate.setValue(null, ::value, value)
```
ilya-g: `delegate.setValue` doesn't use `thisRef` and `property` parameters, so maybe we could provide the way to set value without them, for example by making its `value` a var property. This way we could avoid creating `::value` property reference.

LepilkinaElena: If I understood your suggestion right way, I have a concern that ValueOrigin isn't set right or I should set it also here. It seems to have all in one place, isn't it?

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,173 @@
...
+ * @property owner parser which owns current entity.
+ */
+abstract class CLIEntity<TResult> internal constructor(internal val owner: CLIEntityWrapper) {
+    /**
+     * CLI element value.
+     */
+    internal lateinit var cliElement: ArgumentValueDelegate<TResult>
+    /**
+     * Wrapper  for element - read only property.
+     * Needed to close set of variable [cliElement].
```
ilya-g: `cliElement` is non-public, so the reference leads nowhere.

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Common single argument instance.
+ */
+abstract class AbstractSingleArgument<T: Any, TResult> internal constructor(owner: CLIEntityWrapper): Argument<TResult>(owner) {
+    /**
+     * Check descriptor for this kind of argument.
+     */
+    internal fun checkDescriptor(descriptor: ArgDescriptor<*, *>) {
+        if (descriptor.number == null || descriptor.number > 1) {
+            error("Argument with single value can't be initialized with descriptor for multiple values.")
```
ilya-g: Could this be a user error or it's just our internal assertion?

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Set default value for argument.
+ *
+ * @param value default value.
+ */
+fun <T: Any, TResult> AbstractSingleArgument<T, TResult>.default(value: T): SingleArgument<T> {
+    val newArgument = with((cliElement as ParsingValue<T, T>).descriptor as ArgDescriptor) {
+        if (required) {
+            printWarning("You can use optional(), because option with default value is defined.")
```
ilya-g: So to avoid this warning, a user always has to write `.optional().default(...)`?

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
LepilkinaElena: That's internal assertion, I don't know if we want to make library real MPP. If so JS has now assert method

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
ilya-g: We can introduce `fun failAssertion(message) = throw AssertionError(message)` and use it just so the intent is clear.

LepilkinaElena: Fixed

endorsedLibraries/kotlinx.cli/src/main/kotlin/kotlinx/cli/Arguments.kt
LepilkinaElena: Fixed

PR #3218, opened 2019-07-25
Added verbose mode to follow benchmarks process

LepilkinaElena:


performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -46,11 +46,23 @@ abstract class Launcher {
         }
     }
 
+    enum class LogLevel { DEBUG, OFF }
+
+    class Logger(val level: LogLevel = LogLevel.OFF) {
+         fun log(message: String, messageLevel: LogLevel = LogLevel.DEBUG) {
+            if (messageLevel == level) {
+                printlnStderr("[$level] $message")
```
msink: Maybe also print current time?

msink: And maybe print single '.' every attempt - because `cinterop/stringToKotlin` takes about minute, so single line in a hour is not very informative.

msink:
On Windows:
```
> Task :performance:cinterop:linkBenchmarkReleaseExecutableNative FAILED
e: F:\src\kotlin-native\performance\shared\src\main\kotlin-native\org\jetbrains\benchmarksLauncher\Utils.kt: (55, 9): Unresolved reference: gettimeofday
e: F:\src\kotlin-native\performance\shared\src\main\kotlin-native\org\jetbrains\benchmarksLauncher\Utils.kt: (57, 21): Type mismatch: inferred type is Int but Long was expected
```

LepilkinaElena:
Yes, thank you, I saw problem on our CI on Windows. There should be `mingw_gettimeofday `. I'll fix

msink:
1) Second error on Windows not fixed, fix:
```
--- a/performance/shared/src/main/kotlin-native/mingw/org/jetbrains/benchmarksLauncher/Utils.kt
+++ b/performance/shared/src/main/kotlin-native/mingw/org/jetbrains/benchmarksLauncher/Utils.kt
@@ -25,7 +25,7 @@ actual fun currentTime() =
             val timeVal = alloc<timeval>()
             mingw_gettimeofday(timeVal.ptr, null)
             val sec = alloc<LongVar>()
-            sec.value = timeVal.tv_sec
+            sec.value = timeVal.tv_sec.convert()
             val nowtm = localtime(sec.ptr)
             var timeBuffer = ByteArray(1024)
             strftime(timeBuffer.refTo(0), timeBuffer.size.toULong(), "%H:%M:%S", nowtm)
```

2) I think warm-up phase should print '.' too,
"Warm up iterations for benchmark stringToKotlin" still run 20 minutes with no visible effects.

LepilkinaElena:
Thank you, I pushed to check this branch on Windows machine.

PR #3222, opened 2019-07-25
Array.fill implementation

ilya-g:
https://youtrack.jetbrains.com/issue/KT-32359
Corresponding common part review: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3160

### DIFF
--- a/runtime/src/main/cpp/Arrays.cpp
+++ b/runtime/src/main/cpp/Arrays.cpp
@@ -23,12 +23,25 @@
 #include "Natives.h"
 #include "Types.h"
 
+extern "C" void checkRangeIndexes(KInt from, KInt to, KInt size);
+
 namespace {
 
 ALWAYS_INLINE inline void mutabilityCheck(KConstRef thiz) {
   // TODO: optimize it!
   if (thiz->container()->frozen()) {
     ThrowInvalidMutabilityException(thiz);
+  }
+}
+
+template<typename T>
+inline void fillImpl(KRef thiz, KInt fromIndex, KInt toIndex, T value) {
+  ArrayHeader* array = thiz->array();
+  checkRangeIndexes(fromIndex, toIndex, array->count_);
+  mutabilityCheck(thiz);
+  T* address = PrimitiveArrayAddressOfElementAt<T>(array, fromIndex);
+  for (KInt index = fromIndex; index < toIndex; ++index) {
+    *address++ = value;
   }
 }
 
@@ -102,9 +115,7 @@ KInt Kotlin_Array_getArrayLength(KConstRef thiz) {
 
 void Kotlin_Array_fillImpl(KRef thiz, KInt fromIndex, KInt toIndex, KRef value) {
   ArrayHeader* array = thiz->array();
-  if (fromIndex < 0 || toIndex < fromIndex || toIndex > array->count_) {
-    ThrowArrayIndexOutOfBoundsException();
-  }
+  checkRangeIndexes(fromIndex, toIndex, array->count_);
   mutabilityCheck(thiz);
   for (KInt index = fromIndex; index < toIndex; ++index) {
     UpdateHeapRef(ArrayAddressOfElementAt(array, index), value);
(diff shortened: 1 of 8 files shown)


runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
```diff
@@ -75,10 +76,42 @@ internal fun <E> Array<E>.resetAt(index: Int) {
 
 @SymbolName("Kotlin_Array_fillImpl")
 @PointsTo(0b01000, 0, 0, 0b00001) // <array> points to <value>, <value> points to <array>.
```
ilya-g: What should I specify here, if I want to change parameter order to `array, value, from, to`?

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
```diff
@@ -75,10 +76,42 @@ internal fun <E> Array<E>.resetAt(index: Int) {
...
+
+@SymbolName("Kotlin_FloatArray_fillImpl")
+internal external fun arrayFill(array: FloatArray, fromIndex: Int, toIndex: Int, value: Float)
+
+@SymbolName("Kotlin_BooleanArray_fillImpl")
+internal external fun arrayFill(array: BooleanArray, fromIndex: Int, toIndex: Int, value: Boolean)
+
+@ExportForCppRuntime
+internal fun checkRangeIndexes(fromIndex: Int, toIndex: Int, size: Int) {
+//    println("$fromIndex, $toIndex, $size")
```
JakeWharton: nit: remove?

ilya-g: good catch

runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt
olonho: Just remove PointsTo, escape analysis need to be reworked.

ilya-g: Ok, good to know. I've decided not to change it though.

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -32,6 +38,16 @@ ALWAYS_INLINE inline void mutabilityCheck(KConstRef thiz) {
   }
 }
 
+template<typename T>
+inline void fillImpl(KRef thiz, KInt fromIndex, KInt toIndex, T value) {
+  ArrayHeader* array = thiz->array();
+  checkRangeIndexes(fromIndex, toIndex, array->count_);
+  mutabilityCheck(thiz);
+  for (KInt index = fromIndex; index < toIndex; ++index) {
+    *PrimitiveArrayAddressOfElementAt<T>(array, index) = value;
```
olonho: Maybe sliding pointer instead?
I.e.
```
void fillImpl(KRef thiz, KInt fromIndex, KInt toIndex, T value) {
  ArrayHeader* array = thiz->array();
  checkRangeIndexes(fromIndex, toIndex, array->count_);
  mutabilityCheck(thiz);
  T* address = PrimitiveArrayAddressOfElementAt<T>(array, index);
  for (KInt index = fromIndex; index < toIndex; ++index) {
    *address++ = value;
  }
}```

File: runtime/src/main/cpp/Arrays.cpp [hunk 4]
```diff
@@ -23,6 +23,12 @@
 #include "Natives.h"
 #include "Types.h"
 
+extern "C" {
+
+void checkRangeIndexes(KInt from, KInt to, KInt size);
```
olonho: Just `extern "C" void checkRangeIndexes(KInt from, KInt to, KInt size);` would do.

runtime/src/main/kotlin/generated/_CollectionsNative.kt
```diff
@@ -10,6 +10,8 @@ package kotlin.collections
 // See: https://github.com/JetBrains/kotlin/tree/master/libraries/stdlib
 //
 
+import kotlin.ranges.contains
+import kotlin.ranges.reversed
```
olonho: Why those imports needed?

runtime/src/main/cpp/Arrays.cpp
ilya-g: Just copied that from `IntArray.fill` implementation. I will try pointer increment instead as you're proposing.

runtime/src/main/kotlin/generated/_CollectionsNative.kt
ilya-g: Otherwise `Iterable` extensions `contains` and `reversed` from `kotlin.collections` are preferred even for ranges using the "same package" priority since we're in `kotlin.collections` package here, see https://youtrack.jetbrains.com/issue/KT-30633 for details

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -32,6 +34,17 @@ ALWAYS_INLINE inline void mutabilityCheck(KConstRef thiz) {
   }
 }
 
+template<typename T>
+inline void fillImpl(KRef thiz, KInt fromIndex, KInt toIndex, T value) {
+  ArrayHeader* array = thiz->array();
+  checkRangeIndexes(fromIndex, toIndex, array->count_);
+  mutabilityCheck(thiz);
+  T* address = PrimitiveArrayAddressOfElementAt<T>(array, fromIndex);
```
ilya-g: Is this operation well defined when `fromIndex == array.size` and thus address is pointing past the last array element, given that it's not dereferenced in the loop below?

olonho: Yes, I believe so.

PR #3223, opened 2019-07-26
Made phase switcher aware of compiler args

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -357,6 +357,9 @@ val toplevelPhase: CompilerPhase<*, Unit, Unit> = namedUnitPhase(
                 linkPhase
 )
 
+internal fun PhaseConfig.switchSoft(phase: AnyNamedPhase, onOff: Boolean) =
```
SvyatoslavScherbina: Isn't it the same as `disableUnless`?
Such a name could be more understandable, and such a function could be implemented in a more understandable way, e.g.
```kotlin
internal fun PhaseConfig.disableUnless(phase: AnyNamedPhase, condition: Boolean) {
    if (!condition) disable(phase)
}
```

PR #3225, opened 2019-07-26
Bellard-Pi benchmark

msink:
Draft of new benchmark


LepilkinaElena:
If you start new subset main of program should include list of benchmarks and running launcher. Example https://github.com/JetBrains/kotlin-native/blob/master/performance/cinterop/src/main/kotlin/main.kt

Benchmark should be in another function, not in `main`.
Also, please, avoid output operations in benchmarks, these operations are very unstable and make performance results suspicious.

performance/numerical/src/nativeInterop/cinterop/pi.def
```diff
@@ -0,0 +1 @@
+package = pi
```
olonho: What is the point of that?

msink: It was just blindly copied from my old behchmark.
What should be here?

msink:
I'm not sure - should it better be added as separate subset, or as part of `ring` ?

LepilkinaElena:
I think it's better to separate it, because `ring` is more connected with simple Kotlin opportunities. Here idea is different use c libraries and analog on clean Kotlin. We have no set with such idea.

msink:
Well, Kotlin/Native test works.
Now - how to setup same for Kotlin/Cinterop and Kotlin/JVM ?

performance/numerical/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.konan.target.HostManager
+
+plugins {
+    id("benchmarking")
+}
+
+benchmark {
+    applicationName = "BellardPi"
+    commonSrcDirs = listOf("../../tools/benchmarks/shared/src", "src/main/kotlin", "../shared/src/main/kotlin")
```
LepilkinaElena: To build jvm version you now have to add endorsed library that is accessible from K/N, but not in JVM 
```
applicationName = "BellardPi"
    commonSrcDirs = listOf("../../tools/benchmarks/shared/src", "src/main/kotlin", "../shared/src/main/kotlin", "../../endorsedLibraries/kliopt/src/main/kotlin")
    jvmSrcDirs = listOf("../shared/src/main/kotlin-jvm", "../../endorsedLibraries/kliopt/src/main/kotlin-jvm")
    nativeSrcDirs = listOf("../shared/src/main/kotlin-native/common", "../../endorsedLibraries/kliopt/src/main/kotlin-native")
```

LepilkinaElena:
For running Kotlin/JVM, please, make suggested changes. I fetched your branch and with these changes JVM version is built and started to run (I'm waiting result).

What do you mean under Kotlin/Cinterop. `cinterop` is just subset of our benchmarks to measure such feature as interop with C.

LepilkinaElena:
After asked changes, I'm ready to merge this PR.
@olonho, I registered this benchmark under current name in our system, so it should be safe to merge.

msink:
> What do you mean under Kotlin/Cinterop

Well - ideally there should be two flavors of Kotlin/Native tests - one compiled from Kotlin source, on one compiled by clang from C and called using cinterop.

And third flavor - for Kotlin/JVM.

LepilkinaElena:
Ok, I understood, but then they should be benchmarks with different in names. Difference is in build and sources, but not in way to run.

msink:
Last commit adds Cinterop test, but somehow broke  test for JVM.
I do not understand what it wants.

msink:
Well, JVM build fixed by adding expect/actual declarations.

Now, results:
JVM: 7600000
Clang: 7100000
Konan: 4200000

Weird.

msink:
So now it should be ready for merge, and you can continue to tune it as you want.

LepilkinaElena:
Thank you for your work.

Sorry for delay, I was on vacation.
I have small notes connected with names. It will be difficult to compare when names are different. May be let's stay for JVM and Konan `BellardPi` and for Cinterop choose another name  `BellardPiCinterop`? 
Also you call directory with new benchmark set as `numerical`, then  `applicationName` in `build.gradle.kts` better to be `Numerical`.

LepilkinaElena:
Also about weird results, did you mean that konan is faster than version with interop?
I agree that it's strange. I checked on your separate repository, changed kotlin/native compiler on last build from branch. For cinterop version I got similar results, for konan version in your separate repo they are more similar on truth. I'll explore.

msink:
Renamed.
About resulst - yes, I expected more or less same results for all 3 tests...

LepilkinaElena:
It seems that I understood why result for `Konan` is wrong. I think reason is added some time ago dead code elimination. I'll add mechanism to escape result of benchmark being unneeded.

PR #3231, opened 2019-07-30
[codegen][debug info] initialise end of location info with start if it wasn't provided

vvlevchenko:


backend.native/tests/interop/objc_with_initializer/objc_test.kt
```diff
@@ -0,0 +1,7 @@
+import objc_misc.*
```
SvyatoslavScherbina: This test doesn't reproduce the original issue.

SvyatoslavScherbina: (Does reproduce if compiled with `-g`).

PR #3233, opened 2019-08-01
Improve stack traces in optimized -Xg0 binaries.

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmAttributes.kt
```diff
@@ -0,0 +1,50 @@
...
+    return false
+}
+
+private fun shouldEnforceFramePointer(context: Context): Boolean = when (context.config.target.family) {
+    Family.OSX, Family.IOS -> context.shouldContainLocationDebugInfo()
+    Family.LINUX, Family.MINGW, Family.ANDROID, Family.WASM, Family.ZEPHYR -> false
+}
+
+private fun enforceFramePointer(llvmFunction: LLVMValueRef) {
+    addLlvmFunctionStringAttribute(llvmFunction, "no-frame-pointer-elim", "true")
```
sbogolepov: `LLVMAddTargetDependentFunctionAttr`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmAttributes.kt
```diff
@@ -0,0 +1,50 @@
...
+            return true
+        }
+    }
+
+    return false
+}
+
+private fun shouldEnforceFramePointer(context: Context): Boolean = when (context.config.target.family) {
+    Family.OSX, Family.IOS -> context.shouldContainLocationDebugInfo()
+    Family.LINUX, Family.MINGW, Family.ANDROID, Family.WASM, Family.ZEPHYR -> false
```
sbogolepov: We have another place were we may set `no-frame-pointer-elim` attribute:
https://github.com/JetBrains/kotlin-native/blob/72350cb03017f63ead970c6a1a782679226a31d0/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt#L376
Maybe unify these checks?

runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -121,12 +121,12 @@ extern "C" {
...
 // TODO: this implementation is just a hack, e.g. the result is inexact;
 // however it is better to have an inexact stacktrace than not to have any.
-OBJ_GETTER0(GetCurrentStackTrace) {
+NO_INLINE OBJ_GETTER0(Kotlin_getCurrentStackTrace) {
 #if OMIT_BACKTRACE
   return AllocArrayInstance(theNativePtrArrayTypeInfo, 0, OBJ_RESULT);
 #else
-  // Skips first 3 elements as irrelevant.
-  constexpr int kSkipFrames = 3;
+  // Skips first 2 elements as irrelevant.
```
sbogolepov: It may be a good idea explain what frames we're dropping to reduce cognitive load :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmAttributes.kt
SvyatoslavScherbina: Nice shortcut, thanks!

PR #3262, opened 2019-08-19
Limited C++ interop provided via plain C wrappers and cinterop mechan…

knebekaizer:
…ism.

Add `Language = C++` to .def file or `compilerOpts = -x c++` to command line or def file.

* C++ class:
  + Virtual and non-virtual methods
  * Static methods mapped as companion ones
  * Fields
  * Static fields as companion ones
  * Constructors are mapped to `__init__(args)` companion methods
  * Destructor is mapped to `__destroy__(this)` companion method
* LValueReference parameters and return value internally handled as pointers
* Namespaces provided as simple mangled class name. It works but awfully ugly. Shall be fixed by mapping to packages.
* Nested C++ classes: same as namespaces, simple mangling. Shall be fixed.
* Access modifiers: only public members exposed to Kotlin. Anonymous namespaces are silently ignored.

* C++ object with nontrivial copy constructor may work incorrect when used as by value parameter or return typr.
"Nontrivial" in this context relates to objects which behavior depends on memory location. Most of other non-POD objects may be handled well.
In fact, "by value" return type is mapped to CValue which is immutrable movable block. Particularly, non-const methods invoked with CValue receiver
via useContents mechanism run on temporary object and therefore does not modify the original copy. TBD.

* Operators are not supported yet (silently ignored)
* LVReference is mapped to CPointer<T>? which is incorrect (should be notNull). This may cause segmentation fault in case of null would be sent as a parameter. TBD
* const overload not supported and cause compilation error. That is, two class methods with the same signature (`const` and `non-const`) can't be compiled. The same for function parameters: if two functions differ only in `const*` modifier of parameter, this will cause "conflicting overloads" error. TBD.
* C++ lambda type is not supported yet.
* Member pointer, member reference, rvalue reference and some other types are not wupported.
* Inheritance is not implemented yet. C++-style callbacks (overriding virtual method in Kotlin) may be implemented via plain C bridge (this can be done by hand as a workaround). TBD.


msink:
Well, tried this on Windows GDI+

Simplest header:
```c
#ifndef __GDIPLUS_MEM_H
#define __GDIPLUS_MEM_H
#if __GNUC__ >=3
#pragma GCC system_header
#endif

#ifdef __cplusplus
namespace DllExports {
extern "C" {
#endif

VOID* WINGDIPAPI GdipAlloc(size_t);
VOID WINGDIPAPI GdipFree(VOID*);

#ifdef __cplusplus
}  /* extern "C" */
}  /* namespace DllExports */
#endif

#endif /* __GDIPLUS_MEM_H */
```
currently compiles to:

```kt
fun GdipAlloc(arg0: size_t): COpaquePointer? {
    return interpretCPointer<COpaque>(kniBridge5(arg0))
}

fun GdipFree(arg0: CValuesRef<*>?): kotlin.Unit {
    memScoped {
        return kniBridge6(arg0?.getPointer(memScope).rawValue)
    }
}
```
But when switched to C++ mode:
``` patch
--- a/platformLibs/src/platform/mingw/gdiplus.def
+++ b/platformLibs/src/platform/mingw/gdiplus.def
@@ -1,4 +1,5 @@
 depends = windows posix
+language = C++
 package = platform.gdiplus
 headers = wtypes.h minwindef.h wtypes.h gdiplus/gdiplus.h
 headerFilter = gdiplus/**
```
It becomes:
```kt
@Deprecated("Unable to import this declaration", level = DeprecationLevel.ERROR)
fun Gdiplus__DllExports____GdipAlloc(arg0: size_t): COpaquePointer? = throw UnsupportedOperationException()

@Deprecated("Unable to import this declaration", level = DeprecationLevel.ERROR)
fun Gdiplus__DllExports____GdipFree(arg0: CValuesRef<*>?): Unit = throw UnsupportedOperationException()
```

knebekaizer:
> Simplest header:
Looks like the generated C++ stubs are not compilable. However, this header itself is not self-sufficient (does not include anything), so it's not compilable for me too. What is the command line in your case? I may need the whole environment, included precompiled headers, macros etc, to be able to reproduce the issue. 
My guess is that additional defines specific for C++ build shall be provided in your case. Essentially, all g++ command line options which shall be used to compile one-liner `include "myheader.h"` : `g++ -c myoneliner.cpp -D<whatever> -I<whatever> ...`. That is, required C++ defines may differ with C ones.

msink:
Looks that `extern "C"` breaks things:

Reproducer:
```patch
--- a/samples/cpptest/src/native/features.h
+++ b/samples/cpptest/src/native/features.h
@@ -209,3 +209,18 @@ struct Smth {
     IntTmplStruct intTmplStruct;
 };
 */
+
+#ifdef __cplusplus
+namespace Gdiplus {
+namespace DllExports {
+extern "C" {
+#endif
+
+void* __stdcall GdipAlloc(int);
+void __stdcall GdipFree(void*);
+
+#ifdef __cplusplus
+}
+}
+}
+#endif
```

msink:
Also, maybe there should be some way to skip `Gdiplus__` prefix from everywhere, because it already imported to `platform.gdiplus` package?

knebekaizer:
Namespace support seems to be incomplete and shall be fixed. I'll take a look. However, namespaces to be reimplemented soon (now namespace is just a mangled prefix but shall be mapped to kotlin package) and this will break backward compatibility anyway.  Let's see if this make sense to cancel now and wait until namespace rework is completed.

jurmous:
This MR is paused and waiting for the namespace rework? Where can we follow the namespace rework?

knebekaizer:
Research only prototype. Postponed.

PR #3266, opened 2019-08-20
Rework entry point selector.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/EntryPoint.kt
```diff
@@ -0,0 +1,93 @@
...
+    entryPointDescriptor.bind(entryPoint)
+    entryPoint.annotations += buildSimpleAnnotation(context.irBuiltIns,
+            actualMain.startOffset, actualMain.startOffset,
+            context.ir.symbols.exportForCppRuntime.owner, "Konan_start")
+
+    val builder = context.createIrBuilder(entryPoint.symbol)
+    entryPoint.body = builder.irBlockBody(entryPoint) {
+        +IrTryImpl(
+                startOffset = actualMain.startOffset,
+                endOffset   = actualMain.startOffset,
```
SvyatoslavScherbina: Is this intended?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -316,6 +318,9 @@ internal val dependenciesLowerPhase = SameTypeNamedPhaseWrapper(
                         }
                 input.files += files
 
+                if (context.config.produce == CompilerOutputKind.PROGRAM)
```
SvyatoslavScherbina: Adding entry point seems unrelated to lowering dependencies. Maybe extract explicit phase for this?

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -180,4 +180,13 @@ internal fun <T> listOfInternal(vararg elements: T): List<T> {
     for (i in 0 until elements.size)
         result.add(elements[i])
     return result
-}
\ No newline at end of file
+}
+
+
+@PublishedApi
```
SvyatoslavScherbina: Why not have a single function here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/EntryPoint.kt
olonho: Yes, here and below we just set entry point as if it was in very beginning of the main.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
olonho: Well, not sure if it shall a phase on its own. WDYT?

SvyatoslavScherbina: What do you mean?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -320,6 +322,20 @@ internal val dependenciesLowerPhase = SameTypeNamedPhaseWrapper(
...
 
+internal val entryPointPhase = SameTypeNamedPhaseWrapper(
+        name = "addEntryPoint",
+        description = "Add entry point for program",
+        prerequisite = emptySet(),
+        lower = object : CompilerPhase<Context, IrModuleFragment, IrModuleFragment> {
+            override fun invoke(phaseConfig: PhaseConfig, phaserState: PhaserState<IrModuleFragment>,
+                                context: Context, input: IrModuleFragment): IrModuleFragment {
+                assert(context.config.produce == CompilerOutputKind.PROGRAM)
+                context.ir.symbols.entryPoint!!.owner.file.addChild(makeEntryPoint(context))
```
SvyatoslavScherbina: Btw, this will not work for entry point in a "cached" dependency. I guess we'll need to generate a synthetic file for this.

olonho: Yeah, likely so.

PR #3274, opened 2019-08-22
Remove Gradle plugin docs from the site

ilmat192:


ilmat192:
The doc contain a migration guide so it's better to keep it.

PR #3287, opened 2019-08-28
Added a note on read-only property's setters

artdfel:


OBJC_INTEROP.md
```diff
@@ -67,6 +67,10 @@ as a Swift extension, because Kotlin has no concept of extension constructors.
 
 Kotlin constructors are imported as initializers to Swift/Objective-C. 
 
+### Setters
+
+Reassignable Objective-C properties, that are overrides of the read-only properties, are being represented in Kotlin/Native as read-only properties, with public `set()` function.
```
olonho: Maybe smth like

Writeable Objective-C properties, overriding read-only properties of the superclass are represented as `setFoo` method for the property `foo`.

olonho: Also please add the following (somewhere in inheritance section):

When implementing Objective-C interface with multiple methods sharing the same signature and only differing by selectors (such as `CBCentralManagerDelegateProtocol`), please use `@Suppress("CONFLICTING_OVERLOADS")` annotation on the class implementing such an interface.

artdfel: I've made some additions, please have a look. About conflicting overloads, it was already described in the "Subclassing" section.

OBJC_INTEROP.md
```diff
@@ -67,6 +67,12 @@ as a Swift extension, because Kotlin has no concept of extension constructors.
 
 Kotlin constructors are imported as initializers to Swift/Objective-C. 
 
+### Setters
+
+Objective C provides an ability to override superclasses read-only properties with writeable ones.
```
SvyatoslavScherbina: > superclasses read-only properties

what?

OBJC_INTEROP.md
```diff
@@ -67,6 +67,12 @@ as a Swift extension, because Kotlin has no concept of extension constructors.
...
+These properties are being represented in Kotlin/Native as read-only properties, with generated public setter function.
```
SvyatoslavScherbina: > These properties

Which ones?

PR #3288, opened 2019-08-28
Add "main" as alias to "Konan_main" via LLVMAddAlias

sbogolepov:
It allows to produce programs with embedded bitcode.


olonho:
Guess it's better to avoid doing that and instead generate bitcode that aliases (with `LLVMAddAlias`) `main` to `Konan_main` when creating program and `-nomain` not specified.

PR #3289, opened 2019-08-28
[REVERTME][gradle][plugin][tests] filter out enabled but not supported platforms

vvlevchenko:
see bc1b67678 for details


tools/kotlin-native-gradle-plugin/src/test/groovy/org/jetbrains/kotlin/gradle/plugin/test/PathSpecification.groovy
```diff
@@ -18,6 +18,7 @@ package org.jetbrains.kotlin.gradle.plugin.test
 
 import org.gradle.testkit.runner.TaskOutcome
 import org.jetbrains.kotlin.konan.target.PlatformManager
+import org.jetbrains.kotlin.konan.target.KonanTarget
```
ilmat192: Unused import?

PR #3292, opened 2019-08-29
[codegen] switch scopes for local variable initializers according to their locations provided by inliner.

vvlevchenko:
depends on https://github.com/JetBrains/kotlin/pull/2545

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
@@ -9,6 +9,7 @@ import kotlinx.cinterop.*
 import llvm.*
 import org.jetbrains.kotlin.backend.common.descriptors.allParameters
 import org.jetbrains.kotlin.backend.common.ir.ir2string
+import org.jetbrains.kotlin.backend.common.lower.inline.InlinerExpressionLocationHint
 import org.jetbrains.kotlin.backend.konan.*
 import org.jetbrains.kotlin.backend.konan.descriptors.*
 import org.jetbrains.kotlin.backend.konan.ir.*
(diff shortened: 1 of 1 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -276,6 +278,35 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
         }
     }
 
+    private fun <R> switchTo(irFunction: IrFunction, block: () -> R): R? {
+        val oldCodeContext = currentCodeContext
+
+        try {
+            if ((currentCodeContext as? FunctionScope)?.declaration == irFunction
+                    || (currentCodeContext as? ReturnableBlockScope)?.returnableBlock?.inlineFunctionSymbol?.owner == irFunction
+                    || (currentCodeContext as? FunctionScope)?.declaration == irFunction)
```
homuroll: Same predicate twice?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -276,6 +278,35 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+                return block()
+            currentCodeContext = (currentCodeContext as? InnerScope)?.outerContext ?: return null
```
homuroll: So there is a situation when we don't call `block`? Is it correct? May be better would be to throw an error?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -276,6 +278,35 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+            return switchTo(irFunction, block)
+        } finally {
+            currentCodeContext = oldCodeContext
+        }
+    }
+
+    private fun <R> switchTo(irFile: IrFile, block: () -> R): R? {
```
homuroll: Same questions as above here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
vvlevchenko: correctness of code generation  doesn't degrade, because we process this case in `generateVariable`, but what relates to debug information generation, that yes it'll be inaccurate.

homuroll: Ah, i see.

PR #3295, opened 2019-08-30
Backend: Partially get rid of descriptors in InteropLowering

ilmat192:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -888,8 +892,8 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
                     val target = targetSymbol.owner
                     val signatureTypes = target.allParameters.map { it.type } + target.returnType
 
-                    descriptor.typeParameters.forEachIndexed { index, typeParameterDescriptor ->
-                        val typeArgument = expression.getTypeArgument(typeParameterDescriptor)!!.toKotlinType()
+                    function.typeParameters.forEachIndexed { index, irTypeParameter ->
```
homuroll: Actually, `irTypeParameter.index = index`, I think the situation where `typeParameters` are returned shuffled is just ridiculous.

ilmat192: Fixed

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -260,9 +265,9 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
             IrDeclarationOriginImpl("OVERRIDING_INITIALIZER_BY_CONSTRUCTOR")
 
     private fun IrConstructor.overridesConstructor(other: IrConstructor): Boolean {
-        return this.descriptor.valueParameters.size == other.descriptor.valueParameters.size &&
-                this.descriptor.valueParameters.all {
-                    val otherParameter = other.descriptor.valueParameters[it.index]
+        return this.valueParameters.size == other.valueParameters.size &&
```
SvyatoslavScherbina: Does this particular patch fixes anything?
This predicate is technically related to frontend, so I'm not sure it is worth rewriting in terms of IR.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -301,7 +306,7 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
 
         val outlet = "@${context.interopBuiltIns.objCOutlet.name}"
 
-        if (!descriptor.isVar) {
+        if (!property.isVar) {
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -436,32 +441,32 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
     }
 
     private fun checkKotlinObjCClass(irClass: IrClass) {
-        val kind = irClass.descriptor.kind
+        val kind = irClass.kind
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -436,32 +441,32 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
...
         if (kind != ClassKind.CLASS && kind != ClassKind.OBJECT) {
             context.reportCompilationError(
                     "Only classes are supported as subtypes of Objective-C types",
                     currentFile, irClass
             )
         }
 
-        if (!irClass.descriptor.isFinalClass) {
+        if (!irClass.isFinalClass) {
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -436,32 +441,32 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
...
                     "Non-final Kotlin subclasses of Objective-C classes are not yet supported",
                     currentFile, irClass
             )
         }
 
         var hasObjCClassSupertype = false
-        irClass.descriptor.defaultType.constructor.supertypes.forEach {
-            val descriptor = it.constructor.declarationDescriptor as ClassDescriptor
-            if (!descriptor.isObjCClass()) {
+        irClass.superTypes.forEach {
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -436,32 +441,32 @@ internal class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfo
...
+            val superClass = it.getClass()!!
+            if (!superClass.isObjCClass()) {
                 context.reportCompilationError(
                         "Mixing Kotlin and Objective-C supertypes is not supported",
                         currentFile, irClass
                 )
             }
 
-            if (descriptor.kind == ClassKind.CLASS) {
+            if (superClass.kind == ClassKind.CLASS) {
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -888,8 +892,8 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
                     val target = targetSymbol.owner
                     val signatureTypes = target.allParameters.map { it.type } + target.returnType
 
-                    descriptor.typeParameters.forEachIndexed { index, typeParameterDescriptor ->
-                        val typeArgument = expression.getTypeArgument(typeParameterDescriptor)!!.toKotlinType()
+                    function.typeParameters.indices.forEach { index ->
```
SvyatoslavScherbina: This should also be a part of frontend.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -1122,11 +1126,24 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
 
         return argument.statements.last() as? IrFunctionReference
     }
+
+    val IrValueParameter.isDispatchReceiver: Boolean
+        get() = parent.let {
+            when {
```
SvyatoslavScherbina:
```kotlin
get() = when (val parent = this.parent) {
```

ilmat192: It seems to be impossible to use such a captured variable in _conditions_ of the `when` operator (only in actions):
```
InteropLowering.kt: (1132, 13): Incompatible types: Boolean and IrDeclarationParent
```

SvyatoslavScherbina:
```kotlin
val IrValueParameter.isDispatchReceiver: Boolean
    get() = when (val parent = this.parent) {
       is IrClass -> true
       is IrFunction -> parent.dispatchReceiverParameter == this
       else -> false
    }
```
?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
ilmat192: This patch fixes using the `-Xinclude` option for a library containing a class constructor annotated with `@OverrideInit`. E.g. if the following class is compiled into a klib:
```kotlin
@ExportObjCClass
class ViewController : NSObject {
    @OverrideInit constructor() : super()
}
```
and then passed to the compiler using the `-Xinclude` option, [this](https://github.com/JetBrains/kotlin-native/pull/3295/files/7a70e1e0833aed3c7282108335a5976f2b1e8906#diff-4fbff1fd1ac95615545d03980ba283a4R1008) check based on descriptors fails:
```
error: receiver of 'initBy' must be a 'this' of the constructed class
```

Migrating to IR fixes this issue. I agree that such checks should be done by the frontend. But currently switching to IR here is the simplest way to fix this problem. It's important because I'm going to enable the two-stage compilation of final binaries in the MPP plugin by default.

Actually to fix this issue only aforementioned check should be switched to IR. But migrating to IR other parts of this lowering is usually simple (and I didn't change parts where such migrating is more or less complex) and may help us to avoid similar failures on other checks.

ilmat192: Opened a ticket to save the reproducing steps: https://youtrack.jetbrains.com/issue/KT-33645.

SvyatoslavScherbina: I'm not asking about the entire PR, but only about this particular change in `overridesConstructor`.
What does it fix?

> and may help us to avoid similar failures on other checks.

But will complicate moving this checks to frontend.
I somewhat dislike fixing the code that is not broken.

Also note that the actual bug is likely located in wrapped descriptors implementation.

ilmat192: > What does it fix?

Nothing particular, just a refactoring.

> But will complicate moving this checks to frontend.

Sounds reasonable. Ok, I'll return descriptors in all places you marked as belonging to the frontend.

> Also note that the actual bug is likely located in wrapped descriptors implementation.

As far as I understand the preferred approach in such cases is to migrate from wrapped descriptors to IR instead of fixing wrapped descriptors.

PR #3296, opened 2019-08-30
Use SelectionDAG for `-g` builds

sbogolepov:


konan/konan.properties
```diff
@@ -117,7 +117,7 @@ targetSysRoot.ios_arm64 = target-sysroot-9-ios_arm64
 clangFlags.ios_arm64 = -cc1 -emit-obj -disable-llvm-passes -x ir
 clangNooptFlags.ios_arm64 = -O1
 clangOptFlags.ios_arm64 = -O3
-clangDebugFlags.ios_arm64 = -O0
+clangDebugFlags.ios_arm64 = -O0 -mllvm -fast-isel=false -mllvm -global-isel=false
```
olonho: Better somehow mark it with bug reference and REMOVEME.

konan/konan.properties
```diff
@@ -117,7 +117,11 @@ targetSysRoot.ios_arm64 = target-sysroot-9-ios_arm64
 clangFlags.ios_arm64 = -cc1 -emit-obj -disable-llvm-passes -x ir
 clangNooptFlags.ios_arm64 = -O1
 clangOptFlags.ios_arm64 = -O3
-clangDebugFlags.ios_arm64 = -O0
+# We force LLVM to use SelectionDAG because fast isel may miscompile sign extension.
+# Related:
```
SvyatoslavScherbina: Isn't original LLVM bug related too?

konan/konan.properties
```diff
@@ -117,7 +117,11 @@ targetSysRoot.ios_arm64 = target-sysroot-9-ios_arm64
...
+# https://youtrack.jetbrains.com/issue/KT-33619
+# https://github.com/JetBrains/kotlin-native/issues/3290
+clangDebugFlags.ios_arm64 = -O0 -mllvm -fast-isel=false -mllvm -global-isel=false
```
SvyatoslavScherbina: Why `false` for both of them?

sbogolepov: To force LLVM to use the third one -- SelectionDAG.

konan/konan.properties
sbogolepov: Nope.

konan/konan.properties
olonho: Looks somewhat fragile.

sbogolepov: AFAIK there is no flag to explicitly enable SelectionDAG. 
¯\_(ツ)_/¯

PR #3303, opened 2019-09-02
Preliminary tvOS support

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -183,23 +182,31 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
...
+        }
         contents.append(when (target.family) {
             Family.IOS -> """
                 |    <key>MinimumOSVersion</key>
                 |    <string>$minimumOsVersion</string>
                 |    <key>UIDeviceFamily</key>
                 |    <array>
-                |        <integer>1</integer>
-                |        <integer>2</integer>
+                |       $uiDeviceFamilyValues
```
SvyatoslavScherbina: Will this emit proper indent for second `uiDeviceFamilyValues` line?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -183,23 +182,31 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
...
                 |    </array>
 
                 """.trimMargin()
             Family.OSX -> ""
             else -> error(target)
         })
 
-        if (target == KonanTarget.IOS_ARM64) {
+        if (target.architecture == Architecture.ARM64) {
```
SvyatoslavScherbina: Does Xcode emit `UIRequiredDeviceCapabilities` with `arm64` for tvOS arm64 device framework?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeEmbedding.kt
```diff
@@ -17,31 +19,22 @@ object BitcodeEmbedding {
...
-        Mode.MARKER -> {
-            addEmptyMarker(llvm, "konan_llvm_bitcode", "__LLVM,__bitcode")
-            addEmptyMarker(llvm, "konan_llvm_cmdline", "__LLVM,__cmdline")
+    // TODO: We cannot produce programs due to `-alias` linked flag.
+    private fun KonanConfig.shouldForceBitcodeEmbedding() =
+            target.isTvOs && produce != CompilerOutputKind.PROGRAM
+
+    private val KonanConfig.bitcodeEmbeddingMode: Mode
+        get() = when {
+            shouldForceBitcodeEmbedding() -> if (debug) Mode.MARKER else Mode.FULL
```
SvyatoslavScherbina: Not sure if CLI compiler driver should do this.
Does `clang` enforce bitcode for tvOS?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -18,6 +19,13 @@ internal fun determineLinkerOutput(context: Context): LinkerOutputKind =
             else -> TODO("${context.config.produce} should not reach native linker stage")
         }
 
+// TODO: Move to a more appropriate place.
+internal val KonanTarget.isAppleTarget: Boolean
```
SvyatoslavScherbina: This name doesn't reflect the actual conditions that need to be checked on its use sites.
This is generally more like `IsThisParticularKindOfObjCInteropEnabled`.
I would prefer using old target-family-based approach for now.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -18,6 +19,13 @@ internal fun determineLinkerOutput(context: Context): LinkerOutputKind =
...
+        get() = family == Family.IOS || family == Family.OSX
+
+internal val KonanTarget.isTvOs: Boolean
```
SvyatoslavScherbina: `isTvOS`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
SvyatoslavScherbina: Unless you have short-term plans to add more relevant target families of course.

build-tools/src/main/kotlin/org/jetbrains/kotlin/FrameworkTest.kt
```diff
@@ -131,9 +150,10 @@ open class FrameworkTest : DefaultTask() {
         val bitcodeBuildTool = "${configurables.absoluteAdditionalToolsDir}/bin/bitcode-build-tool"
         val ldPath = "${configurables.absoluteTargetToolchain}/usr/bin/ld"
         val sdk = when (testTarget) {
-            KonanTarget.IOS_X64 -> return // bitcode-build-tool doesn't support iPhone Simulator.
+            KonanTarget.IOS_X64, KonanTarget.TVOS_X64 -> return // bitcode-build-tool doesn't support iPhone Simulator.
```
SvyatoslavScherbina: Comment should be updated too.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -69,6 +69,12 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
                 KonanTarget.IOS_ARM64 ->
                     listOf("-stdlib=libc++", "-arch", "arm64", "-isysroot", absoluteTargetSysRoot, "-miphoneos-version-min=9.0.0")
 
+                KonanTarget.TVOS_ARM64 ->
+                    listOf("-stdlib=libc++", "-arch", "arm64", "-isysroot", absoluteTargetSysRoot, "-mtvos-version-min=12.2")
```
SvyatoslavScherbina: Better use older version.

build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
```diff
@@ -0,0 +1,63 @@
...
+ */
+private fun compareStringsAsVersions(version1: String, version2: String): Int {
+    val version1 = version1.split('.').map { it.toInt() }
+    val version2 = version2.split('.').map { it.toInt() }
+    val minimalLength = min(version1.size, version2.size)
+    for (index in 0 until minimalLength) {
+        if (version1[index] < version2[index]) return -1;
+        if (version1[index] > version2[index]) return 1;
+    }
+    return when {
```
SvyatoslavScherbina: Is it different from 
```kotlin
return version1.size.compareTo(version2.size)
```
?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -136,19 +136,22 @@ open class MacOSBasedLinker(targetProperties: AppleConfigurables)
     private val dsymutil = "$absoluteLlvmHome/bin/llvm-dsymutil"
 
     private fun provideCompilerRtLibrary(libraryName: String): String? {
-        val suffix = if (libraryName.isNotEmpty() && target == KonanTarget.IOS_X64) {
-            "iossim"
+        val prefix = when (val family = target.family) {
```
SvyatoslavScherbina: Why not just have `when` over targets?
This would be simpler and more discoverable.

konan/konan.properties
```diff
@@ -155,6 +155,60 @@ osVersionMinFlagLd.ios_x64 = -ios_simulator_version_min
...
+
+clangFlags.tvos_arm64 = -cc1 -emit-obj -disable-llvm-passes -x ir
+clangNooptFlags.tvos_arm64 = -O1
+clangOptFlags.tvos_arm64 = -O3
+clangDebugFlags.tvos_arm64 = -O0
+clangDynamicFlags.tvos_arm64 =
+
+linkerNoDebugFlags.tvos_arm64 = -S
+linkerDynamicFlags.tvos_arm64 = -dylib
+linkerKonanFlags.tvos_arm64 = -lSystem -lc++ -lobjc -framework Foundation -sdk_version 11.2
```
SvyatoslavScherbina: Different SDK versions for tvos_arm64 and tvos_x64.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -247,15 +248,15 @@ internal val ObjCContainer.classOrProtocol: ObjCClassOrProtocol
  */
 internal fun Type.isStret(target: KonanTarget): Boolean {
     val unwrappedType = this.unwrapTypedefs()
-    return when (target) {
-        KonanTarget.IOS_ARM64 ->
+    return when (target.architecture) {
```
SvyatoslavScherbina: Checking only architecture here is wrong, as targets with same architecture can have different ABIs.

olonho: +1, guess current approach with explicit selection seems better fitting.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
```diff
@@ -21,8 +21,8 @@ class CStubsManager(private val target: KonanTarget) {
         if (stubs.isEmpty()) return null
 
         val compilerOptions = mutableListOf<String>()
-        val sourceFileExtension = when (target.family) {
-            Family.OSX, Family.IOS -> {
+        val sourceFileExtension = when {
+            target.isAppleTarget -> {
```
olonho: Guess previous version was readable enough.

build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
```diff
@@ -0,0 +1,63 @@
+package org.jetbrains.kotlin
+
+import kotlinx.serialization.Serializable
```
olonho: Hmm, is it indeed a good idea to depend on `kotlinx.serialization.` here?

sbogolepov: We already have dependencies on it in `build-utils` (see `ExternalReportUtils.kt`) and I don't think it can cause any trouble.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
sbogolepov: Yes, I think we should introduce `Family.TVOS`, `Family.WATCHOS` because different flavours of `iOS` have different platform libs and runtime dependencies. And `when` over all Apple targets is getting kinda cumbersome.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
sbogolepov: Renamed to `isTvOsBased`. Will be removed with introduction of `Family.TVOS`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeEmbedding.kt
sbogolepov: Agree, it's a responsibility of a build system.
> Does clang enforce bitcode for tvOS?
No.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CStubsManager.kt
sbogolepov: If we introduce `Family.TVOS`, `Family.WATCHOS` then it will be replaced with something like `target.family.isAppleFamily`/ `isDarwinBased`.

olonho: TVOS and WATCHOS are pretty much the same as IOS, so why not just stick this way.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
olonho: You mean platform libraries are deduced from the family in https://github.com/JetBrains/kotlin-native/blob/bcfd196c0498195e3eecd7ef4e91fa2c7f6a194f/platformLibs/build.gradle#L39? This could be worked around (like hack for MIPS zlib), but indeed may be a good idea to use OS family as identity for the set of platform libs.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
olonho: Don't we have `osVersionMin.tvos_x64` and such?

olonho:
What about platform libs?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
SvyatoslavScherbina: No, `isTvOS` is good. I mean that `s` should probably be capital.

sbogolepov:
I suggest to add them separately to make review easier.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
sbogolepov: Yes, that's what I mean.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
sbogolepov: Ok. Anyway, `isTvOsBased` feels a bit better since target is not an OS :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
sbogolepov: Nope, it's incorrect. 👍

build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
```diff
@@ -15,21 +15,17 @@ private fun compareStringsAsVersions(version1: String, version2: String): Int {
...
     val minimalLength = min(version1.size, version2.size)
     for (index in 0 until minimalLength) {
-        if (version1[index] < version2[index]) return -1;
-        if (version1[index] > version2[index]) return 1;
-    }
-    return when {
-        version1.size == version2.size -> 0
-        version1.size < version2.size -> -1
-        else -> 1
+        if (version1[index] < version2[index]) return -1
```
olonho: Why not use compareTo here as well?

sbogolepov: Because we want to continue iteration in case of equality.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -61,22 +67,22 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
                     listOf("-target", targetArg!!, "--sysroot=$absoluteTargetSysRoot", "-Xclang", "-flto-visibility-public-std")
 
                 KonanTarget.MACOS_X64 ->
-                    listOf("--sysroot=$absoluteTargetSysRoot", "-mmacosx-version-min=10.11")
+                    listOf("--sysroot=$absoluteTargetSysRoot", "-mmacosx-version-min=$osVersionMin")
 
                 KonanTarget.IOS_ARM32 ->
-                    listOf("-stdlib=libc++", "-arch", "armv7", "-isysroot", absoluteTargetSysRoot, "-miphoneos-version-min=9.0.0")
+                    listOf("-stdlib=libc++", "-arch", "armv7", "-isysroot", absoluteTargetSysRoot, "-miphoneos-version-min=$osVersionMin")
```
SvyatoslavScherbina: Isn't `osVersionMin` actually differ from `"9.0.0"`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -182,31 +182,31 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
...
         """.trimIndent())
 
-        val uiDeviceFamilyValues = if (target.isTvOs) {
-            """
-                |<integer>3</integer>
-            """.trimIndent()
+        // 1 - iPhone
+        // 2 - iPad
+        // 3 - AppleTV
+        val uiDeviceFamilyValues = if (target.isTvOsBased) {
```
SvyatoslavScherbina: This will anyway turn into `when` over target (or target family) after adding watchOS, making `isTvOsBased` unnecessary. Then why not just use this fact?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -182,31 +182,31 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
...
         contents.append(when (target.family) {
             Family.IOS -> """
                 |    <key>MinimumOSVersion</key>
                 |    <string>$minimumOsVersion</string>
                 |    <key>UIDeviceFamily</key>
                 |    <array>
-                |       $uiDeviceFamilyValues
+                ${uiDeviceFamilyValues.joinToString(separator = "\n") { 
+                "|       <integer>$it</integer>" 
+            }}
```
SvyatoslavScherbina: Weird indent?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -16,6 +16,7 @@
 
 package org.jetbrains.kotlin.native.interop.gen
 
+import org.jetbrains.kotlin.konan.target.Architecture
```
SvyatoslavScherbina: Unused import.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -248,19 +249,24 @@ internal val ObjCContainer.classOrProtocol: ObjCClassOrProtocol
 internal fun Type.isStret(target: KonanTarget): Boolean {
     val unwrappedType = this.unwrapTypedefs()
     return when (target) {
-        KonanTarget.IOS_ARM64 ->
+        KonanTarget.IOS_ARM64,
+        KonanTarget.TVOS_ARM64 ->
```
SvyatoslavScherbina: Btw, have you verified that tvOS arm64 ABI doesn't use stret too?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
```diff
@@ -248,19 +249,24 @@ internal val ObjCContainer.classOrProtocol: ObjCClassOrProtocol
...
             false // On aarch64 stret is never the case, since an implicit argument gets passed on x8.
 
-        KonanTarget.IOS_X64, KonanTarget.MACOS_X64 -> when (unwrappedType) {
+        KonanTarget.IOS_X64,
+        KonanTarget.MACOS_X64,
+        KonanTarget.WATCHOS_X64,
+        KonanTarget.TVOS_X64 -> when (unwrappedType) {
```
SvyatoslavScherbina: Please ensure that watchOS and tvOS simulators use regular macOS x86_64 ABI.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
sbogolepov: Yes.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
sbogolepov: Well, it's 9.0. What kind of problem may it cause?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/ObjCStubs.kt
sbogolepov: Done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -183,21 +186,32 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
...
+                "        <integer>$it</integer>"
+            }
+            contents.append("""
                 |    <key>MinimumOSVersion</key>
                 |    <string>$minimumOsVersion</string>
                 |    <key>UIDeviceFamily</key>
                 |    <array>
-                |        <integer>1</integer>
-                |        <integer>2</integer>
+                |$xmlValues
```
SvyatoslavScherbina: The second line doesn't have `|` prefix. Does this work properly?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeEmbedding.kt
```diff
@@ -17,31 +16,16 @@ object BitcodeEmbedding {
...
     }
 
-    internal fun processModule(llvm: Llvm) = when (llvm.context.config.bitcodeEmbeddingMode) {
-        Mode.NONE -> {}
-        Mode.MARKER -> {
-            addEmptyMarker(llvm, "konan_llvm_bitcode", "__LLVM,__bitcode")
-            addEmptyMarker(llvm, "konan_llvm_cmdline", "__LLVM,__cmdline")
+    private val KonanConfig.bitcodeEmbeddingMode: Mode
+        get() = configuration.get(KonanConfigKeys.BITCODE_EMBEDDING_MODE)!!.also {
+            // TODO: We cannot produce programs due to `-alias` linked flag.
```
olonho: Hmm, so why not fix alias with LLVM API call as suggested in https://github.com/JetBrains/kotlin-native/pull/3288?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
sbogolepov: Yes.

PR #3308, opened 2019-09-03
Add Throwable.getStackTraceAddresses() 

SvyatoslavScherbina:
To allow integration with third-party crash analytic tools.

Also fix Throwable.getStackTrace() implementation.

### DIFF
--- a/runtime/src/main/kotlin/kotlin/Throwable.kt
+++ b/runtime/src/main/kotlin/kotlin/Throwable.kt
@@ -5,6 +5,7 @@
 
 package kotlin
 
+import kotlin.native.concurrent.freeze
 import kotlin.native.internal.ExportForCppRuntime
 import kotlin.native.internal.ExportTypeInfo
 import kotlin.native.internal.NativePtrArray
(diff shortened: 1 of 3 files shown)


runtime/src/main/kotlin/kotlin/Throwable.kt
```diff
@@ -31,7 +31,10 @@ public open class Throwable(open val message: String?, open val cause: Throwable
         getStackTraceStrings(stackTrace)
     }
 
-    public fun getStackTrace(): Array<String> = stackTraceStrings
+    public fun getStackTrace(): Array<String> = stackTraceStrings.copyOf()
```
olonho: Better freeze after taking, if you want it immutable.

runtime/src/main/kotlin/kotlin/native/ThrowableExtensions.kt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+public fun Throwable.getStackTraceAddresses(): List<Long> =
```
ilya-g: Please don't forget to add a kdoc summary for public API.

SvyatoslavScherbina: Thanks for noticing.
See #3364.

PR #3309, opened 2019-09-03
[IR SERIALIZATION] Support new deserializer scheme

romanart:
Goes along with https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3221


olonho:
Is it still relevant?

PR #3312, opened 2019-09-04
[KT-33525] Allow null literal in variadic arguments

knebekaizer:
Pass null as CTypes.voidPtr

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
@@ -45,8 +45,6 @@ import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.types.Variance
 import org.jetbrains.kotlin.util.OperatorNameConventions
-import kotlinx.cinterop.CPointer
-import org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeImpl
 
 internal interface KotlinStubs {
     val irBuiltIns: IrBuiltIns
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.backend.konan.PrimitiveBinaryType
 import org.jetbrains.kotlin.backend.konan.RuntimeNames
 import org.jetbrains.kotlin.backend.konan.ir.*
 import org.jetbrains.kotlin.backend.konan.isObjCMetaClass
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.builtins.UnsignedType
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
@@ -41,6 +42,7 @@ import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.konan.target.Family
 import org.jetbrains.kotlin.konan.target.KonanTarget
 import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.FqNameUnsafe
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.types.Variance
 import org.jetbrains.kotlin.util.OperatorNameConventions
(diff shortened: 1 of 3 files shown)


backend.native/tests/interop/basics/ccallbacksAndVarargs.def
```diff
@@ -43,6 +43,7 @@ struct Args {
     unsigned long long a12;
     enum E a13;
     struct S a14;
+    void* a15; // expectin NULL
```
olonho: expect or expecting

olonho: and dot at the end.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -774,6 +779,12 @@ private fun KotlinStubs.mapType(
     type.isUInt() -> UnsignedValuePassing(type, CTypes.int, CTypes.unsignedInt)
     type.isULong() -> UnsignedValuePassing(type, CTypes.longLong, CTypes.unsignedLongLong)
 
+    type.isNullLiteral() -> if (variadic) {
```
olonho: Maybe better `type.isNullLiteral() && variadic`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -750,6 +752,9 @@ private fun KotlinStubs.mapBlockType(
 private fun KotlinStubs.mapType(type: IrType, retained: Boolean, variadic: Boolean, location: TypeLocation): ValuePassing =
         mapType(type, retained, variadic, location, { reportUnsupportedType(it, type, location) })
 
+private fun IrType.isNullLiteral(): Boolean = this is IrSimpleType && hasQuestionMark
```
SvyatoslavScherbina: This function is misnamed: type can't be null literal, because null literal is an expression (`null`).

knebekaizer: Provide the suggestion please

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -774,17 +774,12 @@ private fun KotlinStubs.mapType(
     type.isFloat() -> TrivialValuePassing(irBuiltIns.floatType, CTypes.float)
     type.isDouble() -> TrivialValuePassing(irBuiltIns.doubleType, CTypes.double)
     type.classifierOrNull == symbols.interopCPointer -> TrivialValuePassing(type, CTypes.voidPtr)
+    type.isNullLiteral() && variadic  -> TrivialValuePassing(type, CTypes.voidPtr)
```
SvyatoslavScherbina: This is not entirely correct, since it likely leads to emitting Kotlin->C bridge with `Nothing?`-typed parameter.
And these bridges are supposed to be extremely simple and support only primitive and pointer types.
So `TrivialValuePassing` should probably be constructed with `CPointer<*>?` type here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
SvyatoslavScherbina: `isNullableNothing` or `isTypeOfNullLiteral`, depending on how exactly you'd like to express the meaning on the call site.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
SvyatoslavScherbina: This code also assumes that `Nothing?` binary representation is the same as of `void*`.
It is currently true, but not desirable to rely on here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -42,10 +42,11 @@ import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.konan.target.Family
 import org.jetbrains.kotlin.konan.target.KonanTarget
 import org.jetbrains.kotlin.name.FqName
-import org.jetbrains.kotlin.name.FqNameUnsafe
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.types.Variance
 import org.jetbrains.kotlin.util.OperatorNameConventions
+import kotlinx.cinterop.CPointer
```
SvyatoslavScherbina: Unused import.

PR #3313, opened 2019-09-04
[kotlin compiler][update] 1.3.60-dev-1867

vvlevchenko:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -268,8 +271,15 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
                         "-DKONAN_NO_MEMMEM",
                         "-DKONAN_NO_CTORS_SECTION=1")
 
+<<<<<<< HEAD
+=======
+            KonanTarget.ANDROID_X64 -> TODO("implement me")
+            KonanTarget.ANDROID_X86 -> TODO("implement me")
+>>>>>>> [REVERTME] filtering out ANDROID_{X86,ARM32}, WATCHOS_{X86,ARM32} from enabled platform till their full support
```
alexander-gorshenev: Oops^^^

PR #3316, opened 2019-09-05
API to allow event processing on any thread.

olonho:


runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -34,14 +34,32 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+        public val currentOrInit: Worker get() {
+            val id = initInternal(true)
+            return if (id != 0) Worker(id) else throw Error("Cannot init as worker")
+        }
+
+        /**
+         * Deinit worker event queue on the current thread. If called in worker context will lead to termination
+         * of the worker.
+         */
+        public fun currentDeinit() {
```
SvyatoslavScherbina: Must be called automatically on thread death.

olonho: Seriously reworked whole worker support and simplified API.

backend.native/tests/runtime/workers/worker5.kt
```diff
@@ -19,4 +17,36 @@ import kotlin.native.concurrent.*
...
+    worker.processQueue()
+    assertEquals(future.state, FutureState.COMPUTED)
+    future.consume {}
+    // After termination request this worker is no longer addressable.
+    assertFailsWith<IllegalStateException> { worker.executeAfter(0, {
+        println("BUG!")
+    }.freeze()) }
+}
+
+fun main() {
```
SvyatoslavScherbina: Maybe add `flags = ['-tr']` to `build.gradle` instead?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -39,14 +39,11 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
-         * [Worker.start].
+         * Return the current worker. Worker context is accessible to any valid Kotlin context,
+         * but only actual active worker produced with [Worker.start] automatically processes execution requests.
+         * For other situations [processQueue] must be called explicitly to process request queue.
          */
-        public val current: Worker? get() {
-            val id = currentInternal()
-            return if (id != 0) Worker(id) else null
-        }
+        public val current: Worker get() = Worker(currentInternal())
```
SvyatoslavScherbina: Btw, `currentInternal()` can still return `0`, for example in deallocation hooks during final GC.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -726,12 +610,218 @@ KNativePtr detachObjectGraphInternal(KInt transferMode, KRef producer) {
...
+
+#endif  // WITH_WORKERS
+
 extern "C" {
 
 KInt Kotlin_Worker_startInternal(KBoolean noErrorReporting) {
   return startWorker(noErrorReporting);
 }
 
+KInt Kotlin_Worker_initInternal(KBoolean noErrorReporting) {
```
SvyatoslavScherbina: Is this used?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -669,6 +544,15 @@ KInt startWorker(KBoolean errorReporting) {
   return -1;
 }
 
+KInt initAsWorker(KBoolean errorReporting) {
```
SvyatoslavScherbina: Are `initAsWorker` and `deinitWorker` used?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
SvyatoslavScherbina: Also when `WITH_WORKERS` is false.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -726,12 +610,218 @@ KNativePtr detachObjectGraphInternal(KInt transferMode, KRef producer) {
...
+  return worker;
+#else
+  return nullptr;
+#endif  // WITH_WORKERS
+}
+
+void WorkerDeinit(Worker* worker) {
+#if WITH_WORKERS
+  ::g_worker = nullptr;
+  theState()->destroyWorkerUnlocked(worker);
```
SvyatoslavScherbina: Two different worker variables involved. Maybe add an assertion to check that the values are the same?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: No for second comment, 
```
KInt currentWorker() {
  ThrowWorkerUnsupported();
  return 0;
}``` will throw an exception in  first case as well.

backend.native/tests/runtime/workers/worker5.kt
olonho: Maybe.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
SvyatoslavScherbina: > No for second comment,

Ok, I see.

> will throw an exception in first case as well.

Could you clarify?

runtime/src/main/cpp/Worker.cpp
olonho: They don't have to, `g_worker` could be zeroed out by TLS deinit already.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: I mean, I fill fix the code, what I just did in 6990ad0c9

backend.native/tests/runtime/workers/worker5.kt
olonho: Turns out, that this will force to change test output to that of test logger, which doesn't look like best idea.

PR #3318, opened 2019-09-06
[KT-32929] Handle all exceptions from objc alloc

knebekaizer:
Catch all and terminate with non-zero code instead of EXC_BAD_ACCESS fault

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -3335,6 +3335,12 @@ Task interopTest(String name, Closure<KonanInteropTest> configureClosure) {
     }
 }
 
+
+standaloneTest("interop_objc_allocException") {
+    expectedExitStatus = 0
+    source = "interop/objc/allocException.kt"
+}
+
 interopTest("interop0") {
     disabled = (project.testTarget == 'wasm32') ||         // No interop for wasm yet.
                (project.testTarget == 'linux_mips32') ||   //  st_uid of '/' is not equal to 0 when using qemu
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -3337,6 +3337,7 @@ Task interopTest(String name, Closure<KonanInteropTest> configureClosure) {
 
 
 standaloneTest("interop_objc_allocException") {
+    disabled = !isAppleTarget(project)
     expectedExitStatus = 0
     source = "interop/objc/allocException.kt"
 }
--- a/backend.native/tests/interop/objc/allocException.kt
+++ b/backend.native/tests/interop/objc/allocException.kt
@@ -20,7 +20,5 @@ fun exc_handler(x: Any?) : Unit {
 fun main() {
     objc_setUncaughtExceptionHandler(staticCFunction(::exc_handler))
 
-    // The following should fail with exception that shall be processed by FilterException and stop at exc_handler
-    // Otherwise, ReportUnhandledException will fail with Segmentation fault (exit code 139) on attempt to printStackTrace
     println(NSJSONSerialization())
 }
--- a/backend.native/tests/interop/objc/allocException.kt
+++ b/backend.native/tests/interop/objc/allocException.kt
@@ -4,23 +4,23 @@
  */
 
 import platform.Foundation.*
+import platform.objc.*
 import kotlin.test.*
 
 import kotlinx.cinterop.*
 import platform.posix.*
 import kotlin.system.exitProcess
 
-fun sigabrt_handler(signum: Int) : Unit {
-    println("Exception handled successfully, signum = $signum")
+fun exc_handler(x: Any?) : Unit {
+    println("Uncaught exception handler")
+    println(x.toString())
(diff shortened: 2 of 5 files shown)


backend.native/tests/build.gradle
```diff
@@ -3335,6 +3335,12 @@ Task interopTest(String name, Closure<KonanInteropTest> configureClosure) {
     }
 }
 
+
+standaloneTest("interop_objc_allocException") {
```
SvyatoslavScherbina: Should be marked as depending on platform libs. See e.g. https://github.com/JetBrains/kotlin-native/blob/ec6b4d6d92d512c3b70786d75fc6ba97d3667cb7/backend.native/tests/build.gradle#L3390

backend.native/tests/build.gradle
```diff
@@ -3335,6 +3335,12 @@ Task interopTest(String name, Closure<KonanInteropTest> configureClosure) {
...
+    expectedExitStatus = 0
```
SvyatoslavScherbina: IIRC this is checked by default.

backend.native/tests/interop/objc/allocException.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.cinterop.*
+import platform.posix.*
+import kotlin.system.exitProcess
+
+fun sigabrt_handler(signum: Int) : Unit {
+    println("Exception handled successfully, signum = $signum")
+    exitProcess(0)
+}
+
+class A : NSJSONSerialization()
```
SvyatoslavScherbina: Subclassing is likely not required here, instantiating `NSJSONSerialization()` should be enough to reproduce the original issue.

backend.native/tests/interop/objc/allocException.kt
```diff
@@ -0,0 +1,26 @@
...
+
+fun main() {
+    signal(SIGABRT, staticCFunction(::sigabrt_handler))
```
SvyatoslavScherbina: Consider using `objc_setUncaughtExceptionHandler` or [`NSSetUncaughtExceptionHandler`](https://developer.apple.com/documentation/foundation/1409609-nssetuncaughtexceptionhandler?language=objc) instead.

backend.native/tests/interop/objc/allocException.kt
```diff
@@ -0,0 +1,26 @@
...
+    println("Uncaught exception handler")
+    println(x.toString())
+    exitProcess(0)
+}
+
+fun main() {
+    objc_setUncaughtExceptionHandler(staticCFunction(::exc_handler))
+
+    // The following should fail with exception that shall be processed by FilterException and stop at exc_handler
+    // Otherwise, ReportUnhandledException will fail with Segmentation fault (exit code 139) on attempt to printStackTrace
```
SvyatoslavScherbina: These implementation details are somewhat unrelated to the test itself.

knebekaizer: 1. Arguable. This comment provides additional context for a test maintainer: the test relies on the fact that FilterException annotation will finish with std::terminate, and std::set_terminate hook is the default ObjC hook. This is not necessary be true: someone may decide to rework KonanTerminateHandler so it does not call oldTerminateHandler (see TerminateWithUnhandledException for example). It may exit with abort() (as TerminateWithUnhandledException does), exit(128) or any other way. Exception processing is still OK but hook has not been invoked and test failed. In that case a  __test__  itself shall be reworked, not the main logic. So the comment provides this kind of clue.
2. Can I ask you to provide all the comments at once, or leave a notice when you are not finished yet, so i won't bother to iterate (i.e. commit) until all feedbacks are collected and processed? The lines we're talking about were in the previous commit too.

SvyatoslavScherbina: > This comment provides additional context for a test maintainer

Please note the huge difference between the behavior description, e.g.
> // Exception thrown by Objective-C `alloc` terminates the program and detected as uncaught with the hook above

and the description of the implementation:
> // The following should fail with exception that shall be processed by FilterException and stop at exc_handler
> // Otherwise, ReportUnhandledException will fail with Segmentation fault (exit code 139) on attempt to printStackTrace

I find the first one better serving the purpose you use as argument.


> The lines we're talking about were in the previous commit too.

This is definitely wrong: these lines have been significantly rewritten in the last commit by this moment: https://github.com/JetBrains/kotlin-native/pull/3318/commits/656a09026c562107010478e88e57244202961c99#diff-022d0c29e3a63ba37d377c862c10c450R23

> Can I ask you to provide all the comments at once, or leave a notice when you are not finished yet

I have to refuse, because
1. This is incompatible with proper review process: what if I notice something later, after I've reviewed the code for the first time? Do you suggest me to discard this comment?
2. There is no much sense in providing minor review comments (like this one) along with major ones (like using `objc_setUncaughtExceptionHandler`), since fixing the latter ones may (and did in this case) also affect the former ones.

PR #3325, opened 2019-09-10
Group static initializers by modules in codegen

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2269,11 +2270,44 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
 
     //-------------------------------------------------------------------------//
     fun appendStaticInitializers() {
-        val ctorName = context.config.moduleId.moduleConstructorName
-        val ctorFunction = LLVMAddFunction(context.llvmModule, ctorName, kVoidFuncType)!!
-        LLVMSetLinkage(ctorFunction, LLVMLinkage.LLVMExternalLinkage)
+        // Null for "current" non-library module:
```
olonho: Not sure what you mean here.

SvyatoslavScherbina: `null` in the list below is used to represent IR that is not loaded from libraries but generated with psi2ir during this compile session.
Currently in our compiler this module depends on `context.librariesWithDependencies`.

olonho: Hmm, could it be somehow documented?

sbogolepov: Maybe add a test for library initialization order?

SvyatoslavScherbina: Maybe later.

PR #3334, opened 2019-09-11
Improve compilation to multiple LLVM modules/object files

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CompilerOutput.kt
```diff
@@ -30,8 +46,11 @@ internal fun produceCStubs(context: Context) {
 }
 
 private fun linkAllDependecies(context: Context, generatedBitcodeFiles: List<String>) {
```
olonho: linkAllDepende*n*cies

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -151,16 +151,17 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
         return resolvedLibraries.filterRoots { (!it.isDefault && !this.purgeUserLibs) || it.isNeededForLink }.getFullList(TopologicalLibraryOrder)
     }
 
-    internal val defaultNativeLibraries: List<String> = mutableListOf<String>().apply {
+    internal val runtimeNativeLibraries: List<String> = mutableListOf<String>().apply {
         add(if (debug) "debug.bc" else "release.bc")
         add(if (memoryModel == MemoryModel.STRICT) "strict.bc" else "relaxed.bc")
-        if (produce == CompilerOutputKind.PROGRAM) {
```
olonho: +1

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
```diff
@@ -399,6 +399,8 @@ internal class Llvm(val context: Context, val llvmModule: LLVMModuleRef) {
         context.config.resolvedLibraries
                 .getFullList(TopologicalLibraryOrder)
                 .filter { (!it.isDefault && !context.config.purgeUserLibs) || imports.bitcodeIsUsed(it) }
+                // TODO: the filter above is incorrect when compiling to multiple LLVM modules.
```
olonho: So what you suggest to do with problem mentioned in TODO?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -380,6 +380,16 @@ internal class ObjCExportCodeGenerator(
 
 }
 
+private fun ObjCExportCodeGenerator.replaceExternalWeakOrCommonGlobal(name: String, value: ConstValue) {
+    val global = staticData.placeGlobal(name, value, isExported = true)
+
+    if (context.llvmModuleSpecification.importsKotlinDeclarationsFromOtherObjectFiles()) {
+        // Note: actually this is required only if global's weak/common definition is in other object file,
+        // but it is simpler to do this for all globals, considering that all usages can't be removed by DCE anyway.
+        context.llvm.usedGlobals += global.llvmGlobal
```
olonho: Looks s bit fishy, aren't we indeed killing DCE wrt globals this way?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -69,7 +69,7 @@ internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
     internal fun generate(codegen: CodeGenerator) {
         if (!target.family.isAppleFamily) return
 
-        if (!context.config.produce.isNativeBinary) return // TODO: emit RTTI to the same modules as classes belong to.
+        if (!context.config.produce.isFinalBinary) return // TODO: emit RTTI to the same modules as classes belong to.
```
olonho: Hmm, why not just do that TODO?

SvyatoslavScherbina: Because actually it's a huge work.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: See the note above.
Globals handled by this function are actually used by the generated code, they can't be removed by DCE without this patch too.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
SvyatoslavScherbina: It mostly depends on how exactly compiling to multiple LLVM modules will be used.
The PR doesn't define this, so it doesn't address the problem.

For example, for build cache we may decide to include the bitcode here unconditionally, i.e. remove the filter above (since it is much like DCE, and compiling to cache is generally supposed to disable DCE).

More general solution may eventually involve getting rid of the prepared bitcode in interop libraries or privately linking it into each LLVM module that uses this bitcode.

SvyatoslavScherbina: Also we could consider finally dropping the concept of default libraries that are supposed to be DCEd (as it brings a lot of issues and complicates the implementation of many components), requiring configuring them explicitly (e.g. with `-l Foundation` in CLI or dependencies in Gradle).

PR #3323, opened 2019-09-09
[watchOS][targets] Add support for 32bit Apple Watch.

sbogolepov:

PR #3346, opened 2019-09-16
Add tests for objc_msgSend* family.

sbogolepov:


backend.native/tests/interop/objc/msg_send/messaging.h
```diff
@@ -0,0 +1,47 @@
...
+typedef struct {
+    float f1;
+    float f2;
+    float f3;
+    float f4;
+    float f5;
+    float f6;
+    float f7;
+    float f8;
+} HomogeneousBig;
```
olonho: SIMD types?

sbogolepov: We don't have support for them yet. What do you meat?

olonho: I just meant, that we need to check for types like int8x16_t in those tests eventually. Another possible kind is bitfields, like `unsigned foo:2`.

backend.native/tests/interop/objc/msg_send/messaging.h
```diff
@@ -0,0 +1,47 @@
...
+typedef struct {
+    float f;
+} SingleFloat;
+
+typedef struct __attribute__((packed)) {
+    char f1;
+    short f2;
+    char f3;
+    char f4;
+} SimplePacked;
```
SvyatoslavScherbina: Maybe add this struct too?
https://github.com/JetBrains/kotlin-native/pull/3323#discussion_r322597676

PR #3347, opened 2019-09-16
Moving performance monitoring system (server and client) to InfluxDB.

LepilkinaElena:


tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+// Need separate classes to describe types, because Int and Double are same in Kotlin/JS.
+sealed class FieldType<T : Any>(val value: T) {
+    override fun toString(): String = value.toString()
+    class InfluxInt(value: Int): FieldType<Int>(value)
+    class InfluxFloat(value: Double): FieldType<Double>(value)
+    class InfluxString(value: Any): FieldType<String>(value.toString())
+    class InfluxBoolean(value: Boolean): FieldType<Boolean>(value)
+}
+
+// Base class for measurements.
```
olonho: Better split in multiple files - classes here seems to be not very related.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+                error("Field ${tag.name} already exists in measurement $name")
+            }
+            tags[tag.name] = tag
+            return tag
+        }
+    }
+
+    val lineProtocol =
+        with(StringBuilder("$name,")) {
+            var prefix = ""
```
olonho: Isn't it `join`?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -139,191 +47,281 @@ data class BuildRegister(val buildId: String, val teamCityUser: String, val team
...
+                branch = request.query.branch
+            }
+        }
+        getBuildsForParameters(branch, target).then { results ->
+            val filteredBuildNumbers = orderedBuildNumbers((results as List<String>), "${request.params.type}")
+            val responseLists = filteredBuildNumbers.map {
+                // Select needed measurements.
+                measurement.select(measurement.all(), (measurement.tag("environment.machine.os") eq target) and
+                        (measurement.field("build.number") eq FieldType.InfluxString(it))).then { dbResponse ->
+                    val measurements = (dbResponse as List<BenchmarkMeasurement>)
```
olonho: Hmm, why such cast, cannot it be at least checked?

tools/performance-server/ui/src/main/kotlin/main.kt
```diff
@@ -153,62 +117,71 @@ fun redirect(url: String) {
...
+                        append("failures: ${currentBuild.failuresNumber}<br>")
+                    }
+                    append("branch: ${currentBuild.branch}<br>")
+                    append("date: ${currentBuild.date}<br>")
+                    append("time: ${currentBuild.formattedStartTime}-${currentBuild.formattedFinishTime}<br>")
+                    append("Commits:<br>")
+                    val commitsList = currentBuild.commits.split(";")
+                    commitsList.forEach {
+                        if (it.isNotEmpty())
+                            append("${it.substringBefore("by").substring(0, 7)} by ${it.substringAfter("by ")}<br>")
```
olonho: `<br/>`?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -139,191 +47,281 @@ data class BuildRegister(val buildId: String, val teamCityUser: String, val team
...
+            if (request.query.branch != undefined) {
+                branch = request.query.branch
+            }
+        }
+
+        // Get golden results to normalize data.
+        val golden = GoldenResultMeasurement()
+        val goldenResults = golden.select(golden.all()).then { results ->
+            val parsedNormalizeResults = mutableMapOf<String, MutableMap<String, Double>>()
+            (results as List<GoldenResultMeasurement>).forEach {
```
olonho: Same, maybe those casts can be made somewhat cleaner?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -139,191 +47,281 @@ data class BuildRegister(val buildId: String, val teamCityUser: String, val team
...
+                parsedNormalizeResults.getOrPut(it.benchmarkName!!,
```
olonho: Shall benchmarkName and benchmarkMetric be nullable then?

tools/performance-server/src/main/kotlin/routes/route.kt
```diff
@@ -1,24 +1,17 @@
...
 import kotlin.js.Date
+import kotlin.js.Promise
 import org.jetbrains.report.json.*
+import org.jetbrains.influxdb.*
 import org.jetbrains.build.Build
+import org.jetbrains.analyzer.*
+import org.jetbrains.report.*
 
 const val teamCityUrl = "https://buildserver.labs.intellij.net/app/rest"
 const val downloadBintrayUrl = "https://dl.bintray.com/content/lepilkinaelena/KotlinNativePerformance"
```
olonho: Is it still needed?

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+    infix fun and(other: WhereExpression) = object : WhereExpression() {
+        override val lineProtocol: String =
+                "WHERE ${this@WhereExpression.lineProtocol.replace("WHERE", "")} AND ${other.lineProtocol.replace("WHERE", "")}"
+    }
+}
+
+// Base class for InfluxDB functions.
+abstract class InfluxFunction<T : Any>(val entity: ColumnEntity<*>): Expression<T>()
+
+// DISTINC InfluxDB function.
```
olonho: DISTINCT

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+    fun fieldsFromInfluxJson(data: JsonElement): List<String> {
+        val points = mutableListOf<String>()
+        var columnsIndexes: Map<String, Int>
+        if (data is JsonObject) {
+            val results = data.getRequiredField("results") as JsonArray
+            results.map {
+                if (it is JsonObject) {
+                    val series = it.getRequiredField("series") as JsonArray
+                    series.map {
+                        if (it is JsonObject) {
```
olonho: What if not? Assert on else path?

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+        checkConnection()
+
+        val queryUrl = "$host:$port/query?db=$databaseName&q=$query"
+        return sendRequest(RequestMethod.GET, queryUrl, user, password, true)
+    }
+
+    // Check that connection with database can be set.
+    private inline fun checkConnection() {
+        if (!::host.isInitialized || !::databaseName.isInitialized) {
+            error("Please, firstly set connection to Influx database to have opportunity to send requests.")
```
olonho: firstly set connection -> establish connection

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
```diff
@@ -0,0 +1,315 @@
...
+
+package org.jetbrains.influxdb
+
+import kotlin.reflect.KProperty
+import kotlin.js.Promise            // TODO - migrate to multiplatform.
+import kotlin.js.json               // TODO - migrate to multiplatform.
+import org.jetbrains.report.json.*
+
+// Connector with InfluxDB.
+object InfluxDBConnector {
```
olonho: Why singleton? Cannot we have multiple connections actually?
Even if so, I'd do logic in class and have separate instance.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/BenchmarkMeasurement.kt
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// List of commits.
+class CommitsList(data: JsonElement): ConvertedFromJson {
+
+    val commits: List<Commit>
+
+    init {
+        if (data !is JsonObject) {
+            error("Commits description is expected to be a json object!")
```
olonho: JSON

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/BenchmarkMeasurement.kt
```diff
@@ -0,0 +1,364 @@
...
+            BenchmarkResult(it.benchmarkName!! + metric.suffix, BenchmarkResult.statusFromString(it.benchmarkStatus!!.value)!!,
+                    it.benchmarkScore!!.value, metric,
+                    it.benchmarkRuntime!!.value, it.benchmarkRepeat!!.value, it.benchmarkWarmup!!.value)
+        }
+        BenchmarksReport(environment, benchmarksList, compiler)
+    }
+}
+
+// Database instance describing benchmark.
+class BenchmarkMeasurement : Measurement("benchmarks") {
```
olonho: Why not make it data class and use predefined https://kotlinlang.org/docs/reference/data-classes.html?

tools/performance-server/ui/src/main/kotlin/main.kt
LepilkinaElena: http://htmlbook.ru/html/br

tools/performance-server/src/main/kotlin/routes/route.kt
LepilkinaElena: Yes, they should, because Tag should provide delegate a nullable value.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/BenchmarkMeasurement.kt
LepilkinaElena: I don't understand what do you want to say? The idea was to have an instance which allows execute all queries. It isn't a data class.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDB.kt
olonho: Any stateful singleton is blocker to composition, so I usually avoid those when could.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/BenchmarkMeasurement.kt
olonho: Then you could make state a data class and have pre-generated `copy` method, manual copying is call for trouble.

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/InfluxDBExpression.kt
```diff
@@ -0,0 +1,28 @@
...
+abstract class Expression<T: Any>() {
+    abstract val lineProtocol: String
+}
+
+// Base class for InfluxDB expression describing condition of selected points.
+abstract class WhereExpression: Expression<String>() {
+    // AND InfluxDb operator to combine sevral conditions.
+    infix fun and(other: WhereExpression) = object : WhereExpression() {
+        override val lineProtocol: String =
+                "WHERE ${this@WhereExpression.lineProtocol.replace("WHERE", "")} AND ${other.lineProtocol.replace("WHERE", "")}"
```
olonho: Looks a bit hacky, maybe conditions are conditions and 'WHERE' shall be created by an explicit `where` statement?

tools/performance-server/shared/src/main/kotlin/org/jetbrains/influxdb/BenchmarkMeasurement.kt
LepilkinaElena: I understand, but data class generates copy method that just call constructor. It isn't enough here, I need all other field be copied.
And I can't replace all fields in constructor, because they are created via delegated properties.

SvyatoslavScherbina:
@LepilkinaElena what is the status of this PR?

LepilkinaElena:
This decision with InfluxDb wasn't suitable for our data structure, so I am rewriting this PR

PR #3357, opened 2019-09-18
Fix CoreSymbolication usage on watchos_x86

SvyatoslavScherbina:


SvyatoslavScherbina:
> and maybe even on 32-bit iOS ARM.

CoreSymbolication is not used (and likely not available) on device targets.

PR #3359, opened 2019-09-18
Add initial support for producing and consuming cache

SvyatoslavScherbina:
Dynamic cache isn't fully functional on Apple targets yet.

### DIFF
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
@@ -90,7 +90,8 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
     }
 
     val K2NativeCompilerArguments.isUsefulWithoutFreeArgs: Boolean
-        get() = listTargets || listPhases || checkDependencies || !includes.isNullOrEmpty()
+        get() = listTargets || listPhases || checkDependencies || !includes.isNullOrEmpty() ||
+                !librariesToCache.isNullOrEmpty()
 
     fun Array<String>?.toNonNullList(): List<String> {
         return this?.asList<String>() ?: listOf<String>()
@@ -211,6 +212,10 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
                 put(LIBRARIES_TO_COVER, arguments.coveredLibraries.toNonNullList())
                 arguments.coverageFile?.let { put(PROFRAW_PATH, it) }
                 put(OBJC_GENERICS, arguments.objcGenerics)
+
+                put(LIBRARIES_TO_CACHE, parseLibrariesToCache(arguments, configuration, outputKind))
+                put(CACHE_DIRECTORIES, arguments.cacheDirectories.toNonNullList())
+                put(CACHED_LIBRARIES, parseCachedLibraries(arguments, configuration))
             }
         }
     }
@@ -298,6 +303,37 @@ private fun selectIncludes(
         emptyList()
     } else {
         includes
+    }
+}
+
+private fun parseCachedLibraries(
+        arguments: K2NativeCompilerArguments,
+        configuration: CompilerConfiguration
+): Map<String, String> = arguments.cachedLibraries?.asList().orEmpty().mapNotNull {
+    val libraryAndCache = it.split(",")
+    if (libraryAndCache.size != 2) {
+        configuration.report(
+                ERROR,
+                "incorrect $CACHED_LIBRARY format: expected '<library>,<cache>', got '$it'"
+        )
+        null
+    } else {
+        libraryAndCache[0] to libraryAndCache[1]
+    }
+}.toMap()
+
+private fun parseLibrariesToCache(
+        arguments: K2NativeCompilerArguments,
+        configuration: CompilerConfiguration,
+        outputKind: CompilerOutputKind
+): List<String> {
+    val input = arguments.librariesToCache?.asList().orEmpty()
+
+    return if (input.isNotEmpty() && !outputKind.isCache) {
+        configuration.report(ERROR, "$MAKE_CACHE can't be used when not producing cache")
+        emptyList()
+    } else {
+        input
     }
 }
 
--- a/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
+++ b/backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
@@ -102,6 +102,22 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     // Make sure to prepend them with -X.
     // Keep the list lexically sorted.
 
+    @Argument(
+            value = "-Xcache-directory",
+            valueDescription = "<path>",
+            description = "Path to the directory containing caches",
+            delimiter = ""
+    )
+    var cacheDirectories: Array<String>? = null
+
+    @Argument(
+            value = CACHED_LIBRARY,
+            valueDescription = "<library path>,<cache path>",
+            description = "Comma-separated paths of a library and its cache",
+            delimiter = ""
+    )
+    var cachedLibraries: Array<String>? = null
+
     @Argument(value="-Xcheck-dependencies", deprecatedName = "--check_dependencies", description = "Check dependencies and download the missing ones")
     var checkDependencies: Boolean = false
 
@@ -138,6 +154,14 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
 
     @Argument(value = "-Xg0", description = "Add light debug information")
     var lightDebug: Boolean = false
+
+    @Argument(
+            value = MAKE_CACHE,
+            valueDescription = "<path>",
+            description = "Path of the library to be compiled to cache",
+            delimiter = ""
+    )
+    var librariesToCache: Array<String>? = null
 
     @Argument(value = "-Xprint-bitcode", deprecatedName = "--print_bitcode", description = "Print llvm bitcode")
     var printBitCode: Boolean = false
@@ -220,3 +244,5 @@ const val EMBED_BITCODE_FLAG = "-Xembed-bitcode"
 const val EMBED_BITCODE_MARKER_FLAG = "-Xembed-bitcode-marker"
 const val STATIC_FRAMEWORK_FLAG = "-Xstatic-framework"
 const val INCLUDE_ARG = "-Xinclude"
+const val CACHED_LIBRARY = "-Xcached-library"
+const val MAKE_CACHE = "-Xmake-cache"
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BuiltInFictitiousFunctionIrClassFactory.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BuiltInFictitiousFunctionIrClassFactory.kt
@@ -48,6 +48,16 @@ internal class BuiltInFictitiousFunctionIrClassFactory(
 
     class FunctionalInterface(val irClass: IrClass, val arity: Int)
 
+    fun buildAllClasses() {
+        val maxArity = 255 // See [BuiltInFictitiousFunctionClassFactory].
+        (0 .. maxArity).forEach { arity ->
+            function(arity)
+            kFunction(arity)
+            suspendFunction(arity)
+            kSuspendFunction(arity)
+        }
+    }
+
     fun function(n: Int) = buildClass(irBuiltIns.builtIns.getFunction(n) as FunctionClassDescriptor)
     fun kFunction(n: Int) = buildClass(reflectionTypes.getKFunction(n) as FunctionClassDescriptor)
     fun suspendFunction(n: Int) = buildClass(irBuiltIns.builtIns.getSuspendFunction(n) as FunctionClassDescriptor)
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.kotlin.backend.konan
+
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.konan.file.File
+import org.jetbrains.kotlin.konan.target.CompilerOutputKind
+import org.jetbrains.kotlin.konan.target.KonanTarget
+import org.jetbrains.kotlin.library.KotlinLibrary
+import org.jetbrains.kotlin.library.resolver.KotlinLibraryResolveResult
+
+class CacheSupport(
+        configuration: CompilerConfiguration,
+        resolvedLibraries: KotlinLibraryResolveResult,
+        target: KonanTarget,
+        produce: CompilerOutputKind
+) {
+    private val allLibraries = resolvedLibraries.getFullList()
+
+    // TODO: consider using [FeaturedLibraries.kt].
+    private val fileToLibrary = allLibraries.associateBy { it.libraryFile }
+
+    internal val cachedLibraries: CachedLibraries = run {
+        val explicitCacheFiles = configuration.get(KonanConfigKeys.CACHED_LIBRARIES)!!
+
+        val explicitCaches = explicitCacheFiles.entries.associate { (libraryPath, cachePath) ->
+            val library = fileToLibrary[File(libraryPath)]
+                    ?: configuration.reportCompilationError("cache not applied: library $libraryPath in $cachePath")
+
+            library to cachePath
+        }
+
+        val implicitCacheDirectories = configuration.get(KonanConfigKeys.CACHE_DIRECTORIES)!!
+                .map {
+                    File(it).takeIf { it.isDirectory }
+                            ?: configuration.reportCompilationError("cache directory $it is not found or not a directory")
+                }
+
+        CachedLibraries(
+                target = target,
+                allLibraries = allLibraries,
+                explicitCaches = explicitCaches,
+                implicitCacheDirectories = implicitCacheDirectories
+        )
+    }
+
+    internal val librariesToCache: Set<KotlinLibrary> = configuration.get(KonanConfigKeys.LIBRARIES_TO_CACHE)!!
+            .map { File(it) }.map {
+                fileToLibrary[it] ?: error("library to cache\n" +
+                        "  ${it.absolutePath}\n" +
+                        "not found among resolved libraries:\n  " +
+                        allLibraries.joinToString("\n  ") { it.libraryFile.absolutePath })
+            }.toSet()
+            .also { if (!produce.isCache) check(it.isEmpty()) }
+
+    init {
+        // Ensure dependencies of every cached library are cached too:
+        resolvedLibraries.getFullList { libraries ->
+            libraries.map { library ->
+                val cache = cachedLibraries.getLibraryCache(library.library)
+                if (cache != null || library.library in librariesToCache) {
+                    library.resolvedDependencies.forEach {
+                        if (!cachedLibraries.isLibraryCached(it.library) && it.library !in librariesToCache) {
+                            val description = if (cache != null) {
+                                "cached (in ${cache.path})"
+                            } else {
+                                "going to be cached"
+                            }
+                            configuration.reportCompilationError(
+                                    "${library.library.libraryName} is $description, " +
+                                            "but its dependency isn't: ${it.library.libraryName}"
+                            )
+                        }
+                    }
+                }
+
+                library
+            }
+        }
+
+        // Ensure not making cache for libraries that are already cached:
+        librariesToCache.forEach {
+            val cache = cachedLibraries.getLibraryCache(it)
+            if (cache != null) {
+                configuration.reportCompilationError("Can't cache library '${it.libraryName}' " +
+                        "that is already cached in '${cache.path}'")
+            }
+        }
+    }
+}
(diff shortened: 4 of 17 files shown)


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -89,8 +89,10 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
         return ExitCode.OK
     }
 
-    val K2NativeCompilerArguments.isUsefulWithoutFreeArgs: Boolean
-        get() = listTargets || listPhases || checkDependencies || !includes.isNullOrEmpty()
+    fun isUsefulWithoutFreeArgs(arguments: K2NativeCompilerArguments, configuration: CompilerConfiguration): Boolean =
+            arguments.listTargets || arguments.listPhases || arguments.checkDependencies ||
+                    !arguments.includes.isNullOrEmpty() ||
+                    configuration.get(KonanConfigKeys.PRODUCE)!!.isCache
```
olonho: As per our discussion, I still believe cache shalln't be usable for compilations . without free args or `-Xinclude`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -145,6 +145,24 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
...
     }
 
+    val usedDynamicCaches: List<String>
+        get() = configuration.get(KonanConfigKeys.USED_DYNAMIC_CACHES)!!
+
+    val usedStaticCaches: List<String>
+        get() = configuration.get(KonanConfigKeys.USED_STATIC_CACHES)!!
+
+    // TODO: is "featured libraries" concept applicable here?
+    //   cache may contain more libraries than required,
```
olonho: Find mapping "one klib -> one cache element` more intuitive, and thus - "cache may contain more libraries than required" shalln't happen, as we pass exactly what we willing to compile.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -145,6 +145,24 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
...
+    //   and it may be convenient to pass all these libraries as -Xcached-library.
+    val cachedLibraries by lazy {
+        val cachedLibrariesFiles = configuration.get(KonanConfigKeys.CACHED_LIBRARIES)!!.toSet()
+        val usesCache = usedDynamicCaches.isNotEmpty() || usedStaticCaches.isNotEmpty()
+        resolvedLibraries.getFullList().filter {
+            usesCache && it.isDefault // For simplicity; default libraries are anyway included to cache by default.
```
olonho: Not sure, if this is good idea. They could be big, and we shall only include them if we were linking them before.

olonho: Also note, that some platform libraries may have non-trivial initializers, so this approach will bring a lot of uneeded init unconditionally.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LlvmModuleSpecificationImpl.kt
```diff
@@ -0,0 +1,31 @@
...
+package org.jetbrains.kotlin.backend.konan
+
+import org.jetbrains.kotlin.backend.konan.descriptors.konanLibrary
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.ir.declarations.IrDeclaration
+import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
+import org.jetbrains.kotlin.ir.util.module
+import org.jetbrains.kotlin.konan.library.KonanLibrary
+
+internal class LlvmModuleSpecificationImpl(private val excludedLibraries: Set<KonanLibrary>) : LlvmModuleSpecification {
```
olonho: This `excludedLibraries` blacklisting mechanism doesn't look reliable, better to use explicit whitelisting approach.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BuiltInFictitiousFunctionIrClassFactory.kt
```diff
@@ -48,6 +48,16 @@ internal class BuiltInFictitiousFunctionIrClassFactory(
 
     class FunctionalInterface(val irClass: IrClass, val arity: Int)
 
+    fun buildAllClasses() {
+        val maxArity = 255 // See [BuiltInFictitiousFunctionClassFactory].
```
olonho: Where is BuiltInFictitiousFunctionClassFactory?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
SvyatoslavScherbina: > Find mapping "one klib -> one cache element` more intuitive

I find this too restrictive. I believe compiler cache support should be more flexible.
For example, in some cases using cache may require user to adjust some build configuration (e.g. embed dynamic cache to application bundle on Apple platforms). Using single cache binary in this case may be easier than using binary per klib.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
SvyatoslavScherbina: > Not sure, if this is good idea.

Well, there are at least two reasons behind this decision.
1. Default libraries are included implicitly. For example, Gradle is unaware of these ones. So any other cache implementation would require compiler to be more complicated in handling default libraries.
2. Consider the case of managing the cache manually. There are quite a lot of platform libraries on Apple platforms, and figuring out which are demanded (and thus should be cached too) requires applying some efforts.

> They could be big, and we shall only include them if we were linking them before.

After migrating interop libraries to metadata this shouldn't be such an issue.

> some platform libraries may have non-trivial initializers

Not sure if this is correct.
Platform libraries are interop libraries. Interop libraries don't have non-trivial Kotlin initializers (and must not have them).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BuiltInFictitiousFunctionIrClassFactory.kt
SvyatoslavScherbina: https://github.com/JetBrains/kotlin/blob/c972bf5882977c8b1ae2684602f3686dd2069f9b/core/descriptors/src/org/jetbrains/kotlin/builtins/functions/BuiltInFictitiousFunctionClassFactory.kt#L48

See also: https://github.com/JetBrains/kotlin/blob/e5a1040dbd3c39e51eb4713cc7485c9bbff947d0/compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/KotlinMangler.kt#L305

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LlvmModuleSpecificationImpl.kt
SvyatoslavScherbina: Why do you think so?
Any whitelisting approach would anyway be based on the "cached libraries" blacklisting mechanism, so it would be as reliable as current approach is.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -346,7 +358,19 @@ internal val entryPointPhase = SameTypeNamedPhaseWrapper(
...
                                 context: Context, input: IrModuleFragment): IrModuleFragment {
                 assert(context.config.produce == CompilerOutputKind.PROGRAM)
-                context.ir.symbols.entryPoint!!.owner.file.addChild(makeEntryPoint(context))
+
+                val originalFile = context.ir.symbols.entryPoint!!.owner.file
+                val originalModule = originalFile.packageFragmentDescriptor.containingDeclaration
+                val file = if (context.llvmModuleSpecification.containsModule(originalModule)) {
+                    originalFile
+                } else {
+                    context.irModule!!.addFile(originalFile.fileEntry, originalFile.fqName)
```
SvyatoslavScherbina: TODO: add comment about the cases being handled here.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -102,6 +102,22 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
...
+            valueDescription = "<path>",
+            description = "Path to the directory containing caches",
+            delimiter = ""
+    )
+    var cacheDirectories: Array<String>? = null
+
+    @Argument(
+            value = CACHED_LIBRARY,
+            valueDescription = "<library path>,<cache path>",
+            description = "Comma-separated paths to the library compiled to cache and to this cache",
```
olonho: Not very clear description with three `to`.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -139,6 +155,14 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-Xg0", description = "Add light debug information")
     var lightDebug: Boolean = false
 
+    @Argument(
+            value = MAKE_CACHE,
+            valueDescription = "<path>",
+            description = "Path to the library to be compiled to cache",
```
olonho: Path to -> Path of

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -97,4 +102,36 @@ internal class Linker(val context: Context) {
...
+    context.librariesWithDependencies.forEach { library ->
+        val currentBinaryContainsLibrary = context.llvmModuleSpecification.containsLibrary(library)
+        val cache = context.config.cachedLibraries.getLibraryCache(library)
+        val libraryIsCached = cache != null
+
+        // Consistency check. Generally guaranteed by implementation.
+        if (currentBinaryContainsLibrary && libraryIsCached) {
+            error("Library ${library.libraryName} is found in both cache and current binary")
+        } else if (!currentBinaryContainsLibrary && !libraryIsCached) {
+            error("Library ${library.libraryName} is not found neither in cache nor in current binary")
```
olonho: Not very good diagnostic if valid relative path is passed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private val allCaches: Map<KonanLibrary, Cache> = allLibraries.mapNotNull { library ->
+        val explicitPath = explicitCaches[library]
+
+        val cache = if (explicitPath != null) {
+            val kind = when {
+                explicitPath.endsWith(target.family.dynamicSuffix) -> Cache.Kind.DYNAMIC
+                explicitPath.endsWith(target.family.staticSuffix) -> Cache.Kind.STATIC
+                else -> error("unexpected cache: $explicitPath")
```
sbogolepov: Message can be more helpful. 
Something like "Expected cache file as static ($target.family.staticSuffix) or dynamic (target.family.dynamicSuffix) library. Got: $explicitPath".

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+            Cache(kind, explicitPath)
+        } else {
+            implicitCacheDirectories.firstNotNullResult { path ->
+                val dir = File(path).takeIf { it.isDirectory } ?: return@firstNotNullResult null
+                val baseName = "${library.uniqueName}-cache"
```
sbogolepov: Maybe add target name to mangling?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
sbogolepov: Also you don't check for file existence in case of explicit path.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+        val cache = if (explicitPath != null) {
+            val kind = when {
+                explicitPath.endsWith(target.family.dynamicSuffix) -> Cache.Kind.DYNAMIC
+                explicitPath.endsWith(target.family.staticSuffix) -> Cache.Kind.STATIC
+                else -> error("unexpected cache: $explicitPath")
+            }
+            Cache(kind, explicitPath)
+        } else {
+            implicitCacheDirectories.firstNotNullResult { path ->
+                val dir = File(path).takeIf { it.isDirectory } ?: return@firstNotNullResult null
```
sbogolepov: If provided path is not a directory then user should be notified about it.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+                val baseName = "${library.uniqueName}-cache"
+                val dynamicFile = dir.child(getArtifactName(baseName, CompilerOutputKind.DYNAMIC_CACHE))
+                val staticFile = dir.child(getArtifactName(baseName, CompilerOutputKind.STATIC_CACHE))
+
+                if (dynamicFile.exists) {
```
sbogolepov: It's not so important, but `when` looks nicer :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
sbogolepov: Maybe add target name to mangling?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+    private fun getArtifactName(baseName: String, kind: CompilerOutputKind) =
+            "${kind.prefix(target)}$baseName${kind.suffix(target)}"
+
+    fun isLibraryCached(library: KonanLibrary): Boolean =
+            getLibraryCache(library) != null
+
+    fun getLibraryCache(library: KonanLibrary): Cache? =
+            allCaches[library]
+
+    val hasStaticCaches = allCaches.values.any {
```
sbogolepov: This value can be set earlier when you iterating over `allLibraries`. On the other hand, this will make code less clean.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
```diff
@@ -0,0 +1,71 @@
...
+package org.jetbrains.kotlin.backend.konan
+
+import org.jetbrains.kotlin.konan.file.File
+import org.jetbrains.kotlin.konan.library.KonanLibrary
+import org.jetbrains.kotlin.konan.target.CompilerOutputKind
+import org.jetbrains.kotlin.konan.target.KonanTarget
+import org.jetbrains.kotlin.library.uniqueName
+import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
+
+internal class CachedLibraries(
```
sbogolepov: Maybe `LibraryCache`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    internal val librariesToCache: Set<KonanLibrary> = run {
+        val files = configuration.get(KonanConfigKeys.LIBRARIES_TO_CACHE)!!.toSet()
+
+        // TODO: consider using [FeaturedLibraries.kt].
+        resolvedLibraries.getFullList()
+                .filter { it.libraryFile.absolutePath in files }
+                .toSet()
+                .also { if (!produce.isCache) check(it.isEmpty()) }
```
sbogolepov: Message can be more helpful than "Check failed."

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
SvyatoslavScherbina: Note that
1) this is _initial_ support
2) this feature is not entirely visible and recommended for end users directly.
I believe this kinda makes using current message ok.

> Also you don't check for file existence in case of explicit path.

See above. Also linker will check this anyway.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
SvyatoslavScherbina: Different targets may have different cache directories instead.
Also I'm not sure that target is the only compiler option to take into account here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CachedLibraries.kt
SvyatoslavScherbina: > This value can be set earlier when you iterating over allLibraries.

What for?

> On the other hand, this will make code less clean.

Exactly.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
```diff
@@ -0,0 +1,76 @@
...
+                }
+
+                library
+            }
+        }
+
+        // Ensure not making cache for libraries that are already cached:
+        librariesToCache.forEach {
+            if (cachedLibraries.isLibraryCached(it)) {
+                configuration.reportCompilationError("Can't cache library that is already cached: ${it.libraryName}")
```
sbogolepov: Show the path to cache file in the message?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
SvyatoslavScherbina: The user-friendly check for this is in `K2Native.kt`.
See also https://github.com/JetBrains/kotlin-native/pull/3359#discussion_r331409348

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OptimizationPipeline.kt
```diff
@@ -149,10 +151,12 @@ internal fun runLlvmOptimizationPipeline(context: Context) {
...
         LLVMAddAnalysisPasses(targetMachine, modulePasses)
-        // Since we are in a "closed world" internalization and global dce
-        // can be safely used to reduce size of a bitcode.
-        LLVMAddInternalizePass(modulePasses, 0)
-        LLVMAddGlobalDCEPass(modulePasses)
+        if (context.llvmModuleSpecification.isFinal) {
+            // Since we are in a "closed world" internalization and global dce
+            // can be safely used to reduce size of a bitcode.
+            LLVMAddInternalizePass(modulePasses, 0)
+            LLVMAddGlobalDCEPass(modulePasses)
```
sbogolepov: GlobalDCE without Internalization may be useful.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LlvmModuleSpecificationImpl.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.jetbrains.kotlin.ir.declarations.IrDeclaration
+import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
+import org.jetbrains.kotlin.ir.util.module
+import org.jetbrains.kotlin.konan.library.KonanLibrary
+
+// TODO: consider making two implementations: for producing cache and anything else.
+internal class LlvmModuleSpecificationImpl(
+        private val cachedLibraries: CachedLibraries,
+        private val producingCache: Boolean,
+        private val librariesToCache: Set<KonanLibrary>
```
sbogolepov: AFAIU, It's always either `producingCache && librariesToCache.isNotEmpty()` or `!producingCache && librariesToCache.isEmpty()`. Maybe remove `producingCache` field?

SvyatoslavScherbina: Correct, but 
1. I prefer this to be more explicit.
2. I consider making two implementation of `LlvmModuleSpecification`, and `producingCache` reflects this.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
SvyatoslavScherbina: What do you mean?

olonho: If path is not absolute, cryptic error was shown, IIRC.

PR #3361, opened 2019-09-18
Gradle plugin for swift benchmarks

LepilkinaElena:


build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,226 @@
...
+        val buildSwift = tasks.create("buildSwift") { task ->
+            task.dependsOn(framework.linkTaskName)
+            task.doLast {
+                val frameworkParentDirPath = framework.outputDirectory.absolutePath
+                val options = listOf("-Xlinker", "-rpath", "-Xlinker", frameworkParentDirPath, "-F", frameworkParentDirPath)
+                compileSwift(project, nativeTarget.konanTarget, benchmark.swiftSources, options,
+                        Paths.get(buildDir.absolutePath, benchmark.applicationName), false)
+            }
+        }
+        framework.linkTask.finalizedBy(buildSwift)
```
ilmat192: We have already added a dependency `buildSwift -> linkTask`. Why do we also need  this `finalizedBy` call?

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,226 @@
...
+
+private val Project.nativeBenchResults: String
+    get() = property("nativeBenchResults") as String
+
+private val Project.compilerArgs: List<String>
+    get() = (findProperty("compilerArgs") as String?)?.split("\\s").orEmpty()
+
+enum class CodeSizeEntity { FRAMEWORK, EXECUTABLE }
+
+open class SwiftBenchmarkExtension @Inject constructor(val project: Project) {
```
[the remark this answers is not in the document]
LepilkinaElena: Yes, I'll commonize in next step.

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,226 @@
...
+                    "codeSize" to getCodeSizeBenchmark(applicationName, codeSizeEntity)
+                )
+
+                val output = createJsonReport(properties)
+                buildDir.resolve(nativeJson).writeText(output)
+            }
+        }
+
+        // JVM run task.
+        val jvmRun = tasks.create("jvmRun", RunJvmTask::class.java) {
```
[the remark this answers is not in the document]
LepilkinaElena: Yes, we need it, because root project expect all subprojects have these run tasks.

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
LepilkinaElena: Because I want to have guarantee that at the end of link I'll have executable.

ilmat192: > at the end of link I'll have executable.
How/where do we start the link process? As far as I understand we never run the framework link task directly. Instead, we run `konanRun` which depends on `buildSwift`. This is true?

LepilkinaElena: We mostly yes, but calling link task isn't prohibited + I have an idea that may be we can have one more only compilation mode for all benchmarks later so I wanted to have same behaviour for all subprojects. Can this both-side dependency cause some problems?

ilmat192: > but calling link task isn't prohibited

Yes, but when a user calls the link task, he expects the framework to be linked and have no expectations about the swift executable. Or he even may not want to link the swift executable. 

> Can this both-side dependency cause some problems?

No, I just wanted to minimize unobvious relations between tasks.

> I wanted to have same behaviour for all subprojects.

I would prefer a separate task, say `buildBenchmark`, that depends on `linkExecutable...` or `buildSwift` tasks in different benchmarks. The `finalizedBy` will work too but it's less obvious.

Also the link tasks generated by the MPP plugin have different names: `linkExecutable...` for executables and `linkFramework` for frameworks, so we still won't be able to refer them uniformly.

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/KotlinNativeBenchmarkingPlugin.kt
```diff
@@ -0,0 +1,102 @@
...
+
+private val Project.jvmBenchResults: String
+    get() = property("jvmBenchResults") as String
+
+open class KotlinNativeBenchmarkExtension @Inject constructor(project: Project) : BenchmarkExtension(project) {
+    var jvmSrcDirs: Collection<Any> = emptyList()
+    var mingwSrcDirs: Collection<Any> = emptyList()
+    var posixSrcDirs: Collection<Any> = emptyList()
+
+    fun BenchmarkExtension.BenchmarkDependencies.jvm(notation: Any) = sourceSets.jvmMain.dependencies {
```
ilmat192: (minor) This method will be available only in Kotlin DSL and only inside the benchmark extension. I.e. this is possible:
```kotlin
benchmark {
    dependencies.jvm(...)
}
```
while this is not:

```kotlin
benchmark.dependencies.jvm(...)
```

But since this plugin is used only by us, this is ok.

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
```diff
@@ -138,89 +81,16 @@ open class SwiftBenchmarkingPlugin: Plugin<Project> {
...
+            getCodeSizeBenchmark(applicationName,
+                    if (benchmark.useCodeSize == CodeSizeEntity.FRAMEWORK)
+                        File("${framework.outputFile.absolutePath}/$nativeFrameworkName").canonicalPath
+                    else
+                        nativeExecutable
+            )
 
-        const val BENCHMARKING_GROUP = "benchmarking"
-    }
+    override fun Project.getCompilerFlags(nativeTarget: KotlinNativeTarget) =
```
ilmat192: Why this is empty for Swift benchmarks?

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -116,25 +101,31 @@ open class BenchmarkExtension @Inject constructor(val project: Project) {
 /**
  * A plugin configuring a benchmark Kotlin/Native project.
  */
-open class BenchmarkingPlugin: Plugin<Project> {
+abstract class BenchmarkingPlugin: Plugin<Project> {
+    protected abstract val Project.nativeExecutable: String
+    protected abstract val Project.nativeLinkTask: Task
+    protected abstract val Project.benchmark: BenchmarkExtension
+    protected abstract val Project.jvmParameters: Pair<Int, String>?
```
ilmat192: May be use a data class instead of `Pair`?

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/SwiftBenchmarkingPlugin.kt
LepilkinaElena: We are interested only in flags that influence

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -217,99 +191,132 @@ open class BenchmarkingPlugin: Plugin<Project> {
...
+                    println("JVM run is unsupported")
+                }
             }
         }
+    }
 
-        // Native report task.
-        val konanJsonReport = tasks.create("konanJsonReport") {
+    protected open fun Project.getCompilerFlags(nativeTarget: KotlinNativeTarget) =
+            nativeTarget.compilations.main.kotlinOptions.freeCompilerArgs.map { "\"$it\"" }
```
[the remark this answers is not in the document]
LepilkinaElena: I want to control what flags are coming from binary. What flags are returned from binary. I need only -g/-opt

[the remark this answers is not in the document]
LepilkinaElena: I think we may be need to rework this to get flags from mode + free flags to get combinations `-g -opt`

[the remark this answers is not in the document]
ilmat192: > What flags are returned from binary? I need only -g/-opt.

`nativeBinary.freeCompilerArgs` doesn't return `-g`/`-opt` so we do need to look at corresponding binary properties too. Also some flags that may influence performance may be specified manually in a build script. E.g. `-Xg0` or disabling some optimizations (e.g. devirtualization).

PR #3365, opened 2019-09-19
Fix KT-33824, ignore FreezableAtomicReference building condensation

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2262,56 +2287,67 @@ void freezeAcyclic(ContainerHeader* rootContainer, ContainerHeaderSet* newlyFroz
...
+    reversedEdges.emplace(currentContainer, KStdVector<ContainerHeader*>(0));
+    traverseContainerReferredObjects(currentContainer, [current, currentContainer, &queue, &reversedEdges](ObjHeader* obj) {
           ContainerHeader* objContainer = obj->container();
           if (canFreeze(objContainer)) {
             if (objContainer->marked())
-              queue.push_back(objContainer);
-            reversedEdges.emplace(objContainer, KStdVector<ContainerHeader*>(0)).first->second.push_back(current);
+              queue.push_back(obj);
+            // We ignore FreezableAtomicsReference during condensation, to avoid KT-33824.
+            if (!isFreezableAtomic(current) && !isFreezableAtomic(obj))
```
SvyatoslavScherbina: `!isFreezableAtomic(obj)` is likely not required here.
Removing it will make this part more consistent with `depthFirstTraversal`.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2262,56 +2287,67 @@ void freezeAcyclic(ContainerHeader* rootContainer, ContainerHeaderSet* newlyFroz
...
     auto& component = *it;
     int internalRefsCount = 0;
     int totalCount = 0;
     for (auto* container : component) {
+      RuntimeAssert(!isAggregatingFrozenContainer(container), "Must not be called on such containers");
       totalCount += container->refCount();
+#if TRACE_MEMORY
+     if (isFreezableAtomic(container)) {
+       RuntimeAssert(component.size() == 1, "Must be trivial condensation");
+       continue;
```
SvyatoslavScherbina: Isn't this required if `TRACE_MEMORY` is not true?

backend.native/tests/runtime/workers/freeze6.kt
```diff
@@ -30,5 +34,126 @@ fun ensureNeverFrozenFailsTarget(){
...
+
+    return ref1
+}
+
+@Test
+fun ensureFreezableHandlesCycles2() {
+    val ref = createRef2()
+    kotlin.native.internal.GC.collect()
+
+    assertTrue(ref.value.toString().length > 0)
```
SvyatoslavScherbina: Not sure if checking that `ref1` is not corrupted makes sense here, because it wouldn't be a part of condensation in any case.

runtime/src/main/cpp/Memory.cpp
olonho: It just makes transposed graph explicit in fact that we shalln't follow edge both from and to freezable atomic, and if we'll do an assert will trigger.

runtime/src/main/cpp/Memory.cpp
olonho: It is just an excessive check, not sure if it makes sense in non-tracing case.

SvyatoslavScherbina: I guess the tests are failing due to this.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: >  we shalln't follow edge ... to freezable atomic

This is redundant. It is enough to properly ignore edges _from_ freezable atomics in the entire condensation algorithm.

> and if we'll do an assert will trigger.

I don't think so.

olonho: I meant `RuntimeAssert(it != reversedEdges->end(), "unknown node during condensation building");`.

SvyatoslavScherbina: I'm still not sure that removing the discussed check will trigger this assertion. Can you provide an example?

PR #3368, opened 2019-09-20
Memory leak checker API.

olonho:


runtime/src/main/kotlin/kotlin/native/Platform.kt
```diff
@@ -81,6 +81,14 @@ public object Platform {
     public val isDebugBinary: Boolean
         get() = Platform_isDebugBinary()
 
+    /**
+     * If the memory leak checker is activated, by default true in debug mode, false in release.
+     * When memory leak checker is activated, and leak detected on process termination process
+     * is terminated with non-zero exit code.
```
SvyatoslavScherbina: Is it?

runtime/src/main/kotlin/kotlin/native/Platform.kt
```diff
@@ -81,6 +81,14 @@ public object Platform {
...
+     */
+    public var hasMemoryLeakChecker: Boolean
```
SvyatoslavScherbina: The name doesn't correspond to the behaviour.
Maybe `isMemoryLeakCheckerEnabled`?

runtime/src/main/kotlin/kotlin/native/Platform.kt
olonho: Oh, forgot to push update doc.

SvyatoslavScherbina: Error message without termination is somewhat indiscoverable (including leaks in Kotlin/Native tests). Why not terminate?

olonho: Well, OK, will update message as well.

runtime/src/main/kotlin/kotlin/native/Platform.kt
```diff
@@ -82,11 +82,11 @@ public object Platform {
         get() = Platform_isDebugBinary()
 
     /**
-     * If the memory leak checker is activated, by default true in debug mode, false in release.
-     * When memory leak checker is activated, and leak detected on process termination process
-     * is terminated with non-zero exit code.
+     * If the memory leak checker is activated, by default `true` in debug mode, `false` in release.
+     * When memory leak checker is activated, and leak is detected during last Kotlin/Native runtime
+     * deinitialization process - error message with leak information is printed.
```
SvyatoslavScherbina: > during last Kotlin/Native runtime deinitialization process

This is unclear. Maybe align the terms with workers?

runtime/src/main/kotlin/kotlin/native/Platform.kt
olonho: Renamed.

runtime/src/main/kotlin/kotlin/native/Platform.kt
olonho: What would you suggest?

SvyatoslavScherbina: Workers seem to use "Kotlin context": https://github.com/JetBrains/kotlin-native/blob/a98e5c799cd7612b014b6bc926c899922eea2453/runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt#L42

I suggest using the same phrase here, maybe described more verbosely (like "workers and platform threads ran Kotlin code"). Or maybe keep only the description, without "Kotlin context".

PR #3372, opened 2019-09-23
[codegen][debug info][type] use `ObjHeader *` as a type for type parameterized variables

vvlevchenko:
KT-33789


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/DebugUtils.kt
```diff
@@ -167,25 +183,7 @@ internal fun IrType.dwarfType(context: Context, targetData: LLVMTargetDataRef):
...
-                            flags         = DWARF.flagsForwardDeclaration,
-                            derivedFrom   = null,
-                            elements      = null,
-                            elementsCount = 0,
-                            refPlace      = null)!! as DITypeOpaqueRef
-                    dwarfPointerType(context, type)
-                }
-                this.isTypeParameter() -> //TODO: Type parameter,  how to deal with if?
-                    debugInfoBaseType(context, targetData, this.toString(), llvmType(context), encoding(context).value.toInt())
+                classOrNull != null || this.isTypeParameter()-> context.debugInfo.objHeaderPointerType!!
```
olonho: Space before ->.

PR #3379, opened 2019-09-25
Changed samples to use new MPP plugin flag

LepilkinaElena:


build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
```diff
@@ -173,7 +173,8 @@ open class BenchmarkingPlugin: Plugin<Project> {
 
     private fun Project.configureNativeTarget(hostPreset: KotlinNativeTargetPreset) {
         kotlin.targetFromPreset(hostPreset, NATIVE_TARGET_NAME) {
-            compilations.getByName("main").kotlinOptions.freeCompilerArgs = project.compilerArgs + listOf("-l", "kotlinx-cli")
+            compilations.getByName("main").kotlinOptions.freeCompilerArgs = project.compilerArgs
+            compilations.getByName("main").enableEndorsedLibs = true
```
ilmat192: It's possible to write just `compilations["main"]`

ilmat192:
Looks like MPP plugin version needs to be updated.

LepilkinaElena:
> Looks like MPP plugin version needs to be updated.

ilmat192:
For samples too :)

build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/BenchmarkingPlugin.kt
LepilkinaElena: It'll be rewritten when I merge review with base class for plugins.

PR #3394, opened 2019-09-30
Use export-dynamic with wasm-ld

knebekaizer:
Fix for 'TypeError: instance.exports.Konan_js_main is not a function'


olonho:
All WASM tests pass now.

PR #3397, opened 2019-09-30
Mark Kotlin_Any_equals and Kotlin_Any_hashCode as ALWAYS_INLINE

sbogolepov:


runtime/src/main/cpp/Natives.cpp
```diff
@@ -29,11 +29,11 @@
 extern "C" {
 
 // Any.kt
-KBoolean Kotlin_Any_equals(KConstRef thiz, KConstRef other) {
+ALWAYS_INLINE KBoolean Kotlin_Any_equals(KConstRef thiz, KConstRef other) {
```
SvyatoslavScherbina: Why not make these ones intrinsic?

SvyatoslavScherbina: Disregard this. Intrinsics probably don't work with overriding.

PR #3403, opened 2019-10-01
Enable code coverage on Linux and MinGW

sbogolepov:


build-tools/src/main/kotlin/org/jetbrains/kotlin/Utils.kt
```diff
@@ -54,6 +54,12 @@ val Project.globalTestArgs: List<String>
 fun Project.platformManager() = findProperty("platformManager") as PlatformManager
 fun Project.testTarget() = findProperty("target") as KonanTarget
 
+fun KonanTarget.supportsCodeCoverage(): Boolean =
+        this == KonanTarget.MINGW_X64 ||
```
olonho: Duplication with CoverageManager.kt? Shalln't be in shared?

sbogolepov: I wasn't sure that there is place to share such logic. But now I remembered `shared` package.

PR #3405, opened 2019-10-01
[codegen] do not export reified inline functions (KT-34044)

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -123,6 +123,10 @@ private fun isExportedFunction(descriptor: FunctionDescriptor): Boolean {
     descriptor.allParameters.forEach {
         if (it.type.isGeneric()) return false
     }
+    descriptor.typeParameters.forEach{
```
olonho: Maybe we can just filter method with any type parameters and replace check above instead?

vvlevchenko: sounds good. could you please take a look in next commit?

PR #3408, opened 2019-10-01
Zipping sources of endorsed libraries for IDE

LepilkinaElena:


build.gradle
```diff
@@ -216,9 +216,39 @@ task distNativeSources(type: Zip) {
...
+    include('**/*.kt')
+
+    from(project(':endorsedLibraries').file('build/sources-common'))
+}
+
+task distEndorsedNativeSources(type: Zip) {
+    dependsOn ":endorsedLibraries:endorsedLibsNativeSources"
+
+    destinationDirectory = file("$distDir/sources")
+    archiveFileName = "kotlin-endorsed-native-sources.zip"
```
ddolovov: No, no. Please put the source code of every single library in a separate ZIP file. Please use pattern `LIBRARY-sources.zip` for the ZIP file name.

build.gradle
```diff
@@ -216,9 +216,39 @@ task distNativeSources(type: Zip) {
...
+
+    includeEmptyDirs = false
+    include('**/*.kt')
+
+    from(project(':endorsedLibraries').file('build/sources-native'))
```
ddolovov: I think referring anything in `build` directory is a dangerous practice. You may refer Gradle tasks, source sets, etc - any entity that Gradle knows how to build and knows how to check if it should be invalidated and re-built. But `build` directory can be treated as always holding stale data.

ilmat192: > You may refer Gradle tasks, source sets, etc - any entity that Gradle knows how to build and knows how to check if it should be invalidated and re-built.

I think that using Gradle's dependency mechanism in the same manner as it's used for stdlib sources is the best choice. I've prototyped something similar for native sources: https://github.com/JetBrains/kotlin-native/commit/eb0f1b1489a479a93b534269e59659c35160129b.

LepilkinaElena: Thanks, but as Dmitry said that there should be separated zip archives for each library, there is no necessity in any coping, so I made all easier

PR #3414, opened 2019-10-02
Unify compilation pipelines

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OptimizationPipeline.kt
```diff
@@ -92,6 +98,8 @@ private class LlvmPipelineConfiguration(context: Context) {
     }
 
     val sizeLevel: Int = when {
+        target is KonanTarget.ZEPHYR ||
+        target == KonanTarget.WASM32 -> 3
```
alexander-gorshenev: What are those magic numbers? 100, 3?

PR #3433, opened 2019-10-07
[KLIB] Fix uniqId API

romanart:
Goes along with https://github.com/JetBrains/kotlin/tree/rr/romart/proto_scheme_opts

Fix in native support to support new klib file API


romanart:
cc: @vvlevchenko

alexander-gorshenev:
Merged within the #3443

PR #3440, opened 2019-10-09
Support ObjCExport with dynamic cache

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -390,13 +430,23 @@ internal class ObjCExportCodeGenerator(
...
+    } else {
+        context.llvmImports.add(origin)
+        val global = staticData.placeGlobal(name, value, isExported = true)
 
-    if (context.llvmModuleSpecification.importsKotlinDeclarationsFromOtherObjectFiles()) {
-        // Note: actually this is required only if global's weak/common definition is in other object file,
-        // but it is simpler to do this for all globals, considering that all usages can't be removed by DCE anyway.
-        context.llvm.usedGlobals += global.llvmGlobal
+        if (context.llvmModuleSpecification.importsKotlinDeclarationsFromOtherObjectFiles()) {
+            // Note: actually this is required only if global's weak/common definition is in other object file,
```
olonho: other -> another

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -315,6 +315,8 @@ - (instancetype)copyWithZone:(NSZone *)zone {
 __attribute__((weak)) const ObjCTypeAdapter** Kotlin_ObjCExport_sortedProtocolAdapters = nullptr;
 __attribute__((weak)) int Kotlin_ObjCExport_sortedProtocolAdaptersNum = 0;
 
+__attribute__((weak)) bool Kotlin_ObjCExport_initTypeAdapters = false;
```
olonho: Not very trivial state management in Kotlin_ObjCExport_initTypeAdapters. Could it be const, as only seems to be changed during linkage, and not runtime?

SvyatoslavScherbina: Actually it is changed in runtime, handled with https://github.com/JetBrains/kotlin-native/pull/3440/files#diff-70fa52f175b3e9befea519e771fa08e2R440

PR #3441, opened 2019-10-09
Move exported runtime Obj-C classes from dynamic cache to final binary

SvyatoslavScherbina:
Otherwise Swift/Obj-C code requires linking with cache directly too.


runtime/build.gradle
```diff
@@ -67,6 +68,13 @@ targetList.each { targetName ->
         srcRoot file('src/profile_runtime')
         target targetName
     }
+
+    task ("${targetName}ObjC", type: CompileCppToBitcode) {
```
olonho: Nice!

runtime/src/main/cpp/ObjCExportCollectionUtils.mm
```diff
@@ -0,0 +1,269 @@
...
+
+static inline KInt objCSizeToKotlinOrThrow(NSUInteger size) {
+  if (size > std::numeric_limits<KInt>::max()) {
+    Kotlin_ObjCExport_ThrowCollectionTooLarge();
+  }
+
+  return size;
+}
+
+// Imported to ObjCExportUtils.kt:
```
olonho: Exported to?

PR #3475, opened 2019-10-18
Use toString as debugger preview on all objects

koush:
This fixes fix cyclic structures stringification and slow debugger stepping when using this script (in VSCode or any lldb environment).
The debugger preview also behaves more like it would under the JVM (toString),
The existing toString also causes large byte arrays to take forever to evaluate.

I understand this script is probably intended for use in the command line, where it may be tedious to inspect member variables. So, maybe this isn't appropriate. A typical IDE using this script will automatically query those members.

### DIFF
--- a/llvmDebugInfoC/src/scripts/konan_lldb.py
+++ b/llvmDebugInfoC/src/scripts/konan_lldb.py
@@ -132,6 +132,25 @@ class KonanHelperProvider(lldb.SBSyntheticValueProvider):
             valobj.GetType().GetBasicType(lldb.eBasicTypeBool)
         ]
 
+        fallback = valobj.GetValue()
+        buff_len = evaluate(
+            '(int)Konan_DebugObjectToUtf8Array({}, (char *)Konan_DebugBuffer(), (int)Konan_DebugBufferSize());'.format(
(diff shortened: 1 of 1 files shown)


koush:
Am using the codelldb plugin for vscode:
https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb

with the following launch.json config snipet:

```json
            "postRunCommands": [
                "command script import /Volumes/Dev/scratch/konan_lldb.py"
            ],
```

TlsTests_kt_—_scratch

kpgalligan:
I've had a similar one out there for a while. It currently works for AppCode and Xcode with slightly different behavior. Haven't looked at VSCode, though. https://github.com/JetBrains/kotlin-native/pull/2972 Xcode "plugin" is overdue for a refresh, so may review lldb as well.

vvlevchenko:
Thanks for your contribution, we'd preferred to keep operation with synthetic objects to stay usable with scripting on top python bindings, but for summary needs we've reduced amount of data gathered throw python API and number of corresponding expression evaluations e.g. here d4028ad13a4f39fcc8b0978d9e320f67a732f607.

PR #3479, opened 2019-10-21
Workaround clang crash in cinterop tool (KT-34467)

ilmat192:


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Utils.kt
```diff
@@ -257,6 +257,9 @@ fun Compilation.copy(
         language = language
 )
 
+fun Compilation.copyWithArgsForPCH(): Compilation =
```
sbogolepov: A comment that describes a motivation for this function will be useful.

backend.native/tests/build.gradle
```diff
@@ -3318,6 +3318,14 @@ if (PlatformInfo.isAppleTarget(project)) {
         it.headers "$projectDir/interop/objc/msg_send/messaging.h"
         it.linkerOpts "-L$buildDir", "-lobjcmessaging"
     }
+
+    createInterop("objcKt34467") {
+        it.defFile 'interop/objc/kt34467/module_library.def'
+        def moduleMap = file("interop/objc/kt34467/module_library.modulemap").absolutePath
+        it.compilerOpts "-fmodule-map-file=$moduleMap"
+        // We need to import CoreFoundation.h but there is no need to run interop for it.
```
SvyatoslavScherbina: Can we make the test import our own header?
Making the test dependent on platform libs is somewhat fragile.

ilmat192: Fixed.

PR #3481, opened 2019-10-21
Improve Gradle DSL in docs

ilmat192:


FAQ.md
```diff
@@ -16,13 +16,19 @@ garbage.
 ### Q: How do I create a shared library?
 
 A: Use the `-produce dynamic` compiler switch, or `binaries.sharedLib()` in Gradle, i.e.
+
+<div class="sample" markdown="1" theme="idea" mode="groovy">
+
```
olonho: Maybe use Kotlin DSL, not Groovy here.

ilmat192: Fixed.

p7nov: You can also consider multi-language samples with Kotlin-Groovy switch as in [Using Gradle](https://kotlinlang.org/docs/reference/using-gradle.html).
Link to source: https://github.com/JetBrains/kotlin-web-site/blob/master/pages/docs/reference/using-gradle.md

PR #3486, opened 2019-10-22
Add Xcode version check

ilmat192:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Apple.kt
```diff
@@ -63,19 +64,35 @@ class AppleConfigurablesImpl(
...
+                val currentXcodeVersion = xcode.version
+                checkXcodeVersion(minimalXcodeVersion, currentXcodeVersion)
             }
         }
 
         XcodePartsProvider.Local(xcode)
     }
 
+    private fun checkXcodeVersion(minimalVersion: String, currentVersion: String) {
+        val minimalVersionParts = minimalVersion.split('.').map { it.toInt() }
```
olonho: Cannot Xcode version contain non-number parts?

konan/konan.properties
```diff
@@ -25,8 +25,8 @@ dependencyProfiles = default alt
 airplaneMode = false
 # if true, ignores Xcode version check.
 ignoreXcodeVersionCheck = false
-# Disabled until decided better mechanism.
-# useFixedXcodeVersion = 11.0
+
+minimalXcodeVersion=11.0
```
olonho: IIRC we agreed that there's a minimal version for building compiler and for using it.  Most important one is using part.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Apple.kt
```diff
@@ -63,19 +64,35 @@ class AppleConfigurablesImpl(
...
+
+        for (i in 0 until size) {
+            val currentPart = currentVersionParts.getOrElse(i) { 0 }
+            val minimalPart = minimalVersionParts.getOrElse(i) { 0 }
+
+            when {
+                currentPart > minimalPart -> return
+                currentPart < minimalPart ->
+                    error("Unsupported Xcode version. Consider updating Xcode. " +
+                        "Minimal supported Xcode version is $minimalVersion, got $currentVersion.")
```
olonho: Minimal supported Xcode version is $minimalVersion, got $currentVersion.->  Unsupported Xcode version $currentVersion, minimal supported version is $minimalVersion.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Apple.kt
ilmat192: I've checked a beta version and it reports `11.2`. Just in case made the check more stable.

konan/konan.properties
ilmat192: This commit is right about check at the user side.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Apple.kt
ilmat192: Fixed.

PR #3493, opened 2019-10-24
Implement park() and name APIs for Worker.

olonho:


runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -35,8 +35,10 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
          * better to use non-blocking IO combined with more lightweight coroutines.
          *
          * @param errorReporting controls if an uncaught exceptions in the worker will be printed out
+         * @param name defines
          */
-        public fun start(errorReporting: Boolean = true): Worker = Worker(startInternal(errorReporting))
+        public fun start(errorReporting: Boolean = true, name: String? = null): Worker
```
SvyatoslavScherbina: This will likely break binary compatibility between libraries and stdlib, which is undesirable for 1.3.60 EAP 2-3.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -35,8 +35,10 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
          * better to use non-blocking IO combined with more lightweight coroutines.
          *
          * @param errorReporting controls if an uncaught exceptions in the worker will be printed out
+         * @param name defines
```
SvyatoslavScherbina: defines what?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
 
     /**
-     * Process pending job(s) on the queue of this worker, returns `true` if something was processed
-     * and `false` otherwise. Note that jobs scheduled with [executeAfter] using non-zero timeout are
+     * Process pending job(s) on the queue of this worker.
+     * Note that jobs scheduled with [executeAfter] using non-zero timeout are
      * not processed this way. If termination request arrives while processing the queue via this API,
      * worker is marked as terminated and will exit once the current request is done with.
+     *
+     * @throws [IllegalArgumentException] if this request is executed on non-current [Worker].
```
SvyatoslavScherbina: `IllegalStateException`?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * @return `true` if request(s) was processed and `false` otherwise.
      */
     public fun processQueue(): Boolean = processQueueInternal(id)
 
+    /**
+     * Park execution of the current worker until a new request arrives or timeout specified in
+     * [timeoutMicroseconds] elapsed. If [processQueue] is true, pending queue elements are processed the same way,
+     * as specified in [processQueue] function.
+     *
+     * @param timeoutMicroseconds defines how long to park worker if no requests arrive.
```
SvyatoslavScherbina: Can it be nanoseconds, as requested in [KT-34570](https://youtrack.jetbrains.com/issue/KT-34570)?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
      * worker is marked as terminated and will exit once the current request is done with.
+     *
+     * @throws [IllegalArgumentException] if this request is executed on non-current [Worker].
+     * @return `true` if request(s) was processed and `false` otherwise.
      */
     public fun processQueue(): Boolean = processQueueInternal(id)
 
+    /**
+     * Park execution of the current worker until a new request arrives or timeout specified in
+     * [timeoutMicroseconds] elapsed. If [processQueue] is true, pending queue elements are processed the same way,
```
SvyatoslavScherbina: > `[processQueue]`

`[process]`

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * as specified in [processQueue] function.
+     *
+     * @param timeoutMicroseconds defines how long to park worker if no requests arrive.
+     * @param process defines if arrived request(s) shall be processed.
+     * @throws [IllegalArgumentException] if this request is executed on non-current [Worker].
+     * @return `true` if request(s) was processed and `false` otherwise if [process] was `true`.
+     * @return `true` if request arrived and `false` if timeout happens if [process] was `false`.
+     */
+    public fun park(timeoutMicroseconds: Long = 0, process: Boolean = true): Boolean =
```
SvyatoslavScherbina: Does default value for `timeoutMicroseconds` make any sense?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: It could, but not sure if nanoseconds are proper granularity for those APIs. Can ask @elizarov on that.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+            parkInternal(id, timeoutMicroseconds, process)
+
+    /**
+     * Name or the worker, as specified in [Worker.start] or "worker $id" by default,
```
SvyatoslavScherbina: > Name or the worker

of

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,17 +107,45 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+    /**
+     * Park execution of the current worker until a new request arrives or timeout specified in
+     * [timeoutMicroseconds] elapsed. If [processQueue] is true, pending queue elements are processed the same way,
+     * as specified in [processQueue] function.
+     *
+     * @param timeoutMicroseconds defines how long to park worker if no requests arrive.
+     * @param process defines if arrived request(s) shall be processed.
+     * @throws [IllegalArgumentException] if this request is executed on non-current [Worker].
+     * @return `true` if request(s) was processed and `false` otherwise if [process] was `true`.
+     * @return `true` if request arrived and `false` if timeout happens if [process] was `false`.
```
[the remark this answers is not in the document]
olonho: Both are correct, just with different value of `process` argument. Not that I like this API much, maybe there are better ideas.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -35,22 +35,26 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
          * better to use non-blocking IO combined with more lightweight coroutines.
          *
          * @param errorReporting controls if an uncaught exceptions in the worker will be printed out
+         * @param name defines the optional name of this worker, if none - default naming is used.
+         * @returns worker object, usable across multiple concurrent contexts.
```
SvyatoslavScherbina: > @returns

`@return`
Same below.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Already fixed in feedback fix commit.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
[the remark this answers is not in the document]
SvyatoslavScherbina: Note that default kdoc parser doesn't seem to handle two `@return` clauses anyway.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,51 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * @throws [IllegalStateException] if this request is executed on non-current [Worker].
      */
-    override public fun toString(): String = "worker $id"
+    public fun park(timeoutMicroseconds: Long, process: Boolean = false): Boolean =
+            parkInternal(id, timeoutMicroseconds, process)
+
+    /**
+     * Name of the worker, as specified in [Worker.start] or "worker $id" by default,
+     *
+     * @throws [IllegalArgumentException] if this request is executed on invalid worker.
```
SvyatoslavScherbina: `IllegalStateException`?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -721,10 +759,14 @@ KLong Worker::checkDelayedLocked() {
...
-void Worker::waitForQueueLocked() {
+bool Worker::waitForQueueLocked(KLong timeoutMicroseconds) {
   while (queue_.size() == 0) {
     KLong closestToRun = checkDelayedLocked();
-    if (closestToRun == 0) continue;
+    if (closestToRun == 0) {
+        continue;
+    }
+    if (timeoutMicroseconds >= 0)
+        closestToRun = timeoutMicroseconds < closestToRun || closestToRun < 0 ? timeoutMicroseconds : closestToRun;
```
SvyatoslavScherbina:
```kotlin

fun main() {
    println(Worker.current.park(2000))
}
```

Finishes much faster than 2000 ms. Is this expected?

Btw, how spurious wakeups are handled here?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
elizarov: I’d prefer nanoseconds here. It would be easier to port code between JVM and native then.

elizarov: Also a question ob why it is an instance method of Worker and what happens if I call it from another worker? Which one gets parked?

olonho: * @throws [IllegalStateException] if this request is executed on non-current [Worker].

olonho: OK, can switch to nanoseconds, although it doesn't look entirely correct.

runtime/src/main/cpp/Worker.cpp
olonho: Argument is microseconds, as explicitly stated in the name.

olonho: Spurious wakeups are not handled, ant not sure which way/if we shall handle them. If we woke up earlier in timed wait, we'll just check appropriate conditions on that moment. The only questionable part is unlimited wait, but it's also doesn't seem critical.

olonho:
```nike@~/kotlin/kotlin-native
>cat park.kt 

fun main() {
    println(Worker.current.park(2000000))
}
nike@~/kotlin/kotlin-native
>time ./p.kexe
false

real	0m2.013s
user	0m0.005s
sys	0m0.004s
```

olonho: With new version with timeout in nanoseconds  it has to be 2_000_000_000, of course.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -736,7 +778,30 @@ void Worker::waitForQueueLocked() {
     } else {
       pthread_cond_wait(&cond_, &lock_);
     }
+    if (timeoutMicroseconds >= 0) return queue_.size() != 0;
```
SvyatoslavScherbina:
```kotlin
  
fun main() {
    Worker.current.executeAfter(10*1000, {
        println("task")
    }.freeze())
    println(Worker.current.park(1000*1000*1000, process = true))
}
```

Prints `false`.

runtime/src/main/cpp/Worker.cpp
SvyatoslavScherbina: > Argument is microseconds, as explicitly stated in the name.
 
Sorry, my bad.

SvyatoslavScherbina: > Spurious wakeups are not handled, and not sure which way/if we shall handle them.

For example, by waiting in a loop till a particular moment in time (preferably using monotonic clock if available).

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,70 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     *
+     * @throws [IllegalStateException] if this request is executed on non-current [Worker].
+     * @return `true` if request(s) was processed and `false` otherwise.
      */
     public fun processQueue(): Boolean = processQueueInternal(id)
 
     /**
-     * String representation of this worker.
+     * Park execution of the current worker until a new request arrives or timeout specified in
+     * [timeoutMicroseconds] elapsed. If [process] is true, pending queue elements are processed the same way,
```
SvyatoslavScherbina: > `[timeoutMicroseconds]`

nanoseconds.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,70 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * as specified in [processQueue] function.
```
SvyatoslavScherbina: Not exactly, because `processQueue` processes single queue element.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,70 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
     public fun asCPointer() : COpaquePointer? = id.toLong().toCPointer()
+}
+
+/**
+ * Executes [block] in new [Worker], by starting a worker, calling provided [block] and terminating
+ * worker after block is completed. Note that this operaation is pretty heavyweight, use preconfigured
+ * worker or worker pool if need to execute it frequently.
+ *
+ * @param name of the started worker.
+ * @param reportErrors controls if uncaught errors in worker to be reported.
```
SvyatoslavScherbina: Name is inconsistent with `errorReporting` in `[Worker.start]`.

backend.native/tests/runtime/workers/worker4.kt
```diff
@@ -9,14 +9,65 @@ import kotlin.test.*
...
+            assertTrue(Worker.current.park(10_000_000_000, true))
+            assertEquals(3, counter.value)
+        }.freeze())
+
+        executeAfter(0, {
+            counter.increment()
+        }.freeze())
+
+        while (counter.value == 1) {
+            Worker.current.park(1_000_000)
```
SvyatoslavScherbina: This never happens.

runtime/src/main/cpp/Worker.cpp
olonho: Not sure if it is indeed required/important for this functionality.

SvyatoslavScherbina: Can current implementation wait longer than specified if spurious wakeups happen?

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: No, I will avoid using microseconds, otherwise API gets really crazy when `executeAfter` takes microseonds and this one nanoseconds.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
olonho: Reverted to us.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,70 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
      * Convert worker to a COpaquePointer value that could be passed via native void* pointer.
      * Can be used as an argument of [Worker.fromCPointer].
+     *
+     * @return worker identifier as C pointer.
      */
     public fun asCPointer() : COpaquePointer? = id.toLong().toCPointer()
+}
+
+/**
+ * Executes [block] in new [Worker], by starting a worker, calling provided [block] and terminating
```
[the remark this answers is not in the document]
olonho: Clarified.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,70 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * as specified in [processQueue] function.
+     *
+     * @param timeoutNanoseconds defines how long to park worker if no requests arrive.
+     * @param process defines if arrived request(s) shall be processed.
+     * @return if [process] is `true`: if request(s) was processed `true` and `false` otherwise.
+     *   if [process] is `false`:` true` if request(s) has arrived and `false` if timeout happens.
+     * @throws [IllegalStateException] if this request is executed on non-current [Worker].
+     */
+    public fun park(timeoutNanoseconds: Long, process: Boolean = false): Boolean =
+            parkInternal(id, timeoutNanoseconds / 1000, process)
```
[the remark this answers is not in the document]
olonho: Fixed anyway.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -736,7 +782,34 @@ void Worker::waitForQueueLocked() {
...
+  {
+    Locker locker(&lock_);
+    if (terminated_) {
+        return false;
+    }
+    auto arrived = false;
+    KLong remaining = timeoutMicroseconds;
+    do {
+      arrived = waitForQueueLocked(remaining, &remaining);
+    } while (remaining != 0 && !arrived);
```
SvyatoslavScherbina:
```kotlin

fun main() {
    val worker = Worker.start()
    val main = Worker.current 

    worker.executeAfter(1000, {
        main.executeAfter(1, {
            println("Task")
        }.freeze())
    }.freeze())

    println(main.park(1000 * 1000, process = true))
}
```

Prints `false`.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,71 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     *
+     * @return worker identifier as C pointer.
      */
     public fun asCPointer() : COpaquePointer? = id.toLong().toCPointer()
+}
+
+/**
+ * Executes [block] with new [Worker] as resource, by starting the new worker, calling provided [block]
+ * (in current context) with newly started worker as [this] and terminating worker after the block completes.
+ * Note that this operaation is pretty heavyweight, use preconfigured worker or worker pool if need to
```
SvyatoslavScherbina: > operaation

backend.native/tests/runtime/workers/worker4.kt
olonho: No, it could be either 0 or 1, depending on if `assertTrue(Worker.current.processQueue())` executed job with increment scheduled line above or not.

runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -103,21 +109,71 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
...
+     * [timeoutMicroseconds] elapsed. If [process] is true, pending queue elements are processed,
+     * including delayed requests. Note that multiple requests could be processed this way.
+     *
+     * @param timeoutMicroseconds defines how long to park worker if no requests arrive.
+     * @param process defines if arrived request(s) shall be processed.
+     * @return if [process] is `true`: if request(s) was processed `true` and `false` otherwise.
+     *   if [process] is `false`:` true` if request(s) has arrived and `false` if timeout happens.
+     * @throws [IllegalStateException] if this request is executed on non-current [Worker].
+     */
+    public fun park(timeoutMicroseconds: Long, process: Boolean = false): Boolean =
```
SvyatoslavScherbina: Doesn't check that `timeoutMicroseconds >= 0`.

backend.native/tests/runtime/workers/worker4.kt
```diff
@@ -9,14 +9,88 @@ import kotlin.test.*
...
+}
+
+@Test fun runTest2() {
+    withWorker {
+        val counter = AtomicInt(0)
+
+        executeAfter(0, {
+            assertTrue(Worker.current.park(10_000_000, false))
+            assertEquals(counter.value, 0)
+            assertTrue(Worker.current.processQueue())
```
SvyatoslavScherbina: Is it guaranteed? What if this task gets executed before `executeAfter` below?

PR #3496, opened 2019-10-25
[platformLibs] added missed libs to mingw platform.windows

msink:


platformLibs/src/platform/mingw/windows.def
```diff
@@ -3,6 +3,9 @@ headers = wtypes.h minwindef.h windows.h commctrl.h dwmapi.h shlobj.h shlwapi.h
     urlmon.h usp10.h uxtheme.h vfw.h wininet.h winsock2.h ws2tcpip.h ws2def.h
 compilerOpts = -DUNICODE -DWINVER=0x0601 -D_WIN32_WINNT=0x0601 -DWINAPI_FAMILY=3 -DOEMRESOURCE \
     -Wno-incompatible-pointer-types -Wno-deprecated-declarations
-linkerOpts = -lcomctl32 -lcrypt32 -lshlwapi -lshell32 -limm32 -lusp10 -lwininet -lgdi32 -luser32 -lkernel32 -lbcrypt -luuid
+linkerOpts = -lcomctl32 -lcrypt32 -lshlwapi -lshell32 -limm32 -lusp10 -lwininet -lgdi32 -luser32 -lkernel32 -lbcrypt -luuid \
```
olonho: Not sure if it's a good idea. Would be better to split libraries into separate modules. It's already ugly enough.

msink: Maybe, it was just simplest possible solution.

Then, how it should look ideally - one module for `.h` file?
For example, if Misrosoft's documentation for [SCardConnectA](https://docs.microsoft.com/en-us/windows/win32/api/winscard/nf-winscard-scardconnecta) says "Header: winscard.h" - in K/N it should be in `platform.winscard` instead of `platform.windows` ?

msink:
Continued in #3502

PR #3498, opened 2019-10-25
simd-128 basic support

knebekaizer:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -149,30 +149,31 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
...
-            "kotlin.CharArray"          to int16Type,
-            "kotlin.ShortArray"         to int16Type,
-            "kotlin.IntArray"           to int32Type,
-            "kotlin.LongArray"          to int64Type,
-            "kotlin.FloatArray"         to floatType,
-            "kotlin.DoubleArray"        to doubleType,
-            "kotlin.BooleanArray"       to int8Type,
-            "kotlin.String"             to int16Type,
-            "kotlin.native.ImmutableBlob" to int8Type,
+            "kotlin.ByteArray"          to LLVMInt8Type()!!,
```
olonho: How it relates to your change?

samples/simd/simd_lib.c
```diff
@@ -0,0 +1,76 @@
+#include "simd_lib.h"
```
olonho: Better use SIMD types from platform APIs in `uikit` or `objc` samples.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmUtils.kt
```diff
@@ -110,6 +110,7 @@ internal val int64Type get() = LLVMInt64TypeInContext(llvmContext)!!
 internal val int8TypePtr get() = pointerType(int8Type)
 internal val floatType get() = LLVMFloatTypeInContext(llvmContext)!!
 internal val doubleType get() = LLVMDoubleTypeInContext(llvmContext)!!
+internal val vectorType get() = LLVMVectorType(LLVMFloatTypeInContext(llvmContext), 4)!!
```
olonho: vector16Type?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt
```diff
@@ -207,6 +207,7 @@ private fun ObjCExportMapper.bridgeType(
                 KonanPrimitiveType.FLOAT -> ObjCValueType.FLOAT
                 KonanPrimitiveType.DOUBLE -> ObjCValueType.DOUBLE
                 KonanPrimitiveType.NON_NULL_NATIVE_PTR -> ObjCValueType.POINTER
+                KonanPrimitiveType.VECTOR16 -> ObjCValueType.LONG_LONG  // FIXME
```
olonho: `simd_float4`?

runtime/src/main/cpp/Types.cpp
```diff
@@ -93,4 +93,10 @@ bool IsSubInterface(const TypeInfo* thiz, const TypeInfo* other) {
   return false;
 }
 
+typedef float vFloat __attribute__ ((__vector_size__ (16)));
```
SvyatoslavScherbina: Maybe align naming with these types: https://github.com/JetBrains/kotlin-native/blob/aac2de48fcde013ae0d4f74ebd3347c5198cc388/runtime/src/main/cpp/Types.h#L41?

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -499,6 +499,19 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
                 spelling = clang_getTypeSpelling(type).convertAndDispose().dropConstQualifier()
         )
 
+        CXType_Vector -> {
+            val elementType = convertType(clang_getElementType(type))
+            val size = clang_Type_getSizeOf(type)
+            val elemSize = clang_Type_getSizeOf(clang_getElementType(type))
+            val length = size / elemSize
+            val spelling = type.name // ex: __attribute__((__vector_size__(4 * sizeof(const float)))) const float
```
SvyatoslavScherbina: `dropConstQualifier` hack may also be required here as above: https://github.com/JetBrains/kotlin-native/blob/e1b3fefd3333fa730bb00174b7eb24c90822f6a0/Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt#L499

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -756,6 +756,14 @@ private fun KotlinStubs.mapType(type: IrType, retained: Boolean, variadic: Boole
 private fun IrType.isTypeOfNullLiteral(): Boolean = this is IrSimpleType && hasQuestionMark
         && classifier.isClassWithFqName(KotlinBuiltIns.FQ_NAMES.nothing)
 
+private fun IrType.isVector(): Boolean {
+    if (this is IrSimpleType) {
+        val fqName = FqName("kotlinx.cinterop.NativeVector").toUnsafe()  // FIXME
+        return classifier.isClassWithFqName(fqName)
```
SvyatoslavScherbina: Nullability (`this.hasQuestionMark`) must also be checked.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -954,6 +964,18 @@ private class BooleanValuePassing(override val cType: CType, private val irBuilt
     override fun cToBridged(expression: String): String = cBridgeType.cast(expression)
 }
 
+private class VectorTypePassing(private val kotlinClass: IrClass, val cType: CType) : KotlinToCArgumentPassing {
```
SvyatoslavScherbina: Seems unused.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmUtils.kt
SvyatoslavScherbina: `floatType` can be used instead of `LLVMFloatTypeInContext(llvmContext)`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -172,7 +172,8 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
             floatType to 6,
             doubleType to 7,
             kInt8Ptr to 8,
-            int1Type to 9
+            int1Type to 9,
+            vectorType to 10
```
SvyatoslavScherbina: Should be also supported in runtime debug support.
See [comment above](https://github.com/JetBrains/kotlin-native/blob/e1b3fefd3333fa730bb00174b7eb24c90822f6a0/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt#L165).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/KonanDefaultParameterInjector.kt
```diff
@@ -29,6 +28,7 @@ internal class KonanDefaultParameterInjector(private val konanContext: KonanBack
             PrimitiveBinaryType.FLOAT -> IrConstImpl.float(startOffset, endOffset, type, 0.0F)
             PrimitiveBinaryType.DOUBLE -> IrConstImpl.double(startOffset, endOffset, type, 0.0)
             PrimitiveBinaryType.POINTER -> irCall(startOffset, endOffset, symbols.getNativeNullPtr.owner, emptyList())
+            PrimitiveBinaryType.VECTOR16 -> IrConstImpl.double(startOffset, endOffset, type, 0.0) // FIXME
```
SvyatoslavScherbina: It makes sense to add a test for this. Something like
```kotlin
fun foo(v: NativeVector = vectorOf(1.0f, 2.0f, 3.0f, 4.0f)) = v
foo()
```
should cover this line.

samples/simd/simd_lib.c
SvyatoslavScherbina: > Better use SIMD types from platform APIs in uikit or objc samples.

This would clutter these samples with unrelated SIMD types.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -1026,6 +1026,7 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
             KonanPrimitiveType.FLOAT -> "${prefix}_KFloat"
             KonanPrimitiveType.DOUBLE -> "${prefix}_KDouble"
             KonanPrimitiveType.NON_NULL_NATIVE_PTR -> "void*"
+            KonanPrimitiveType.VECTOR16 -> "${prefix}_KVector"
```
SvyatoslavScherbina: This is likely not enough, `*_KVector` type should also be defined.
Please add a test for this (see https://github.com/JetBrains/kotlin-native/tree/e1b3fefd3333fa730bb00174b7eb24c90822f6a0/backend.native/tests/produce_dynamic/simple).

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -499,6 +499,19 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
                 spelling = clang_getTypeSpelling(type).convertAndDispose().dropConstQualifier()
         )
 
+        CXType_Vector -> {
+            val elementType = convertType(clang_getElementType(type))
+            val size = clang_Type_getSizeOf(type)
+            val elemSize = clang_Type_getSizeOf(clang_getElementType(type))
+            val length = size / elemSize
```
SvyatoslavScherbina: Maybe add some checks that `size` is a multiple of `elemSize`?

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,20 @@ external fun CPointer<*>.getRawValue(): NativePtr
 
 internal fun CPointer<*>.cPointerToString() = "CPointer(raw=$rawValue)"
 
+
+public final class NativeVector private constructor()
```
SvyatoslavScherbina: This class should also have `equals`/`hashCode`/`toString` implemented (and tested).
`==` equality should be supported too, see https://github.com/JetBrains/kotlin-native/blob/e1b3fefd3333fa730bb00174b7eb24c90822f6a0/runtime/src/main/kotlin/kotlin/native/internal/Intrinsics.kt#L22

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,20 @@ external fun CPointer<*>.getRawValue(): NativePtr
...
+
+public class VectorVarOf<T : NativeVector>(rawPtr: NativePtr) : CVariable(rawPtr)
```
SvyatoslavScherbina: Generally names of `*Var(Of)` types should be prefixed with names of corresponding value classes, e.g. `NativeVectorVarOf`.

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Types.kt
```diff
@@ -470,4 +470,5 @@ public typealias CArrayPointerVar<T> = CPointerVar<T>
 /**
  * The C function.
  */
-public class CFunction<T : Function<*>>(rawPtr: NativePtr) : CPointed(rawPtr)
\ No newline at end of file
+public class CFunction<T : Function<*>>(rawPtr: NativePtr) : CPointed(rawPtr)
+
```
SvyatoslavScherbina: Redundant line.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,20 @@ external fun CPointer<*>.getRawValue(): NativePtr
...
+
+public typealias VectorVar = VectorVarOf<NativeVector>
```
SvyatoslavScherbina: Should also be included here:
https://github.com/JetBrains/kotlin-native/blob/e1b3fefd3333fa730bb00174b7eb24c90822f6a0/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrType.kt#L90
cc @sbogolepov

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/Mappings.kt
```diff
@@ -75,6 +75,9 @@ fun PrimitiveType.getKotlinType(declarationMapper: DeclarationMapper): KotlinCla
         else -> TODO(this.toString())
     }
 
+    is VectorType -> {
+        KotlinTypes.nativeVector
```
SvyatoslavScherbina: It makes sense to add an assertion checking that this is 16-byte vector.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/SimpleBridgeGeneratorImpl.kt
```diff
@@ -61,6 +62,7 @@ class SimpleBridgeGeneratorImpl(
             BridgedType.ULONG -> "uint64_t"
             BridgedType.FLOAT -> "float"
             BridgedType.DOUBLE -> "double"
+            BridgedType.NATIVE_VECTOR -> "NOT USED???" // "float __attribute__ ((__vector_size__ (16)))"
```
SvyatoslavScherbina: If it is actually not used, then this should throw an exception instead.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,20 @@ external fun CPointer<*>.getRawValue(): NativePtr
...
+
+public var <T : NativeVector> VectorVarOf<T>.value: T
+    get() = nativeMemUtils.getVector(this) as T
+    set(value) = nativeMemUtils.putVector(this, value)
+
+@SymbolName("Kotlin_Vector_of")
+external fun vectorOf(f0: Float, f1: Float, f2: Float, f3: Float): NativeVector
```
SvyatoslavScherbina: Element accessors are likely required too.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
sbogolepov: Correct. Please note, that the name of `typealias` should changed be `NativeVectorVar` because this code relies on such naming convention.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
olonho: Only getters, I presume.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -81,6 +81,16 @@ public final class NativeVector private constructor() {
 
     @TypedIntrinsic(IntrinsicType.EXTRACT_ELEMENT)
     external fun getULong(index: Int): ULong
+
+    public override fun toString() = "(${getFloat(0)}, ${getFloat(1)}, ${getFloat(2)}, ${getFloat(3)})"
+
+    // Not as good for floating types
+    public fun equals(other: NativeVector): Boolean =
```
olonho: `NativeVector128`?

samples/simd/simd_lib.c
```diff
@@ -68,7 +62,7 @@ int main(void) {
 	printVFloat(ceil);
 
 	vFloat f2 = {1, 3.162, 10, 31};
-	vFloat lg = vlog10f(f2);
+//	vFloat lg = vlog10f(f2);
```
olonho: ?

samples/simd/simd_lib.h
```diff
@@ -10,25 +10,14 @@
...
-//typedef __attribute__((__ext_vector_type__(2))) float simd_float2;
 typedef __attribute__((__ext_vector_type__(2))) float simd_float2;
-//typedef __attribute__((__ext_vector_type__(2))) const float simd_float2_c;
-simd_float2 sf2;
-
-typedef const float CFloat;
-typedef CFloat __attribute__ ((__vector_size__ (16))) CFloatV;
-CFloatV const CFloat_var = {1., 3.162, 1., 31.};
-const float __attribute__ ((__vector_size__ (16)))    CFloat_var2  = {1., 3.162, 1., 31.};
+simd_float2 sf2;  // not supported
```
olonho: You mean top level globals are not supported?

samples/simd/simd_test.kt
```diff
@@ -3,37 +3,27 @@ import kotlin.test.*
...
-//    val dBox: Box<Double> = Box<Double>(21.0)
-////    println(dBox.value)
-//    val v = vectorOf(42f, 1f, 2f, 3f)
-//    val box: Box<NativeVector> = Box<NativeVector>(v)
-////    println(box.value)
-////    printVFloat(box.value)
-//}
 
-fun testOOB() {
+fun test_OOB() {
```
olonho: `assertFailsWith<ArrayIndexOutOfBoundsException> {}`

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -504,16 +504,16 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
...
+            assert(elemSize * elementCount == size)
 
             // Spelling example: `__attribute__((__vector_size__(4 * sizeof(float)))) const float`
             // Re-generate spelling removing constness and typedefs to limit number of variants for bridge generator
             // Supposed to be the same (i.e. natively compatible) as clang_getTypeSpelling(type) aka type.name
             val spelling = "__attribute__((__vector_size__($size))) ${clang_getCanonicalType(elementCXType).name}"
 
             if (size == 16L) {
-                VectorType(elementType, length, spelling)
+                VectorType(elementType, elementCount.toInt(), spelling)
```
olonho: `VectorType128`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
```diff
@@ -611,6 +631,14 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
         }
     }
 
+    private fun FunctionGenerationContext.emitThrowIfOOB(index: LLVMValueRef, size: LLVMValueRef) {
+        ifThen(icmpUGe(index, size)) {
+            val throwArithExc = codegen.llvmFunction(context.ir.symbols.throwIndexOutOfBoundsException.owner)
```
olonho: `throwIndexOutOfBoundsException`, no abbreviations.

samples/simd/simd_test.kt
```diff
@@ -40,9 +66,19 @@ fun main() {
...
+    println(f2.getByte(0))
+    println(f2.getUByte(1))
+    println(f2.getUByte(2))
+    println(f2.getByte(3))
+
     val lg = vlog10f(f2)
     printVFloat(lg)
 //    runTest()
 
+    test_Accelerate()
```
olonho: Inconsistent test naming.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/Mappings.kt
olonho: And better mention that it's 128 bits in the name.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/SimpleBridgeGeneratorImpl.kt
```diff
@@ -62,7 +62,7 @@ class SimpleBridgeGeneratorImpl(
             BridgedType.ULONG -> "uint64_t"
             BridgedType.FLOAT -> "float"
             BridgedType.DOUBLE -> "double"
-            BridgedType.NATIVE_VECTOR -> "NOT USED???" // "float __attribute__ ((__vector_size__ (16)))"
+            BridgedType.NATIVE_VECTOR -> TODO() // "float __attribute__ ((__vector_size__ (16)))"
```
olonho: Here too.

samples/simd/simd_wrapper.def
```diff
@@ -0,0 +1,4 @@
+headers = simd_lib.h
+headerFilter = simd_lib.h
+#language = C
```
olonho: No commented out code.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -499,6 +499,26 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
...
+            val elementType = convertType(elementCXType)
+            val size = clang_Type_getSizeOf(type)
+            val elemSize = clang_Type_getSizeOf(elementCXType)
+            val elementCount = size / elemSize
+            assert(elemSize * elementCount == size)
+
+            // Spelling example: `__attribute__((__vector_size__(4 * sizeof(float)))) const float`
+            // Re-generate spelling removing constness and typedefs to limit number of variants for bridge generator
+            // Supposed to be the same (i.e. natively compatible) as clang_getTypeSpelling(type) aka type.name
+            val spelling = "__attribute__((__vector_size__($size))) ${clang_getCanonicalType(elementCXType).name}"
```
SvyatoslavScherbina: Does `clang_getCanonicalType` remove `const` modifier?
Is there a test that ensures it does?

samples/simd/simd_lib.h
knebekaizer: Nope. The canonical type, which is gcc extension  __ext_vector_size__, is not supported. Clang provides it with kind = CXType_Unexposed instead of CXType_Vector. This should be investigated further, because clang actually supports this extension internally.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
knebekaizer: dropConstQualifier won't do the trick, but we can generate spelling from the components

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
knebekaizer: Done.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/SimpleBridgeGeneratorImpl.kt
knebekaizer: Done. Dead code which can't be easily removed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
knebekaizer: Done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmUtils.kt
knebekaizer: Done.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
knebekaizer: clang_getCanonicalType unwrap typedefs only, never remove qualifiers
Const (as well as volatile) in this context is a qualifier of the vector itself. Element type seems never having qualifiers. Even when the element type is declared with typedef the canonical type of element does not have qualifier. Makes sense, because elements can't have different qualifiers, so it is always propagated to the vector itself.
AFAIK this is not ruled by the C standard, and probably is implementation specific of clang parser.

samples/simd/simd_lib.c
olonho: I mean use SIMD types from Apple APIs.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,59 @@ external fun CPointer<*>.getRawValue(): NativePtr
...
+    @TypedIntrinsic(IntrinsicType.EXTRACT_ELEMENT)
+    external fun getUByte(index: Int): UByte
+
+    @TypedIntrinsic(IntrinsicType.EXTRACT_ELEMENT)
+    external fun getUInt(index: Int): UInt
+
+    @TypedIntrinsic(IntrinsicType.EXTRACT_ELEMENT)
+    external fun getULong(index: Int): ULong
+
+    public override fun toString() = "(${getFloat(0)}, ${getFloat(1)}, ${getFloat(2)}, ${getFloat(3)})"
```
olonho: `getInt().toString(16)`

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
olonho: Also `vectorOf(i0: Int, i1: Int, i2: Int, i3: Int): NativeVector128`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -89,7 +89,9 @@ private fun ContextUtils.createClassBodyType(name: String, fields: List<IrField>
 
     val classType = LLVMStructCreateNamed(LLVMGetModuleContext(context.llvmModule), name)!!
 
-    LLVMStructSetBody(classType, fieldTypes.toCValues(), fieldTypes.size, 0)
+    val hasBigAlignment = fields.any { LLVMABIAlignmentOfType(context.llvm.runtime.targetData, getLLVMType(it.type)) > 8 }
```
SvyatoslavScherbina: It makes sense to add a brief comment describing this decision.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
```diff
@@ -486,11 +491,15 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
...
             llvm.LLVMTypeKind.LLVMFloatTypeKind, llvm.LLVMTypeKind.LLVMDoubleTypeKind -> {
-                val numBits = llvm.LLVMSizeOfTypeInBits(codegen.llvmTargetData, first.type).toInt()
-                val integerType = LLVMIntTypeInContext(llvmContext, numBits)!!
+                val integerType = LLVMIntTypeInContext(llvmContext, first.type.sizeInBits())!!
                 icmpEq(bitcast(integerType, first), bitcast(integerType, second))
             }
             llvm.LLVMTypeKind.LLVMIntegerTypeKind, llvm.LLVMTypeKind.LLVMPointerTypeKind -> icmpEq(first, second)
+            LLVMTypeKind.LLVMVectorTypeKind -> {
+                assert(first.type.sizeInBits() == 128)
+                val integerType = LLVMInt128TypeInContext(llvmContext)!!
```
SvyatoslavScherbina: Isn't `LLVMInt128TypeInContext(llvmContext)` the same as `LLVMIntTypeInContext(llvmContext, 128)`?
If it is, why not just add `LLVMTypeKind.LLVMVectorTypeKind` to the first branch?

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
```diff
@@ -56,6 +56,59 @@ external fun CPointer<*>.getRawValue(): NativePtr
...
+    public fun equals(other: NativeVector): Boolean =
+            getLong(0) == other.getLong(0) && getLong(1) == other.getLong(1)
+
+    public override fun equals(other: Any?): Boolean =
+            other is NativeVector && this.equals(other)
+
+    override fun hashCode(): Int {
+        val x0 = getLong(0)
+        val x1 = getLong(1)
+        return 31 * (x0 xor (x0 shr 32)).toInt() + (x1 xor (x1 shr 32)).toInt()
```
SvyatoslavScherbina: Replacing `(x0 xor (x0 shr 32)).toInt()` by `x0.hashCode()` (which is almost the same) would make this code simpler.

samples/simd/simd_test.kt
```diff
@@ -1,87 +0,0 @@
-import kotlinx.cinterop.*
```
SvyatoslavScherbina: I guess we still need some tests.
Considering that `NativeVector` becomes first-class type with "intrinsified" implementation, I guess we need 
both 
1. Interop-less tests for `NativeVector` type itself 
2. Interop-specific tests for passing vectors between Kotlin and C.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeTypes.kt
SvyatoslavScherbina: I believe float-specific representation is quite usable too. Can we include both representations to `toString()` result?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/KonanDefaultParameterInjector.kt
SvyatoslavScherbina: (Apparently `type` can't be `NativeVector` anymore since https://github.com/JetBrains/kotlin/commit/924fa6f9d3d16ef8e615ca639ba97d5646493aaa#diff-02427743f847bc86578ed06272b961f7R471, so we've decided to add such a test and throw an exception here).

SvyatoslavScherbina:
Note that passing struct with vector fields should be explicitly supported [in StructRendering.kt](https://github.com/JetBrains/kotlin-native/blob/master/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StructRendering.kt#L65).

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
```diff
@@ -187,6 +187,7 @@ object KotlinTypes {
     val map by CollectionClassifier
 
     val nativePtr by InteropType
+    val vector128 by InteropType
```
SvyatoslavScherbina: `Vector128` type is not in this (`kotlinx.cinterop`) package anymore.

runtime/src/main/kotlin/kotlin/native/simd.kt
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+import kotlin.native.internal.*
+import kotlinx.cinterop.*
```
SvyatoslavScherbina: Is this import used?
Importing `kotlinx.cinterop.*` to `stdlib` is kinda wrong.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
```diff
@@ -485,9 +490,10 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
         assert (first.type == second.type) { "Types are different: '${llvmtype2string(first.type)}' and '${llvmtype2string(second.type)}'" }
 
         return when (val typeKind = LLVMGetTypeKind(first.type)) {
-            llvm.LLVMTypeKind.LLVMFloatTypeKind, llvm.LLVMTypeKind.LLVMDoubleTypeKind -> {
-                val numBits = llvm.LLVMSizeOfTypeInBits(codegen.llvmTargetData, first.type).toInt()
-                val integerType = LLVMIntTypeInContext(llvmContext, numBits)!!
+            llvm.LLVMTypeKind.LLVMFloatTypeKind, llvm.LLVMTypeKind.LLVMDoubleTypeKind,
+            LLVMTypeKind.LLVMVectorTypeKind -> {
+                // TODO LLVMIntTypeInContext works for 128 bits but may return null for longer types. Provide meaningful diag message instead of NPE
```
SvyatoslavScherbina: I'm not sure `LLVMIntTypeInContext` it can actually return `null`. According to its source code, it never does.

olonho:
Have you tested it on some ARM32 machine, such as Raspberry Pi or older iPhone?

sbogolepov:
Did you check `objc_msgSend*`? See [tests](https://github.com/JetBrains/kotlin-native/tree/master/backend.native/tests/interop/objc/msg_send) for the reference.

knebekaizer:
> Have you tested it on some ARM32 machine, such as Raspberry Pi or older iPhone?
Do you mean LLVMIntTypeInContext, or LLVMVectorType itself? Nope. LLVM implementation is platform agnostic and the only limit is NumBits < (1<<24). When backend compiler (platform specific) may fail for some platform it will not impact LLVM type availability at this point. So we have to trust in LLVM conceptual consistency: if a platform does not have floating point numbers, or int128_t, or vector type - whatever, it's up to backend compiler to provide emulation layer for *all* LLVM types.

sbogolepov:
I think it's still useful to check these cases because LLVM may contain bugs.

olonho:
Just the overall code? It includes potential alignment issues, for example.

libclangext/src/main/cpp/ExtVector.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <cassert>
+
+#include <clang/AST/Type.h>
+#include <clang-c/Index.h>
+
+
+using namespace clang;
+
+
+static inline QualType GetQualType(CXType CT) {
```
SvyatoslavScherbina: Why not reuse `unwrapCXType` from ClangExt.cpp instead?

libclangext/src/main/cpp/ExtVector.cpp
```diff
@@ -0,0 +1,42 @@
...
+  return QualType::getFromOpaquePtr(CT.data[0]);
+}
+
+static inline CXTranslationUnit GetTU(CXType CT) {
```
SvyatoslavScherbina: Why not reuse `getTranslationUnit` from ClangExt.cpp instead?

SvyatoslavScherbina: Also it seems unused.

backend.native/tests/runtime/basic/simd.kt
```diff
@@ -0,0 +1,77 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package runtime.basic.simd
+
+import kotlin.test.*
+
+
+@Test fun runTest() {
+    testSetGet()
```
SvyatoslavScherbina: It makes sense to add `@Test` to all these functions instead of calling them directly.

backend.native/tests/runtime/basic/simd.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    // type punning: set the variable to another runtime type
+    val a = arrayOf(1f, 3.162f, 10f, 31f)
+    v4any = vectorOf(a[0], a[1], a[2], a[3])
+    (0 until 4).forEach { assertEquals(a[it], v4any.getFloatAt(it)) }
+}
+
+fun testString() {
+    val v4i = vectorOf(100, 1024, Int.MAX_VALUE, Int.MIN_VALUE)
+    assertEquals("(0x64, 0x400, 0x7fffffff, 0x-80000000)", v4i.toString())
```
SvyatoslavScherbina: `0x-80000000` is somewhat incorrect.

backend.native/tests/runtime/basic/simd.kt
```diff
@@ -0,0 +1,77 @@
...
+}
+
+fun testOOB() {
+    val f = vectorOf(1f, 3.162f, 10f, 31f)
+    f.getByteAt(15)
+    var result = "PASSED"
+    try {
+        println("getFloatAt(4) should fail: ${f.getFloatAt(4)}")
+        result = "FAILED"
+    } catch (e: IndexOutOfBoundsException) {}
```
SvyatoslavScherbina: This may help to improve the code here: https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html

backend.native/tests/runtime/basic/simd.kt
```diff
@@ -0,0 +1,77 @@
...
+        println(f.getFloatAt(4))
+        result = "FAILED"
+    } catch (e: IndexOutOfBoundsException) {}
+    assertEquals("PASSED", result)
+}
+
+fun testEquals() {
+    var v1 = vectorOf(-1f, 0f, 0f, -7f)
+    var v2 = vectorOf(1f, 4f, 3f, 7f)
+    assertEquals(false, v1 == v2)
```
SvyatoslavScherbina: `assertNotEquals(v1, v2)`.

backend.native/tests/runtime/basic/simd.kt
```diff
@@ -0,0 +1,77 @@
...
+    val h1 = vectorOf(1f, 4f, 3f, 7f).hashCode()
+    val h2 = vectorOf(3f, 7f, 1f, 4f).hashCode()
+    assertEquals(false, h1 == h2)
+
+    val i = 654687
+    assertEquals(true, vectorOf(0, 0, i, 0).hashCode() == i.hashCode())  // exploit little endianness
+    assertEquals(true, vectorOf(1, 0, 0, 0).hashCode() == vectorOf(0, 0, 31, 0).hashCode())
+}
+
+fun testDefaultValue(v: Vector128 = vectorOf(1.0f, 2.0f, 3.0f, 4.0f)) = v
```
SvyatoslavScherbina: This function is not called.

backend.native/tests/runtime/basic/simd.kt
knebekaizer: Looks ugly indeed. But this is what Int.toString(16) does. Let's have it fixed first as the assertion is aligned with that behavior.

Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -522,7 +522,7 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
             val size = clang_Type_getSizeOf(type)
             val elemSize = clang_Type_getSizeOf(elementCXType)
             val elementCount = clang_getNumElements(type)
-            assert(elemSize * elementCount == size)
+            assert(size >= elemSize * elementCount && size % elemSize == 0L)
```
SvyatoslavScherbina: Could you provide an example violating previous version of assertion?

knebekaizer: simd_float3 and similar has internal padding and sizeof == 16

libclangext/src/main/cpp/ExtVector.cpp
knebekaizer: Calling getFromOpaquePtr direrctly would be even simpler. However, the intention is to make copy-paste in clang_isExtVectorType to be easily comparable with prototype in Type.cpp. So I keep a snippet intact,  including naming.
Second: unwrapCXType declared as static, which implies two options: either refactor cpp, h and build script to reuse declaration; or add content to ClangExt.cpp. Refactoring efforts are hardly paid back in this case and both options contradict the intention to remove a kludge as soon as libclang is upgraded to 59 version (clang 9).

libclangext/src/main/cpp/ExtVector.cpp
knebekaizer: Removed.

knebekaizer:
Rebased manually because of a lot of changes

backend.native/tests/runtime/basic/simd.kt
knebekaizer: Following the dominating style of other tests in the folder(s) in this case. Refactoring (modernize) the whole test suite "all at once" may be suggested.

SvyatoslavScherbina: The dominating style here is to have `main` function.
It doesn't make much sense to have `@Test` annotation but only on a single function.

SvyatoslavScherbina:
I guess remaining issues should be fixed with follow-up PRs.

samples/simd/simd.kt
```diff
@@ -0,0 +1,58 @@
+import kotlinx.cinterop.*
```
SvyatoslavScherbina: I guess build scripts are also required.

PR #3499, opened 2019-10-25
Fix Objective-C block to kotlin.Function* dynamic conversion

SvyatoslavScherbina:
after migrating to synthetic function classes.

### DIFF
--- a/runtime/src/main/cpp/ObjCExport.mm
+++ b/runtime/src/main/cpp/ObjCExport.mm
@@ -440,6 +440,9 @@ static const char* getBlockEncoding(id block) {
 
 // Note: replaced by compiler in appropriate compilation modes.
 __attribute__((weak)) convertReferenceFromObjC* Kotlin_ObjCExport_blockToFunctionConverters = nullptr;
+__attribute__((weak)) int Kotlin_ObjCExport_blockToFunctionConverters_size = 0;
+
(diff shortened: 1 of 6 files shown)


runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -440,6 +440,9 @@ static OBJ_GETTER(boxedBooleanToKotlinImp, NSNumber* self, SEL cmd) {
 
 // Note: replaced by compiler in appropriate compilation modes.
 __attribute__((weak)) convertReferenceFromObjC* Kotlin_ObjCExport_blockToFunctionConverters = nullptr;
+__attribute__((weak)) int Kotlin_ObjCExport_blockToFunctionConverters_size = 0;
+
+extern "C" id objc_retainBlock(id self);
```
olonho: Why block retention is different from objects?

SvyatoslavScherbina: There stack-allocated blocks, which are not exactly reference-counted. `objc_retain` on these blocks is no-op.

Stack-allocated blocks can be copied to heap, a copy is reference-counted and thus "escapable". `objc_retainBlock` on stack-allocated block performs such copying, and the result has RC=1; for heap-allocated blocks it simply increments RC.

PR #3500, opened 2019-10-25
[platformLibs] mingw platform.posix cleanup

msink:


platformLibs/src/platform/mingw/posix.def
```diff
@@ -1,13 +1,144 @@
...
+    vadefs.h \
+    wchar.h \
+    wctype.h
+excludedFunctions = __WSAFDIsSet
 noStringConversion = send sendto recv
-linkerOpts = -lWs2_32
+linkerOpts = -lws2_32
 
 ---
+WINSOCK_API_LINKAGE SOCKET WSAAPI accept(SOCKET s,struct sockaddr *addr,int *addrlen);
```
olonho: Why so?

msink:
Because it's much less ugly way to include `winsock2` to parser but exclude from export.

msink:
Currently windows-specific parts of `winsock2` still leak ko `platform.posix` - `WSA...` functions and structures, `UUID` definition.
After this path they moved to `platform.windows`, as expected.

msink:
POSIX-compatible function definitions were copy-pasted from `winsock2.h` because  POSIX API is stable, really they are the same in `winsock` and `winsock2`, and unlikely will ever changed.

PR #3502, opened 2019-10-26
[platformLibs] mingw windows cleanup

msink:


msink:
`windows.kt` size already reduced from 10MB to 5MB

platformLibs/src/platform/mingw/winscard.def
```diff
@@ -0,0 +1,15 @@
+package = platform.winscard
+depends = windows
```
micolous: Add: `noStringConversion = SCardListReaders`

This returns a `msz` (multi-string) which is a list of null terminated strings, then terminated by a null.

platformLibs/src/platform/mingw/winscard.def
```diff
@@ -0,0 +1,15 @@
+package = platform.winscard
```
micolous: These _all_ should be `platform.windows.winscard` or similar.

Despite the name, `winscard.h` exists on non-Windows platforms, and it has [subtle differences between Linux (which uses pcsclite) and macOS (which is based on a modified version of pcsclite) APIs](https://pcsclite.apdu.fr/api/group__API.html#details).

[It needs quite significant hand-holding to fix that...](https://github.com/micolous/kotlin-pcsc/)

micolous:
Another alternative to this is include stuff that is actually tested, and exclude anything that is not. 

The `SCardListReaders` issue (this returns a multi-string, but Kotlin tries String conversion on it) is one of many such methods in the Win32 API, and I feel like one is better off explicitly _not_ declaring these unless it is known to work.

platformLibs/src/platform/mingw/winscard.def
msink: Then same should be for `SCardLocateCards`.
Done.

platformLibs/src/platform/mingw/winscard.def
msink: I'm not sure.
There always can be subtle differnces, even on `mingw_x86` (32-bit) and `mingw_x64` (64-bit) platforms.

msink:
I don't know any, this is relatively "conservarive" change - 5MB in `windows.kt`
It can potentially be reduced to 2.2MB if define `WIN32_LEAN_AND_MEAN`

platformLibs/src/platform/mingw/posix.def
```diff
@@ -1,11 +1,114 @@
 package = platform.posix
-headers = assert.h complex.h ctype.h dirent.h errno.h fcntl.h \
-    fenv.h float.h ftw.h getopt.h inttypes.h libgen.h limits.h \
-    locale.h math.h memory.h pthread.h sched.h search.h semaphore.h \
-    setjmp.h signal.h stdint.h stdio.h stdlib.h string.h \
-    time.h uchar.h unistd.h utime.h wchar.h wctype.h
+
```
olonho: We likely need matching compiler options with other headers.

msink: You mean same as in `windows.def`?
```
compilerOpts = \
    -DUNICODE \
    -DWINVER=0x0601 \
    -D_WIN32_WINNT=0x0601 \
    -DWINAPI_FAMILY=3 \
    -DOEMRESOURCE \
    -Wno-incompatible-pointer-types \
    -Wno-deprecated-declarations
```
Tried it and recompile - almost no change in `posix.kt`
``` diff
diff -ur a/posix.kt b/posix.kt
--- a/posix.kt	2019-10-30 12:58:39.175066200 +0500
+++ b/posix.kt	2019-10-30 13:48:24.090793700 +0500
@@ -6512,8 +6512,6 @@
 
 const val __MSVCRT_VERSION__: Int = 1792
 
-const val _WIN32_WINNT: Int = 1537
-
 const val MINGW_HAS_SECURE_API: Int = 1
 
 const val __STDC_SECURE_LIB__: Int = 200411
@@ -8062,9 +8060,9 @@
 
 const val CLOCK_THREAD_CPUTIME_ID: Int = 3
 
-const val _INC_CRT_UNICODE_MACROS: Int = 2
+const val _INC_CRT_UNICODE_MACROS: Int = 1
 
-val __MINGW_PROCNAMEEXT_AW: String get() = "A"
+val __MINGW_PROCNAMEEXT_AW: String get() = "W"
 
 const val INVALID_SOCKET: SOCKET = 18446744073709551615u
 
@@ -8314,7 +8312,7 @@
 
 const val SO_PROTOCOL_INFOW: Int = 8197
 
-const val SO_PROTOCOL_INFO: Int = 8196
+const val SO_PROTOCOL_INFO: Int = 8197
 
 const val PVD_CONFIG: Int = 12289
 
```

msink:
Maybe better do a little less "safe" decomposition, but with `windows.kt` size 2.2 MBytes?
Because in my experience 5 MBytes is still too big for IDEA autocompletion.

PR #3507, opened 2019-10-28
Improve ObjC block dynamic conversion generation when using cache

SvyatoslavScherbina:
Generate conversions along with stdlib to guarantee all synthesized
function classes get covered.

### DIFF
--- a/runtime/src/main/cpp/ObjCExport.mm
+++ b/runtime/src/main/cpp/ObjCExport.mm
@@ -438,9 +438,9 @@ static const char* getBlockEncoding(id block) {
       reinterpret_cast<struct Block_descriptor_1_without_helpers*>(literal->descriptor)->signature;
 }
 
-// Note: replaced by compiler in appropriate compilation modes.
-__attribute__((weak)) convertReferenceFromObjC* Kotlin_ObjCExport_blockToFunctionConverters = nullptr;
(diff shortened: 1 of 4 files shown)


runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -438,9 +438,9 @@ static OBJ_GETTER(boxedBooleanToKotlinImp, NSNumber* self, SEL cmd) {
       reinterpret_cast<struct Block_descriptor_1_without_helpers*>(literal->descriptor)->signature;
 }
 
-// Note: replaced by compiler in appropriate compilation modes.
-__attribute__((weak)) convertReferenceFromObjC* Kotlin_ObjCExport_blockToFunctionConverters = nullptr;
-__attribute__((weak)) int Kotlin_ObjCExport_blockToFunctionConverters_size = 0;
+// Note: defined by compiler.
+extern "C" convertReferenceFromObjC* Kotlin_ObjCExport_blockToFunctionConverters;
```
olonho: Maybe put all `extern "C"` in single clause?

SvyatoslavScherbina: Unfortunately
```c++
extern "C" {
    int x;
}
```
is not the same as
```c++
extern "C" int x;
```
in C++.

PR #3520, opened 2019-10-30
Mingw platform.windows cleanup (try #2)

msink:


olonho:
Great job, but we need somehow insure that it works. What tests it got?

PR #3532, opened 2019-11-01
Get free options from binaries

LepilkinaElena:


build-tools/src/main/kotlin/org/jetbrains/kotlin/benchmark/KotlinNativeBenchmarkingPlugin.kt
```diff
@@ -109,6 +109,9 @@ open class KotlinNativeBenchmarkingPlugin: BenchmarkingPlugin() {
         project.configureJVMTarget()
     }
 
+    override fun getCompilerFlags(project: Project, nativeTarget: KotlinNativeTarget) =
```
ilmat192: What about framework args in the Swift benchmarking plugin?

PR #3533, opened 2019-11-01
Prepare interop driver for a metadata-only mode.

sbogolepov:


utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/InteropCompiler.kt
```diff
@@ -69,6 +69,10 @@ fun invokeInterop(flavor: String, args: Array<String>): Array<String> {
     }
 
     val cinteropArgsToCompiler = interop(flavor, args + additionalArgs, additionalProperties)
+    // There is no need in compiler invocation if we're generating only metadata.
```
SvyatoslavScherbina: Why not return `null` from `interop(...)` above instead?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -270,31 +278,48 @@ private fun processCLib(args: Array<String>, additionalArgs: Map<String, Any> =
...
+            val compilerCmd = arrayOf(compiler, *compilerArgs,
+                    "-c", outCFile.absolutePath, "-o", outOFile.absolutePath)
+            runCmd(compilerCmd, verbose)
+
+            val outLib = File(nativeLibsDir, System.mapLibraryName(libName))
+            val linkerCmd = arrayOf(linker,
+                    outOFile.absolutePath, "-shared", "-o", outLib.absolutePath,
+                    *linkerOpts)
+            runCmd(linkerCmd, verbose)
+            outOFile.absolutePath
```
SvyatoslavScherbina: Shouldn't it be more like `outLib.absolutePath`?

sbogolepov: Yeah, it’s a typo.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/GenerationMode.kt
```diff
@@ -0,0 +1,13 @@
+package org.jetbrains.kotlin.native.interop.gen.jvm
```
olonho: Copyright.

PR #3535, opened 2019-11-01
[platformLibs] moved bunch of headers to platform.posix

msink:
from POSIX standard  (IEEE 1003.1)


SvyatoslavScherbina:
@msink Thank you for your contributions! I'm sorry we couldn't accept them and failed to communicate this.

The PRs (#3500, #3520 and this one) address important issues -- the current state of platform libraries is totally a mess.
And, I believe, you put a lot of effort into fixing this here.

Unfortunately, changing anything in platform libs is complicated:
* We have to keep source and binary compatibility, and don't have any tools that would fix or mitigate the related consequences of these changes.
* We don't have proper test infrastructure for platform libs, so we have to stick to keeping this as is, just to avoid possible regressions.

PR #3538, opened 2019-11-02
User-side library generator.

olonho:

PR #3547, opened 2019-11-05
https://github.com/JetBrains/kotlin-native/pull/3538 followup.

olonho:


build.gradle
```diff
@@ -356,7 +356,8 @@ task distDef(type: Copy) {
     destinationDir distDir
 
     from(project("platformLibs").file("src/platform")) {
-        into('klib/platformDef')
+        into('konan/platformDef')
+        exclude '**/*.sh'
```
SvyatoslavScherbina: Why not `include '**/*.def'` instead?

build.gradle
```diff
@@ -461,6 +462,7 @@ task crossDist {
 }
 
 task bundle(type: (isWindows()) ? Zip : Tar) {
+    dependsOn("distDef")
```
SvyatoslavScherbina: This is likely required for both `bundle` and `bundleRestricted`.

utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/GeneratePlatformLibraries.kt
```diff
@@ -83,6 +83,7 @@ private fun generatePlatformLibraries(target: String, inputDirectory: File, outp
         if (!saveTemps)  {
             File("$outputDirectory/build-${def.name}").deleteRecursively()
         }
+        File("$outputDirectory/clangModulesCache").deleteRecursively()
```
SvyatoslavScherbina: The idea behind module cache here was to reuse it for all platform libraries.

olonho: Yeah, moved removal below, PTAL.

PR #3554, opened 2019-11-07
Catch exception in MSVC embedding scenario.

olonho:


runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -291,16 +299,32 @@ static void KonanTerminateHandler() {
   }
 }
 
+#if KONAN_WINDOWS
+#include <windows.h>
+
+LONG WINAPI KonanUnhandledExceptionFilter(EXCEPTION_POINTERS* ExceptionInfo) {
+  if (currentException != nullptr) {
+    auto e = currentException;
```
SvyatoslavScherbina: What happens if the following sequence of events happens?
1.  Kotlin exception is thrown and caught.
2.  Non-Kotlin exception is thrown and not caught, causing `KonanUnhandledExceptionFilter` to be called.

benasher44:
I remember we discussed how you weren't happy with this solution, so you're going to think on it.

However, I remember we discussed how this could be implemented on our own for now as a workaround. How is that exactly again? I'm looking at this code, and I can only remember the way that involves adding `try { } catch () {}` to all of our entry points 😬

benasher44:
I spoke with @josephdwyer about it, and we're a bit lost on how to implement something like this without another hook inside of `ThrowException`. Can you provide a hint on how to get this working without a patch to K/N? 🙏

SvyatoslavScherbina:
Abandoned.

PR #3555, opened 2019-11-07
Fix isolation between dynamic Objective-C classes in multiple frameworks

SvyatoslavScherbina:
Also add test and improve handling duplicates for `@ExportObjCClass`.

### DIFF
--- a/backend.native/tests/framework/multiple/multiple.swift
+++ b/backend.native/tests/framework/multiple/multiple.swift
@@ -25,7 +25,7 @@ func testInteraction() throws {
     try assertEquals(actual: SecondKt.getFortyTwoFrom(i2: I1Impl()), expected: 42)
 }
 
-func testIsolation() throws {
+func testIsolation1() throws {
     try assertFalse(SecondKt.isUnit(obj: FirstKt.getUnit()))
 
     // Ensure frameworks don't share the same runtime (state):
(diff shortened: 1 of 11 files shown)


runtime/src/main/cpp/ObjCInterop.h
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+#ifndef RUNTIME_OBJCINTEROP_H
+#define RUNTIME_OBJCINTEROP_H
+
+#if KONAN_OBJC_INTEROP
+
+extern "C" const char* Kotlin_ObjCInterop_uniquePrefix;
```
olonho: Maybe function instead?

backend.native/tests/framework/multiple/multiple.swift
```diff
@@ -36,14 +36,31 @@ func testIsolation() throws {
     try assertTrue(Second.RuntimeState().consumeChange())
 }
 
+func testIsolation2() throws {
+    try assertEquals(actual: FirstKt.getI1().getFortyTwo(), expected: 42)
+    try assertEquals(actual: SecondKt.getI2().getFortyTwo(), expected: 42)
+}
+
+func testIsolation3() throws {
+/*
```
olonho: Maybe `#if 0`?

PR #3560, opened 2019-11-08
Yet another dedescriptorization step

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -77,6 +78,14 @@ internal fun IrSimpleFunction.resolveFakeOverride(allowAbstract: Boolean = false
     return realSupers.first { allowAbstract || it.modality != Modality.ABSTRACT }
 }
 
+internal fun IrSimpleFunction.overrides(superFunctionSymbol: IrSimpleFunctionSymbol): Boolean {
+    if (superFunctionSymbol in overriddenSymbols) return true
+    return overriddenSymbols.any { it.owner.overrides(superFunctionSymbol) }
```
SvyatoslavScherbina: The implementation is kinda exponential time in worst case.
I believe general utility method shouldn't be like this.
Its usage in `InteropLowering.kt` can be replaced by a simple name-based check, so we probably can avoid adding this method at all for now.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -254,6 +263,11 @@ fun IrConstructorCall.getStringValue(name: String): String {
     return (getValueArgument(parameter.index) as IrConst<String>).value
 }
 
+fun <T> IrConstructorCall.getArgumentValueOrNull(name: String): T? {
```
SvyatoslavScherbina: I guess the name of this function should clearly indicate that it is applicable to annotations only.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -839,11 +834,11 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
 
         expression.transformChildrenVoid(this)
         builder.at(expression)
-        val descriptor = expression.descriptor.original
         val function = expression.symbol.owner
 
-        if (descriptor == interop.nativePointedRawPtrGetter ||
-                OverridingUtil.overrides(descriptor, interop.nativePointedRawPtrGetter, false)) {
+        // TODO: this is fragile.
```
SvyatoslavScherbina: Yes it is.
What I meant is to check here that class derives from `NativePointed` and that this function is getter of property named `rawPtr`.
In this case this property overrides `NativePointed.rawPtr`.

Btw, it is possible that overriding check is not required anymore, after this patch: https://github.com/JetBrains/kotlin-native/commit/2b00850164d6aa5680248ac9ed7f3d35ec5ef1f9#diff-87031cea584971a666c72def5450f827R27

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -836,9 +837,8 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
         builder.at(expression)
         val function = expression.symbol.owner
 
-        // TODO: this is fragile.
-        if (function.isGetter && function.parent is IrClass
-                && function.name == symbols.interopNativePointedRawPtrGetter.owner.name) {
+        if ((function as? IrSimpleFunction)?.resolveFakeOverrideMaybeAbstract()?.symbol
+                == symbols.interopNativePointedGetRawPointer) {
```
SvyatoslavScherbina: `interopNativePointedRawPtrGetter`.

homuroll: 🤦‍♂

PR #3565, opened 2019-11-11
Avoid fake DCE root for irrelevant platforms.

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanIrFileSerializer.kt
```diff
@@ -4,18 +4,22 @@ import org.jetbrains.kotlin.backend.common.LoggingContext
...
 class KonanIrFileSerializer(
+        val target: KonanTarget,
         logger: LoggingContext,
         declarationTable: DeclarationTable,
         bodiesOnlyForInlines: Boolean = false
 ): IrFileSerializer(logger, declarationTable, bodiesOnlyForInlines) {
 
     override fun backendSpecificExplicitRoot(declaration: IrFunction) =
-            declaration.annotations.hasAnnotation(RuntimeNames.exportForCppRuntime)
+            if (declaration.neverRetainFor(target)) false else
```
alexander-gorshenev: That makes IR serialization different for differing targets. 
Currently `stdlib` shares a single IR representation between targets.
That would make it incorrect for some of the targets.
We need a bit more flexible solution. Let's discuss tomorrow.

alexander-gorshenev: The solution should better be platform (js, jvm, native) aware. 
Because between platforms there can be differences similar to those between targets.

PR #3566, opened 2019-11-12
[Coverage] Support conditional expressions

shadrina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/coverage/CoverageRegionCollector.kt
```diff
@@ -84,163 +83,54 @@ private class IrFunctionRegionsCollector(
...
+            val condition = it.condition
+            val result = it.result
 
-    private fun collectWhen(irWhen: IrWhen) {
-        irWhen.branches.forEach { branch ->
-            // Do not record location for else branch since it doesn't look correct.
-            if (branch.condition !is IrConst<*>) {
-                collectRegions(branch.condition)
+            if (it is IrElseBranch) {
+                val previous = branches[branches.lastIndex - 1]
```
sbogolepov:
```kotlin
fun main() {
    when {
        else -> println(":(")
    }
}
```

sbogolepov: Let's use `result.startOffset` instead.

shadrina: Added the sample to tests.

PR #3575, opened 2019-11-13
Added more control over starting point

kavanmevada:


olonho:
Please use pointer arithmetics instead.

kavanmevada:
> Please use pointer arithmetics instead.

I'm some methods like mmap returns COpaquePointer. Its really need to read only some portion of memory without doing lots of reinterpretation and extra stuff if method gives more flexibility.

kavanmevada:
btw I really liked COpaquePointer.readBytes save me from allocing in memScope lots of time.
It keep me in kotlin side, pointer arithmetic feels like still I'm writing in 80's.
I felt like it should also help others.

PR #3581, opened 2019-11-14
Avoid bitcast from/to function types returning simd to respect the ca…

knebekaizer:
…lling conventions difference

### DIFF
--- a/runtime/src/main/cpp/Types.cpp
+++ b/runtime/src/main/cpp/Types.cpp
@@ -97,8 +97,15 @@ KVector4f Kotlin_Vector4f_of(KFloat f0, KFloat f1, KFloat f2, KFloat f3) {
 	return {f0, f1, f2, f3};
 }
 
-KVector4i32 Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
-	return {f0, f1, f2, f3};
+/*
+ * In the current design all simd types are mapped internally to floating type, e.g. <4 x float>.
+ * However, some platforms (ex. arm32) have different calling convention for <4 x float> and <4 x i32>.
+ * To avoid illegal bitcast from/to function types the following function
+ * return type MUST be <4 x float> and explicit type cast is done on the variable type.
+ */
+KVector4f Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
+	KVector4i32 v4i = {f0, f1, f2, f3};
+	return (KVector4f)v4i;
 }
 
 }  // extern "C"
--- a/runtime/src/main/cpp/Types.cpp
+++ b/runtime/src/main/cpp/Types.cpp
@@ -104,7 +104,7 @@ KVector4f Kotlin_Vector4f_of(KFloat f0, KFloat f1, KFloat f2, KFloat f3) {
  * return type MUST be <4 x float> and explicit type cast is done on the variable type.
  */
 KVector4f Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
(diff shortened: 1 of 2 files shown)


runtime/src/main/cpp/Types.cpp
```diff
@@ -97,8 +97,15 @@ KVector4f Kotlin_Vector4f_of(KFloat f0, KFloat f1, KFloat f2, KFloat f3) {
...
-KVector4i32 Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
-	return {f0, f1, f2, f3};
+/*
+ * In the current design all simd types are mapped internally to floating type, e.g. <4 x float>.
+ * However, some platforms (ex. arm32) have different calling convention for <4 x float> and <4 x i32>.
+ * To avoid illegal bitcast from/to function types the following function
+ * return type MUST be <4 x float> and explicit type cast is done on the variable type.
+ */
+KVector4f Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
+	KVector4i32 v4i = {f0, f1, f2, f3};
```
olonho: Why not just make `Kotlin_Vector128_ofInt` and `Kotlin_Vector128_ofFloat`?

runtime/src/main/cpp/Types.cpp
```diff
@@ -97,8 +97,15 @@ KVector4f Kotlin_Vector4f_of(KFloat f0, KFloat f1, KFloat f2, KFloat f3) {
 	return {f0, f1, f2, f3};
 }
 
-KVector4i32 Kotlin_Vector4i32_of(KInt f0, KInt f1, KInt f2, KInt f3) {
-	return {f0, f1, f2, f3};
+/*
+ * In the current design all simd types are mapped internally to floating type, e.g. <4 x float>.
+ * However, some platforms (ex. arm32) have different calling convention for <4 x float> and <4 x i32>.
+ * To avoid illegal bitcast from/to function types the following function
+ * return type MUST be <4 x float> and explicit type cast is done on the variable type.
```
SvyatoslavScherbina: I guess the same issue should be reproducible with `cinterop` when calling a function having int vector in signature.

runtime/src/main/cpp/Types.cpp
knebekaizer: > get rid of KVector4i32 completely
What you mean? The only way to create vector of ints in C is: declare it as vector of ints (KVector4i32). Otherwise `KVector4f x = {i0, i1, i2, i3}` will cast int to floats at right hand side, which provides wrong result. Finally I shall cast result to <4 x float> and return.
The name `Vector4i32_of` is aligned with intention to create vector <4 x i32>, not <4 x float>. We supposed to get <4 x i32> in kotlin, then cast and store it in suitable storage type (which happens to be <4 x float>). However, LLVM code looks like:
```
%6 = invoke <4 x float> bitcast (<4 x i32> (i32, i32, i32, i32)* @Kotlin_Vector4i32_of to <4 x float> (i32, i32, i32, i32)*)(i32 1, i32 2, i32 3, i32 4)
```
which is not as expected ant seems to be a bug in intrinsic generation (bitcast shall be applied to result type, not to a function type).  The quick patch here is  a reasonable workaround for that. Let's consider intrinsic fix separately.

runtime/src/main/cpp/Types.cpp
knebekaizer: You're right. Looking at that.

runtime/src/main/cpp/Types.cpp
olonho: I mean just have single vector type (matching to KVector4f) and use in all the runtime (C++ and Kotlin) only such a type.

PR #3591, opened 2019-11-19
Run tests on a ios device

PavelPunegov:


backend.native/tests/build.gradle
```diff
@@ -3805,6 +3805,7 @@ if (isAppleTarget(project)) {
                     dependsOn ":${targetName}CrossDistRuntime", ':commonDistRuntime', ':distCompiler'
                 }
 
+                extraOpts "-Xembed-bitcode"
```
SvyatoslavScherbina: Why is this required?

PavelPunegov: It looks like it should be wether "-Xembed-bitcode" or "-Xembed-bitcode-marker" because swift code compiles with the same options. And dyld doesn't load these frameworks built without one of these options.

build-tools/src/main/kotlin/org/jetbrains/kotlin/KonanTestExecutable.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.kotlin
+
+import org.gradle.api.Action
+import org.gradle.api.Task
+
+interface KonanTestExecutable : Task {
+    val executable: String
+    var doBeforeRun: Action<in Task>?
+    var doBeforeBuild: Action<in Task>?
```
SvyatoslavScherbina: This is somewhat unreadable.
Do I understand correctly that its usages rely on some kind of Groovy/Gradle convention, which makes `doBeforeBuild { ... }` mean assigning this lambda to `doBeforeBuild`? If so, this definitely requires some comments added.

PavelPunegov: Yep, I'm going to add a doc to this. As soon as most tests are executed and built in different tasks, there should be a some kind of convention to add actions to be done before build or before the run but after the build. The order is the following:  `doBeforeBuild {} -> build test executable -> doBeforeRun { } -> run test executable`

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,238 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+    into(Paths.get(project.testOutputRoot, "launcher"))
+}
+
+fun KonanTestExecutable.xcodeBuild() {
+    this.doBeforeRun = Action {
+        val signIdentity = project.findProperty("sign_identity") as? String ?: "iPhone Developer"
+        val developmentTeam = project.findProperty("development_team") as? String
+        val xcProject = Paths.get(project.testOutputRoot, "launcher")
+
+        val shellScript: String = // language=Bash
```
sbogolepov: I think this code may be cleaner.
1. Use raw string literals to avoid escaping `"`.
2. `listOf()` instead of `it += `

PavelPunegov: 1. But it still required to escape `$` that escapes via `${"$"}` that isn't much better.
2. You still have to do `it += listOf()` in the when.  I'll join the first cp and += dSym part

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+    private val xcProject = Paths.get(project.testOutputRoot, "launcher")
+
+    private val idb = project.findProperty("idb_path") as? String ?: "idb"
+
+    private val deviceName = project.findProperty("device_name") as? String
+
+    override fun execute(action: Action<in ExecSpec>): ExecResult? {
+        val udid = targetUDID()
+        println("Found device UDID: $udid")
+        install(udid, "build/KonanTestLauncher.ipa")
```
ilmat192: Understanding a location of this `build` directory is complicated. May be specify the path more explicitly, say `xcProject.resolve("KonanTestLauncher/build/KonanTestLauncher.ipa")`?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+            execSpec.executable = "lldb"
+            execSpec.args = commands + "-b" + "-o" + "command script import ${pythonScript()}" +
+                    "-o" + ("process launch" +
+                        (execSpec.args.takeUnless { it.isEmpty() }
+                                ?.let { " -- ${it.joinToString(" ")}" }
+                                ?: "")) +
+                    "-o" + "get_exit_code" +
+                    "-k" + "get_exit_code" +
+                    "-k" + "exit -1"
+            savedOut = execSpec.standardOutput
```
ilmat192: Am I understand right that we save `execSpec.standardOutput` because a user may set it in the `action`? If yes, may be add some comment about it?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+            execSpec.standardOutput = out
+        }
+        out.toString()
+                .also { if (project.verboseTest) println(it) }
+                .split("\n")
+                .run { drop(indexOfFirst { s -> s.startsWith("(lldb) process launch") } + 1) }
```
ilmat192: Do we really need this `run { drop(...)}` here? Why not just `drop(...)`?

It seems that we also can replace `drop(indexOfFirst { ... } + 1)` with `dropWhile { !it.startsWith(...) }.drop(1)` but I'm not sure that it would be more readable.

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+                .run {
```
ilmat192: The same question about `run { ... }`

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+
+        project.exec {
+            it.workingDir = xcProject.toFile()
+            it.commandLine = listOf(idb, "uninstall", "--udid", udid, bundleId)
+            it.standardOutput = out
+            it.errorOutput = out
+            it.isIgnoreExitValue = true
+        }
+        println(out.toString())
+    }
```
ilmat192: Shouldn't we check the exit value here as we do in other similar methods?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+
+    private fun startDebugServer(udid: String, bundleId: String): String {
+        val out = ByteArrayOutputStream()
+
+        val result = project.exec {
+            it.workingDir = xcProject.toFile()
+            it.commandLine = listOf(idb, "debugserver", "start", "--udid", udid, bundleId)
+            it.standardOutput = out
+            it.errorOutput = out
```
ilmat192: We check the exit code on our own below. Shouldn't we set `isIgnoreExitValue = true` here?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+        }
+        check(result.exitValue == 0) { "Failed to start debug server: $out" }
+        return out.toString()
+    }
+}
+
+fun KonanTestExecutable.xcodeBuild() {
```
ilmat192: IMHO it may worth renaming this method to something like `configureXcodeBuild`. Just `xcodeBuild` may make an impression that the build is preformed right now while it will be performed only before task execution.

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+    this.doBeforeRun = Action {
+        val signIdentity = project.findProperty("sign_identity") as? String ?: "iPhone Developer"
+        val developmentTeam = project.findProperty("development_team") as? String
+        val xcProject = Paths.get(project.testOutputRoot, "launcher")
+
+        val shellScript: String = // language=Bash
+                mutableListOf("""
+                        # Copy executable to the build dir.
```
ilmat192: Why do we need to put the test executable into the Xcode project directory (`KonanTestLauncher/build/...`) before coping it to the `TARGET_BUILD_DIR`? Isn't it simpler to copy it directly from the original place as we do for frameworks?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+                        COPY_TO="${"$"}TARGET_BUILD_DIR/${"$"}EXECUTABLE_PATH"
+                        cp "${"$"}PROJECT_DIR/KonanTestLauncher/build/${"$"}TARGET_NAME.kexe" "${"$"}COPY_TO"
+                        # copy dSYM if it exists
+                        DSYM_DIR="${"$"}PROJECT_DIR/KonanTestLauncher/build/${"$"}TARGET_NAME.kexe.dSYM"
+                        if [ -d "${"$"}DSYM_DIR" ]; then
+                            cp -r "${"$"}DSYM_DIR" "${"$"}TARGET_BUILD_DIR/${"$"}EXECUTABLE_FOLDER_PATH/"
+                        fi
+                    """.trimIndent()).also {
+                    when (this) {
```
ilmat192: Why not just `if`?

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
```diff
@@ -311,3 +320,233 @@ private fun sshExecutor(project: Project) : ExecutorService = object : ExecutorS
...
+
+        xcProject.resolve("KonanTestLauncher.xcodeproj/project.pbxproj")
+                .toFile()
+                .apply {
+                    val text = readLines().joinToString("\n") {
+                        when {
+                            it.contains("CODE_SIGN_IDENTITY") ->
+                                it.replaceAfter("= ", "\"$signIdentity\";")
+                            it.contains("DEVELOPMENT_TEAM") || it.contains("DevelopmentTeam") ->
+                                it.replaceAfter("= ", "$developmentTeam;")
```
ilmat192: If `development_team` isn't set, `null` is written here. May be add a check separate check that `development_team` is specified?

build-tools/src/main/kotlin/org/jetbrains/kotlin/KonanTestExecutable.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Action that configures task or does some workload before the test will be executed.
+     * Could be done as a first step in the test or just as a `doFirst` action in the test task.
+     */
+    var doBeforeRun: Action<in Task>?
+
+    /**
+     * Action that configures task or does some workload before the test will be built.
+     * Depending on the test task impelmentation this action is done before the build task
```
ilmat192: Typo: `impelmentation` -> `implementation`

backend.native/tests/iosLauncher/KonanTestLauncher.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,381 @@
+// !$*UTF8*$!
```
ilmat192: May be add a `.gitignore` to the Xcode project to exclude user data? See samples: https://github.com/JetBrains/kotlin-native/blob/master/samples/uikit/.gitignore.

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
PavelPunegov: Yes. I'll add a comment here as it's a really tricky part

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
PavelPunegov: I'm going to add other tests like `Konan(Interop|Link|Dynamic)Test` tests. I've tried but forgot to replace `when` with several cases with `if`

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
PavelPunegov: We should but I've experienced a failures here with `idb` being unable to uninstall the app. And after the failure it was able to install an app in the next test. So I removed the check

backend.native/tests/iosLauncher/KonanTestLauncher.xcodeproj/project.pbxproj
PavelPunegov: This is a template project that should not be opened there. It rather should be opened in `./test.output/launcher` as it's a ready for the build there.

ilmat192: But what if we want to change something in this template project? Editing a `pbxproj` file is complicated and manual filtering userdata from committed files is boring.

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecutorService.kt
PavelPunegov: It's not needed. Rewritten

PR #3598, opened 2019-11-21
Fixed some Apple platform libs

homuroll:


platformLibs/src/platform/ios/UIKit.def
```diff
@@ -5,3 +5,4 @@ modules = UIKit
 
 compilerOpts = -framework UIKit
 linkerOpts = -framework UIKit
+excludedFunctions = UISceneErrorDomain
```
SvyatoslavScherbina: I can't find this function in `UIKIt`. Where is it located?

platformLibs/src/platform/ios/builtin.def
```diff
@@ -2,6 +2,9 @@ depends = posix
 package = platform.builtin
 headerFilter = **
 language = C
+
+excludedFunctions.ios_arm32 = builtin_clear_cache
+excludedFunctions.ios_arm64 = builtin_clear_cache
```
SvyatoslavScherbina: What are these ones?
Which binary contains definitions of these functions on ios_x64?

platformLibs/src/platform/ios/UIKit.def
SvyatoslavScherbina: I see, `excludedFunctions` filters globals too.

platformLibs/src/platform/ios/builtin.def
homuroll: This function is defined below in the same def file. It calls `__builtin___clear_cache`, which is for some reason missing on ios_armXX.

platformLibs/src/platform/ios/UIKit.def
homuroll: `UIKit.framework/Headers/UISceneDefinitions.h`

platformLibs/src/platform/osx/darwin.def
```diff
@@ -33,4 +33,4 @@ headerFilter = **
 excludedFunctions = __tg_promote
 
 compilerOpts = -D_XOPEN_SOURCE -DSHARED_LIBBIND -D_DARWIN_NO_64_BIT_INODE -DSYSCTL_DEF_ENABLED
-linkerOpts = -ldl -lz
+linkerOpts = -ldl -lz -lcurses -undefined dynamic_lookup
```
SvyatoslavScherbina: This affects linkage of all libraries and Kotlin code.
I believe we should find better solution if possible.
Would supporting regexp in `excludedFunctions` help?

platformLibs/src/platform/ios/darwin.def
```diff
@@ -29,10 +29,15 @@ headers = AppleTextureEncoder.h AssertMacros.h Availability.h AvailabilityIntern
...
+excludedFunctions = __tg_promote KERNEL_AUDIT_TOKEN KERNEL_SECURITY_TOKEN \
+                    asl_decode_buffer asl_encode_buffer asl_remove_index \
+                    averunnable dbm_forder pidlock ttyaction ttylock ttymsg ttyunlock uuid_generate_early_random
+
+excludedFunctions.ios_arm32 = _Unwind_Backtrace _Unwind_FindEnclosingFunction _Unwind_Find_FDE \
+                              _Unwind_SjLj_ForcedUnwind __deregister_frame __register_frame
 
 compilerOpts = -D_XOPEN_SOURCE -DSYSCTL_DEF_ENABLED
-linkerOpts = -ldl -lz
+linkerOpts = -ldl -lz -lbz2 -lcompression -late
```
SvyatoslavScherbina: Isn't the same required for macOS?

platformLibs/src/platform/osx/darwin.def
olonho: Yeah, but maybe not in this PR, but in the followup where we'll use static libs.

platformLibs/src/platform/ios/darwin.def
homuroll: Indeed.

platformLibs/src/platform/osx/darwin.def
homuroll: Actually, after adding `bz2`, `compression`, `ate` and `bsm` libs only a couple of dozens of functions remain unresolved and they can be explicitly added to `excludedFunctions`.

platformLibs/src/platform/ios/CoreAudioTypes.def
```diff
@@ -4,4 +4,3 @@ package = platform.CoreAudioTypes
 modules = CoreAudioTypes
 
 compilerOpts = -framework CoreAudioTypes
-linkerOpts = -framework CoreAudioTypes
```
SvyatoslavScherbina: + watchOS, tvOS

Issue #3594, opened 2019-11-20
Update to 1.3.60, got some error


VincentChen1212:
Hi,
When I update k/m to 1.3.60
I got two Error

First:
On iOS Build
```
Task :linkDebugFrameworkIOSArm64 
warning: linking module flags 'SDK Version': IDs have conflicting values ('[2 x i32] [i32 13, i32 2]' from /var/folders/ds/2v9832g575jblpl14n7p3nxh0000gn/T/protocols3108039485156635603.bc with '[2 x i32] [i32 13, i32 0]' from out) 
```

Second:
On Android Run
```
dlopen("/data/app/.../lib/arm64/libAndroid.so", RTLD_LAZY) failed: dlopen failed: cannot locate symbol "__cxa_throw" referenced by "/data/app/.../lib/arm64/libAndroid.so"... 
```

But 1.3.50 is OK.
I don't know what's wrong

Please Help me, Thanks

olonho:
On iOS it is just a warning, not an error. 

Android error is strange, could you please provide info on the Android version and device where you see this problem, it could be real.

olonho:
Try to add smth like

``
diff --git a/konan/konan.properties b/konan/konan.properties
index 7bda4d7d4..c1164556a 100644
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -496,7 +496,7 @@ clangOptFlags.android_arm32 = -O3 -ffunction-sections
 linkerNoDebugFlags.android_arm32 = -Wl,-S
 clangNooptFlags.android_arm32 = -O1
 targetSysRoot.android_arm32 = target-sysroot-1-android_ndk
-linkerKonanFlags.android_arm32 = -lm -lc++_static -landroid -llog -latomic
+linkerKonanFlags.android_arm32 = -lm -lc++_static -lc++abi -landroid -llog -latomic
 
 # Android ARM64, based on NDK.
 targetToolchain.macos_x64-android_arm64 = target-toolchain-2-osx-android_ndk
@@ -523,7 +523,7 @@ clangFlags.android_arm64 = -cc1 -target-cpu cortex-a57 -emit-obj -disable-llvm-p
 clangOptFlags.android_arm64 = -O3 -ffunction-sections
 clangNooptFlags.android_arm64 = -O1
 targetSysRoot.android_arm64 = target-sysroot-1-android_ndk
-linkerKonanFlags.android_arm64 = -lm -lc++_static -landroid -llog -latomic
[...]

PR #3599, opened 2019-11-21
Fix C++ ABI libraries on Android.

olonho:


olonho:
Fixes https://github.com/JetBrains/kotlin-native/issues/3594

PR #3602, opened 2019-11-21
Interop support of struct containing simd

knebekaizer:


knebekaizer:
> Did you actually tested those on arm32 and arm64 devices?

This is a fix of binding only, so the calling convention for integer simd type is still broken on arm32.  Should be another change.

PR #3616, opened 2019-11-26
Shared cyclic garbage detector.

olonho:


olonho:
TODO: add non-trivial tests.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -689,6 +698,31 @@ inline container_size_t objectSize(const ObjHeader* obj) {
   return alignUp(size, kObjectAlignment);
 }
 
+template <typename func>
+inline void traverseObjectFields(ObjHeader* obj, func process) {
```
SvyatoslavScherbina: The entire implementation of this function is quite similar to `traverseContainerObjectFields `. Is it possible to extract the common part?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -897,14 +931,30 @@ void freeAggregatingFrozenContainer(ContainerHeader* container) {
...
+      if (previous != nullptr)
+        previous->meta_object()->LeakDetector.next_ = next;
+      if (next != nullptr)
+        next->meta_object()->LeakDetector.previous_ = previous;
+      if (obj == g_leakCheckerGlobalList) {
+        g_leakCheckerGlobalList = next;
+      }
+      unlock(&g_leakCheckerGlobalLock);
+   }
+   if (obj->has_meta_object()) {
```
SvyatoslavScherbina: Wrong indents.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -897,14 +931,30 @@ void freeAggregatingFrozenContainer(ContainerHeader* container) {
 
 void runDeallocationHooks(ContainerHeader* container) {
   ObjHeader* obj = reinterpret_cast<ObjHeader*>(container + 1);
-
   for (int index = 0; index < container->objectCount(); index++) {
-    if (obj->has_meta_object()) {
-      ObjHeader::destroyMetaObject(&obj->typeInfoOrMeta_);
+    if (KonanNeedDebugInfo && g_checkLeaks && (obj->type_info()->flags_ & TF_LEAK_DETECTOR_CANDIDATE) != 0) {
```
SvyatoslavScherbina: `g_checkLeaks` here is typically `true`, so it may make sense to check type info flag first to reduce the performance cost of the entire check.

SvyatoslavScherbina: Also, if the object doesn't have a meta object, then this entire `if` doesn't seem to have any sense.
So it probably can be moved under `if (obj->has_meta_object()) {` below.

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -78,6 +78,19 @@ object GC {
...
 
+    /**
+     * Detect cyclic references going via atomic references and return list of cycle-inducing objects
+     * or `null` if the leak detector is not available.
+     */
+    @SymbolName("Kotlin_native_internal_GC_detectCycles")
+    external fun detectCycles(): Array<Any>?
+
+    /**
+     * Find a cycle starting from the given object, `null` if there are no cycles.
```
SvyatoslavScherbina: "Cycle" is somewhat ambiguous. Maybe add "reference"?
Also not sure if a cycle has any start.

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -78,6 +78,19 @@ object GC {
         get() = getTuneThreshold()
         set(value) = setTuneThreshold(value)
 
+    /**
+     * Detect cyclic references going via atomic references and return list of cycle-inducing objects
+     * or `null` if the leak detector is not available.
```
SvyatoslavScherbina: So a user getting `null` from this function would have no idea on which is the case: no leaks or unavailable leak detector.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2504,6 +2567,99 @@ void shareAny(ObjHeader* obj) {
...
+  KRefSet seen;
+  KRefListDeque queue;
+  KRefDeque toVisit;
+  KRefList path;
+  traverseReferredObjects(root, [&toVisit](ObjHeader* obj) { toVisit.push_front(obj); });
+  bool isFound = false;
+  while (!toVisit.empty() && !isFound) {
+    KRef current = toVisit.front();
+    toVisit.pop_front();
+    // Do BFS path search.
```
SvyatoslavScherbina: Is this a BFS?
It adds and takes vertices from the same end of the queue.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2504,6 +2567,99 @@ void shareAny(ObjHeader* obj) {
...
+  for (auto* it: cyclic) {
+    UpdateHeapRef(place++, it);
+  }
+  for (auto* root: rootset) {
+    DisposeStablePointer(root);
+  }
+  RETURN_OBJ(result->obj());
+}
+
+OBJ_GETTER(findCycle, KRef root) {
```
SvyatoslavScherbina: Does it report aggregated cycles too?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2504,6 +2567,99 @@ void shareAny(ObjHeader* obj) {
...
+  traverseReferredObjects(root, [&toVisit](ObjHeader* obj) { toVisit.push_front(obj); });
+  bool isFound = false;
+  while (!toVisit.empty() && !isFound) {
+    KRef current = toVisit.front();
+    toVisit.pop_front();
+    // Do BFS path search.
+    KRefList first;
+    first.push_back(current);
+    queue.emplace_back(first);
+    seen.clear();
```
SvyatoslavScherbina: Isn't the entire algorithm `O(N^2)` then?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2504,6 +2567,99 @@ void shareAny(ObjHeader* obj) {
   container->makeShared();
 }
 
+OBJ_GETTER0(detectCyclicReferences) {
+  // Collect rootset, hold references to simplify remaining code.
+  KRefList rootset;
+  lock(&g_leakCheckerGlobalLock);
+  auto* candidate = g_leakCheckerGlobalList;
+  while (candidate != nullptr) {
+    CreateStablePointer(candidate);
```
SvyatoslavScherbina: Looks much like an abuse of `CreateStablePointer`. Why not use `AddRef` directly instead?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2504,6 +2567,99 @@ void shareAny(ObjHeader* obj) {
...
+      if (current == root) seenToRoot = true;
+      // TODO: racy against concurrent mutators.
+      if (seen.count(current) == 0) {
+        traverseReferredObjects(current, [&toVisit](ObjHeader* obj) {
+           toVisit.push_front(obj);
+        });
+        seen.insert(current);
+      }
+    }
+    if (seenToRoot && cyclic.count(root) == 0) {
```
SvyatoslavScherbina: Why check for `cyclic.count(root) == 0`?

runtime/src/main/cpp/Memory.h
```diff
@@ -315,16 +315,27 @@ ALWAYS_INLINE bool hasPointerBits(T* ptr, unsigned bits) {
...
+  struct {
+    // Strong reference to the counter object.
+    ObjHeader* counter_;
+  } WeakReference;
+  struct {
+    // Leak detector's previous list element.
+    ObjHeader* previous_;
+    // Leak detector's next list element.
+    ObjHeader* next_;
+  } LeakDetector;
```
SvyatoslavScherbina: This makes Objective-C interoperability consume even more memory.

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
olonho: Added reference to appropriate docs.

runtime/src/main/cpp/Memory.cpp
olonho: Typo.

runtime/src/main/cpp/Memory.cpp
olonho: No, it finds some cycle including the given element.

runtime/src/main/cpp/Memory.cpp
olonho: N is small is most cases, and this code is not meant to be high performance.

runtime/src/main/cpp/Memory.cpp
olonho: Well, can do it either way.

runtime/src/main/cpp/Memory.cpp
olonho: Debug code, removed.

runtime/src/main/cpp/Memory.h
olonho: Indeed.

olonho: We may want to completely separate debug and release runtimes bitcodes, and make those debug-only.

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -108,4 +110,7 @@ object GC {
 
     @SymbolName("Kotlin_native_internal_GC_setTuneThreshold")
     private external fun setTuneThreshold(value: Boolean)
+
+    @SymbolName("Kotlin_native_internal_GC_getHasLeakDetector")
+    private external fun getHasLeakDetector(): Boolean
```
SvyatoslavScherbina: Dead code?

backend.native/tests/build.gradle
```diff
@@ -2720,6 +2720,12 @@ standaloneTest("memory_only_gc") {
     source = "runtime/memory/only_gc.kt"
 }
 
+standaloneTest("leak_detector") {
+    disabled = !isAppleTarget(project) || project.globalTestArgs.contains('-opt')
```
SvyatoslavScherbina: > `!isAppleTarget(project)`

Why so?

runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -80,13 +80,15 @@ object GC {
...
-     * or `null` if the leak detector is not available.
+     * or `null` if the leak detector is not available. Use [Platform.isMemoryLeakCheckerActive] to check
+     * leak detector availability.
      */
     @SymbolName("Kotlin_native_internal_GC_detectCycles")
     external fun detectCycles(): Array<Any>?
 
     /**
-     * Find a cycle starting from the given object, `null` if there are no cycles.
+     * Find a reference cycle including from the given object, `null` if there are no cycles or if leak
```
SvyatoslavScherbina: > `null` if there are no cycles or if leak detector is not avaialble.

Is this true? `findCycle` doesn't seem to require `g_checkLeaks`.

SvyatoslavScherbina: Also typos:
> avaialble
> cycle including from the given object

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: I mean that this function probably finds non-leaking cycles too. Is this true?

runtime/src/main/cpp/Memory.h
SvyatoslavScherbina: Or meta object allocator could reserve extra room for these fields in debug mode.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -104,6 +109,10 @@ volatile int aliveMemoryStatesCount = 0;
 
 KBoolean g_checkLeaks = KonanNeedDebugInfo;
 
+// Only used by the leak detector.
+KRef g_leakCheckerGlobalList = nullptr;
+KInt g_leakCheckerGlobalLock = 0;
```
knebekaizer: * Global visibility is not needed here. Consider RAII + internal linkage. 
* leakCheckerGlobalList shall be accessed in the locked scope only, so make it obvious.
```
class LeakCheckerGlobalLocker {
  LeakCheckerGlobalLocker() { lock(lock_); }
  ~LeakCheckerGlobalLocker() { unlock(lock_); }
  KRef checkerGlobalList() const { return leakCheckerGlobalList_; } // still may escape, but better than nothing
private:
  LeakCheckerGlobalLocker(const LeakCheckerGlobalLocker&) = delete;
  LeakCheckerGlobalLocker(LeakCheckerGlobalLocker&&) = delete;
  static KInt lock_;
  static KRef leakCheckerGlobalList_;
};
```

runtime/src/main/cpp/Memory.cpp
```diff
@@ -689,24 +698,38 @@ inline container_size_t objectSize(const ObjHeader* obj) {
...
-    } else {
-      ArrayHeader* array = obj->array();
-      for (int index = 0; index < array->count_; index++) {
-        process(ArrayAddressOfElementAt(array, index));
-      }
-    }
+    traverseObjectFields(obj, process);
     obj = reinterpret_cast<ObjHeader*>(
       reinterpret_cast<uintptr_t>(obj) + objectSize(obj));
   }
```
knebekaizer: Now, the only difference between `traverseContainerObjectFields` and `traverseContainerReferredObjects` is that the latter adds instrumenting (null check). 
1. Consider better naming to reflect that.
2. Or revise usage of non-checked version: guess all the call sites add their own checks anyway. Do we really need both?
3. if two policies are definitely needed than make the intention clear, e.g.:

```
template <typename func> auto checked(func process) {
  return [=](auto location) { if (*location) process(*location); };
}
...
traverseObjectFields(obj, checked(process))
```
4. However, a particular instrumenting policy seems to be a property of (or deducted from) the first parameter type, rather than the call site property. For example, can we get rid of null check for array of not nullable elements? Simple overriding may be good enough.

runtime/src/main/cpp/Memory.cpp
olonho: It is not global, it's anonymous C++ namespace.

olonho: Yes, we could do RAII here, although it's not too important.

runtime/src/main/cpp/Memory.cpp
olonho: Not sure if it is related to this issue.

PR #3617, opened 2019-11-26
Update to Gradle 6.0.1 and drop the middle Gradle plugin

ilmat192:


msink:
Publising to bintray is broken with Gradle 6.0.1 : https://github.com/gradle/gradle/issues/11412
Workaround: add `systemProp.org.gradle.internal.publish.checksums.insecure=true` to your `gradle.properties` file

ilmat192:
Thank you for the report! We've discussed this problem and it seems that we have no technical ability to fix this and need to wait for Bintray's answer.

PR #3618, opened 2019-11-26
Refactor psi2irPhase in Native compiler to support a new way to enable compiler plugins

sandwwraith:
See https://github.com/JetBrains/kotlin/commit/3e18350a1f4ffe295756dc68250d5d7bf78278cb .


vvlevchenko:
merged in #3626

PR #3621, opened 2019-11-26
Char-based enum shall be mapped to integer-based

knebekaizer:


Interop/Indexer/src/main/kotlin/org/jetbrains/kotlin/native/interop/indexer/Indexer.kt
```diff
@@ -267,7 +267,11 @@ internal class NativeIndexImpl(val library: NativeLibrary, val verbose: Boolean
             val cursorType = clang_getCursorType(cursor)
             val typeSpelling = clang_getTypeSpelling(cursorType).convertAndDispose()
 
-            val baseType = convertType(clang_getEnumDeclIntegerType(cursor))
+            val canonicalBase = clang_getCanonicalType(clang_getEnumDeclIntegerType(cursor))
+            val baseType = when (canonicalBase.kind) {
+                CXType_Char_S, CXType_Char_U -> IntegerType(1, false, canonicalBase.name.dropConstQualifier())
```
SvyatoslavScherbina: This makes char-based enums somewhat special.
The example of (somewhat special) case that is handled not entirely correct here:
```c
typedef char MyChar;
enum MyCharEnum : MyChar { ZERO };
```

`MyCharEnum` should be based on `MyChar` in Kotlin too.

I believe it should be possible to fix `char`-based enums by supporting `CharType` in some new cases in stub generator properly (IIRC, this would be making integer literals).

knebekaizer: This is an intended trade-off.
Typedef is declared out of scope of enum and may be used in many different contexts. We want MyCharEnum base type to be IntegerType, but typedef type is already parsed and mapped to Kotlin without relation to Enum and therefore has base type CharType. Moreover it may be used in other context such as `typedef MyChar* CString` and then `void foo(CString s)`. I don't see a way to provide two typedefs with different underlying type to avoid such conflicts.
OTOH `enum MyCharEnum : MyChar` is kinda of edge case, so decision was made in favor of a common case: use underlying type directly if enum base type is typedefed char.
As you can see, any other (non-edge) case such as 
```
typedef signed char MyChar;
enum MyCharEnum : MyChar { ZERO };
```
will produce Kotlin declaration based on MyChar as expected.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
```diff
@@ -221,7 +221,11 @@ class StubsBuildingContextImpl(
         get() = stubIrContext.getKotlinName(this)
 
     override fun tryCreateIntegralStub(type: Type, value: Long): IntegralConstantStub? {
-        val integerType = type.unwrapTypedefs() as? IntegerType ?: return null
+        val integerType = when (val unwrappedType = type.unwrapTypedefs()) {
+            is IntegerType -> unwrappedType as IntegerType
```
SvyatoslavScherbina: The cast is likely not required here.

PR #3623, opened 2019-11-28
A native counterpart for plumbing metadata and ir versions to klib manifest

alexander-gorshenev:


alexander-gorshenev:
This PR has been integrated through mirror.

PR #3625, opened 2019-11-28
Local escape analysis

LepilkinaElena:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -439,13 +446,70 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                val className = irClass.fqNameForIrSerialization.asString()
+                assert(className in arrayClasses)
+                allocArrayOnStack(className, count)
+            } else {
+                call(context.llvm.allocArrayFunction, listOf(codegen.typeInfoValue(irClass), count),
+                        lifetime, exceptionHandler)
+            }
+
+    fun setTypeInfoForLocalObject(objectHeader: LLVMValueRef) {
+        val typeInfo = structGep(objectHeader, 0)
```
homuroll: Here and below would be nice to see the name of the field being accessed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -439,13 +446,70 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                val stackSlot = alloca(context.llvmDeclarations.forClass(irClass).bodyType)
+                val objectHeader = structGep(stackSlot, 0)
+                setTypeInfoForLocalObject(objectHeader)
+                objectHeader
+            } else {
+                allocInstance(codegen.typeInfoForAllocation(irClass), lifetime)
+            }
+
+    // TODO: find better place?
+    val arrayClasses = mapOf(
```
homuroll: arrayClasses -> arrayToElementType

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -504,9 +504,9 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
     fun allocArrayOnStack(arrayTypeName: String, count: LLVMValueRef): LLVMValueRef {
         val arraySlot = alloca(localArrayType(arrayTypeName, extractConstUnsignedInt(count).toInt()))
         // Set array size in ArrayHeader.
-        val arrayHeaderSlot = structGep(arraySlot, 0)
+        val arrayHeaderSlot = structGep(arraySlot, 0, "arrayHeaderField")
```
homuroll: I meant exactly the name of a field from the declaration in runtime. With that we can grep it if want to find all usages.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/LocalEscapeAnalysis.kt
```diff
@@ -74,6 +74,11 @@ internal object LocalEscapeAnalysis {
                         } ?: node.callee.irFunction?.isExternal?.let {
                             !(it || node.callee.irFunction.isBuiltInOperator)
```
homuroll: `isBuiltInOperator` is not needed anymore, isn't it? Since we're replacing call to `get` with `ArrayGet`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/LocalEscapeAnalysis.kt
```diff
@@ -74,6 +74,11 @@ internal object LocalEscapeAnalysis {
...
                         } ?: true
+                        // Connect with all arguments that return value pointsTo.
+                        val returnPointsToMask = node.callee.pointsTo?.elementAtOrNull(node.callee.parameters.size) ?: 0
+                        if (returnPointsToMask and (1 shl index) != 0) {
```
homuroll: Here should be generic code connecting all possible nodes (see usage of `pointsTo` in global escape analysis).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
LepilkinaElena: There is no declaration in runtime for this case. It's generated class for local array. In other cases I'll fix.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/LocalEscapeAnalysis.kt
LepilkinaElena: Do any other builtinOperator exist? I amn'tt sure and as far as there is `isBuiltinOperator` https://github.com/JetBrains/kotlin-native/blob/d6b20598e82be6a35c003fcfacc7dbda38a897be/backend.native/compiler/ir/backend.native/src/org/jetbains/kotlin/backend/konan/optimizations/DataFlowIR.kt#L600 I left it.

PR #3635, opened 2019-12-03
Exclude stdlib and platform libs when building cache for tests

SvyatoslavScherbina:


build-tools/src/main/kotlin/org/jetbrains/kotlin/CacheTesting.kt
```diff
@@ -51,6 +51,7 @@ fun configureCacheTesting(project: Project): CacheTesting? {
                 "-p", cacheKind.visibleName,
                 "-o", cacheFile,
                 "-Xmake-cache=$stdlib",
+                "-nodefaultlibs", "-nostdlib",
```
olonho: `-no-default-libs` and , `-nodefaultlibs` is deprecated form.

PR #3645, opened 2019-12-06
Allow stateless and constant objects compile-time init .

olonho:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -18,6 +18,39 @@ import org.jetbrains.kotlin.ir.util.*
...
+    PERMANENT,
+    THREAD_LOCAL,
+    SHARED
+}
+
+private val IrConstructor.isAnyConstructorDelegation: Boolean
+    get() {
+        val statements = this.body?.statements ?: return false
+        if (statements.size != 2) return false
+        if (statements[1] !is IrReturnImpl) return false
```
SvyatoslavScherbina: The returned value should also be checked.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -18,6 +18,39 @@ import org.jetbrains.kotlin.ir.util.*
...
+
+// TODO: shall we memoize that property?
+internal fun IrClass.hasConstStateAndNoSideEffects(context: Context): Boolean {
+    val fields = context.getLayoutBuilder(this).fields
+    if (this.hasAnnotation(KonanFqNames.canBePrecreated)) {
+        assert(fields.all { it.isFinal && (it.type.isPrimitiveType() || it.type.isString()) &&
+                it.initializer?.expression is IrConst<*> })
+        return true
+    }
+    return fields.isEmpty() && this.constructors.all { it.isAnyConstructorDelegation }
```
SvyatoslavScherbina: What if this class is cached and its IR is just a stub?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -847,8 +883,13 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
             }
         }
 
+        if (irClass.hasConstStateAndNoSideEffects(context)) {
+            return loadSlot(codegen.getObjectInstanceStorage(irClass, ObjectStorageKind.PERMANENT), false)
```
SvyatoslavScherbina: What if the class has `@ThreadLocal` annotation?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -738,9 +737,33 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+                context.llvm.staticData.createConstKotlinObject(declaration,
+                        *computeFields(declaration)).llvm else codegen.kNullObjHeaderPtr)
+        }
+    }
+
+    private fun computeFields(declaration: IrClass): Array<ConstValue> {
+        val fields = context.getLayoutBuilder(declaration).fields
+        return Array(fields.size) { index ->
+            val initializer = fields[index].initializer!!.expression as IrConst<*>
+            when (initializer.kind) {
```
SvyatoslavScherbina: Why not reuse `evaluateConst`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/DFGBuilder.kt
```diff
@@ -230,7 +230,8 @@ internal class ModuleDFGBuilder(val context: Context, val irModule: IrModuleFrag
                         println("Analysing global field ${declaration.descriptor}")
                         println("IR: ${ir2stringWhole(declaration)}")
                     }
-                    analyze(declaration, IrSetFieldImpl(it.startOffset, it.endOffset, declaration.symbol, null,
+                    if (it.expression !is IrConst<*>)
```
SvyatoslavScherbina: The entire `visitField` looks intended for global fields. Maybe adjust the entry condition instead then?
Otherwise debug output will be incorrect.

runtime/src/main/kotlin/kotlin/native/internal/Annotations.kt
```diff
@@ -115,4 +115,13 @@ annotation class Independent
  */
 @Target(AnnotationTarget.FUNCTION)
 @Retention(AnnotationRetention.BINARY)
-@PublishedApi internal annotation class FilterExceptions
\ No newline at end of file
+@PublishedApi internal annotation class FilterExceptions
+
+/**
+ * Indicates that given top level signleton object can be created in compile time and thus
+ * members access doesn't need to have to use init barrier and allow better optimizations for
```
SvyatoslavScherbina: > doesn't need to have to use

Kinda overcomplicated.

runtime/src/main/kotlin/kotlin/native/internal/NativePtr.kt
```diff
@@ -12,6 +12,7 @@ external fun getNativeNullPtr(): NativePtr
 
 class NativePtr @PublishedApi internal constructor(private val value: NonNullNativePtr?) {
     companion object {
+        // TODO: make it properly precreated, maybe use an intrinsic for that.
```
SvyatoslavScherbina: `getNativeNullPtr()` _is_ an intrinsic already.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -18,6 +18,39 @@ import org.jetbrains.kotlin.ir.util.*
...
+        if (!constructor.constructedClass.isAny()) return false
+        return true
+    }
+
+// TODO: shall we memoize that property?
+internal fun IrClass.hasConstStateAndNoSideEffects(context: Context): Boolean {
+    val fields = context.getLayoutBuilder(this).fields
+    if (this.hasAnnotation(KonanFqNames.canBePrecreated)) {
+        assert(fields.all { it.isFinal && (it.type.isPrimitiveType() || it.type.isString()) &&
+                it.initializer?.expression is IrConst<*> })
```
homuroll: If `initializer == null` the cast will fail.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
```diff
@@ -73,11 +76,19 @@ internal class InitializersLowering(val context: CommonBackendContext) : ClassLo
...
                                             context.irBuiltIns.unitType,
                                             STATEMENT_ORIGIN_ANONYMOUS_INITIALIZER)))
                     )
-                    declaration.initializer = null
+
+                    // We shall keep initializer for constants for compile-time instantiation.
+                    declaration.initializer =
+                            if (initExpression is IrConst<*> &&
+                                    (initExpression.type.isPrimitiveType() || initExpression.type.isString())) {
+                                IrExpressionBodyImpl(initExpression.copy())
```
homuroll: What this `IrExpression` for? What's wrong with just `initExpression.copy()`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/DFGBuilder.kt
homuroll: Yeah, here implicitly being used the fact of erasing of all initializers for non-global fields. Probably better to check that explicitly by reading the field's owner and skipping if it's a class.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/DataFlowIR.kt
```diff
@@ -485,7 +486,10 @@ internal object DataFlowIR {
                 }
 
                 override fun visitField(declaration: IrField) {
-                    declaration.initializer?.let { mapFunction(declaration) }
+                   declaration.initializer?.let {
```
homuroll: Same as above.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Hmm, are ctors required to return `Unit`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Don't think it makes sense to do have thread local storage in such a case.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: None of the types above are nullable.

runtime/src/main/kotlin/kotlin/native/internal/NativePtr.kt
olonho: We need to come up with the way to make native NULL easily foldable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Why there's a difference?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: Not really.
```kotlin
class Test {
    constructor() {
        return println(42)
    }
}
```

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: External IR stubs may omit fields as non-public interface.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: I mean that presence of `@ThreadLocal` is handled inconsistently here and in `LlvmDeclarations.kt`. The latter assumes that `@ThreadLocal object Foo {}` has thread-local instance field, while the former decides it is global.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: IR for return is `IrReturn`, not `IrReturnImpl`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -18,6 +18,39 @@ import org.jetbrains.kotlin.ir.util.*
...
+
+private val IrConstructor.isAnyConstructorDelegation: Boolean
+    get() {
+        val statements = this.body?.statements ?: return false
+        if (statements.size != 2) return false
+        if (statements[1] !is IrReturnImpl) return false
+        val constructorCall = statements[0] as? IrDelegatingConstructorCallImpl ?: return false
+        val constructor = constructorCall.symbol.owner as? IrConstructor ?: return false
+        if (!constructor.constructedClass.isAny()) return false
+        return true
```
SvyatoslavScherbina: Pattern matching here is not correct for IR that is not lowered (e.g. for IR of cached library).

olonho: Indeed, only left it for debug builds.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -18,6 +18,34 @@ import org.jetbrains.kotlin.ir.util.*
...
+    SHARED
+}
+
+private fun IrConstructor.isAnyConstructorDelegation(context: Context): Boolean {
+        val statements = this.body?.statements ?: return false
+        if (statements.size != 2) return false
+        val lastStatement = statements[1]
+        if (lastStatement !is IrReturn ||
+                (lastStatement.value as? IrGetObjectValue)?.symbol != context.irBuiltIns.unitClass) return false
+        val constructorCall = statements[0] as? IrDelegatingConstructorCallImpl ?: return false
```
SvyatoslavScherbina: `IrDelegatingConstructorCall`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -47,6 +46,7 @@ internal val IrField.storageClass: FieldStorage get() {
     // TODO: Is this correct?
     val annotations = correspondingPropertySymbol?.owner?.annotations ?: annotations
     return when {
+        annotations.hasAnnotation(KonanFqNames.canBePrecreated) -> FieldStorage.SHARED
```
SvyatoslavScherbina: How does `@CanBeCreated` work on fields? What is it supposed to mean?

olonho: Nothing, just incorrect place for the check, will fix.

PR #3657, opened 2019-12-10
Fixed -make-cache option and tests on caches

homuroll:


build-tools/src/main/kotlin/org/jetbrains/kotlin/CacheTesting.kt
```diff
@@ -49,7 +49,7 @@ fun configureCacheTesting(project: Project): CacheTesting? {
         it.commandLine(
                 "$dist/bin/konanc",
                 "-p", cacheKind.visibleName,
-                "-o", cacheFile,
+                "-o", "stlib-cache",
```
SvyatoslavScherbina: How does this work?
AFAIK, this command would create output in working directory of this `Exec` task. Is it the same as `cacheDir`?

homuroll: Ooops, my bad.

PR #3659, opened 2019-12-11
Separate bits combination in object header for local variables on stack

LepilkinaElena:


runtime/src/main/cpp/Memory.h
```diff
@@ -360,6 +360,11 @@ struct ObjHeader {
     return (reinterpret_cast<MetaObjHeader*>(clearPointerBits(typeInfoOrMeta_, OBJECT_TAG_MASK)))->container_;
   }
 
+  inline bool local() const {
+    unsigned bits = getPointerBits(typeInfoOrMeta_, OBJECT_TAG_MASK);
+    return (bits & (OBJECT_TAG_PERMANENT_CONTAINER | OBJECT_TAG_NONTRIVIAL_CONTAINER)) == 3;
```
olonho: A bit better to use `(OBJECT_TAG_PERMANENT_CONTAINER | OBJECT_TAG_NONTRIVIAL_CONTAINER)` instead of `3` to make intentions clear.

PR #3661, opened 2019-12-11
klib expect actual

alexander-gorshenev:
This is a basic expect and actual support in klibs.
It is needed to be able to reliably write and read expect and actual declaration to klib metadata and klib ir.
Use -Xklib-mpp to turn it on.
The IR part is rather limited for now.


alexander-gorshenev:
The corresponding Kotlin review is here: 
https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3625

PR #3667, opened 2019-12-12
[Interop] Add classes and constructors to metadata-based libraries

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/InteropLibraryCreation.kt
```diff
@@ -44,12 +48,44 @@ fun createInteropLibrary(arguments: LibraryCreationArguments) {
...
 }
+
+// TODO: Consider adding it to kotlinx-metadata-klib.
+class ChunkingWriteStrategy(
+        private val classesChunkSize: Int = 128,
+        private val packagesChunkSize: Int = 128
+) : KlibModuleFragmentWriteStrategy {
+
+    override fun processPackageParts(parts: List<KmModuleFragment>): List<KmModuleFragment> {
+        val fqName = parts.firstOrNull { it.fqName != null }?.fqName
```
SvyatoslavScherbina: Maybe ensure that all `fqName`s are the same?
Btw, why `fqName` can be`nul`? Is it guaranteed that there is at least one non-null `fqName` among `parts`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -341,6 +344,11 @@ sealed class PropertyAccessor : FunctionalStub {
                 val constant: ConstantStub? = null
         ) : Getter()
 
+        class ConstructorParameter(
+                val constructorParameter: FunctionParameterStub,
+                override val annotations: List<AnnotationStub> = emptyList()
+        ) : Getter()
```
SvyatoslavScherbina: This code can be read as "constructor parameter is a getter". Is this intended? What does this mean?

SvyatoslavScherbina: Ah, I see. Maybe `GetConstructorParameter`?
Also, it is initializer, and you can't actually use a constructor parameter in a getter.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
```diff
@@ -229,72 +231,73 @@ enum class VisibilityModifier {
...
-            override val functions: List<FunctionalStub> = emptyList(),
             override val simpleContainers: List<SimpleStubContainer> = emptyList()
-    ) : ClassStub()
+    ) : ClassStub() {
+        override val functions: List<FunctionalStub> = constructors + methods
+    }
 
     class Companion(
+            override val classifier: Classifier,
+            constructors: List<ConstructorStub> = emptyList(),
```
SvyatoslavScherbina: Why not synthesize the constructor here? Isn't it the same for all `companion object`s?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -61,4 +61,15 @@ val StubContainer.defaultMemberModality: MemberStubModality
...
+
+/**
+ * Returns constructor that should be rendered in class header.
+ * Note that while companion object has a primary constructor, it is not visible to user.
+ */
+val ClassStub.explicitPrimaryConstructor: ConstructorStub?
+    get() = when (this) {
+        is ClassStub.Simple -> functions.filterIsInstance<ConstructorStub>().firstOrNull(ConstructorStub::isPrimary)
+        is ClassStub.Companion -> null
+        is ClassStub.Enum -> functions.filterIsInstance<ConstructorStub>().firstOrNull(ConstructorStub::isPrimary)
```
SvyatoslavScherbina: Can we join this branch with the first one?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -61,4 +61,15 @@ val StubContainer.defaultMemberModality: MemberStubModality
         }
         is ClassStub.Companion -> MemberStubModality.FINAL
         is ClassStub.Enum -> MemberStubModality.FINAL
+    }
+
+/**
+ * Returns constructor that should be rendered in class header.
+ * Note that while companion object has a primary constructor, it is not visible to user.
```
SvyatoslavScherbina: Maybe synthesize a constructor when serializing a companion object then?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrUniqIdProvider.kt
```diff
@@ -51,11 +47,49 @@ internal class StubIrUniqIdProvider {
...
+        is StubOrigin.Enum -> origin.enum.uniqueSymbolName
+        else -> null
+    }
+
+    fun uniqIdForClass(classStub: ClassStub): UniqId = with (mangler) {
+        when (classStub) {
+            is ClassStub.Simple,
+            is ClassStub.Enum -> uniqSymbolNameForClass(classStub.origin)
+            is ClassStub.Companion -> "${context.prefix}#Companion"
+        } ?: error("Unexpected origin ${classStub.origin} for enum class ${classStub.classifier.fqName}.")
```
SvyatoslavScherbina: > for enum class

Why so?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrUniqIdProvider.kt
```diff
@@ -51,11 +47,49 @@ internal class StubIrUniqIdProvider {
...
+
     /**
      * MSB should be set to 1 for public declarations.
      * @see org.jetbrains.kotlin.ir.util.UniqId
      */
     private fun Long.markAsPublic() = this or (1L shl 63)
 
     private fun String.toUniqId() = UniqId(cityHash64().markAsPublic())
+
+    private fun FunctionalStub.mangleParameters(): String = parameters.joinToString(
```
SvyatoslavScherbina: This method is kinda harmful. It is too general, and doesn't handle actual language-specific overloads (e.g. for Objective-C constructors).
Is it possible to avoid having such a fallback?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -20,313 +21,460 @@ class StubIrMetadataEmitter(
...
+                with (MappingExtensions(data.typeParametersInterner)) {
+                    KmTypeAlias(element.flags, element.alias.topLevelName).also { km ->
+                        km.uniqId = data.uniqIds.uniqIdForTypeAlias(element)
+                        km.underlyingType = element.aliasee.map(shouldExpandTypeAliases = false)
+                        km.expandedType = element.aliasee.map()
+                    }
+                }
+
+        override fun visitFunction(element: FunctionStub, data: VisitingContext) =
+                with (MappingExtensions(data.typeParametersInterner)) {
```
SvyatoslavScherbina: Does this approach handle the case when function type parameters are used in function signature?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrUniqIdProvider.kt
sbogolepov: Agree, it's the most controversial part of the PR. I'll try to rethink the implementation.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/InteropLibraryCreation.kt
sbogolepov: It is guaranteed that all these `fqName` are the same. I'll note it in `processPackageParts` documentation.
> Btw, why fqName can be `null`?
Actually, it [shouldn't](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/klib/src/kotlinx/metadata/klib/KlibModuleMetadata.kt#L112). Will add an assertion.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
sbogolepov: Agreed. What do you think about adding `backingField: ValueStub?` field to `PropertyStub` and adding `PropertyAccessor.Getter.ReadBackingField` instead of `ConstructorParameter `

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: In general, yes. But current implementation is incorrect :)

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -20,313 +21,460 @@ class StubIrMetadataEmitter(
...
+    val FunctionParameterStub.flags: Flags
+        get() = flagsOfNotNull(
+                Flag.HAS_ANNOTATIONS.takeIf { annotations.isNotEmpty() }
+        )
+
+    val ClassStub.flags: Flags
+        get() = flagsOfNotNull(
+                Flag.HAS_ANNOTATIONS.takeIf { annotations.isNotEmpty() },
+                Flag.IS_PUBLIC,
+                Flag.IS_OPEN.takeIf { this is ClassStub.Simple && this.modality == ClassStubModality.OPEN },
```
SvyatoslavScherbina: Does `KmClass` for final class require `IS_FINAL` flag?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -20,313 +21,460 @@ class StubIrMetadataEmitter(
...
+            val uniqIds: StubIrUniqIdProvider,
+            val typeParametersInterner: Interner<TypeParameterStub> = Interner()
+    )
+
+    private val visitor = object : StubIrVisitor<VisitingContext, Any> {
+
+        override fun visitClass(element: ClassStub, data: VisitingContext): List<KmClass> {
+            val classVisitingContext = VisitingContext(
+                    container = element,
+                    uniqIds = data.uniqIds.createChild(element.classifier.relativeFqName),
```
SvyatoslavScherbina: Is the entire `relativeFqName` required here?
Shouldn't it be just a last segment in such a name?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIr.kt
SvyatoslavScherbina: I guess `GetConstructorParameter` is ok for now, so maybe later.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: No.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -20,313 +21,461 @@ class StubIrMetadataEmitter(
...
+                        km.uniqId = data.uniqIds.uniqIdForTypeAlias(element)
+                        km.underlyingType = element.aliasee.map(shouldExpandTypeAliases = false)
+                        km.expandedType = element.aliasee.map()
+                    }
+                }
+
+        override fun visitFunction(element: FunctionStub, data: VisitingContext) =
+                with (MappingExtensions(data.typeParametersInterner)) {
+                    KmFunction(element.flags, element.name).also { km ->
+                        element.typeParameters.mapTo(km.typeParameters) { it.map() }
```
SvyatoslavScherbina: Btw, Kotlin allows referencing type parameter before it is defined like this:
```kotlin
fun <T : U, U> foo() {}
```

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -20,313 +21,461 @@ class StubIrMetadataEmitter(
...
+        get() = flagsOfNotNull(
+                Flag.HAS_ANNOTATIONS.takeIf { annotations.isNotEmpty() }
+        )
+
+    val ClassStub.flags: Flags
+        get() = flagsOfNotNull(
+                Flag.HAS_ANNOTATIONS.takeIf { annotations.isNotEmpty() },
+                Flag.IS_PUBLIC,
+                Flag.IS_OPEN.takeIf { this is ClassStub.Simple && this.modality == ClassStubModality.OPEN },
+                Flag.IS_FINAL.takeIf { this is ClassStub.Simple && this.modality == ClassStubModality.NONE },
```
SvyatoslavScherbina: Why not reuse `modality.flags` then?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: 😧

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: It will require additional refactoring that can be performed separately.

PR #3678, opened 2019-12-13
Rework TLS access.

olonho:


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -30,7 +30,7 @@ struct RuntimeState {
   volatile int executionStatus;
 };
 
-typedef void (*Initializer)(int initialize);
+typedef void (*Initializer)(int initialize, MemoryState* memory);
```
SvyatoslavScherbina: Is it actually required to pass `MemoryState` explicitly here?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1751,6 +1757,7 @@ void deinitMemory(MemoryState* memoryState) {
   konanDestructInstance(memoryState->toFree);
   konanDestructInstance(memoryState->roots);
   konanDestructInstance(memoryState->toRelease);
+  konanDestructInstance(memoryState->tlsMap);
```
SvyatoslavScherbina: Maybe add an assertion checking it is empty?

runtime/src/main/cpp/Memory.h
```diff
@@ -81,6 +81,8 @@ typedef enum {
 
 typedef uint32_t container_size_t;
 
+struct InitNode;
```
SvyatoslavScherbina: Dead code?

runtime/src/main/cpp/Memory.h
```diff
@@ -548,6 +550,13 @@ void MutationCheck(ObjHeader* obj);
 void FreezeSubgraph(ObjHeader* obj);
 // Ensure this object shall block freezing.
 void EnsureNeverFrozen(ObjHeader* obj);
+// Add TLS object storage, called by the generated code.
+void AddTLSRecord(MemoryState* memory, int module, int size) RUNTIME_NOTHROW;
```
SvyatoslavScherbina: Note that using module as a key here is somewhat too restricting.
For example, it is incompatible with fine-grained incremental compilation.
Maybe at least rename `module` to `key` or `id`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -277,6 +277,12 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
     var serializedIr: SerializedIrModule? = null
     var dataFlowGraph: ByteArray? = null
 
+    val moduleId by lazy {
+        irModule!!.name.asString().localHash.value.toInt()
```
SvyatoslavScherbina: Using an address of a private LLVM global as a TLS map key is kinda more flexible:
1.  It would work when compiling only a part of a module (e.g. proper incremental compilation).
2.  It would make TLS code generation more self-contained (e.g. independent from IR).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -15,12 +15,13 @@ import org.jetbrains.kotlin.ir.declarations.*
...
 import org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid
 import org.jetbrains.kotlin.ir.visitors.acceptChildrenVoid
+import org.jetbrains.kotlin.ir.visitors.acceptVoid
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 
 internal fun createLlvmDeclarations(context: Context): LlvmDeclarations {
     val generator = DeclarationsGeneratorVisitor(context)
-    context.ir.irModule.acceptChildrenVoid(generator)
+    context.ir.irModule.acceptVoid(generator)
```
SvyatoslavScherbina: What is this patch for?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -262,35 +263,113 @@ private class DeclarationsGeneratorVisitor(override val context: Context) :
...
-                        "kshadowobjref:" + qualifyInternalName(irClass)
+        val addressGetterName = if (isExported) {
+            irClass.objectInstanceAddressSymbolName
+        } else {
+            null
+        }
+
+        // If object is imported - access it via getter functions.
+        if (isExternal(irClass)) {
+            assert(valueGetterName != null)
```
SvyatoslavScherbina: Is this code reachable?

SvyatoslavScherbina: Note that `LlvmDeclarations` is explicitly designed to generate declarations for IR to be compiled. It is not supposed to cover any "external" IR declarations.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -935,39 +895,16 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
-        val newValue = call(initFunction, args, Lifetime.GLOBAL, exceptionHandler)
-        val bbInitResult = currentBlock
-        br(bbExit)
-
-        positionAtEnd(bbExit)
-        val valuePhi = phi(codegen.getLLVMType(irClass.defaultType))
-        addPhiIncoming(valuePhi, bbCurrent to objectVal, bbInitResult to newValue)
-
-        return valuePhi
+        return context.llvmDeclarations.forSingleton(irClass).instanceGetter.getValue(this,
```
SvyatoslavScherbina: Doesn't work for cache.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -262,35 +263,113 @@ private class DeclarationsGeneratorVisitor(override val context: Context) :
...
-                    addGlobal(shadowSymbolName, getLLVMType(irClass.defaultType), isExported = isExported, threadLocal = true)
                 }
+            }
+            return SingletonLlvmDeclarations(instanceGetter, addressGetter)
+        }
 
-        instanceShadowFieldRef?.let { LLVMSetInitializer(it, kNullObjHeaderPtr) }
+        val threadLocal = irClass.storageKind(context) == ObjectStorageKind.THREAD_LOCAL
+        val symbolName = "kobjref:" + qualifyInternalName(irClass)
+        val instanceAddress = addKotlinGlobal(symbolName, getLLVMType(irClass.defaultType), threadLocal = threadLocal)
```
SvyatoslavScherbina: It shouldn't be public.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -262,35 +263,113 @@ private class DeclarationsGeneratorVisitor(override val context: Context) :
...
+
+        val instanceShadowAddress = if (threadLocal) null else {
+            val shadowSymbolName = "kshadowobjref:" + qualifyInternalName(irClass)
+            addKotlinGlobal(shadowSymbolName, getLLVMType(irClass.defaultType), threadLocal = true)
```
SvyatoslavScherbina: Probably not required for permanent objects.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -935,39 +895,16 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
                 )
             }
         }
+
         val storageKind = irClass.storageKind(context)
-        if (storageKind == ObjectStorageKind.PERMANENT) {
-            return loadSlot(codegen.getObjectInstanceStorage(irClass, storageKind), false)
+        when (storageKind) {
+            ObjectStorageKind.SHARED -> context.llvm.sharedObjects += irClass
+            ObjectStorageKind.THREAD_LOCAL -> context.llvm.objects += irClass
```
SvyatoslavScherbina: Doesn't seem required anymore.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
```diff
@@ -262,35 +263,113 @@ private class DeclarationsGeneratorVisitor(override val context: Context) :
...
+        }
+
+        val instanceGetter = object: ValueAccess() {
+            override fun prepare(codegen: CodeGenerator) {
+                if (valueGetterName != null) {
+                    generateFunction(codegen, LLVMFunctionType(kObjHeaderPtr, cValuesOf(kObjHeaderPtrPtr), 1, 0)!!,
```
SvyatoslavScherbina: Generally `LlvmDeclarations` machinery isn't supposed to generate any _definitions_ (including function bodies). So it makes sense to move all these new instruction generators to `IrToBitcode` to keep it consistent with the existing code.

runtime/src/main/cpp/Memory.h
olonho: Well, this depends on how incremental we are, if codegen phase is being run on module as a whole - it shalln't be a problem.Think it's better describe the current contract in the signature, and fix it when working on IC.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/LlvmDeclarations.kt
olonho: Remnant of code where we have to run per-module init.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Well, it has to be rethought a bit.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -935,12 +892,33 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
                 )
             }
         }
+
+        // If object is imported - access it via getter function.
+        if (isExternal(irClass)) {
+            val valueGetterName = irClass.objectInstanceGetterSymbolName
+            val valueGetterFunction = LLVMGetNamedFunction(context.llvmModule, valueGetterName) ?:
```
SvyatoslavScherbina: Is `context.llvm.externalFunction` applicable here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -935,12 +891,33 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                LLVMAddFunction(context.llvmModule, valueGetterName,
+                        LLVMFunctionType(kObjHeaderPtr, cValuesOf(kObjHeaderPtrPtr), 1, 0))
```
SvyatoslavScherbina: `functionType` function is generally a "kotlinized" shortcut for `LLVMFunctionType`.

runtime/src/main/cpp/Runtime.cpp
olonho: I guess so, at least for deinit case, so decided to make it symmetrical.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Function above mentions in comments that it works with C++ functions and does related signext magic here, so decided to avoid using it here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Changed properly.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: It is actually used for Kotlin functions.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -751,10 +771,24 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+            val singleton = context.llvmDeclarations.forSingleton(declaration)
+            val access = singleton.instanceStorage
+            if (access is GlobalAddressAccess) {
+                LLVMSetInitializer(access.getAddress(null), if (declaration.storageKind(context) == ObjectStorageKind.PERMANENT)
+                    context.llvm.staticData.createConstKotlinObject(declaration,
+                            *computeFields(declaration)).llvm else codegen.kNullObjHeaderPtr)
+            }
+            // If can be exported and can be instantiated.
+            if (declaration.isExported() &&
+                    declaration.constructors.singleOrNull() { it.valueParameters.size == 0 } != null) {
```
SvyatoslavScherbina: Isn't this true for any `object` declaration?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -935,12 +891,33 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                        listOf(),
+                        resultLifetime = Lifetime.GLOBAL,
+                        exceptionHandler = exceptionHandler)
+        }
+
         val storageKind = irClass.storageKind(context)
-        if (storageKind == ObjectStorageKind.PERMANENT) {
-            return loadSlot(codegen.getObjectInstanceStorage(irClass, storageKind), false)
+        when (storageKind) {
+            ObjectStorageKind.SHARED -> context.llvm.sharedObjects += irClass
```
SvyatoslavScherbina: Somewhat harmful for DCE.
Now global deinitializer zeroes storage for all defined `public` singletons, even for unused ones.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -779,16 +813,16 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
             val type = codegen.getLLVMType(declaration.type)
-            val globalProperty = context.llvmDeclarations.forStaticField(declaration).storage
+            val globalPropertyAccess = context.llvmDeclarations.forStaticField(declaration).storageAddressAccess
             val initializer = declaration.initializer?.expression as? IrConst<*>
-            if (initializer != null)
-                LLVMSetInitializer(globalProperty, evaluateExpression(initializer))
-            else
-                LLVMSetInitializer(globalProperty, LLVMConstNull(type))
+            val globalProperty = (globalPropertyAccess as? GlobalAddressAccess)?.getAddress(null)
+            if (globalProperty != null) {
```
SvyatoslavScherbina:
```kotlin
@ThreadLocal
val x = "test"

fun main() {
    println(x)
}
```

->

```
Segmentation fault: 11
```

runtime/src/main/cpp/Runtime.cpp
SvyatoslavScherbina: IIRC, `::memoryState` is (and must be) available during e.g. `Initializer(DEINIT_THREAD_LOCAL_GLOBALS)`.

Question is: what happens if `memory != ::memoryState`? I guess these two must actually be equal. For me this means that `memory` parameter is kinda redundant here. At least until `Initializer` ensures that `memory` and `::memoryState` are consistent.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: No, that's why nightlies use to fall.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
olonho: Doesn't look like big deal.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
SvyatoslavScherbina: Could you provide an example of `object` declaration with `declaration.constructors.singleOrNull() { it.valueParameters.size == 0 } == null`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: Good catch, will fix. Strange that we have no tests on that.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -407,41 +417,50 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-                                if (it.initializer?.expression !is IrConst<*>?) {
-                                   if (it.storageKind == FieldStorageKind.THREAD_LOCAL) {
-                                       val initialization = evaluateExpression(it.initializer!!.expression)
-                                       val address = context.llvmDeclarations.forStaticField(it).storage
-                                       storeAny(initialization, address, false)
-                                    }
+                            .forEach { irField ->
+                                val expression = irField.initializer?.expression
+                                if (irField.storageKind == FieldStorageKind.THREAD_LOCAL) {
+                                    val initialization = evaluateExpression(irField.initializer!!.expression)
```
SvyatoslavScherbina: So it becomes suboptimal for constant-initialized thread-local non-reference fields.

olonho: Guess on many platform it's pretty much the same code executed by the system linker.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
olonho: PMed.

PR #3696, opened 2019-12-18
Klib expect actual 1.3.70

alexander-gorshenev:
This is a native counterpart of Kotlin fix for 1.3.70.
Initial support of expect/actuals in klibs.


alexander-gorshenev:
Merged by other means

PR #3697, opened 2019-12-18
Fix platform libs generation for watchos_arm64

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/ToolConfig.kt
```diff
@@ -39,7 +39,12 @@ class ToolConfig(userProvidedTargetName: String?, flavor: KotlinPlatform) {
     fun downloadDependencies() = platform.downloadDependencies()
 
     val defaultCompilerOpts =
-            platform.clang.targetLibclangArgs.toList()
+            // We compile with -O2 because Clang may insert inline asm in bitcode at -O0.
```
olonho: Maybe not the best place to add this option. Maybe move to some other place?

PR #3704, opened 2019-12-19
Mimalloc allocator integration

LepilkinaElena:


build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileCToBitcode.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.gradle.api.Action
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+import java.io.File
+import javax.inject.Inject
+
+open class CompileCToBitcode @Inject constructor(@InputDirectory val srcRoot: File,
```
olonho: Are you sure we don't have this functionality already?

runtime/build.gradle
```diff
@@ -29,6 +33,11 @@ targetList.each { targetName ->
         linkerArgs project.file("../common/build/$targetName/hash.bc").path
     }
 
+    tasks.create("${targetName}Mimalloc", CompileCToBitcode, file('src/mimalloc'), "mimalloc", targetName).configure {
+        excludeFiles.addAll(["**/alloc-override*.c", "**/page-queue.c", "**/static.c"])
+        excludedTargets.addAll(['wasm32', 'linux_mips32', 'linux_mipsel32', 'zephyr'])
```
olonho: Maybe explicit opt-in instead of opt-out?

runtime/src/opt_alloc/cpp/AllocImpl.cpp
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+#include <stdlib.h>
+#include <stdio.h>
+// Note that only C++ part of the runtime goes via those functions, Kotlin uses specialized versions.
```
olonho: Are you sure this comment is relevant here?

runtime/src/mimalloc/c/alloc-override.c
```diff
@@ -0,0 +1,197 @@
...
+#if (defined(__GNUC__) || defined(__clang__)) && !defined(__MACH__)
+#pragma GCC visibility push(default)
+#endif
+
+// ------------------------------------------------------
+// Override new/delete
+// This is not really necessary as they usually call
+// malloc/free anyway, but it improves performance.
+// ------------------------------------------------------
+#ifdef __cplusplus
```
olonho: Do we really need that? We override allocators anyway.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileCToBitcode.kt
LepilkinaElena: There is similar for compiling C++. it's possible to combine both to one and make language as property. Should I combine?

runtime/src/mimalloc/c/include/mimalloc-atomic.h
```diff
@@ -0,0 +1,241 @@
...
+      (defined(__x86_64__) || defined(__i386__) || defined(__arm__) || defined(__aarch64__))
+#if defined(__x86_64__) || defined(__i386__)
+  static inline void mi_atomic_yield(void) {
+    asm volatile ("pause" ::: "memory");
+  }
+#elif defined(__aarch64__)
+  static inline void mi_atomic_yield(void) {
+    asm volatile("yield");
+  }
+// <Jetbrains> Fix for armv6.
```
olonho: Better have if marked with some define, i.e. `#if KONAN_MI_MALLOC`.

runtime/build.gradle
LepilkinaElena: I did excluded list because it's smaller, I want to do this next as extension property of target, but I didn't want waiting for updating kotlin version.

runtime/src/mimalloc/c/alloc-aligned.c
```diff
@@ -0,0 +1,204 @@
+/* ----------------------------------------------------------------------------
```
olonho: Better add unmodified allocator in one commit and all our changes in the separate one, to make sure we can update the allocator code.

runtime/src/mimalloc/c/alloc-override.c
LepilkinaElena: This file isn't really needed, it's excluded from our build, but I save it for future updates of versions of allocator.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileCToBitcode.kt
olonho: Yes

runtime/build.gradle
olonho: You can have an extension property in K/N. We shall opt-in only to platforms we tested allocator upon.

LepilkinaElena: If I placed it in K/N I can't use it both in K/N and Gradle.
I'll have to repeat same twice.

LepilkinaElena: I haven't managed to make it as extension, can't get it from gradle then.

LepilkinaElena: What I  plan https://github.com/JetBrains/kotlin/pull/2911
https://github.com/JetBrains/kotlin-native/pull/3709

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @OutputFile
+    val outFile = File(targetDir, "${folderName}.bc")
+
+    @TaskAction
+    fun compile() {
+        if (target in excludedTargets) return
+        objDir.mkdirs()
+        val plugin = project.convention.getPlugin(ExecClang::class.java)
+        var executable = "clang++"
```
olonho: More readable and Kotlin-style would be 
```val (executable, defaultFlags, srcFilesPatterns) = if (language == Language.C) 
   Triple("clang",  "...", "...") else  Triple("clang++",  "...", "...")
```

runtime/build.gradle
```diff
@@ -24,63 +25,52 @@ targetList.each { targetName ->
...
-        name "launcher"
-        srcRoot file('src/launcher')
-        target targetName
-        includeRuntime(delegate)
+    tasks.create("${targetName}Mimalloc", CompileToBitcode, file('src/mimalloc'), "mimalloc", targetName).configure {
+        language = CompileToBitcode.Language.C
+        excludeFiles.addAll(["**/alloc-override*.c", "**/page-queue.c", "**/static.c"])
+        excludedTargets.addAll(['wasm32', 'linux_mips32', 'linux_mipsel32', 'zephyr'])
+        srcDir = new File(srcRoot, "c")
+        compilerArgs.add("-DKONAN_MI_MALLOC")
```
olonho: Better `-DKONAN_MI_MALLOC=1`.

runtime/src/mimalloc/c/alloc-override.c
olonho: Better use `#if !KONAN_MI_MALLOC` here to make sure it's not accidentally get included into the build during refactorings.

runtime/src/mimalloc/c/include/mimalloc-internal.h
```diff
@@ -457,14 +457,27 @@ static inline uintptr_t _mi_thread_id(void) mi_attr_noexcept {
...
 static inline uintptr_t _mi_thread_id(void) mi_attr_noexcept {
   uintptr_t tid;
   #if defined(__i386__)
   __asm__("movl %%gs:0, %0" : "=r" (tid) : : );  // 32-bit always uses GS
   #elif defined(__MACH__)
-  __asm__("movq %%gs:0, %0" : "=r" (tid) : : );  // x86_64 macOS uses GS
+  #if defined(KONAN_MI_MALLOC)
+  #include <TargetConditionals.h>
+    #if TARGET_OS_EMBEDDED // iOS/tvOS/watchOS devices.
+      tid = pthread_self();
```
olonho: At least on 64-bit iOS x18 seems to be TLS register (https://stackoverflow.com/questions/28109826/arm64-using-gas-on-ios), but not sure if we rely on that here or not.

LepilkinaElena: Thank you

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTargetExtenstions.kt
```diff
@@ -4,4 +4,20 @@ fun KonanTarget.supportsCodeCoverage(): Boolean =
         this == KonanTarget.MINGW_X64 ||
         this == KonanTarget.LINUX_X64 ||
         this == KonanTarget.MACOS_X64 ||
-        this == KonanTarget.IOS_X64
\ No newline at end of file
+        this == KonanTarget.IOS_X64
+
+fun KonanTarget.supportsOptimizedAllocator(): Boolean =
```
sbogolepov: `supportsOptimizedAllocator` is a bit unclear. Maybe `supportsMimallocAllocatior`?

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
```diff
@@ -0,0 +1,96 @@
...
+        val (executable, defaultFlags, srcFilesPatterns) =
+                if (language == Language.C)
+                    Triple("clang",
+                    commonFlags + listOf("-std=gnu11", "-O3", "-Wall", "-Wextra", "-Wno-unknown-pragmas",
+                            "-ftls-model=initial-exec"),
+                            listOf("**/*.c"))
+                else
+                    Triple("clang++",
+                            commonFlags + listOf("-std=c++14", "-Werror", "-O2") +
+                                    if (!HostManager().targetByName(target).isMINGW) listOf("-fPIC") else emptyList(),
```
sbogolepov: `listOfNotNull("-std=c++14", "-Werror", "-O2", "-fPIC".takeIf { !HostManager().targetByName(target).isMINGW })` is a bit more compact.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
```diff
@@ -0,0 +1,96 @@
...
+    val outFile = File(targetDir, "${folderName}.bc")
+
+    @TaskAction
+    fun compile() {
+        if (target in excludedTargets) return
+        objDir.mkdirs()
+        val plugin = project.convention.getPlugin(ExecClang::class.java)
+        val commonFlags = listOf("-c", "-emit-llvm", "-I$headersDir")
+        val (executable, defaultFlags, srcFilesPatterns) =
+                if (language == Language.C)
```
sbogolepov: Exhaustive `when` over `language` will be more robust.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
```diff
@@ -0,0 +1,96 @@
...
+                    Triple("clang",
+                    commonFlags + listOf("-std=gnu11", "-O3", "-Wall", "-Wextra", "-Wno-unknown-pragmas",
```
sbogolepov: Why `-O3` for C and `-O2` for C++? Also why there is no `-fPIC` here?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTargetExtenstions.kt
LepilkinaElena: I don't want to connect this with exact allocator, there is possibility that we change or add another one.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
LepilkinaElena: I tried to save options that are set in build scripts for allocator, it's already tuned. I thought about `-fPIC`, and I understood it wasn't used because of some assembly parts for some platforms. Assembly code  isn't written in proper way for position independent code.
If you see real problems with this I need to turn on and test again all platforms it influences.

LepilkinaElena: It would be easier to update versions if we save settings, because there are some hacks to speed up allocator.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTargetExtenstions.kt
sbogolepov: The problem is that the set of supported targets is determined by allocator itself, not target. The set for `jemalloc` or any other allocator may be different. 
Another possible solution is to add enum of alternative allocators and its entry to `supportsOptimizedAllocator`.

sbogolepov: Another good thing is to describe why other targets are not supported.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
sbogolepov: > I tried to save options that are set in build scripts for allocator, it's already tuned.

Then it is better to mention it with a comment.

> If you see real problems with this I need to turn on and test again all platforms it influences.

There was an Android target that wasn't compiling because of `-fPIC` absence, but I don't remember which one :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -111,11 +104,25 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
 
     val shouldCoverSources = configuration.getBoolean(KonanConfigKeys.COVERAGE)
     val shouldCoverLibraries = !configuration.getList(KonanConfigKeys.LIBRARIES_TO_COVER).isNullOrEmpty()
+    // TODO: Replace to KonanTarget in Kotlin repo as extension property.
+    private val targetsWithoutOptAllocMode = listOf("wasm32", "linux_mips32", "linux_mipsel32", "zephyr")
```
sbogolepov: Dead code.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -153,6 +153,7 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
                 put(LIGHT_DEBUG, arguments.lightDebug)
                 put(STATIC_FRAMEWORK, selectFrameworkType(configuration, arguments, outputKind))
                 put(OVERRIDE_CLANG_OPTIONS, arguments.clangOptions.toNonNullList())
+                put(ALLOCATION_MODE, arguments.allocationMode)
```
sbogolepov: "Allocation mode" is somewhat unclear term. Maybe use more common "allocator" instead?

backend.native/tests/build.gradle
```diff
@@ -317,8 +317,8 @@ def createTestTasks(File testRoot, Class<Task> taskType, Closure taskConfigurati
 
 void dependsOnPlatformLibs(Task t) {
     if (!useCustomDist) {
-        def testTarget = project.testTarget
-        if (testTarget != null && testTarget != project.hostName) {
+        def testTarget = project.target
```
sbogolepov: Why this change is needed?

LepilkinaElena: It was for running raspberry-pi tests. @PavelPunegov should commit sane changes on master.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
LepilkinaElena: I built some projects with native Android, if we meet a problem then will be a time to decide. I think without exact use case it's better to save default options set.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTargetExtenstions.kt
LepilkinaElena: Yes, it's a good idea with enum, but now it would be strange to have enumeration with one allocator. I don't want to show name of allocator in compiler options. Now we have just std and opt. It seems that same names everywhere are better.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt
```diff
@@ -116,6 +109,18 @@ class KonanConfig(val project: Project, val configuration: CompilerConfiguration
         add(if (debug) "debug.bc" else "release.bc")
         add(if (memoryModel == MemoryModel.STRICT) "strict.bc" else "relaxed.bc")
         if (shouldCoverLibraries || shouldCoverSources) add("profileRuntime.bc")
+        if (configuration.get(KonanConfigKeys.ALLOCATION_MODE) == "mimalloc") {
+            if (!target.supportsMimallocAllocator()) {
+                configuration.report(CompilerMessageSeverity.STRONG_WARNING,
+                        "Optimized allocation mode isn't supported on target ${target.name}. Used standard mode.")
```
sbogolepov: Optimized allocation mode -> "mimalloc allocator".

PR #3709, opened 2019-12-20
Use property of KonanTarget to filter targets which shouldn't use optimized allocator

LepilkinaElena:


LepilkinaElena:
Next step of https://github.com/JetBrains/kotlin-native/pull/3704

runtime/build.gradle
```diff
@@ -32,7 +33,8 @@ targetList.each { targetName ->
     tasks.create("${targetName}Mimalloc", CompileToBitcode, file('src/mimalloc'), "mimalloc", targetName).configure {
         language = CompileToBitcode.Language.C
         excludeFiles.addAll(["**/alloc-override*.c", "**/page-queue.c", "**/static.c"])
-        excludedTargets.addAll(['wasm32', 'linux_mips32', 'linux_mipsel32', 'zephyr'])
+        if (!new HostManager().targetByName(targetName).supportsOptimizedAllocator)
```
olonho: It doesn't look right at all. Knowledge if allocator is supported leaks to the wrong place. Please make it extension/function/whatever in K/N repo.

PR #3711, opened 2019-12-21
Fixes for cache bundling

ilmat192:


tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanCacheTask.kt
```diff
@@ -20,11 +18,16 @@ open class KonanCacheTask: DefaultTask() {
...
-    lateinit var cacheDirectory: File
+    lateinit var cacheRoot: File
 
     @get:Input
     lateinit var target: String
 
+    @get:Internal
+    // TODO: Reuse NativeCacheKind from Big Kotlin plugin when it is available.
+    val cacheDirectory: File
+        get() = cacheRoot.resolve("$target-g$cacheKind")
```
olonho: Is it consistent with the current naming which is `$target-g`?

ilmat192: What do you mean by consistency in this case? If you speak about compatibility with the current naming, then no, this patch requires the [Igor's fix](https://github.com/JetBrains/kotlin/commit/ce2446f80fac6ec088eedd32c7b2e6cfae012e8c) to be merged.

PR #3733, opened 2019-12-27
Do not use zipTree in CollisionDetector task

ilya-g:
This avoids unpacking all the dependency jars and significantly speeds up the task.


LepilkinaElena:
Thank you! Checked locally, works fine.

PR #3734, opened 2019-12-30
[Interop] Fix staticCFunction with a reference to C function.

sbogolepov:
This commit makes possible things like `staticCFunction(::malloc)`.
The problem is two different `staticCFunction(::malloc)` will produce different results.


SvyatoslavScherbina:
Actually this approach is still useful to support explicit operation of taking pointer to original C function itself.
E.g. `::foo.cptr`, which is `COpaquePointer` and semantically different from `staticCFunction(::foo)`.
Somewhat important for pointers to variadic C functions.

PR #3738, opened 2019-12-30
Update performance tasks: added caches to default mode for running compile-only benchmark + replace deprecated flag

LepilkinaElena:


build-tools/src/main/kotlin/org/jetbrains/kotlin/MPPTools.kt
```diff
@@ -128,14 +129,19 @@ fun mergeReports(reports: List<File>): String {
...
-fun getCompileOnlyBenchmarksOpts(project: Project, defaultCompilerOpts: List<String>) =
-        (project.findProperty("nativeBuildType") as String?)?.let {
-            if (it.equals("RELEASE", true))
-                listOf("-opt")
-            else if (it.equals("DEBUG", true))
-                listOf("-g")
-            else listOf()
-        } ?: defaultCompilerOpts
+fun getCompileOnlyBenchmarksOpts(project: Project, defaultCompilerOpts: List<String>): List<String> {
+    val dist = project.file(project.findProperty("kotlin.native.home") ?: "dist")
```
PavelPunegov: There is a project externsion property `kotlinNativeDist` that can be used here to reduce amount of same properties to be searched for.

build-tools/src/main/kotlin/org/jetbrains/kotlin/MPPTools.kt
```diff
@@ -128,14 +129,19 @@ fun mergeReports(reports: List<File>): String {
...
+    val useCache =  project.findProperty("disableCompilerCaches")?.let { false } ?: true
```
PavelPunegov: Looks like the same as ` = ! project.hasProperty("disableCompilerCaches")`

build-tools/src/main/kotlin/org/jetbrains/kotlin/MPPTools.kt
LepilkinaElena: I can't use it, because 1) there is an old deprecated name, 2) I need guarantee that I don't use `konan.home`, because it's part of flow with multiplatform plugin and I should have same behaviour.

PR #3742, opened 2020-01-02
Shared cyclic garbage collector

olonho:


runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -78,6 +84,14 @@ object GC {
         get() = getTuneThreshold()
         set(value) = setTuneThreshold(value)
 
+
+    /**
+     * If cyclic collector for atomic references to be deployed.
+     */
+    var cyclicCollector: Boolean
```
homuroll: Why collector is of Boolean type?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -823,6 +827,8 @@ bool Worker::park(KLong timeoutMicroseconds, bool process) {
 }
 
 JobKind Worker::processQueueElement(bool blocking) {
+  // Allow GC safepointing check.
+  GC_RendezvouzCallback(this);
```
homuroll: RendezvouS

runtime/src/main/cpp/Worker.cpp
```diff
@@ -259,11 +259,14 @@ class State {
   }
 
   Worker* addWorkerUnlocked(bool errorReporting, KRef customName) {
```
homuroll: Why the function is called *****Unlocked while there is a lock used in it?

runtime/src/main/cpp/Worker.cpp
```diff
@@ -259,11 +259,14 @@ class State {
...
-    Locker locker(&lock_);
-    Worker* worker = konanConstructInstance<Worker>(nextWorkerId(), errorReporting,
-        customName);
-    if (worker == nullptr) return nullptr;
-    workers_[worker->id()] = worker;
+    Worker* worker = nullptr;
```
homuroll: Why the function is called *****Unlocked while there is a lock used in it?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+           });
+           if (atomicGet(&mutatedAtomics_) != 0) {
+             COLLECTOR_LOG("restarted during visiting collect\n")
+             restartCount++;
+             goto restart;
+           }
+           while (toVisit.size() > 0)  {
+             auto* obj = toVisit.front();
+             toVisit.pop_front();
+             COLLECTOR_LOG("visit %p\n", obj);
```
SvyatoslavScherbina: Despite the atomic mutation notification, dangling pointer is still possible here.

Consider the following test case:
```kotlin

fun createRef(): AtomicReference<Any?> {
    return AtomicReference(Any().freeze())
}

fun zeroRef(ref: AtomicReference<Any?>) {
    ref.value = null
}
  
fun main() {
    val atomic = createRef()
    kotlin.native.internal.GC.collectCyclic()
    Worker.current.park(1000 * 1000)
    zeroRef(atomic)
}
```
If `sleep(2)` was inserted above this line, then the object would be visited here after being freed.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+               if (ref != nullptr && visited.count(ref) == 0) {
+                 toVisit.push_back(ref);
+               }
+             });
+           }
+         }
+         // Now find all elements with external references, and mark objects reachable from them as non suitable
+         // for collection by setting their side reference count to -1.
+         toVisit.clear();
+         for (auto it: sideRefCounts_) {
```
SvyatoslavScherbina: The entire loop is not properly synchronized with `objContainer->refCount()` mutation.
For example:
```kotlin
  
fun createRef(): AtomicReference<Any?> {
    val atomic1 = AtomicReference<Any?>(null)
    val atomic2 = AtomicReference<Any?>(null)
    atomic1.value = atomic2
    atomic2.value = atomic1
    return atomic1
}

class Holder(var value: AtomicReference<Any?>) {
    fun switch() {
        value = value.value as AtomicReference<Any?>
    }
}

fun createHolder() = Holder(createRef())

fun main() {
    val holder = createHolder()
    kotlin.native.internal.GC.collectCyclic()
    Worker.current.park(1500 * 1000)
    holder.switch()
    Worker.current.park(1500 * 1000)
    Worker.start().executeAfter(0L, {
    }.freeze())
    Worker.current.park(1000)
    println("VALUE: " + holder.value.value)
}  
```
If `sleep(1)` was added to the beginning of the loop's body, then this code would print `VALUE: ... -> null`, indicating that one of the references got nullified.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -823,6 +827,8 @@ bool Worker::park(KLong timeoutMicroseconds, bool process) {
 }
 
 JobKind Worker::processQueueElement(bool blocking) {
+  // Allow GC safepointing check.
```
SvyatoslavScherbina: Not sure if current approach is essentially related to safe points.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+        traverseObjectFields(it, [](ObjHeader** location) {
+          ZeroHeapRef(location);
+        });
+      }
+      toRelease_.clear();
+      atomicSet(&pendingRelease_, 0);
+    }
+  }
+
+  void rendezvouz(void* worker) {
```
SvyatoslavScherbina: Have you considered the option to run the code below in specific dedicated thread?
E.g. the GC thread itself (which would then require `::memoryState` initialized).
This could simplify interaction with GC thread and make the entire algorithm more independent (including its guarantees).

runtime/src/main/cpp/Worker.cpp
```diff
@@ -283,6 +286,7 @@ class State {
         workers_.erase(it);
       }
     }
+    GC_UnregisterWorker(worker);
```
SvyatoslavScherbina: So the worker gets unregistered when it is already invalid.
Doesn't probably matter with current "registry" implementation, but doesn't look correct semantically anyway.
Also it is not symmetrical to `addWorkerUnlocked` which registers the worker in GC only after it becomes fully valid.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+    // TODO: we can only add root when collector is not processing, which looks like a limitation,
+    //  instead we can add elements to the side buffer or have a separate lock for that.
+    Locker lock(&lock_);
+    rootset_.insert(obj);
+  }
+
+  void removeRoot(ObjHeader* obj) {
+    COLLECTOR_LOG("remove root %p\n", obj);
+    // Note that we can only remove root when the collector is not processing.
+    Locker lock(&lock_);
```
SvyatoslavScherbina: Can this block the main thread?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+         atomicSet(&gcRunning_, 1);
+         restartCount = 0;
+        restart:
+         if (restartCount > 10) {
+           COLLECTOR_LOG("wait for some time to avoid GC trashing\n");
+           struct timeval tv;
+           struct timespec ts;
+           long long nsDelta = 1000LL * 1000LL;
+           ts.tv_nsec = (tv.tv_usec * 1000LL + nsDelta) % 1000000000LL;
+           ts.tv_sec = (tv.tv_sec * 1000000000LL + nsDelta) / 1000000000LL;
```
SvyatoslavScherbina: Is this correct?
Isn't `(tv.tv_sec * 1000000000LL + nsDelta) / 1000000000LL` always equal to `tv.tv_sec`?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+               toVisit.push_back(ref);
+               COLLECTOR_LOG("adding %p for visiting\n", ref);
+             }
+           });
+           if (atomicGet(&mutatedAtomics_) != 0) {
+             COLLECTOR_LOG("restarted during visiting collect\n")
+             restartCount++;
+             goto restart;
+           }
+           while (toVisit.size() > 0)  {
```
SvyatoslavScherbina: `sideRefCounts_` computation doesn't seem correct to me:
The following code produces the leak:
```kotlin

class Node(val value: Any?)

fun createGarbage() {
    val atomic = AtomicReference<Any?>(null)
    atomic.value = Pair(atomic, Node(atomic)).freeze()
}

fun main() {
    createGarbage()
}
```

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+             auto* obj = toVisit.front();
+             toVisit.pop_front();
+             COLLECTOR_LOG("visit %p\n", obj);
+             auto* objContainer = obj->container();
+             if (objContainer == nullptr) continue;  // Permanent object.
+             if (!objContainer->frozen()) continue;
```
SvyatoslavScherbina: Can `objContainer->frozen()` be actually `false`? If not, then this should be an assertion.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+           }
+         }
+         // Now find all elements with external references, and mark objects reachable from them as non suitable
+         // for collection by setting their side reference count to -1.
+         toVisit.clear();
+         for (auto it: sideRefCounts_) {
+           auto* obj = it.first;
+           auto* objContainer = obj->container();
+           if (objContainer == nullptr) continue;  // Permanent object.
+           if (it.second != objContainer->refCount()) {
```
SvyatoslavScherbina: Maybe add an assertion that `it.second <= objContainer->refCount()`?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,435 @@
...
+         CHECK_CALL(pthread_cond_wait(&cond_, &lock_), "Cannot wait collector condition")
+         if (!shallRunCollector_) continue;
+         atomicSet(&gcRunning_, 1);
+         restartCount = 0;
+        restart:
+         if (restartCount > 10) {
+           COLLECTOR_LOG("wait for some time to avoid GC trashing\n");
+           struct timeval tv;
+           struct timespec ts;
+           long long nsDelta = 1000LL * 1000LL;
```
SvyatoslavScherbina: Maybe use larger timeout for later attempts?

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Yes, but actually I find that initiating GC checks bound to the event processing seems better reflect event-driven program behavior.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Yes, modifications to atomic rootset are blocking wrt GC lock, see no way of fixing it.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: This and other races likely to be fixed by adding sync point taking GC lock in local GC, if GC is in progress.

runtime/src/main/cpp/Worker.cpp
olonho: Reworked.

runtime/src/main/cpp/Worker.cpp
olonho: What do you mean by `invalid`? Object is still correct, it just cannot be discovered by id.

runtime/src/main/cpp/CyclicCollector.cpp
Takhion: Memory-wise, how big does the rootset data structure get, realistically?

Would it be possible to post all modifications as events in a non-blocking queue, and rebuild the state on the side, asynchronously, to avoid any locking whatsoever?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,472 @@
...
+           ts.tv_nsec = (tv.tv_usec * 1000LL + nsDelta) % 1000000000LL;
+           ts.tv_sec = tv.tv_sec;
```
SvyatoslavScherbina: What I [meant](https://github.com/JetBrains/kotlin-native/pull/3742#discussion_r370590727) is that it should probably take `tv_usec` into account.
The same bug can likely be found with other `pthread_cond_timedwait` usages across our codebase.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -361,6 +361,11 @@ class CyclicCollector {
     CHECK_CALL(pthread_cond_signal(&cond_), "Cannot signal collector")
   }
 
+  void localGC() {
+    // We just need to take GC lock here, to avoid release of object we walk on.
+    Locker locker(&lock_);
```
SvyatoslavScherbina: This and other cases of taking GC lock could block the UI thread for long.
Isn't it enough to notify GC with a flag like in `mutateRoot`?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,472 @@
...
+               goto restart;
+             }
+             if (visited.count(obj) == 0) {
+               visited.insert(obj);
+               traverseObjectFields(obj, [&toVisit, &visited, obj](ObjHeader** location) {
+                 ObjHeader* ref = *location;
+                 if (ref != nullptr) {
+                   // We shall not account for edges inside the same frozen container, unless it originates
+                   // from an atomic reference.
+                   bool shallCount = isAtomicReference(obj) || (obj->container() != ref->container());
```
SvyatoslavScherbina: Why not increment `sideRefCounts_` here instead?
(Of course after placing roots to `toVisit` instead of traversing their fields [here](https://github.com/JetBrains/kotlin-native/pull/3742/files#diff-2f35eb64c8651572916c7aeb9cea20d7R199)).
This seems to be simpler.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Linear by size of all data rooted in atomic references. Not sure about feasibility of the proposal, it's somewhat too heavyweight to make the whole system transactional.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Yeah, will fix all such place in separate PR.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,447 @@
...
+#endif
+
+/**
+ * Theory of operations:
+ *
+ * Kotlin/Native runtime has concurrent cyclic garbage collection for the shared mutable objects,
+ * such as `AtomicReference` and `FreezableAtomicReference` instances (further known as the atomic rootset).
+ * We perform such analysis by iterating over the transitive closure of the atomic rootset, and computing
+ * aggregated inner reference counter for rootset elements over this transitive closure.
+ * Collector runs in its own thread and is started by an explicit request or after certain time interval since last
```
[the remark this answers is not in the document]
olonho: There's API to activate collector, could be called from the callback.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Yeah, will fix there separately.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Maybe, but first I want to be provably correct.

olonho: Added TODO.

runtime/src/main/cpp/CyclicCollector.cpp
Takhion: But the only part we're interested in accessing lock-free are the _pointers/references_ to the data, not the data itself, no? Wouldn't that be a lot smaller? Unless those are not stable...?

runtime/src/main/cpp/CyclicCollector.cpp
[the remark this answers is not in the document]
Takhion: Can users invoke the API directly, or do we have to provide such implementation of the callback?

runtime/src/main/cpp/Worker.cpp
olonho: Just a convention.

runtime/src/main/cpp/CyclicCollector.cpp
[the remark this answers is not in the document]
olonho: Sure, `kotlin.native.internal.GC.collectCyclic()`.

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Well, you could probably try prototype to describe this idea, I'm not seeing how it could work so far.

runtime/src/main/cpp/CyclicCollector.cpp
SvyatoslavScherbina: A similar race may still be possible when using local variable instead of `Holder`, since local GC's stack and buffer processing is not actually mutually exclusive with this loop.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,460 @@
...
+    if (shallCollectGarbage_) return true;
+    auto delta = tick - atomicGet(&lastTick_);
+    if (delta > 10 || delta < 0) {
+      auto currentTimestampUs = konan::getTimeMicros();
+      if (currentTimestampUs - atomicGet(&lastTimestampUs_) > 10000) {
+        // TODO: maybe take another lock.
+        Locker locker(&lock_);
+        lastTick_ = currentTick_;
+        lastTimestampUs_ = currentTimestampUs;
+        shallCollectGarbage_ = true;
```
SvyatoslavScherbina: What is this flag for? Is it ever set to `false`?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,460 @@
...
+         atomicSet(&gcRunning_, 1);
+         restartCount = 0;
+        restart:
+         if (restartCount > 10) {
+           COLLECTOR_LOG("wait for some time to avoid GC trashing\n");
+           struct timeval tv;
+           struct timespec ts;
+           long long nsDelta = 1000LL * 1000LL * (restartCount - 10);
+           ts.tv_nsec = (tv.tv_usec * 1000LL + nsDelta) % 1000000000LL;
+           ts.tv_sec = (tv.tv_sec * 1000000000LL + tv.tv_usec * 1000000LL + nsDelta) / 1000000000LL ;
```
SvyatoslavScherbina: > `tv.tv_usec * 1000000LL`

Why `1000000LL`?

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,460 @@
...
+         }
+         atomicSet(&mutatedAtomics_, 0);
+         visited.clear();
+         toVisit.clear();
+         sideRefCounts_.clear();
+         for (auto* root: rootset_) {
+           // We only care about frozen values here, as only they could become part of shared cycles.
+           if (!root->container()->frozen()) continue;
+           COLLECTOR_LOG("process root %p\n", root);
+           traverseObjectFields(root, [&toVisit, &visited](ObjHeader** location) {
```
SvyatoslavScherbina: Why not just put `root` into `toVisit`?

SvyatoslavScherbina:
```kotlin
  
class Node(val value: Any?)

fun createRoot(): AtomicReference<Any?> {
    val ref1 = AtomicReference<Any?>(null)
    val ref2 = AtomicReference<Any?>(null)

    ref1.value = Node(ref2).freeze()
    ref2.value = Any().freeze()

    return ref1
}

fun main() {
    val ref1 = createRoot()

    kotlin.native.internal.GC.collectCyclic()
    Worker.current.park(500 * 1000L)

    Worker.start().executeAfter(0L, {}.freeze())
    Worker.current.park(500 * 1000L) 

    val node = ref1.value as Node
    val ref2 = node.value as AtomicReference<Any?>
    println(ref2.value)
}
```
->
`null`

runtime/src/main/cpp/CyclicCollector.cpp
olonho: Indeed, that's microseconds, not milliseconds.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,450 @@
...
+           if (objContainer == nullptr) continue;  // Permanent object.
+           RuntimeCheck(objContainer->shareable(), "Must be shareable");
+           sideRefCounts_[obj] = -1;
+           visited.insert(obj);
+           if (atomicGet(&mutatedAtomics_) != 0) {
+             COLLECTOR_LOG("restarted during reachable visit\n")
+             restartCount++;
+             goto restart;
+           }
+           traverseObjectFields(obj, [&toVisit, &visited](ObjHeader** location) {
```
SvyatoslavScherbina: One more data race (reproducible with sleeps as in https://github.com/JetBrains/kotlin-native/commit/e0d016b13287ad1c02d12c3e575de823475f5fea):
```kotlin
  
class Node(val value: Any?)

fun createRoot(): AtomicReference<Any?> {
    val ref1 = AtomicReference<Any?>(null)
    val ref2 = AtomicReference<Any?>(null)

    ref1.value = Node(ref2).freeze()
    ref2.value = Any().freeze()

    return ref1
}

fun main() {
    val ref1 = createRoot()
    kotlin.native.internal.GC.collectCyclic()

    Worker.current.park(100 * 1000L)
    val node = ref1.value as Node
    ref1.value = null
    Worker.current.park(500 * 1000L)

    Worker.start().executeAfter(0L, {
    }.freeze())
    Worker.current.park(500 * 1000L)

    val ref2 = node.value as AtomicReference<Any?>
    println("VALUE: ${ref2.value}")
}
```

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,450 @@
...
+                ObjHeader* ref = *location;
+                if (ref != nullptr) {
+                  COLLECTOR_LOG("object field %p in %p\n", ref, obj);
+                  // We shall not account for edges inside the same frozen container, unless it originates
+                  // from an atomic reference.
+                  if (isAtomicReference(obj) || (obj->container() != ref->container())) {
+                    COLLECTOR_LOG("counting %p -> %p\n", obj, ref)
+                    sideRefCounts_[ref]++;
+                  } else {
+                    COLLECTOR_LOG("not counting %p -> %p\n", obj, ref)
```
SvyatoslavScherbina: All traversed objects should be present in `sideRefCounts_`.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,450 @@
...
+  int gcRunning_;
+  int mutatedAtomics_;
+  int pendingRelease_;
+  bool shallRunCollector_;
+  bool terminateCollector_;
+  int32_t currentTick_;
+  int32_t lastTick_;
+  int64_t lastTimestampUs_;
+  void* mainWorker_;
+  KStdUnorderedMap<ObjHeader*, int> sideRefCounts_;
```
SvyatoslavScherbina: Side ref counts should be per-container.

olonho: No, as container can be changed during freezing.

SvyatoslavScherbina: Leak reproducer:
```kotlin

fun array(size: Int) = Array<Any?>(size, { null })

fun createGarbage() {
    val ref = AtomicReference<Any?>(null)
    val obj1 = array(2)
    val obj2 = array(1)
    val obj3 = array(2)

    obj1[0] = obj2
    obj1[1] = obj3

    obj2[0] = obj3

    obj3[0] = obj2
    obj3[1] = ref

    ref.value = obj1.freeze()
}

fun main() {
    createGarbage()
}
```

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,475 @@
...
+           auto* objContainer = obj->container();
+           if (objContainer == nullptr) continue;  // Permanent object.
+           int refCount;
+           // If object is in aggregated container - sum up RC for all elements.
+           if (objContainer->objectCount() != 1) {
+             RuntimeAssert(objContainer->frozen(), "Must be frozen aggregate");
+             ContainerHeader** subContainer = reinterpret_cast<ContainerHeader**>(objContainer + 1);
+             refCount = 0;
+             for (int i = 0; i < objContainer->objectCount(); ++i) {
+                 auto* componentObj = reinterpret_cast<ObjHeader*>((*subContainer) + 1);
```
SvyatoslavScherbina: Isn't it the same for every `i`?

olonho: Indeed :).

runtime/src/main/cpp/CyclicCollector.cpp
SvyatoslavScherbina: Kinda reproducer with https://github.com/JetBrains/kotlin-native/commit/765358085afc70a705d83624595e6ca7a1d48e7d 

Doesn't fire in 100% of cases because depends on order of elements in hash table.

```kotlin
import kotlin.native.internal.GC
  
class Node(val value: Any?)

fun createNode1(): Node {
    val ref = AtomicReference<Any?>(null)
    val node2 = Node(ref)
    val node1 = Node(node2)
    ref.value = node1.freeze()

    return node1
}

fun getNode2(): Node {
    val node1 = createNode1()
    GC.collect()

    return node1.value as Node
}   

fun main() {
    val worker = Worker.start()
    
    val node2 = getNode2()
    worker.executeAfter(10 * 1000L, { GC.collectCyclic() }.freeze())

    GC.collect()

    Worker.current.park(50 * 1000L)

    worker.execute(TransferMode.SAFE, {}, {}).result
    
    val ref = node2.value as AtomicReference<Any?>
    check(ref.value != null)
}
```

backend.native/tests/runtime/memory/cycle_collector.kt
```diff
@@ -0,0 +1,187 @@
...
+
+fun getNode2(): Holder {
+    val node1 = createNode1()
+    GC.collect()
+
+    return node1.other as Holder
+}
+
+fun test9() {
+    val worker = Worker.start()
```
SvyatoslavScherbina: This worker isn't terminated.
IIRC, in this case leak detector is skipped during program termination, which is undesirable, considering the tests for leaks above.
Same for `test7`.

runtime/src/main/cpp/CyclicCollector.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+inline bool isAtomicReference(ObjHeader* obj) {
+  return (obj->type_info()->flags_ & TF_LEAK_DETECTOR_CANDIDATE) != 0;
+}
+
+#define CHECK_CALL(call, message) RuntimeCheck((call) == 0, message)
+
+class CyclicCollector {
+  pthread_mutex_t lock_;
+  pthread_mutex_t timestampLock_;
```
SvyatoslavScherbina: It may require initialization.

PR #3751, opened 2020-01-10
[debugger][binding] fix issue with resolving theStringTypeInfo symbol

vvlevchenko:


llvmDebugInfoC/src/scripts/konan_lldb.py
```diff
@@ -49,11 +49,23 @@ def evaluate(expr):
...
+        address = candidate.GetStartAddress().GetLoadAddress(target)
+        log(lambda: "{} {:#x}".format(name, address))
+        return address
+
 def is_instance_of(addr, typeinfo):
     return evaluate("(bool)IsInstance({}, {})".format(addr, typeinfo)).GetValue() == "true"
 
 def is_string_or_array(value):
-    return evaluate("(bool)IsInstance({0}, theStringTypeInfo) ? 1 : ((int)Konan_DebugIsArray({0}) ? 2 : 0)".format(lldb_val_to_ptr(value))).unsigned
+    return evaluate("(bool)IsInstance({0}, {1}) ? 1 : ((int)Konan_DebugIsArray({0}) ? 2 : 0)".format(lldb_val_to_ptr(value), _symbol_loaded_address('ktype:kotlin.String'))).unsigned
```
kishmakov: As I get it, initially `val to ptr` was checked for being of type `theStringTypeInfo` (String?), but now it is checked against address of kotlin.String. Does it prevously process strings as general objects because of this?

vvlevchenko: Well theStringTypeInfo points exactly to ktype:kotlin.String, I found that with new lldb for 'theStringTypeInfo' returned two entries in debugger(in new lldb) (w/o any reasons in DI we have generate only one and binary form contains only one) that confuse evaluation but doesn't break with error. new implementation uses more prone variant of symbol resolving taken from llvm examples for resolving globals (they're also removes duplicates like it done in __symbol_loaded_address)

PR #3762, opened 2020-01-17
Updated kotlinx.cli to release version

LepilkinaElena:


LepilkinaElena:
> I hope we'll automate updating sources with git submodules or some another way soon.

I hope that we'll create right mechanism of endorsed libraries.

PR #3765, opened 2020-01-20
Fixes of mimalloc warnings

LepilkinaElena:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -243,7 +243,7 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     var clangOptions: Array<String>? = null
 
     @Argument(value="-Xallocator", valueDescription = "std | mimalloc", description = "Allocator used in runtime")
-    var allocator: String = "std"
+    var allocator: String = "mimalloc"
```
LepilkinaElena: I'll revert this. Need for testing of mimalloc on TC

PR #3768, opened 2020-01-21
Avoid Valgrind message due to glibc bug.

olonho:


runtime/src/main/cpp/Porting.cpp
```diff
@@ -140,7 +140,18 @@ static void onThreadExitCallback(void* value) {
   pthread_setspecific(terminationKey, nullptr);
 }
 
+#if KONAN_LINUX
+static pthread_key_t dummyKey;
+#endif
 static void onThreadExitInit() {
+#if KONAN_LINUX
+  // Due to glibc bug we have to create first key as dummy, to avoid
+  // conflicts with dlfcn error key.
```
SvyatoslavScherbina: This comment is imprecise.

PR #3770, opened 2020-01-21
Rename Clock to TimeSource

ilya-g:
Not yet in master of native, but already in kotlin-mirror/master.

Please update kotlin-compiler dependency version to at least 1.3.70-eap-104 when it is ready.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,15 +15,15 @@
 #
 
 # A version of the Kotlin compiler that is used to build Kotlin/Native.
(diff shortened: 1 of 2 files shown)


LepilkinaElena:
If there is no problem during updating could you update version in this PR then we can check tests results ?

LepilkinaElena:
Or it can be done by @knebekaizer when he updates versions for next eap.

LepilkinaElena:
Thank you for updating

PR #3773, opened 2020-01-22
Speed up codeGen phase in debug mode with caches

LepilkinaElena:


LepilkinaElena:
Second commit can be omitted, it provides additional speed up in 1 second during compilation of videoplayer

PR #3774, opened 2020-01-22
Fixed bug in devirtualization + test

homuroll:
The bug was not taking into account implicit runtime DFG edges generated by associated objects


homuroll:
https://youtrack.jetbrains.com/issue/KT-35940

PR #3779, opened 2020-01-23
Fixed static framework linker options

homuroll:
When using static caches we need to extract linker options
from entire DAG of touched native libraries

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
@@ -21,6 +21,8 @@ import org.jetbrains.kotlin.ir.util.isReal
 import org.jetbrains.kotlin.konan.library.KonanLibrary
 import org.jetbrains.kotlin.konan.target.KonanTarget
 import org.jetbrains.kotlin.library.KotlinLibrary
+import org.jetbrains.kotlin.library.uniqueName
+import org.jetbrains.kotlin.library.unresolvedDependencies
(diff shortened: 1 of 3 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
```diff
@@ -394,6 +396,31 @@ internal class Llvm(val context: Context, val llvmModule: LLVMModuleRef) {
 
     }
 
+    // The closure of nativeDependenciesToLink.
+    val allNativeDependencies: List<KonanLibrary> by lazy {
```
SvyatoslavScherbina: Is it essentially different from https://github.com/JetBrains/kotlin-native/blob/0338b5aa0fbd55ffa28cabd41f257e830a678431/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt#L48 ?

homuroll: I tried not to collect linker options from not actually used libraries, it might be an overkill, I don't really know, but it seems to me like it's a cleaner way.

SvyatoslavScherbina: The cleanest way to solve this problem is to use the same policy and (preferably) the same code as used for solving the completely semantically equivalent problem.

The code I referred builds linker options for dynamic framework, the code patched in this PR builds linker options for static framework, which is nearly the same.
It would be quite weird to do nearly the same completely differently.

PR #3780, opened 2020-01-24
Optimize atomics access by keeping thread affinity info an atomic.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,65 @@ OBJ_GETTER(initSharedInstance,
...
-OBJ_GETTER(readHeapRefLocked, ObjHeader** location, int32_t* spinlock) {
+OBJ_GETTER(readHeapRefLocked, ObjHeader** location, int32_t* spinlock, int32_t* cookie) {
   MEMORY_LOG("ReadHeapRefLocked: %p\n", location)
   lock(spinlock);
   ObjHeader* value = *location;
-  auto* container = value ? value->container() : nullptr;
-  if (container != nullptr)
-    incrementRC<true>(container);
-  unlock(spinlock);
+  bool shallRemember = *cookie != COOKIE();
```
SvyatoslavScherbina: How does `*cookie == COOKIE()` guarantee that `rememberNewContainer` is not required?

SvyatoslavScherbina: (Discussed offline, `COOKIE()` may require "invalidation").

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,81 @@ OBJ_GETTER(initSharedInstance,
...
+  UpdateReturnRef(OBJ_RESULT, oldValue);
   unlock(spinlock);
 
-  UpdateReturnRef(OBJ_RESULT, oldValue);
   if (shallRelease) {
     // No need to rememberNewContainer() on this path, as if `oldValue` is not null - it is explicitly released
     // anyway, and thus can not escape GC.
     ReleaseHeapRef(oldValue);
+  } else {
+    if (IsStrictMemoryModel && oldValue != nullptr && shallRemember) {
```
SvyatoslavScherbina: `oldValue` can't be `null` if `shallRemember` is `true`.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,81 @@ OBJ_GETTER(initSharedInstance,
...
+ * repeating read operation of the same value do not lead to the repeating rememberNewContainer() operation.
+ * We must invalidate cookie after the local GC, as otherwise fact that container of the `value` is retained
+ * may change, if the last reference to the value read is lost during GC and we re-read same value from
+ * the same atomic reference. Thus we also include last GC timestamp into the cookie.
+ */
+inline int32_t computeCookie(ObjHeader* value) {
+  auto* state = memoryState;
+  auto timestamp = state->lastGcTimestamp;
+  return (static_cast<int32_t>(reinterpret_cast<intptr_t>(state))) ^
+    (static_cast<int32_t>(reinterpret_cast<intptr_t>(value))) ^
```
SvyatoslavScherbina: Is it actually required to include `value` to cookie?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,81 @@ OBJ_GETTER(initSharedInstance,
...
+/**
+ * We keep thread affinity and reference value based cookie in the atomic references, so that
+ * repeating read operation of the same value do not lead to the repeating rememberNewContainer() operation.
+ * We must invalidate cookie after the local GC, as otherwise fact that container of the `value` is retained
+ * may change, if the last reference to the value read is lost during GC and we re-read same value from
+ * the same atomic reference. Thus we also include last GC timestamp into the cookie.
+ */
+inline int32_t computeCookie(ObjHeader* value) {
+  auto* state = memoryState;
+  auto timestamp = state->lastGcTimestamp;
```
SvyatoslavScherbina: Timestamp isn't reliable much as a unique-per-mutation-cycle value. Why not use a dedicated counter for this?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,81 @@ OBJ_GETTER(initSharedInstance,
...
+  bool shallRemember = *cookie != realCookie;
+  if (shallRemember) *cookie = realCookie;
   UpdateReturnRef(OBJ_RESULT, value);
-  if (value != nullptr)
-    ReleaseHeapRef(value);
+  unlock(spinlock);
+#if USE_GC
+  if (IsStrictMemoryModel && shallRemember && value != nullptr) {
+    auto* container = value->container();
+    rememberNewContainer(container);
```
SvyatoslavScherbina: Isn't it too late to do this after `unlock`? Another thread might have already nullified the reference and removed its value.

runtime/src/main/cpp/Memory.cpp
olonho: That's indeed may not be needed. I was thinking about scenario where we read value, write new value and then read the new value from the same thread (and second read doesn't remember the new container), but indeed such a container must be already remembered as we managed to write it.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2056,81 @@ OBJ_GETTER(initSharedInstance,
...
   bool shallRelease = false;
+  bool shallRemember = false;
   // We do not use UpdateRef() here to avoid having ReleaseRef() on return slot under the lock.
   if (oldValue == expectedValue) {
     SetHeapRef(location, newValue);
     shallRelease = oldValue != nullptr;
   } else {
-    if (IsStrictMemoryModel && oldValue != nullptr)
-      rememberNewContainer(oldValue->container());
+    if (IsStrictMemoryModel && oldValue != nullptr) {
```
SvyatoslavScherbina: `oldValue == expectedValue` case probably requires setting cookie too.

runtime/src/main/cpp/Memory.cpp
olonho: Given its microsecond resolution it shall work as such, but indeed could make it a counter.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2056,51 +2058,78 @@ OBJ_GETTER(initSharedInstance,
...
-  UpdateReturnRef(OBJ_RESULT, oldValue);
   if (shallRelease) {
     // No need to rememberNewContainer() on this path, as if `oldValue` is not null - it is explicitly released
     // anyway, and thus can not escape GC.
     ReleaseHeapRef(oldValue);
+  } else {
+    if (IsStrictMemoryModel && shallRemember) {
+      // `oldValue` is guaranteed to never be null in such a case.
+      RuntimeAssert(oldValue != nullptr, "Must not be null");
+      rememberNewContainer(oldValue->container());
```
SvyatoslavScherbina: Same as https://github.com/JetBrains/kotlin-native/pull/3780#discussion_r371716261

olonho: Simplified.

Issue #3723, opened 2019-12-25
runtime assert: Object must not be tagged


msink:
Trying to port `ObserverMap` from Android Jetpack Compose to K/N - get weird error:
`runtime assert: Object must not be tagged`
and process with unit-tests terminated abnormally.

implementation: https://github.com/msink/kotlin-libui/blob/ObserverMap/compose-runtime/src/nativeMain/kotlin/ObserverMap.kt
unit-test: https://github.com/msink/kotlin-libui/blob/ObserverMap/compose-runtime/src/commonTest/kotlin/ObserverMapTests.kt

What is wrong here?

msink:
Fixed by #3781

msink:
@olonho Will #3781 be in 1.3.70?

kylejbrock:
This seems to still be a problem in 1.3.71.  Just migrated from 1.3.61 and now my app is crashing with this error.

PR #3781, opened 2020-01-24
Allow weak references to permanent objects.

olonho:


olonho:
Shall fix https://github.com/JetBrains/kotlin-native/issues/3723

PR #3783, opened 2020-01-24
Added addressOf for Pinned<CharSequence>

LepilkinaElena:


Interop/Runtime/src/native/kotlin/kotlinx/cinterop/Pinning.kt
```diff
@@ -47,6 +47,9 @@ inline fun <T : Any, R> T.usePinned(block: (Pinned<T>) -> R): R {
 fun Pinned<ByteArray>.addressOf(index: Int): CPointer<ByteVar> = this.get().addressOfElement(index)
 fun ByteArray.refTo(index: Int): CValuesRef<ByteVar> = this.usingPinned { addressOf(index) }
 
+fun Pinned<CharSequence>.addressOf(index: Int): CPointer<ByteVar> = this.get().addressOfElement(index)
```
olonho: `String`, not `CharSequence`, seems there's no way to do that for generic char sequence, as we must know implementation details.

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -661,6 +661,15 @@ KNativePtr Kotlin_Arrays_getByteArrayAddressOfElement(KRef thiz, KInt index) {
   return AddressOfElementAt<KByte>(array, index);
 }
 
+KNativePtr Kotlin_Arrays_getStringAddressOfElement (KRef thiz, KInt index) {
+  ArrayHeader* array = thiz->array();
+  if (index < 0 || index >= array->count_) {
+    ThrowArrayIndexOutOfBoundsException();
+  }
+
+  return AddressOfElementAt<KByte>(array, index);
```
olonho: KChar?

PR #3788, opened 2020-01-27
Metadata-based interop enhancements

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/interop/IrProviderForInteropStubs.kt
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private fun referenceSuperClassifiers(descriptor: ClassDescriptor) {
+        val symbolTable = declarationStubGenerator.symbolTable
+        descriptor.getAllSuperClassifiers().forEach { symbolTable.referenceClassifier(it) }
+    }
+
+    private fun collectObjCClasses(klass: IrClass) {
+        val descriptor = klass.descriptor
+        // Force-load types to be able to generate RTTI for the given class.
```
SvyatoslavScherbina: Is RTTI really required for Obj-C classes? It shouldn't be.

sbogolepov: IIUC, it is needed when we perform instance checking.

PR #3797, opened 2020-01-28
Drop unsupported experimental coroutines API

ilya-g:
There will be no expect counterpart for it in common stdlib anymore

[#KT-36083](https://youtrack.jetbrains.com/issue/KT-36083)


olonho:
Note that it fails due to some other problem, maybe rebase.

ilya-g:
@olonho It fails because it's targeted to `kotlin-mirror\master` branch. The composite build is ok: https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_SanityTestKotlinNativeLinuxComposite/63394176?buildTab=changes

PR #3804, opened 2020-01-30
Fix benchmarks with Swift interop after enabling -Xobjc-generics

LepilkinaElena:


performance/swiftinterop/swiftSrc/benchmarks.swift
```diff
@@ -88,11 +88,11 @@ class SwiftInteropBenchmarks {
     }
 
     func initMultigraph(size: Int) {
-        multigraph = Multigraph()
+        multigraph = Multigraph<NSNumber>()
         for i in 1...size {
-            multigraph.addEdge(from: i, to: i + 1, cost: SimpleCost(cost: (i + 1) % i ))
-            multigraph.addEdge(from: i, to: i * 2, cost: SimpleCost(cost: (i * 2) % i))
-            multigraph.addEdge(from: i + 5, to: i, cost: SimpleCost(cost: (i + 5) % i))
+            multigraph.addEdge(from: NSNumber(value: i), to: NSNumber(value: i + 1), cost: SimpleCost(cost: (i + 1) % i ))
```
olonho: Ugly looking code, now way to make conversion implicit?

LepilkinaElena: I don't know such conversion. Problem as I understood is that Int is a structure and I can't use it(or NSInteger which is its alias) in generics.

PR #3809, opened 2020-01-31
[IR] Supported conversions to fun interfaces

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -134,6 +137,21 @@ internal val buildCExportsPhase = konanUnitPhase(
         prerequisite = setOf(createSymbolTablePhase)
 )
 
+object KonanGeneratorExtensions : GeneratorExtensions() {
+    override val samConversion: SamConversion
+        get() = KonanSamConversion
+
+    open class KonanSamConversion : SamConversion() {
+        override fun isPlatformSamType(type: KotlinType): Boolean =
+                JavaSingleAbstractMethodUtils.isSamType(type)
```
olonho: Is indeed correct with Native?

olonho: Maybe just check for `Function`, `SuspendFunction` here?

PR #3818, opened 2020-02-03
Rework debugging interface to allow coexisting for multiple frameworks.

olonho:


runtime/src/debug/cpp/KDebug.cpp
```diff
@@ -64,16 +64,16 @@ constexpr int runtimeTypeAlignment[] = {
...
 }
 
-RUNTIME_USED int Konan_DebugBufferSize() {
+RUNTIME_USED RUNTIME_WEAK int Konan_DebugBufferSize() {
   return sizeof(debugBuffer);
 }
 
 // Auxilary function which can be called by developer/debugger to inspect an object.
-RUNTIME_USED KInt Konan_DebugObjectToUtf8Array(KRef obj, char* buffer, KInt bufferSize) {
+RUNTIME_USED RUNTIME_WEAK KInt Konan_DebugObjectToUtf8Array(KRef obj, char* buffer, KInt bufferSize) {
```
SvyatoslavScherbina: Debugger still uses this API, so now it allows an object of one runtime to leak into other runtime.

olonho: Now it's up to bindings if use legacy API or rely upon the new API. So this is runtime part of the mechanism.

SvyatoslavScherbina: Legacy API turns out to be unsafe.
Why not make it safe by implementing public API through function pointers from  "extended info"?

runtime/src/debug/cpp/KDebug.cpp
```diff
@@ -188,6 +207,119 @@ RUNTIME_USED const char* Konan_DebugGetTypeName(KRef obj) {
...
+
+RUNTIME_USED RUNTIME_WEAK int Konan_DebugBufferSizeWithObject(KRef obj) {
+  auto* impl = getImpl<int (*)(KRef)>(obj, DO_DebugBufferSizeWithObject);
+  if (impl == nullptr) return 0;
+  return impl(obj);
+}
+
+// Auxilary function which can be called by developer/debugger to inspect an object.
+RUNTIME_USED RUNTIME_WEAK int Konan_DebugObjectToUtf8Array(KRef obj, char* buffer, int bufferSize) {
+  auto* impl = getImpl<int (*)(KRef, char*, int)>(obj, DO_DebugObjectToUtf8Array);
```
SvyatoslavScherbina: `KInt`?

SvyatoslavScherbina: Same below.

olonho: Switched all to explicit `int32_t`.

runtime/src/debug/cpp/KDebug.cpp
```diff
@@ -188,6 +207,119 @@ RUNTIME_USED const char* Konan_DebugGetTypeName(KRef obj) {
...
+}
+
+RUNTIME_USED RUNTIME_WEAK const char* Konan_DebugGetTypeName(KRef obj) {
+  auto* impl = getImpl<const char* (*)(KRef)>(obj, DO_DebugGetTypeName);
+  if (impl == nullptr) return nullptr;
+  return impl(obj);
+}
+
+RUNTIME_USED RUNTIME_WEAK
+void* Konan_DebugGetOperation(KRef obj, /* Konan_DebugOperation */ int32_t operation) {
```
SvyatoslavScherbina: Is this public debug API actually required?

olonho: Mostly for the completness of the debug API, but indeed, we may not expose it and enum.

runtime/src/debug/cpp/KDebug.cpp
```diff
@@ -188,6 +223,98 @@ RUNTIME_USED const char* Konan_DebugGetTypeName(KRef obj) {
...
+}
+
+RUNTIME_USED RUNTIME_WEAK const char* Konan_DebugGetTypeName(KRef obj) {
+  auto* impl = getImpl<const char* (*)(KRef)>(obj, DO_DebugGetTypeName);
+  if (impl == nullptr) return nullptr;
+  return impl(obj);
+}
+
+// IMPORTANT: when updating, fix `debugOperationsSize` in RTTIGenerator.kt.
+RUNTIME_USED RUNTIME_WEAK
```
SvyatoslavScherbina: Why are these attributes required?

olonho: So that we could use them from Kotlin side and not conflict on dups.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -395,6 +395,30 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
...
+    private val debugRuntimeOrNull: LLVMModuleRef? by lazy {
+        context.config.runtimeNativeLibraries.singleOrNull { it.endsWith("debug.bc")}?.let {
+            parseBitcodeFile(it)
+        }
+    }
+
+    private val debugOperations: ConstValue by lazy {
+        if (debugRuntimeOrNull != null) {
+            val external = LLVMGetNamedGlobal(debugRuntimeOrNull, "Konan_debugOperationsList")!!
+            val local = LLVMAddGlobal(context.llvmModule, LLVMTypeOf(external),"Konan_debugOperationsList")!!
```
SvyatoslavScherbina: `LLVMTypeOf(external)` is not the type of the global itself, it is the type of the pointer to it.
As a result, `Konan_debugOperationsList` is declared here as `external global [14 x i8*]*`, while originally defined as `global [14 x i8*]`.

This is also likely the cause of "LLVMGetArrayLength() doesn't work as expected" below.

runtime/src/debug/cpp/KDebug.cpp
SvyatoslavScherbina: What do you mean by "Kotlin side"? Generated code?
There are plenty of runtime functions that aren't `RUNTIME_USED RUNTIME_WEAK` and can be properly used from generated code without conflicts.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
SvyatoslavScherbina: Type of the global added here is still wrong.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -395,6 +395,30 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
     private fun mapRuntimeType(type: LLVMTypeRef): Int =
             runtimeTypeMap[type] ?: throw Error("Unmapped type: ${llvmtype2string(type)}")
 
+    private val debugRuntimeOrNull: LLVMModuleRef? by lazy {
+        context.config.runtimeNativeLibraries.singleOrNull { it.endsWith("debug.bc")}?.let {
+            parseBitcodeFile(it)
```
SvyatoslavScherbina: Should be disposed.

SvyatoslavScherbina: Alternatively we can avoid loading this module at all.
`debugOperationsSize` could be stored in a runtime-defined struct containing debug operations array (or a pointer to it) too.
Then it would be enough to have a single pointer in `ExtendedTypeInfo` (which can also be considered an optimization).

runtime/src/debug/cpp/KDebug.cpp
olonho: OK, if needed we can easily add it.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: Not sure if it's important, but fixed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
olonho: Prefer disposing solution.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -105,6 +105,9 @@ internal class RTTIGeneratorVisitor(context: Context) : IrElementVisitorVoid {
         }
     }
 
+    fun dispose() {
+        generator.dispose()
```
SvyatoslavScherbina: It's not enough since there are more instances of `RTTIGenerator`.

PR #3819, opened 2020-02-03
Remove obsolete leak detector.

olonho:


runtime/src/main/kotlin/kotlin/native/internal/GC.kt
```diff
@@ -92,24 +92,6 @@ object GC {
...
 
-    /**
-     * Detect cyclic references going via atomic references and return list of cycle-inducing objects
-     * or `null` if the leak detector is not available. Use [Platform.isMemoryLeakCheckerActive] to check
-     * leak detector availability.
-     * Note that cycle detector requires reference graph stability, thus it may not work as
-     * expected or even crash for mutating graphs.
-     */
-    @SymbolName("Kotlin_native_internal_GC_detectCycles")
-    external fun detectCycles(): Array<Any>?
```
SvyatoslavScherbina: `Atomics.kt` still mentions this API.

runtime/src/main/kotlin/kotlin/native/concurrent/Atomics.kt
```diff
@@ -216,8 +216,7 @@ private fun debugString(value: Any?): String {
 /**
  * An atomic reference to a frozen Kotlin object. Can be used in concurrent scenarious
  * but frequently shall be of nullable type and be zeroed out once no longer needed.
- * Otherwise memory leak could happen. To detect such leaks [kotlin.native.internal.GC.detectCycles]
- * in debug mode could be helpful.
+ * Asynchronous cycle collector takes care of cyclic references of that kind.
```
SvyatoslavScherbina: Is "shall be of nullable type and be zeroed out once no longer needed" above left intentionally?

olonho: Yes, it kind of suggest people to think about cleaning up resources.

PR #3820, opened 2020-02-04
Omit RTTI generation for external Obj-C classes

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -97,9 +97,9 @@ internal class RTTIGeneratorVisitor(context: Context) : IrElementVisitorVoid {
 
     override fun visitClass(declaration: IrClass) {
         super.visitClass(declaration)
-
-        generator.generate(declaration)
-
+        if (declaration.requiresRtti()) {
```
SvyatoslavScherbina: How do these classes happen to be found in this IR module in the first place?
Shouldn't these classes be among external IR stubs which aren't supposed to go through our lowerings and code generation?

sbogolepov: Right now external Obj-C classes are usual classes that can be found in the IR module. There are other classes (for example, `CStructVar`) which are defined in the module and should be skipped during RTTI generation (I’ll support them later).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -195,11 +195,13 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
 
         val instanceSize = getInstanceSize(bodyType, className)
 
-        val superTypeOrAny = irClass.getSuperClassNotAny() ?: context.ir.symbols.any.owner
+        val superTypeOrAny = irClass.getSuperClassNotAny().let { superType ->
+            if (superType == null || !superType.requiresRtti()) context.ir.symbols.any.owner else superType
```
SvyatoslavScherbina: What is this required for?
It would be cleaner to avoid generating `irClass` too, and leave this code crashing if `irClass.requiresRtti()`, but one of its super types doesn't.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CodeGenerationInfo.kt
```diff
@@ -15,3 +15,17 @@ internal fun IrClass.isNonGeneratedAnnotation(): Boolean =
                 !this.annotations.hasAnnotation(serialInfoAnnotationFqName)
 
 private val serialInfoAnnotationFqName = FqName("kotlinx.serialization.SerialInfo")
+
+/**
+ * We don't need to generate RTTI in some cases, e.g. Objective-C external classes.
+ */
+internal fun IrClass.requiresRtti(): Boolean = when {
+    // TODO: Support more cases
+    this.isExternalObjCClass() -> false
```
SvyatoslavScherbina: What about Kotlin subclasses of Objective-C classes? Do they actually require any RTTI?
If so, this probably should be fixed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
sbogolepov: For now, It is required for Kotlin subclasses of Objective-C classes. I'll try to eliminate RTTI for them, and if it work out, fix this code.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/ClassLayoutBuilder.kt
```diff
@@ -278,7 +278,7 @@ internal class ClassLayoutBuilder(val irClass: IrClass, val context: Context) {
         val superVtableEntries = if (irClass.isSpecialClassWithNoSupertypes()) {
             emptyList()
         } else {
-            val superClass = irClass.getSuperClassNotAny() ?: context.ir.symbols.any.owner
+            val superClass = irClass.getRuntimeSuperClassNotAny() ?: context.ir.symbols.any.owner
```
SvyatoslavScherbina: Looks unpredictable and thus harmful. Is this actually required?

sbogolepov: On the one hand, it’s indeed looks kinda unclear. On the other, there are several places that require uniform approach. I’m not sure, is there a better approach that doesn’t require a lot of refactoring.

SvyatoslavScherbina: The thing is: the list of these places is not well-defined.
This partucular place isn't supposed to work properly with `superClass` different from actual super class.
Generally type info for Kotlin subclass of Obj-C class is a hack, and it should continue being local. Patching only one (unused) field of actual TypeInfo is local enough. Hacking different non-trivial fragments of compiler code is not so local.

PR #3821, opened 2020-02-04
Skip C overloads

knebekaizer:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
```diff
@@ -146,6 +148,9 @@ class StubsBuildingContextImpl(
 
     private var theCounter = 0
 
+    private val uniqFunctions = mutableSetOf<String>()
+    override fun isOverloading(name: String) = !uniqFunctions.add(name) // TODO params & return type
```
olonho: Format  is
``
// TODO: sentence with dot at the end.
``

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrElementBuilders.kt
```diff
@@ -412,8 +412,12 @@ internal class EnumStubBuilder(
 
 internal class FunctionStubBuilder(
         override val context: StubsBuildingContext,
-        private val func: FunctionDecl
+        private val func: FunctionDecl,
+        private val skipOverloads: Boolean = false
 ) : StubElementBuilder {
+
+    private fun isOverloading(name: String): Boolean = context.platform == KotlinPlatform.NATIVE && context.isOverloading(name)
```
olonho: Doesn't it makes impression that we support overloads in JVM case.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrBuilder.kt
sbogolepov: Why not to pass `FunctionDecl` directly?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrElementBuilders.kt
```diff
@@ -466,6 +470,8 @@ internal class FunctionStubBuilder(
             context.mirror(func.returnType).argType
         }.toStubIrType()
 
+        if (skipOverloads && isOverloading(func.name))
```
sbogolepov: Do we guarantee that two consequent `cinterop` invocations will return the same result?

knebekaizer: @sbogolepov  Don't catch the question: nothing is shared between  `cinterop` invocations.
Do you suggest to move `skipOverloads` to `buildingContext` or `InteropConfiguration`?

sbogolepov: I'm talking about an order of incoming declarations. Is it possible that on the first invocation of `cinterop` the input is `[A, B, C]` and on the second one it is `[B, C, A]`?

sbogolepov: > Do you suggest to move skipOverloads to buildingContext or InteropConfiguration?
I think it's fine where it is.

PR #3822, opened 2020-02-04
Rework `@Throws`

SvyatoslavScherbina:
It now enables propagation only for instances of listed classes.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
@@ -22,6 +22,8 @@ import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.descriptors.konan.CompiledKlibModuleOrigin
 import org.jetbrains.kotlin.descriptors.konan.CurrentKlibModuleOrigin
 import org.jetbrains.kotlin.ir.declarations.*
+import org.jetbrains.kotlin.ir.expressions.IrClassReference
+import org.jetbrains.kotlin.ir.expressions.IrVararg
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.*
@@ -375,6 +377,9 @@ internal class ObjCExportCodeGenerator(
 
     internal val directMethodAdapters = mutableMapOf<DirectAdapterRequest, ObjCToKotlinMethodAdapter>()
 
+    internal val exceptionTypeInfoArrays = mutableMapOf<IrFunction, ConstPointer>()
+    internal val typeInfoArrays = mutableMapOf<Set<IrClass>, ConstPointer>()
+
     inner class ObjCToKotlinMethodAdapter(
             selector: String,
             encoding: String,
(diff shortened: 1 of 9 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -825,6 +837,29 @@ private fun ObjCExportCodeGenerator.generateObjCImp(
     ret(genReturnValueOnSuccess(returnType))
 }
 
+private fun ObjCExportCodeGenerator.generateExceptionTypeInfoArray(baseMethod: IrFunction): LLVMValueRef =
+        exceptionTypeInfoArrays.getOrPut(baseMethod) {
+            val throwsAnnotation = findThrowsAnnotation(baseMethod)!!
+            val types = (throwsAnnotation.getValueArgument(0) as IrVararg).elements
```
olonho: Why those casts are safe?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -825,6 +837,29 @@ private fun ObjCExportCodeGenerator.generateObjCImp(
...
+        typeInfoArrays.getOrPut(types) {
+            val typeInfos = types.map { with(codegen) { it.typeInfoPtr } } + NullPointer(codegen.kTypeInfo)
+            codegen.staticData.placeGlobalConstArray("", codegen.kTypeInfoPtr, typeInfos)
+        }
+
+private fun findThrowsAnnotation(baseMethod: IrFunction): IrConstructorCall? =
+        if (baseMethod !is IrSimpleFunction || baseMethod.overriddenSymbols.isEmpty()) {
+            baseMethod.annotations.findAnnotation(KonanFqNames.throws)
+        } else {
+            // Note: frontend ensures that all topmost overridden methods have equal @Throws annotations.
```
olonho: Can/shall we assert for that condition?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: Frontend and psi2ir are supposed to guarantee this.
I will make cast failures more readable though.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: Technically adding such an assertion should be easy.
However, considering binary compatibility (e.g. linking with other versions of libraries), it is actually possible that such a condition isn't met.
We don't have much choice in this case: either report compile (link?) error or handle this silently.
And I would strongly prefer to avoid reporting any errors from compiler, at least because we don't have proper stable infrastructure (in IR linker) to do this reliably. WDYT?

olonho: OK, let's keep it silently processed so far. Maybe warning?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: Somewhat improved.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: Reporting warning makes sense, however we still don't have proper infrastructure for this. Generally I believe we should cease reporting compiler diagnostics from backend the way we are doing it now.
I expect this particular case to be quite rare, so it should be ok to postpone adding such a warning. Especially considering that I'm not completely sure about the details of new `@Throws` design, so it can change before 1.4.

PR #3824, opened 2020-02-05
Fix external tests

PavelPunegov:
* Don't run tests targeted to API < 1.4
* Fix importing classes in packages replacements during external test processing

### DIFF
--- a/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
+++ b/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
@@ -370,11 +370,12 @@ class RunExternalTestGroup extends OldKonanTest {
 
         def packagePattern = ~/(?m)^\s*package\s+(${fullQualified}*)/
         def boxPattern = ~/(?m)fun\s+box\s*\(\s*\)/
-        def classPattern = ~/.*(class|object|enum)\s+(${identifier}*).*/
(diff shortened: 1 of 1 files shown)


build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -542,6 +544,11 @@ fun runTest() {
             return false
         }
 
+        def apiVersion = findLinesWithPrefixesRemoved(text, '// !API_VERSION: ')
+        if (apiVersion.size() != 0 && !apiVersion.contains("1.4")) {
```
ilmat192: Shouldn't we update the same check for language version above?

olonho: Maybe rewrite that logic in Kotlin finally?

PavelPunegov: I'm going to rewrite this and there is something already done in this way in another branch. But this fix is just to fix tests failures now.

As for language_version, I don't think this is necessary as tests pass.

ilmat192: > As for language_version, I don't think this is necessary as tests pass.

But currently we exclude all tests with LANGUAGE_VERSION=1.4, don't we?

PR #3826, opened 2020-02-05
[StubIr] Globals and macros

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -445,7 +445,7 @@ class StubIrTextEmitter(
     }
 
     private fun renderValueUsage(value: ValueStub): String = when (value) {
-        is StringConstantStub -> value.value
+        is StringConstantStub -> value.value.quoteAsKotlinLiteral()
```
SvyatoslavScherbina: How did this work before?

sbogolepov: Quotes were added in `MacroConstantStubBuilder`. We don't need these quotes in metadata mode, so they were moved to `StubIrTextEmitter`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -820,6 +820,10 @@ private fun KotlinStubs.mapType(
                 ?: reportUnsupportedType("not a structure or too complex"))
     }
 
+    type.classOrNull?.isSubtypeOfClass(symbols.nativePointed) == true -> {
+        ByRefValuePassing(symbols.nativePtrType)
```
SvyatoslavScherbina: Isn't `TrivialValuePassing(type, CTypes.voidPtr)` enough?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -1045,6 +1049,34 @@ private class StructValuePassing(private val kotlinClass: IrClass, override val
 
 }
 
+/**
+ * To avoid memory copying, we get global struct by passing it through the bridge by pointer
```
SvyatoslavScherbina: > To avoid memory copying

IIRC, the original motivation was to simplify access to struct's fields (`CValue` has problems with that).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
sbogolepov: We need to perform conversion in `bridgedToKotlin` to dereference incoming pointer.

PR #3829, opened 2020-02-06
Improve library generator

ilmat192:


PavelPunegov:
It seems that it breaks some platform libs dependency according to interop tests on Mac and Simulator

ilmat192:
It changes contents of a bundle. Now the default distribution archive doesn't contain platform libraries and tests that require platform libs fail. I'll change corresponding build configurations.

utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/GeneratePlatformLibraries.kt
```diff
@@ -9,46 +9,175 @@ import org.jetbrains.kotlin.cli.bc.K2Native
...
+import org.jetbrains.kotlin.native.interop.tool.CommonInteropArguments.Companion.DEFAULT_MODE
+import org.jetbrains.kotlin.native.interop.tool.CommonInteropArguments.Companion.MODE_METADATA
+import org.jetbrains.kotlin.native.interop.tool.CommonInteropArguments.Companion.MODE_SOURCECODE
+import java.io.PrintWriter
+import java.io.StringWriter
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.system.exitProcess
+import java.io.File as JFile
 
+private class Logger(val level: Level = Level.NORMAL) {
```
sbogolepov: Can we reuse existing logging logic if it exists? If not, it worth moving this class to a separate file so other users could reuse it.

ilmat192: I didn't find any simple logger suitable for this tool. So, let's postpone this refactoring.

alexander-gorshenev: compiler/util-io/src/org/jetbrains/kotlin/util/WithLogger.kt could be useful or could be not.

PR #3836, opened 2020-02-06
Fail fast in Regex.findAll on an invalid startIndex #KT-28356

qurbonzoda:


runtime/src/main/kotlin/kotlin/text/Regex.kt
```diff
@@ -155,10 +155,11 @@ public actual class Regex internal constructor(internal val nativePattern: Patte
      *
      * @param startIndex An index to start search with, by default 0. Must be not less than zero and not greater than `input.length()`
      * @return An instance of [MatchResult] if match was found or `null` otherwise.
+     * @throws IndexOutOfBoundsException if [startIndex] is less than zero or greater than the length of the [input] char sequence.
      */
     actual fun find(input: CharSequence, startIndex: Int): MatchResult? {
         if (startIndex < 0 || startIndex > input.length) {
-            throw IndexOutOfBoundsException("Start index out of bounds: $startIndex")
+            throw IndexOutOfBoundsException("Start index out of bounds: $startIndex, input length: ${input.length}")
```
ilya-g: "... is out of bounds"

PR #3837, opened 2020-02-06
samples/tetris build cleanup

msink:
Support build for multiple targets at the same time.

### DIFF
--- a/samples/tetris/build.gradle.kts
+++ b/samples/tetris/build.gradle.kts
@@ -85,6 +85,19 @@ kotlin {
                     )
                     presets["linuxArm32Hfp"] -> linkerOpts("-lSDL2")
                 }
+
+                val distTaskName = linkTaskName.replaceFirst("link", "dist")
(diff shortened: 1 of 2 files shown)


samples/tetris/build.gradle.kts
```diff
@@ -1,42 +1,79 @@
...
+        sourceSets["${targetName}Main"].apply {
+            kotlin.srcDir("src/tetrisMain/kotlin")
+        }
+
         binaries {
             executable {
                 entryPoint = "sample.tetris.main"
+
+                // Compile Windows Resources
+                if (konanTarget.family == org.jetbrains.kotlin.konan.target.Family.MINGW) {
```
ilmat192: `KonanTarget` is an internal thing. It does appear in buildscripts of some our libraries but it's better to minimize its usage. Could you check `preset` instead of `konanTarget` here and below?

msink: Done

PR #3839, opened 2020-02-07
Added caching of stdlib for dist tasks that includes building stdlib

LepilkinaElena:


build.gradle
```diff
@@ -397,7 +397,7 @@ task listDist(type: Exec) {
 }
 
 task distRuntime(type: Copy) {
-    dependsOn "${hostName}CrossDistRuntime"
+    dependsOn "${hostName}CrossDistRuntimeAndCache"
```
ilmat192: This dependency seems inconsistent. Currently just `distRuntime` builds caches while `crossDistRuntime` and `<target>CrossDistRuntime` don't.

PR #3840, opened 2020-02-07
[Interop][Metadata] Mangle global properties names

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -146,7 +159,8 @@ internal class ModuleMetadataEmitter(
 
         override fun visitProperty(element: PropertyStub, data: VisitingContext) =
                 with (MappingExtensions(data.typeParametersInterner)) {
-                    KmProperty(element.flags, element.name, element.getterFlags, element.setterFlags).also { km ->
+                    val name = getPropertyNameInScope(element.name, data.container)
+                    KmProperty(element.flags, name, element.getterFlags, element.setterFlags).also { km ->
```
olonho: `km` is not meaningful name, it shall be a meaningful noun.

sbogolepov: In general, I agree that `km` is an awful name. But in this particular case it was chosen because all declarations from `kotlinx.metadata` are prefixed with `Km`. Thus, `km` makes it clear that it is a reference to `KmProperty`, `KmFunction`, `KmWhatever`.

olonho: km -> metadataDescriptor?

sbogolepov: Maybe. But this is clearly out of scope of the PR (because the whole file should be refactored) and can be done separately.

PR #3841, opened 2020-02-07
Improved concurrency docs.

olonho:


CONCURRENCY.md
```diff
@@ -149,3 +178,32 @@ the following mechanisms to prevent the unintended sharing of state via global o
...
+
+<a name="atomic_references"></a>
+### Atomic primitives and references
+
+ Kotlin/Native standard library provides primitives for safe working with concurrently mutable data, namely
+`AtomicInt`, `AtomicLong`, `AtomicNativePtr`, `AtomicReference` and `FreezableAtomicReference` in the package
+`kotlin.native.concurrent`.
+Atomic primitives allows concurrency-safe update operations, such as increment, decrement and compare-and-swap,
+along with value setters and getters. Atomic primitives are considered always frozen by the runtime, and their fields cannot
+be updated with the regular field mutation, but the value can be changed using dedicated operations, so it is possible
```
SvyatoslavScherbina: > cannot be updated with the regular field mutation

`atomicInt.value` mutation with getter and setter is not distinguishable from "regular field mutation" at call site.
Moreover, `atomicInt.value += 1` is still possible, and it is not too different from "regular field mutation" by its behaviour.

CONCURRENCY.md
```diff
@@ -149,3 +178,32 @@ the following mechanisms to prevent the unintended sharing of state via global o
...
+to perform concurrent-safe global counters and similar data structures.
+
+  Some algorithms require shared mutable references across the multiple workers, for example global mutable configuration
+could be implemented as an immutable instance of properties list atomically replaced with the new version on
+configuration update. This way no concurrency issues may appear and at the same time configuration could be
```
SvyatoslavScherbina: > This way no concurrency issues may appear

Not sure if this is correct. See example [above](https://github.com/JetBrains/kotlin-native/pull/3841/files#r377067054).

olonho: Well, we could use a bit different wording.

PR #3852, opened 2020-02-11
Workaround Swift bug triggering #3825

### LINKED ISSUE #3825: Crash when throws keyword is added to swift functions used in kotlin
Hey,

I stumbled about this in the docs:

„Note that the opposite reversed translation is not implemented yet: Swift/Objective-C error-throwing methods aren't imported to Kotlin as exception-throwing.“
https://kotlinlang.org/docs/reference/native/objc_interop.html#errors-and-exceptions

I tried this, and it is not simply ignored, instead it crashes every time, if the function has a closure parameter:
`func functionWithCallback(callback: @escaping (KotlinInt, String) -> Void) throws {} //This throws produces a crash Thread 1: EXC_BAD_ACCESS (code=1, address=0x10)`
If the function has no parameter it works:
`func function() throws {} //this works`

I call it like:
```
let someKotlinInterfaceImplementationInSwift = SomeInterfaceImplementationInSwift()
let someKotlinClass = SomeKotlinClass(dependencieImplementedInSwift: someKotlinInterfaceImplementationInSwift)
someKotlinClass.someFunctionImplementedInKotlin()
```
 ```
Thread 1 Queue : com.apple.main-thread (serial)
 #0    0x0000000106c215ff in thunk for @escaping @callee_unowned @convention(block) (@unowned GreetingInt, @unowned NSString) -> () ()
 #1    0x0000000106c213bc in SomeInterfaceImplementationInSwift.functionWithCallback(callback:) at /Users/timb/projects/mpp-example-exceptions/iosApp/iosApp/SomeInterfaceImplementation.swift:13
 #2    0x0000000106c214c5 in @objc SomeInterfaceImplementationInSwift.functionWithCallback(callback:) ()
 #3    0x0000000106f9e7d8 in ___lldb_unnamed_symbol3$$greeting ()
 #4    0x0000000106f992be in kfun:org.greeting.SomeKotlinClass.someFunctionImplementedInKotlin() at /Users/timb/projects/mpp-example-exceptions/greeting/src/commonMain/kotlin/SomeKotlinClass.kt:6
 #5    0x0000000106f9e6dd in objc2kotlin.65 at /Users/timb/projects/mpp-example-exceptions/greeting/src/iosMain/kotlin/<compiler-generated>:1
 #6    0x0000000106c20510 in ViewController.viewDidLoad() at /Users/timb/projects/mpp-example-exceptions/iosApp/iosApp/ViewController.swift:21
```

 I made a simple reproducer: https://github.com/Ti-m/mpp-example/tree/problem-with-exceptions

### PR DESCRIPTION
**Pull request by SvyatoslavScherbina**
Avoid placing Objective-C error out parameter right after block parameters.
Fix #3825.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt
@@ -316,10 +316,21 @@ private fun ObjCExportMapper.bridgeMethodImpl(descriptor: FunctionDescriptor): M
 
     val returnBridge = bridgeReturnType(descriptor, convertExceptionsToErrors)
     if (convertExceptionsToErrors) {
-        valueParameters += MethodBridgeValueParameter.ErrorOutParameter
+        // Add error out parameter before tail block parameters. The convention allows this.
+        // Placing it after would trigger https://github.com/JetBrains/kotlin-native/issues/3825
+        val tailBlocksCount = valueParameters.reversed().takeWhile { it.isBlockPointer() }.count()
+        valueParameters.add(valueParameters.size - tailBlocksCount, MethodBridgeValueParameter.ErrorOutParameter)
     }
 
     return MethodBridge(returnBridge, receiver, valueParameters)
+}
+
+private fun MethodBridgeValueParameter.isBlockPointer(): Boolean = when (this) {
+    is MethodBridgeValueParameter.Mapped -> when (this.bridge) {
+        ReferenceBridge, is ValueTypeBridge -> false
+        is BlockPointerBridge -> true
+    }
+    MethodBridgeValueParameter.ErrorOutParameter -> false
 }
 
 internal fun ObjCExportMapper.bridgePropertyType(descriptor: PropertyDescriptor): TypeBridge {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt [hunk 1]
```diff
@@ -316,12 +316,23 @@ private fun ObjCExportMapper.bridgeMethodImpl(descriptor: FunctionDescriptor): M
 
     val returnBridge = bridgeReturnType(descriptor, convertExceptionsToErrors)
     if (convertExceptionsToErrors) {
-        valueParameters += MethodBridgeValueParameter.ErrorOutParameter
+        // Add error out parameter before tail block parameters. The convention allows this.
+        // Placing it after would trigger https://github.com/JetBrains/kotlin-native/issues/3825
```
olonho: Do we know Swift bug number here?

backend.native/tests/framework/values/values.kt
```diff
@@ -935,4 +936,43 @@ class TestStringConversion {
     lateinit var str: Any
 }
 
-fun foo(a: kotlin.native.concurrent.AtomicReference<*>) {}
\ No newline at end of file
+fun foo(a: kotlin.native.concurrent.AtomicReference<*>) {}
+
+interface GH3825 {
```
olonho: This values.kt gets awfully big. Cannot we split it, or at least add new tests to separate files?

backend.native/tests/framework/values/values.swift
```diff
@@ -1197,6 +1197,46 @@ func testStringConversion() throws {
     try test2()
 }
 
+class GH3825SwiftImpl : GH3825 {
```
olonho: Same here, can it be split?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportMapper.kt
SvyatoslavScherbina: I've reported it and improved the comment.

backend.native/tests/framework/values/values.kt
SvyatoslavScherbina: Indeed.
Actually I am going to improve this test infrastructure and start adding tests to separate files with next PR.

PR #3853, opened 2020-02-11
[codegen][debug info] use single compile unit for compilation

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
```diff
@@ -34,6 +35,16 @@ internal val contextLLVMSetupPhase = makeKonanModuleOpPhase(
             val llvmModule = LLVMModuleCreateWithNameInContext("out", llvmContext)!!
             context.llvmModule = llvmModule
             context.debugInfo.builder = LLVMCreateDIBuilder(llvmModule)
+            context.debugInfo.compilationUnit = if (context.shouldContainLocationDebugInfo()) DICreateCompilationUnit(
+                    builder = context.debugInfo.builder,
+                    lang = DWARF.language(context.config),
+                    File = "kotlin-file-${Random.nextLong()}",
```
olonho: Making compilation results non-repeatable doesn't look like a good idea. What we actually need here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -31,6 +31,7 @@ import org.jetbrains.kotlin.konan.target.CompilerOutputKind
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.resolve.descriptorUtil.classId
 import org.jetbrains.kotlin.resolve.descriptorUtil.module
+import kotlin.random.Random
```
olonho: Unused?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
vvlevchenko: We need uniq name for compilation unit to prevent debug info clashing. This usually happens with in compilation with libraries have interception of used files.

SvyatoslavScherbina: Generally we have a lot of information that can be used to produce name unique enough.
For example, resulting binary path, or stable hash of compiler inputs etc.

SvyatoslavScherbina: Compiler inputs are not unique enough though.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
```diff
@@ -33,6 +34,17 @@ internal val contextLLVMSetupPhase = makeKonanModuleOpPhase(
             val llvmModule = LLVMModuleCreateWithNameInContext("out", llvmContext)!!
             context.llvmModule = llvmModule
             context.debugInfo.builder = LLVMCreateDIBuilder(llvmModule)
+
+            context.debugInfo.compilationUnit = if (context.shouldContainLocationDebugInfo()) DICreateCompilationUnit(
+                    builder = context.debugInfo.builder,
+                    lang = DWARF.language(context.config),
+                    File = File(context.config.outputFile).absolutePath,
```
SvyatoslavScherbina: We could also split this to short file name and directory path (if this makes any sense).

vvlevchenko: I've double checked and added comment why we can't split file and directory names

PR #3861, opened 2020-02-13
Rework lazy due to cyclic collector.

olonho:


backend.native/tests/runtime/workers/lazy1.kt
```diff
@@ -9,14 +9,40 @@ import kotlin.test.*
 
 import kotlin.native.concurrent.*
 
-class Leak {
-    val leak by lazy { this }
+class SelfReference {
```
SvyatoslavScherbina: Can a `lazy {}` leak now when global cyclic collector is disabled?

olonho: Yes, as expected.

runtime/src/main/kotlin/kotlin/native/concurrent/Lazy.kt
```diff
@@ -6,55 +6,49 @@
...
                 }
             } else {
-                var result: Any? = value_
-                if (result === UNINITIALIZED) {
-                    result = initializer_!!()
-                    if (isFrozen)
-                        throw InvalidMutabilityException("$this got frozen during lazy evaluation" )
-                    value_ = result
-                    initializer_ = null
+                var result: Any? = value_.value
```
SvyatoslavScherbina: The entire code block has little difference from the one under the lock above.
Isn't it possible to extract it to a function instead?

PR #3862, opened 2020-02-13
Fix lazy in presense of atomic references.

olonho:


backend.native/tests/runtime/workers/lazy1.kt
```diff
@@ -13,10 +13,43 @@ class Leak {
...
+@Test fun runTest1() {
+    assertFailsWith<IllegalStateException> {
+        println(SelfReference().recursion)
+    }
+    assertFailsWith<IllegalStateException> {
+        println(SelfReference().freeze().recursion)
+    }
+}
+
+// Disabled, as requires cyclic collector.
```
SvyatoslavScherbina: I'm afraid this PR would do more harm than good.
I expect number of users having memory leaks in 1.3.70 because of it. Not sure if users have encountered the incorrect behaviour too often.

olonho: They would have leaks only in very convoluted cases which didn't work before (throwing an exception).

SvyatoslavScherbina: I see.

PR #3865, opened 2020-02-13
Fix GC deinit race leading to bogus leak reports.

olonho:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -1747,12 +1747,17 @@ MemoryState* initMemory() {
...
+
 void deinitMemory(MemoryState* memoryState) {
+  atomicAdd(&pendingDeinit, 1);
 #if USE_GC
   bool lastMemoryState = atomicAdd(&aliveMemoryStatesCount, -1) == 0;
   if (lastMemoryState) {
    garbageCollect(memoryState, true);
 #if USE_CYCLIC_GC
+   // We have to ensure that all other deinit sequences are complete.
+   while (atomicGet(&pendingDeinit) > 1) {}
```
SvyatoslavScherbina: Can this loop hang?
For example, if a new memory state has been initialized after decrementing `aliveMemoryStatesCount` above, and then reached `deinitMemory` as "last memory state" too.

PR #3867, opened 2020-02-13
Document NaN propagating behaviour of maxOf/minOf functions for floats

qurbonzoda:


qurbonzoda:
It was decided to have separate discussion before cementing this behavoir

PR #3869, opened 2020-02-14
[Interop] Fixes for metadata mode

sbogolepov:


backend.native/tests/interop/basics/cstructs.def
```diff
@@ -0,0 +1,30 @@
...
+    R, G, B
+};
+
+struct Complex {
+    unsigned int ui;
+    KVector4f vec4f;
+    Trivial t;
+    struct Complex* next;
+    enum E e;
+};
```
SvyatoslavScherbina: + array-typed field

PR #3875, opened 2020-02-14
Add unique prefix to platform libraries.

olonho:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -282,7 +285,9 @@ private fun processCLib(args: Array<String>, additionalArgs: Map<String, Any> =
         _, oldValue, newValue ->
             warn("The package value `$oldValue` specified in .def file is overridden with explicit $newValue")
     }
-    def.manifestAddendProperties["interop"] = "true"
+    def.manifestAddendProperties[KLIB_PROPERTY_INTEROP] = "true"
+    if (outKtPkg.startsWith("platform."))
+        def.manifestAddendProperties[KLIB_PROPERTY_UNIQUE_NAME] = "$platformPrefixNoPlatform${outKtPkg}"
```
alexander-gorshenev: May be make it a cli flag?

utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/GeneratePlatformLibraries.kt
```diff
@@ -10,6 +10,8 @@ import org.jetbrains.kotlin.konan.file.File
 import java.util.concurrent.*
 import kotlinx.cli.*
 
+const val platformPrefix = "org.jetbrains.konan.platform."
```
ilmat192: (minor) May be move this constant into the `shared` project? In this case it will be available in both `platformLibs/build.gradle` and `GeneratePlatformLibraries.kt`.

sbogolepov: `konan` -> `kotlin.native`?

olonho: Not sure if using java keyword is a good idea here.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
olonho: Not sure what is the point in having that outside of platform libraries context.

ilmat192: But now this logic is applied to all interop libraries, isn't it? What if a user creates a library with a `project.something` package?

ilmat192:
Closing since this change was included into https://github.com/JetBrains/kotlin-native/pull/3829.

PR #3878, opened 2020-02-17
Add vararg overloads for maxOf/minOf functions #KT-33906

qurbonzoda:


runtime/src/main/kotlin/generated/_ComparisonsNative.kt
```diff
@@ -136,6 +136,76 @@ public actual inline fun maxOf(a: Double, b: Double, c: Double): Double {
     return maxOf(a, maxOf(b, c))
 }
 
+/**
+ * Returns the greater of given values.
+ */
+@SinceKotlin("1.4")
+public actual fun <T : Comparable<T>> maxOf(a: T, vararg other: T): T {
```
olonho: What about unsigned versions?

qurbonzoda: Unsigned versions are in common source set.
See the PR for big Kotlin: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3810

PR #3879, opened 2020-02-17
Only file names in exception backtrace.

olonho:


SvyatoslavScherbina:
Short names are ambiguous. This can break parsing stack traces in IDE, for example.
Closing this PR for now until we either implement more sophisticated approach here or reconsider IDE part.

PR #3886, opened 2020-02-19
Disable shared cyclic garbage collector until the crash is fixed

SvyatoslavScherbina:


SvyatoslavScherbina:
Known crashes are fixed in #3892, closing this PR for now.

PR #3903, opened 2020-02-24
Remove null terminator from Strings returned by toKStringFromUtf32

Dominaezzz:
Remove the null terminator from `String`s returned by `toKStringFromUtf32`.

The loop should check for the null terminator *before* incrementing `toIndex`.
Found by @z3ntu in https://kotlinlang.slack.com/archives/C3SGXARS6/p1582582248013000 .


SvyatoslavScherbina:
Thank you!

PR #3906, opened 2020-02-25
Added option to add compiler options for kotlinc compiler call in cin…

LepilkinaElena:
…terop


utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/InteropCompiler.kt
```diff
@@ -67,7 +67,8 @@ fun invokeInterop(flavor: String, args: Array<String>): Array<String>? {
         repos.flatMap { listOf("-repo", it) } +
         (if (noDefaultLibs) arrayOf("-$NODEFAULTLIBS") else emptyArray()) +
         (if (noEndorsedLibs) arrayOf("-$NOENDORSEDLIBS") else emptyArray()) +
-        (if (purgeUserLibs) arrayOf("-$PURGE_USER_LIBS") else emptyArray())
+        (if (purgeUserLibs) arrayOf("-$PURGE_USER_LIBS") else emptyArray()) +
+        (if (arguments.kotlincOption.isNotEmpty()) arguments.kotlincOption else emptyList())
```
SvyatoslavScherbina: Isn't this the same as `arguments.kotlincOption`?

PR #3909, opened 2020-02-25
[Interop] fix constants

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -881,6 +882,29 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
...
+        if (!function.descriptor.module.isFromInteropLibrary()) return null
+        if (!function.isGetter) return null
+
+        val constantProperty = (function as? IrSimpleFunction)
+                ?.correspondingPropertySymbol
+                ?.owner
+                ?.takeIf { it.isConst }
+                ?: return null
+
+        return constantProperty.backingField
```
SvyatoslavScherbina: How exactly does this work? Isn't `constantProperty` a lazy IR? If so, why does it have backing field?

sbogolepov: Backing field is generated for constant properties, see https://github.com/JetBrains/kotlin/blob/1f721796b8cc24867b05acc7278a9217cb28173e/compiler/frontend/src/org/jetbrains/kotlin/resolve/resolveUtil.kt#L15.
The initializer for this field is also generated because corresponding descriptor has compile-time initializer (https://github.com/JetBrains/kotlin/blob/ca598b64652aad03294d0cf5d399347e8dc38796/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/lazy/IrLazyField.kt#L62).

PR #3919, opened 2020-02-27
[Interop] Additional simple mangling of special names such as Companion

knebekaizer:


backend.native/tests/interop/objc/smoke.h
```diff
@@ -248,4 +248,28 @@ extern BOOL customStringDeallocated;
...
+
+struct EnumFieldMangleStruct {
+	enum {Companion, Any} smth;
+};
+struct EnumFieldMangleStruct enumMangledStruct = { Any };
+
+struct MyStruct {
+	int Companion;      // simple clash
+	int _Companion;
+	int $_Companion;
```
sbogolepov: Since mangling scheme changed, these cases should be updated as well.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/CodeUtils.kt
```diff
@@ -44,6 +44,20 @@ fun String.asSimpleName(): String = if (this in kotlinKeywords || this.contains(
     this
 }
 
+/**
+Yet another mangler, particularly to avoid secondary clash, e.g. when a property in prototype (interface)
```
SvyatoslavScherbina: Are secondary clashes still relevant here?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/CodeUtils.kt
```diff
@@ -44,6 +44,20 @@ fun String.asSimpleName(): String = if (this in kotlinKeywords || this.contains(
...
+is mangled and that will cause another clash in the class which implements this interface.
+Rationale: keep algorithm simple but use the mangling characters which are rare in normal code, and keep mangling easy readable.
+ */
```
SvyatoslavScherbina: Formatting of the comment is still non-conventional.

backend.native/tests/interop/objc/smoke.h
```diff
@@ -248,4 +248,28 @@ extern BOOL customStringDeallocated;
...
 - (int)instanceMethod;
 + (int)classMethod:(int)first :(int)second;
-@end;
\ No newline at end of file
+@end;
+
+// [KT-36067] cinterop tool fails when there is a structure member named Companion
+
+struct EnumFieldMangleStruct {
+	enum {Companion, Any} smth;
```
SvyatoslavScherbina: Strict enums (i.e. mapped to Kotlin enums) require testing too.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/CodeUtils.kt
knebekaizer: Yes. We try to avoid another clash by applying characters which are not typical for manually written code - that simple. I let this comment to make clear the difference in rationale with `String.mangled`in `ObjCMethod.getKotlinParameterNames`

backend.native/tests/interop/objc/smoke.h
knebekaizer: I tested it separately, as well as global variable, but adding all variants into smoke.h would be a syntax error, and adding more separate test cases may be an overkill for that trivial solution. I consider adding mangler tests to `interop/basics` as enums are not objc specific anymore.

backend.native/tests/interop/objc/smoke.h
knebekaizer: Last three members are examples of name which are _not affected_ by mangling scheme. The case which demostrated "secondary clash" avoidance is removed now, as it is not supported in the simple scheme.

PR #3924, opened 2020-02-28
Pack of small fixes for the new metadata-based interop

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -197,6 +197,15 @@ private fun InteropCallContext.convertCPointerToNativePtr(cPointer: IrExpression
 private fun InteropCallContext.writePointerToMemory(nativePtr: IrExpression, value: IrExpression): IrExpression {
     val valueToWrite = when {
         value.type.isCPointer() -> convertCPointerToNativePtr(value)
+        // Special case of null value
```
SvyatoslavScherbina: This is the reason to avoid relying on actual type (`value.type`) and use expected type (`accessor.valueParameters[0].type`) instead.

PR #3925, opened 2020-02-28
Fix pthread_cond_timedwait usage

projedi:


runtime/src/main/cpp/Worker.cpp
```diff
@@ -413,8 +413,10 @@ class State {
     struct timespec ts;
     gettimeofday(&tv, nullptr);
     KLong nsDelta = millis * 1000LL * 1000LL;
-    ts.tv_nsec = (tv.tv_usec * 1000LL + nsDelta) % 1000000000LL;
-    ts.tv_sec =  (tv.tv_sec * 1000000000LL + nsDelta) / 1000000000LL;
+    constexpr long long nanosecondsInASecond = 1000LL * 1000LL * 1000LL;
+    long long nanoseconds = tv.tv_usec * 1000LL + nsDelta;
+    ts.tv_nsec = nanoseconds % nanosecondsInASecond;
+    ts.tv_sec = tv.tv_sec + nsDelta / nanosecondsInASecond;
```
SvyatoslavScherbina: The entire pattern 
```
timespec ts = gettimeofday() + delta
pthread_cond_timedwait until ts
```
still looks much like a copy-paste. It makes some sense to introduce a function for this.

PR #3926, opened 2020-02-28
Fix for https://youtrack.jetbrains.com/issue/KT-36980

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReturnsInsertionLowering.kt
```diff
@@ -32,14 +31,25 @@ internal class ReturnsInsertionLowering(val context: Context) : FileLoweringPass
...
                     irBuilder.run {
                          (body as IrBlockBody).statements += irReturn(irGetObject(symbols.unit))
                     }
                 }
             }
+
+            override fun visitBlock(expression: IrBlock) {
+                expression.acceptChildrenVoid(this)
+                if (expression !is IrReturnableBlock) return
+                if (expression.inlineFunctionSymbol?.owner?.returnType == context.irBuiltIns.unitType) {
```
SvyatoslavScherbina: Which inline function has `Unit` return type in the test case?

homuroll: The lambda being passed to `let` function.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1818,11 +1818,7 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
         val bbExit = returnableBlockScope.bbExit
         if (bbExit != null) {
             if (!functionGenerationContext.isAfterTerminator()) {                 // TODO should we solve this problem once and for all
-                if (returnableBlockScope.resultPhi != null) {
-                    functionGenerationContext.unreachable()
-                } else {
-                    functionGenerationContext.br(bbExit)
-                }
+                functionGenerationContext.unreachable()
```
SvyatoslavScherbina: Please run external tests to ensure that emitting `unreachable` unconditionally is ok now.

PR #3934, opened 2020-03-03
Detect leaked workers

projedi:


backend.native/tests/runtime/workers/leak_memory_with_worker_termination.kt
```diff
@@ -0,0 +1,9 @@
+import kotlin.native.concurrent.*
+import kotlinx.cinterop.*
+
+fun main() {
+    val worker = Worker.start()
+    Worker.current.park(1000 * 1000);
```
SvyatoslavScherbina: There is a more reliable and efficient approach to ensure the worker is fully initialized: it is enough to schedule the task and wait for its result.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -402,7 +401,8 @@ class ForeignRefManager {
 
       if (hadNoStateInitialized) {
         // Discard the memory state.
-        DeinitMemory(memoryState);
+        if (!DeinitMemory(memoryState))
+          konan::abort();
```
SvyatoslavScherbina: What was moving `abort` out from `DeinitMemory` exactly required for?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1747,19 +1747,13 @@ MemoryState* initMemory() {
...
-  atomicAdd(&pendingDeinit, 1);
+bool deinitMemory(MemoryState* memoryState) {
 #if USE_GC
   bool lastMemoryState = atomicAdd(&aliveMemoryStatesCount, -1) == 0;
-  bool checkLeaks = g_checkLeaks && lastMemoryState;
+  bool checkLeaks = Kotlin_memoryLeakCheckerEnabled() && lastMemoryState;
   if (lastMemoryState) {
    garbageCollect(memoryState, true);
 #if USE_CYCLIC_GC
-   // If there are other pending deinits (rare situation) - just skip the leak checker.
```
SvyatoslavScherbina: Can this race still happen with threads that aren't produced with `Worker.start()`?

runtime/src/main/cpp/Memory.cpp
projedi: That used to be for consistency with `WorkerDeinit`, where I didn't use `abort` in order to run memory leak checker even after leaked workers were found.

However, current implementation won't allow running memory leak checker in that situation anyway, so I restored `abort` here and in `WorkerDeinit`.

runtime/src/main/cpp/Memory.cpp
projedi: True, restored the check and added a comment explaining when this race may happen.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1752,11 +1751,13 @@ void deinitMemory(MemoryState* memoryState) {
...
 #if USE_GC
   bool lastMemoryState = atomicAdd(&aliveMemoryStatesCount, -1) == 0;
-  bool checkLeaks = g_checkLeaks && lastMemoryState;
+  bool checkLeaks = Kotlin_memoryLeakCheckerEnabled() && lastMemoryState;
   if (lastMemoryState) {
    garbageCollect(memoryState, true);
 #if USE_CYCLIC_GC
    // If there are other pending deinits (rare situation) - just skip the leak checker.
+   // This may happen when there're several threads with Kotlin runtimes created
+   // by foreign code, and that code stops those threads simultaneously.
```
SvyatoslavScherbina: Is this still true?

build-tools/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -197,7 +197,10 @@ open class KonanLocalTest : KonanTest() {
     override var testLogger = Logger.SILENT
 
     @Input @Optional
-    var expectedExitStatus = 0
+    var expectedExitStatus: Int? = null
```
SvyatoslavScherbina: @PavelPunegov please review changes to this file.

runtime/src/main/cpp/Memory.cpp
projedi: Yes. Memory leak checker still runs on "foreign" threads termination, it's not bound to `main()` termination.

PR #3939, opened 2020-03-04
Fixing KT-37212

artdfel:
[YT ticket here](https://youtrack.jetbrains.com/issue/KT-37212).
Initializing constant value with an infinity makes the value equal to zero.
Unification of the floating-point constants evaluation with other primitive types fixes it. Seems like the `LLVMConstRealOfString` caused the problem for some reason.

### DIFF
--- a/backend.native/tests/codegen/basics/const_infinity.kt
+++ b/backend.native/tests/codegen/basics/const_infinity.kt
@@ -7,12 +7,15 @@ package codegen.basics.const_infinity
 
 import kotlin.test.*
 
+//Original issue here https://youtrack.jetbrains.com/issue/KT-37212
+@Suppress("DIVISION_BY_ZERO")
 const val fpInfConst = 1.0F / 0.0F
+@Suppress("DIVISION_BY_ZERO")
 val fpInfVal = 1.0F / 0.0F
 
 @Test
 fun runTest() {
-    if (fpInfConst!=Float.POSITIVE_INFINITY) throw Error()
-    if (fpInfVal != Float.POSITIVE_INFINITY) throw Error()
-    if (fpInfConst!=fpInfVal) throw Error()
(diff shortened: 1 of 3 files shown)


sbogolepov:
Nice work! You also need to add a test for this case.
1. Add file to `/Users/jetbrains/IdeaProjects/kotlin-native/backend.native/tests/codegen/basics`. For example, you can call it `kt37212.kt`.
2. Add a test case. (Check equality of top-level const val and non-const val)
3. Add a corresponding Gradle task ([example](https://github.com/JetBrains/kotlin-native/blob/6d65722b6776627cb82604decc01ebd98015cc7c/backend.native/tests/build.gradle#L1289)).
4. Run `:backend.native:tests:{task_name}` Gradle task to verify that everything is OK.

backend.native/tests/codegen/basics/const_infinity.kt
```diff
@@ -0,0 +1,18 @@
...
+package codegen.basics.const_infinity
+
+import kotlin.test.*
+
+const val fpInfConst = 1.0F / 0.0F
+val fpInfVal = 1.0F / 0.0F
+
+@Test
+fun runTest() {
+    if (fpInfConst!=Float.POSITIVE_INFINITY) throw Error()
```
sbogolepov: There is a special [function](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-equals.html) for that. It is used in other tests.

Also it is useful to mention an origin issue in a comment.

Thomas-Vos:
The following prints "test: 0" with 1.3.70 for me. Only happens in release mode for the watchOS simulator (watchosX86). Does this pull request fix this issue as well?

    println("test: " + 5.toDouble().toInt()) // prints 0 instead of 5

artdfel:
@Thomas-Vos no, it seems like this problem is still here. We're looking at it, thanks for the report.

PR #3940, opened 2020-03-04
Add SIZE_BITS and SIZE_BYTES for Double and Float

ilya-g:
Common part is reviewed in https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3918

In addition to new constants SIZE_BITS and SIZE_BYTES, there are properties in `Double/Float`: `POSITIVE_INFINITY`, `NEGATIVE_INFINITY`, `NaN` and so on, that are now constant in common builtins. 
However here, I can't make them constant due to the bug in constant initiazers: https://youtrack.jetbrains.com/issue/KT-37212
~Surprisingly, tests that expect MIN/MAX VALUE to be constants still pass, even though they are not constants in Native.~ - ah, that's because they are already constants in K/N. So KT-37212 is somewhat a blocker, though I think we can live up to RC with this constant inconsistency.

### DIFF
--- a/backend.native/tests/stdlib_external/utils.kt
+++ b/backend.native/tests/stdlib_external/utils.kt
@@ -24,3 +24,6 @@ internal actual inline fun testOnNonJvm6And7(f: () -> Unit) {
 
 actual fun testOnJvm(action: () -> Unit) {}
 actual fun testOnJs(action: () -> Unit) {}
+
+
+public actual val isFloat32RangeEnforced: Boolean get() = true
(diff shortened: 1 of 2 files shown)


ilya-g:
I've temporary included the fix of KT-37212 by @artdfel in my branch until it gets in `kotlin-mirror/master`

SvyatoslavScherbina:
What is the status of this PR?

ilya-g:
@SvyatoslavScherbina  Ready for review in Native, but blocked for some obstacles in the common part.

SvyatoslavScherbina:
@PavelPunegov could you please review? Also note https://github.com/JetBrains/kotlin-native/pull/3940#issuecomment-598174576.

PR #3946, opened 2020-03-05
`String.toBoolean()` should be `String?.toBoolean()` #KT-14119

qurbonzoda:


runtime/src/main/kotlin/kotlin/text/StringNumberConversions.kt
```diff
@@ -52,10 +52,11 @@ external internal fun longToString(value: Long, radix: Int): String
 public actual inline fun Long.toString(radix: Int): String = longToString(this, checkRadix(radix))
 
 /**
- * Returns `true` if the contents of this string is equal to the word "true", ignoring case, and `false` otherwise.
+ * Returns `true` if this string is not `null` and it's contents is equal to the word "true", ignoring case, and `false` otherwise.
```
ilya-g:
```suggestion
 * Returns `true` if this string is not `null` and its content is equal to the word "true", ignoring case, and `false` otherwise.
```
Either "content is" or "contents are". IMO, the first one is better.

PR #3947, opened 2020-03-05
[IR] Fix for https://youtrack.jetbrains.com/issue/KT-37057 + test

homuroll:


backend.native/tests/codegen/inlineClass/customEquals.kt
```diff
@@ -0,0 +1,16 @@
...
+package codegen.inlineClass.customEquals
+
+import kotlin.test.*
+
+private inline class Z(val data: Int) {
+    override fun equals(other: Any?) = other is Z && data % 256 == other.data % 256
+}
+
+@Test fun runTest() {
+    assertEquals(Z(0), Z(256))
```
SvyatoslavScherbina: This test doesn't actually reproduce the original issue with `==`.

backend.native/tests/codegen/inlineClass/customEquals.kt
```diff
@@ -0,0 +1,16 @@
...
+package codegen.inlineClass.customEquals
+
+import kotlin.test.*
+
+private inline class Z(val data: Int) {
+    override fun equals(other: Any?) = other is Z && data % 256 == other.data % 256
+}
+
+@Test fun runTest() {
+    println(Z(0) == Z(256))
```
SvyatoslavScherbina: Using `assertTrue` for this would make this test more self-contained.

PR #3957, opened 2020-03-09
Fix android executables

penn5:
Presently, it is not possible to build an android executable (ELF) file due to the hardcoded presence of the -shared flag. I haven't tested this as my laptop isn't powerful enough to compile it. The test would be to compile a project with the following buildscript:
```
plugins {
    kotlin("multiplatform") version "1.3.70"
}

repositories {
    mavenCentral()
}

kotlin {
    androidNativeX64("android") {
        binaries {
            executable {
                entryPoint = "example.main"
            }
        }
    }
}
```
and see if it generates a .so or a real executable - as the executable{} block was used, it should be an executable, but currently generates a library.

### DIFF
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -497,6 +497,7 @@ quadruple.android_arm32 = arm-linux-androideabi
 clangFlags.android_arm32 = -cc1 -target-cpu arm7tdmi -emit-obj -disable-llvm-optzns -x ir -femulated-tls -mrelocation-model pic
 clangOptFlags.android_arm32 = -O3 -ffunction-sections
 linkerNoDebugFlags.android_arm32 = -Wl,-S
+linkerDynamicFlags.android_arm32 = -shared -Wl,-soname,${File(executable).name}
 clangNooptFlags.android_arm32 = -O1
 targetSysRoot.android_arm32 = target-sysroot-1-android_ndk
 linkerKonanFlags.android_arm32 = -lm -lc++_static -lc++abi -landroid -llog -latomic
@@ -528,6 +529,7 @@ clangNooptFlags.android_arm64 = -O1
 targetSysRoot.android_arm64 = target-sysroot-1-android_ndk
 linkerKonanFlags.android_arm64 = -lm -lc++_static -lc++abi -landroid -llog -latomic
 linkerNoDebugFlags.android_arm64 = -Wl,-S
+linkerDynamicFlags.android_arm64 = -shared -Wl,-soname,${File(executable).name}
 
 # Android X86, based on NDK.
 targetToolchain.macos_x64-android_x86 = target-toolchain-2-osx-android_ndk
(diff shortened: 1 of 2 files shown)


r4zzz4k:
IIRC Android target's "application" is meant to be used for NDK-based apps (`NativeActivity`, `android_main`, that stuff), not for generic Android binaries, which are rarely needed. And you build .so for this case, that's where `-dynamic` comes from.
Not judging whether that's rational decision or not as I didn't have need for both of these cases yet.

SvyatoslavScherbina:
> IIRC Android target's "application" is meant to be used for NDK-based apps (NativeActivity, android_main, that stuff), not for generic Android binaries

This is correct.

penn5:
> > IIRC Android target's "application" is meant to be used for NDK-based apps (NativeActivity, android_main, that stuff), not for generic Android binaries
> 
> This is correct.

So how would I build a native true executable binary? It's essential to my project. If the executable block is used to build a .so, what is the difference with DYNAMIC_LIBRARY supposed to be? From what I can see, the only difference before this PR is the `-Wl,soname=` parameter, which sets the SONAME field in the generated .so, and is not really relevant in this case. Not specifying this parameter (as when you generate an executable) will have little to no effect, and certainly no positive effect. Hence, the only concern here would be backwards compatibility, I think.

SvyatoslavScherbina:
> Not specifying this parameter (as when you generate an executable) will have little to no effect, and certainly no positive effect.

Have you actually checked any of this?

penn5:
> > Not specifying this parameter (as when you generate an executable) will have little to no effect, and certainly no positive effect.
> 
> Have you actually checked any of this?

No. I read the docs for the first time today

penn5:
But according to some old LD docs, the -soname parameter will affect linking, however, if it is used by JNI rather than a normal dynamic linker, as you suggested, I don't think this has any effect. My laptop died so I can't really check anything in practice so everything is from SO or docs.

SvyatoslavScherbina:
>> Have you actually checked any of this?

> No. I read the docs for the first time today

Then there is no much value in this PR yet. Please ensure that your solution works first.

penn5:
> > > Have you actually checked any of this?
> 
> > No. I read the docs for the first time today
> 
> Then there is no much value in this PR yet. Please ensure that your solution works first.

Yep, I understand. I was hoping someone else could test it, as I have no idea when I'll be able to build.

penn5:
It doesn't work. I know why, but there are two options to fix it:
1. Replace the `executable` type to output a real executable
2. Create a new `androidExecutable` target to handle the generation of non-NativeActivity programs
I would err towards the second one, because the first option carries the disadvantage that you can't make non-NativeActivity shared objects. It is probably easier to implement the first option, but the second is fully backwards compatible. However, it would probably reduce maintainability. I don't think I can reasonably implement either of these, so I'm closing this PR, but I'll open an issue.

A short list of things that would have to be changed for the second option:
- Create new `KONAN_ANDROID_EXE` which has the behaviour of `KONAN_LINUX` in the launcher, but returns osFamily Android.
- Create new `androidExecutable` output type in both konan and the gradle plugin
- Make `androidExecutable` output the right file types
- Also probably rework the `androidNative` stuff, as it appears very weird currently.

PR #3959, opened 2020-03-09
Expose validity of objc stubs' descriptors

fkistner:


SvyatoslavScherbina:
My approve here means only that "I can see this PR doesn't break K/N itself". I'm not an expert in Kotlin-IDEA integration, so generally have no idea if this is actually correct.
So feel free to invite any competent reviewer if necessary.

dsavvinov:
It's hard for me to tell whether those changes are correct from IDEA point of view, as I don't know how IDE calls this API. Abstractly speaking, `isValid` calls sound more or less sane for me; but yet again, in CLI environment `isValid` is pretty much a no-op, and IDE it's behaviour highly depends on how frontend has been configured, so it's hard to tell without inspecting actual clients.

Folks, am I right that our IDE plugins do not call this API, and it's needed only for mobile team needs?

SvyatoslavScherbina:
> Folks, am I right that our IDE plugins do not call this API, and it's needed only for mobile team needs?

AFAIK you are right.

fkistner:
> Folks, am I right that our IDE plugins do not call this API, and it's needed only for mobile team needs?
Yes, it's mostly to make things easier for us.
Note that `isValid` is only really relevant for our use-case, where we want to continue translating at some later point in time.
So unless you have concerns about the semantics, it shouldn't really affect anyone else.

SvyatoslavScherbina:
Ok. Should we merge this PR now then?

dsavvinov:
Yes, think we should

SvyatoslavScherbina:
Merged to master. Will be available in 1.4-M2.

PR #3962, opened 2020-03-10
[KLIB] Fix K/N binary interface

romanart:
- make it use return type for binary symbol name

Fix goes along with the corresponding Kotlin one.


vvlevchenko:
cherry-picked

PR #3964, opened 2020-03-10
[Runtime] Implement getValue for all KotlinNumber types [KT-36204]

knebekaizer:


runtime/src/objc/cpp/ObjCExportNumbers.mm
```diff
@@ -270,6 +270,10 @@ @implementation KotlinBoolean {
   BOOL value_;
 }
 
+- (void)getValue:(void *)value {
```
SvyatoslavScherbina: These classes are actually generated with "script" above: https://github.com/JetBrains/kotlin-native/blob/e8ef64764cf8ff30c911d3b37d2d3b433ea9c47c/runtime/src/objc/cpp/ObjCExportNumbers.mm#L103
So it would be great to fix the script too.

knebekaizer: Done

benasher44:
One thing that's worth noting I think is this getValue is deprecated: https://developer.apple.com/documentation/foundation/nsvalue/1415141-getvalue. There's another one that takes a size param that's not deprecated.

SvyatoslavScherbina:
@benasher44 thank you for noticing this!

PR #3966, opened 2020-03-10
Migrate StablePtr to KRefSharedHolder

projedi:
This PR also fixes tests from #3969

### DIFF
--- a/runtime/src/main/cpp/MemoryPrivate.hpp
+++ b/runtime/src/main/cpp/MemoryPrivate.hpp
@@ -36,6 +36,10 @@ void DeinitForeignRef(ObjHeader* object, ForeignRefContext context);
 
 bool IsForeignRefAccessible(ObjHeader* object, ForeignRefContext context);
 
+// If this object is shareable, increments its RC and schedules its decrementing.
+// Should be used when assigning shared objects to a stack variable.
+void InitStackRef(ObjHeader* object);
+
 }  // extern "C"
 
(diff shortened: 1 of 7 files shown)


runtime/src/main/cpp/MemoryPrivate.hpp
```diff
@@ -36,6 +36,10 @@ void DeinitForeignRef(ObjHeader* object, ForeignRefContext context);
 
 bool IsForeignRefAccessible(ObjHeader* object, ForeignRefContext context);
 
+// If this object is shareable, increments its RC and schedules its decrementing.
+// Should be used when assigning shared objects to a stack variable.
+void InitStackRef(ObjHeader* object);
```
SvyatoslavScherbina: I find the name and description of this function too narrow:
1. It specifies too many details about RC manipulation
2. Assigning shared objects to a stack variable is not a problem itself.

Also `InitStackRef` is somewhat confusing here: `ObjHeader* object` is not exactly a stack ref.

We indeed have leaking abstractions here and this code probably requires some refactoring (later).
For now let's try to minimize the effect of this.
Maybe something like `AdoptReferenceFromSharedVariable`?
With description stating that it should be used when reference is read from possibly shared variable and nothing else guarantees the object to be alive long enough.

SvyatoslavScherbina:
Please don't forget to mention that this PR fixes #3969 in the commit message when merging.

PR #3941, opened 2020-03-04
Nullable Array.contentEquals/contentHashCode/contentToString #KT-34161

qurbonzoda:

PR #3970, opened 2020-03-11
Filter out nullable extensions to arrays

PavelPunegov:
Fixes issue found in https://github.com/JetBrains/kotlin-native/pull/3941


ilya-g:
Note that the extensions for non-nullable receivers are made LowPriority and eventually will be deprecated. So it makes sense to stop using them in compiler generated code at some point as well.

SvyatoslavScherbina:
I see. Thanks!

PR #3972, opened 2020-03-12
Made cinterop command line options consistent

LepilkinaElena:


utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/InteropCompiler.kt
```diff
@@ -27,22 +27,17 @@ fun invokeInterop(flavor: String, args: Array<String>): Array<String>? {
     val temporaryFilesDir = arguments.tempDir
 
     val buildDir = File("$outputFileName-build")
-    val generatedDir = File(buildDir, "kotlin")
-    val nativesDir = File(buildDir, "natives")
+    val generatedDir = File(arguments.generated)
+    val nativesDir = File(arguments.natives)
```
SvyatoslavScherbina: Do I understand correctly that `generatedDir` and `nativesDir` aren't located under `buildDir` anymore when calling `cinterop`?

LepilkinaElena: They are under if you use default values. I changed default values, and now default value of these parameters are created like `nativelib-build/kotlin`. The only problem is that if reset output parameter you have to reset these two also.
Unfortuantly, that's impossible now to show dependency on `kotlinx.cli`

LepilkinaElena: But it seems better to me to use such approach than just redefine provided by user arguments

SvyatoslavScherbina: > They are under if you use default values.

Is this true if you use non-default output (which is almost always the case)?

LepilkinaElena: You should reset them if you use not default output. There are predefined parameters from native interop plugin 
```
File getNativeLibsDir() {
        return new File(project.buildDir, "nativelibs/$target")
    }

    File getGeneratedSrcDir() {
        return new File(project.buildDir, "nativeInteropStubs/$name/kotlin")
    }
```

LepilkinaElena: Ok, I see problem in MPP gradle plugin. Will it be ok if I fix there by adding not default parameters?

LepilkinaElena: Or is it better, to change command line to allow to change values of generated and natives only inside buildDir?

LepilkinaElena: https://github.com/JetBrains/kotlin/commit/f5f8dfac708d0aae8d7720d2431b8f642044ca82

SvyatoslavScherbina: What I mean is
```
$ cinterop -def test.def -o test.klib
```
creates **both** `nativelib-build` and `test.klib-build` with your patch. It wasn't acting like this before.

LepilkinaElena: Yes, now you have to add `-generated test.klib/kotlin -natives test.klib/natives`.I made such changes in MPP plugin. Is this critical?

sbogolepov: > Is this critical?

Yes. It makes CLI cumbersome for the default use-case.
BTW, `CommonInteropArguments.generated` is not needed for `metadata` mode.

LepilkinaElena: Ok, added new feature to kotlinx.cli, but I'll wait review from Ilya G. https://github.com/Kotlin/kotlinx.cli/pull/22/files

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -166,12 +185,10 @@ private fun findFilesByGlobs(roots: List<Path>, globs: List<String>): Map<Path,
...
 
-private fun processCLib(args: Array<String>, additionalArgs: Map<String, Any> = mapOf()): Array<String>? {
-    val cinteropArguments = CInteropArguments()
-    cinteropArguments.argParser.parse(args)
-    val ktGenRoot = cinteropArguments.generated
-    val nativeLibsDir = cinteropArguments.natives
-    val flavorName = cinteropArguments.flavor
+private fun processCLib(flavorName: String, cinteropArguments: CInteropArguments,
+                        additionalArgs: Map<String, Any> = mapOf()): Array<String>? {
+    val ktGenRoot = additionalArgs["generated"]!! as String
```
sbogolepov: Please use `getValue` instead.

utilities/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/InteropCompiler.kt
```diff
@@ -28,24 +28,21 @@ fun invokeInterop(flavor: String, args: Array<String>): Array<String>? {
...
     val cstubsName ="cstubs"
     val libraries = arguments.library
     val repos = arguments.repo
     val targetRequest = if (arguments is CInteropArguments) arguments.target
         else (arguments as JSInteropArguments).target
     val target = PlatformManager().targetManager(targetRequest).target
 
-    val cinteropArgsToCompiler = interop(flavor, args + additionalArgs,
+    val cinteropArgsToCompiler = interop(flavor, args,
             listOfNotNull(
```
sbogolepov: All these maps and unsafe casts are looking very python-ish :)
There are better solution in a typed world. For example, you can create a `data` class.

PR #3973, opened 2020-03-12
[Runtime] Use mimalloc aligned allocation for interop types

knebekaizer:


PavelPunegov:
Tests have been rewritten: moved all vector types to a separate test.

backend.native/tests/interop/basics/vectors.kt
```diff
@@ -0,0 +1,35 @@
+import kotlinx.cinterop.*
+import kotlin.native.*
+import kotlin.test.*
+import cvectors.*
+
+fun isWin32() = Platform.osFamily == OsFamily.WINDOWS && Platform.cpuArchitecture == CpuArchitecture.X86
+
+fun main(args: Array<String>) {
+    val mimalloc = args.takeIf { it.size == 1 }?.get(0) == "mimalloc"
```
SvyatoslavScherbina: Where is this argument passed from?

backend.native/tests/interop/basics/vectors.kt
```diff
@@ -0,0 +1,35 @@
...
+    if (mimalloc || !isWin32()) {
+        produceComplex().useContents {
+            assertEquals(vec4f, vectorOf(1.0f, 1.0f, 1.0f, 1.0f))
+            vec4f = vectorOf(0.0f, 0.0f, 0.0f, 0.0f)
+            assertEquals(vec4f, vectorOf(0.0f, 0.0f, 0.0f, 0.0f))
+        }
+    }
+
+    // FIXME: KT-36285
+    if (Platform.osFamily != OsFamily.LINUX && Platform.cpuArchitecture != CpuArchitecture.ARM32) {
```
SvyatoslavScherbina: Shouldn't it be `||` here?

runtime/src/std_alloc/cpp/AllocImpl.cpp
```diff
@@ -8,8 +8,14 @@
 extern "C" {
 // Memory operations.
 void* konan_calloc_impl(size_t n_elements, size_t elem_size) {
- return calloc(n_elements, elem_size);
+  return calloc(n_elements, elem_size);
 }
+
+void* konan_calloc_aligned_impl(size_t count, size_t size, size_t alignment) {
+  // alignment is not supported by std alloc - use mimalloc
+  return calloc(count, size);
```
SvyatoslavScherbina: Could you please provide more context on this?
Isn't `calloc` supposed to return pointer aligned sufficiently for any type?

backend.native/tests/interop/basics/vectors.kt
PavelPunegov: Yes, it should!

backend.native/tests/interop/basics/vectors.kt
PavelPunegov: From the build.gradle

runtime/src/std_alloc/cpp/AllocImpl.cpp
PavelPunegov: Can the `aligned_alloc` be used here https://en.cppreference.com/w/c/memory/aligned_alloc ?

backend.native/tests/build.gradle
```diff
@@ -3591,6 +3609,7 @@ interopTest("interop_structs") {
     disabled = (project.testTarget == 'wasm32') // No interop for wasm yet.
     source = "interop/basics/structs.kt"
     interop = 'cstructs'
+    flags = [ "-Xallocator=mimalloc" ]
```
SvyatoslavScherbina: Probably not required anymore.

PavelPunegov: Yes, this is not required.

runtime/src/std_alloc/cpp/AllocImpl.cpp
knebekaizer: On posix calloc: `The pointer returned if the allocation succeeds shall be suitably aligned so that it may be assigned to a pointer to any type`. I believe that other flavours follow the same rule. This would be enough for typical cases but it does not work on 32-bit platform when we want to align a struct containing SIMD vector. W.r.t. vector types it's not quite clear even for 64-bit: does that specific hardware or virtualization platform implements vector as a first-class type or not, and what particular libc port knows about that. For example, we've seen occasional failure on some AWS installation. 
@SvyatoslavScherbina Would you like this comment to be included in the source or git message (the latter is my preferable)?
It's still questionable if we may want sometime to overwrite alignment to make it _less_ than pointer size, e.g. for a packed struct (we don't rely on that now).

SvyatoslavScherbina: > Would you like this comment to be included in the source or git message (the latter is my preferable)?

I have no preferences.

> his would be enough for typical cases but it does not work on 32-bit platform when we want to align a struct containing SIMD vector.

And that's what I don't get. Doesn't "suitably aligned so that it may be assigned to a pointer to any type" mean that it should be suitably aligned for any pointer type, including pointers to vector types and to structs with vector fields? For example "if you assign the result of `calloc` to some pointer, operations with this pointer should work ok", and "operations" clearly include reading and writing to this pointer.

knebekaizer: On `aligned_alloc`: it may be good enhancement for std allocator option. It can't be mixed with mimalloc indeed. I suggest to take it as a separate patch. The question is: do we want to support std allocator option in future, or just switch to mimalloc as default?

knebekaizer: > it should be suitably aligned for any pointer type, _including pointers to vector types_
The fact is: the latter assumption not necessary works. One may blame the low quality of libc implementation on a specific platform. OTOH what exactly is "any type" here? I would expect "any == any C standard type". Vector is not a C standard type, but compiler- and HW-vendor extension, declared using different non-standard attributes. Strictly conforming `libc` is not required to support vector type alignment. So, the conservative assumption is: non-standard type may need additional care (i.e. manual help of the programmer), particularly in the grey zone between HW, library and compiler.

knebekaizer: Yet another two cents: aligning _any_ allocation  to 16 bytes (by default) on 32-bit platform may be just too expensive, particularly for embedded applications that never use vector types. So, from library perspective it seems pretty reasonable to implement bigger alignment as opt-in.

SvyatoslavScherbina: Yes, that's what I expected. Thank you for the explanation!
Using `aligned_alloc` makes sense.

PR #3978, opened 2020-03-13
Disable leak tests on wasm32

projedi:


backend.native/tests/build.gradle
```diff
@@ -1030,13 +1030,15 @@ task enumIdentity(type: KonanLocalTest) {
 }
 
 standaloneTest("leakWorker") {
+    disabled = project.globalTestArgs.contains('-opt') || (project.testTarget == 'wasm32') // Needs debug build and pthreads.
```
SvyatoslavScherbina: Do these tests pass without both `-opt` and `-g` flags?

SvyatoslavScherbina: (Discussed offline that this test forces `-g` so the check above is ok).

PR #3979, opened 2020-03-13
Printing linker error message from  gradle MPP plugin

LepilkinaElena:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/exec/ExecuteCommand.kt
```diff
@@ -47,23 +48,23 @@ open class Command(initialCommand: List<String>) {
...
+    open fun runProcess() {
         val builder = ProcessBuilder(command)
 
         builder.redirectOutput(Redirect.INHERIT)
         builder.redirectInput(Redirect.INHERIT)
-        builder.redirectError(Redirect.INHERIT)
 
         val process = builder.start()
+
+        val reader = BufferedReader(InputStreamReader(process.errorStream))
```
PavelPunegov: It seems that error stream is now will not be visible if process exits correctly. Where are all warnings printed? Into err or std out?

LepilkinaElena: Before nothing is seen if run with gradle plugin, only if run `konanc` you got all messages. I think how to save both modes.

LepilkinaElena: Warnings are printed in stderr

shared/src/main/kotlin/org/jetbrains/kotlin/konan/exec/ExecuteCommand.kt
```diff
@@ -47,23 +48,23 @@ open class Command(initialCommand: List<String>) {
...
+
         val exitCode = process.waitFor()
-        return exitCode
+        handleExitCode(exitCode, reader.readLines())
```
PavelPunegov: It doesn't return anything, while it's required in Linker.kt

PavelPunegov: See https://buildserver.labs.intellij.net/viewLog.html?buildId=66946137&problemId=1583664

PR #3986, opened 2020-03-16
Fix using cached Kotlin subclasses of Objective-C classes

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -1009,8 +1009,8 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
                 error("type-checking against Kotlin classes inheriting Objective-C meta-classes isn't supported yet")
             }
 
-            val objCDeclarations = context.llvmDeclarations.forClass(irClass).objCDeclarations!!
-            val classPointerGlobal = objCDeclarations.classPointerGlobal.llvmGlobal
+            val classInfo = codegen.kotlinObjCClassInfo(irClass)
+            val classPointerGlobal = load(structGep(classInfo, 12))
```
homuroll: How about adding a comment (like `12 /*createdClass*/`)?

SvyatoslavScherbina: Done.

SvyatoslavScherbina:
> What about tests?

@sbogolepov 
I have tested the case provided by reporter -- it compiles now.
I hope to do more manual testing before freezing 1.3.71. Don't have enough time for automated testing though. If you feel that it is required, I would gratefully accept your help with this.

PR #3996, opened 2020-03-19
Llvm for daemon fixes

homuroll:


backend.native/llvm.def
```diff
@@ -18,7 +18,8 @@ linkerOpts = -fvisibility-inlines-hidden \
     -pedantic -Wno-long-long -Wcovered-switch-default -Wnon-virtual-dtor -Wdelete-non-virtual-dtor \
     -std=c++14 \
     -DNDEBUG -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS \
-    -ldebugInfo -lcoverageMapping
+    -ldebugInfo -lcoverageMapping \
+    -Wl,-exported_symbols_list,llvm.list
```
SvyatoslavScherbina: Not sure if this works on Linux or Windows.
Btw, do we need the similar fix on these platforms?

homuroll: I will check.

backend.native/llvm.list
```diff
@@ -0,0 +1 @@
+_Java_llvm_*
```
SvyatoslavScherbina: Considering that `-exported_symbols_list` doesn't fail when wildcard doesn't match, it may make sense to omit package here (e.g. use `_Java_*`) to ensure this fix keeps working if we change the package name.

PR #3998, opened 2020-03-19
Support suppressed exceptions

ilya-g:
This PR introduces implementations for the following common API:
- `Throwable.addSuppressed(Throwable)`
- `Throwable.suppressedExceptions: List<Throwable>`
- `Throwable.stackTraceToString()`
- makes `printStackTrace` common
https://youtrack.jetbrains.com/issue/KT-35013
https://youtrack.jetbrains.com/issue/KT-37603
https://youtrack.jetbrains.com/issue/KT-38044

Also it refactors stack trace rendering in order to:
- add suppressed exceptions support
- omit common stackframes of causes and suppressed exceptions
- reuse the single stringbuilder for all string concatenations during the exception stacktrace rendering or printout.

Common part is reviewed in https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3982

### DIFF
--- a/runtime/src/main/kotlin/kotlin/Throwable.kt
+++ b/runtime/src/main/kotlin/kotlin/Throwable.kt
@@ -6,6 +6,7 @@
 package kotlin
 
 import kotlin.native.concurrent.freeze
+import kotlin.native.concurrent.isFrozen
 import kotlin.native.internal.ExportForCppRuntime
 import kotlin.native.internal.ExportTypeInfo
 import kotlin.native.internal.NativePtrArray
@@ -75,6 +76,8 @@ public open class Throwable(open val message: String?, open val cause: Throwable
         val s = kClass.qualifiedName ?: kClass.simpleName ?: "Throwable"
         return if (message != null) s + ": " + message.toString() else s
     }
+
+    internal var suppressedExceptionsList: MutableList<Throwable>? = null
 }
 
 @SymbolName("Kotlin_getCurrentStackTrace")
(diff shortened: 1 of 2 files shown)


runtime/src/main/kotlin/kotlin/Throwable.kt
```diff
@@ -42,43 +43,150 @@ public open class Throwable(open val message: String?, open val cause: Throwable
...
+ * - the short description (see [Throwable.toString]) of this throwable;
+ * - the complete stack trace;
+ * - detailed descriptions of the exceptions that were [suppressed][suppressedExceptions] in order to deliver this exception;
+ * - the detailed description of each throwable in the [Throwable.cause] chain.
+ */
+@SinceKotlin("1.4")
+public actual fun Throwable.stackTraceToString(): String = dumpStackTrace()
+
+/**
+ * Prints the [detailed description][Throwable.stackTraceToString] of this throwable to the standard error output.
```
ilya-g, line 161: Currently it prints to the standard output, not the error output, so need to correct docs here.

I think it would be inconsistent being able to output to stderr exceptions only, while we don't expose general API for that in Kotlin.

runtime/src/main/kotlin/kotlin/Throwable.kt
```diff
@@ -42,43 +43,150 @@ public open class Throwable(open val message: String?, open val cause: Throwable
             (0 until stackTrace.size).map { index -> stackTrace[index].toLong() }
 
     /**
-     * Prints the stack trace of this throwable to the standard output.
+     * Prints the [detailed description][Throwable.toStringWithTrace] of this throwable to the standard output.
```
ilya-g, line 46:
```suggestion
     * Prints the [detailed description][Throwable.stackTraceToString] of this throwable to the standard output.
```

ilya-g:
Merged manually

PR #4003, opened 2020-03-20
Fix sharing for instances of Kotlin subclasses of Obj-C classes

SvyatoslavScherbina:
Make Obj-C references to these objects properly shareable,
fix deallocation on other threads.

To achieve that, move Kotlin fields to Kotlin counterpart
and make its lifetime equal to Obj-C counterpart lifetime,
like with KotlinBase.

 #KT-37225 Fixed.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -1006,7 +1006,7 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
             }
 
             val classInfo = codegen.kotlinObjCClassInfo(irClass)
-            val classPointerGlobal = load(structGep(classInfo, 12 /* createdClass */))
+            val classPointerGlobal = load(structGep(classInfo, 11 /* createdClass */))
```
knebekaizer: Consider named const

SvyatoslavScherbina: Done

PR #4008, opened 2020-03-23
Add test for static framework

SvyatoslavScherbina:


build-tools/src/main/kotlin/org/jetbrains/kotlin/FrameworkTest.kt
```diff
@@ -71,11 +82,11 @@ open class FrameworkTest : DefaultTask(), KonanTestExecutable {
 
     private fun buildTestExecutable() {
         val frameworkParentDirPath = "$testOutput/$testName/${project.testTarget.name}"
-        frameworkNames.forEach { frameworkName ->
+        frameworkArtifactNames.forEach { frameworkName ->
             val frameworkPath = "$frameworkParentDirPath/$frameworkName.framework"
             val frameworkBinaryPath = "$frameworkPath/$frameworkName"
             validateBitcodeEmbedding(frameworkBinaryPath)
-            codesign(project, frameworkPath)
+            if (codesign) codesign(project, frameworkPath)
```
PavelPunegov: Does the static framework come as the static library that doesn't require code signing?

SvyatoslavScherbina: I'm actually not sure about this. I disabled code signing because it fails to sign the code that is already signed when running the test for the second time without clean.

But I guess code signing should not be required, because static framework code simply gets included into the (Swift) executable, so code signing here shouldn't be responsibility of the framework.

WDYT?

PavelPunegov: If it's included then it should be sigined as a whole swift executable, so it seems to be ok. Lets merge then and wait for its nightly runs.

PR #4009, opened 2020-03-23
Fix #4002

### LINKED ISSUE #4002: ObjCExport.mm:332: runtime assert:
Hello!

After upgrading to 1.3.70, our iOS app crashes with this error

/Users/teamcity/buildAgent/work/4d622a065c544371/runtime/src/main/cpp/ObjCExport.mm:332: runtime assert: 

Tried digging around but no luck. ☹️

SvyatoslavScherbina:
Fix ObjCExport initialization order when static initializer triggers
early KotlinBase initialization.
For example, _swift_eager_class_initialization can do this.


SvyatoslavScherbina:
Test actually requires #4008

PR #4017, opened 2020-03-24
[Interop][Linker] Skip declarations from cache dependencies

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -209,7 +209,17 @@ internal val psiToIrPhase = konanUnitPhase(
                     moduleDescriptor, symbolTable,
                     config.configuration.languageVersionSettings
             )
-            val irProviderForCEnumsAndCStructs = IrProviderForCEnumAndCStructStubs(generatorContext, interopBuiltIns, symbols)
+
+            val generationMode = if (llvmModuleSpecification.isFinal) {
+                IrProviderForCEnumAndCStructStubs.GenerationMode.WholeWorld
```
SvyatoslavScherbina: Is this true? Final module can consume caches.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -245,9 +255,6 @@ internal val psiToIrPhase = konanUnitPhase(
             if (this.stdlibModule in modulesWithoutDCE) {
                 functionIrClassFactory.buildAllClasses()
             }
-            modulesWithoutDCE
-                    .filter(ModuleDescriptor::isFromInteropLibrary)
-                    .forEach(irProviderForCEnumsAndCStructs::buildAllEnumsAndStructsFrom)
```
SvyatoslavScherbina: But the cache should include all code that could be generated. How do you achieve this now?

sbogolepov: It's a duplication which appeared after merge or rebase some time ago. See https://github.com/JetBrains/kotlin-native/blob/sbogolepov/yet-another-interop-fixes/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt#L225

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
sbogolepov: Currently, yes. IIUC, `LlvmModuleSpecificationImpl.containsDeclaration` will return false for declarations from cached libraries, so it works correctly.
We can generate lazy IR for declarations from cached libraries, but it will be an optimization that should be implemented separately (this PR is bug fix).

SvyatoslavScherbina: > We can generate lazy IR for declarations from cached libraries

If this code generate non-lazy IR for cached libraries, how does it avoid placing it to current module? If it doesn't, then it can end up with same declaration in multiple object files.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -209,7 +209,15 @@ internal val psiToIrPhase = konanUnitPhase(
                     moduleDescriptor, symbolTable,
                     config.configuration.languageVersionSettings
             )
-            val irProviderForCEnumsAndCStructs = IrProviderForCEnumAndCStructStubs(generatorContext, interopBuiltIns, symbols)
+
+            val moduleFilter = if (llvmModuleSpecification.isFinal) {
+                llvmModuleSpecification::containsModule
+            } else {
+                modulesWithoutDCE::contains
```
SvyatoslavScherbina: Isn't it actually the same as `llvmModuleSpecification::containsModule`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/interop/IrProviderForCEnumAndCStructStubs.kt
```diff
@@ -67,12 +72,13 @@ internal class IrProviderForCEnumAndCStructStubs(
...
                 error("Module has already been set")
             field = value
-            value.files += filesMap.values
+            value.files += filesMap.filterKeys { moduleFilter(it.module) }.values
         }
 
     fun canHandleSymbol(symbol: IrSymbol): Boolean {
         if (!symbol.isPublicApi) return false
         if (symbol.signature.run { !IdSignature.Flags.IS_NATIVE_INTEROP_LIBRARY.test() }) return false
+        if (!moduleFilter(symbol.descriptor.module)) return false
```
SvyatoslavScherbina: Do I understand correctly that this is an optimization to use lazy IR instead of generated IR for declarations from cached interop libraries?

sbogolepov: Yes

SvyatoslavScherbina: AFAIK, lazy IR is harmful, because it potentially generates unbound symbols after psi2ir.
Is this consideration applicable here?

SvyatoslavScherbina: Discussed offline. Should be ok.

SvyatoslavScherbina: However it makes sense to add a brief comment mentioning that this line makes IR machinery fall back to lazy IR for this declaration.

PR #4018, opened 2020-03-24
Add performance tests on singletons

projedi:
This also adds "manual" benchmarks that warmup and measure time themselves. This mechanism is useful for benchmarks that can only be run once per process lifetime. In this PR they're used for singleton initialization benchmarks: you cannot initialize singleton more than once.


performance/ring/src/main/kotlin/org/jetbrains/ring/SingletonInitBenchmark.kt
```diff
@@ -0,0 +1,18117 @@
+/*
```
projedi: This file is actually generated by https://github.com/google/googletest/blob/master/googlemock/scripts/pump.py

performance/ring/src/main/kotlin/main.kt
```diff
@@ -174,6 +174,9 @@ class RingLauncher : Launcher() {
                     "ParameterNotNull.invokeEightArgsWithoutNullCheck" to BenchmarkEntryWithInit.create(::ParameterNotNullAssertionBenchmark, { invokeEightArgsWithoutNullCheck() }),
                     "PrimeList.calcDirect" to BenchmarkEntryWithInit.create(::PrimeListBenchmark, { calcDirect() }),
                     "PrimeList.calcEratosthenes" to BenchmarkEntryWithInit.create(::PrimeListBenchmark, { calcEratosthenes() }),
+                    "Singleton.access" to BenchmarkEntryWithInit.create(::SingletonBenchmark, { access() }),
```
LepilkinaElena: I think, in this case it would be better to create a separate benchmarks set.

performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -60,6 +67,72 @@ abstract class Launcher {
...
+                                       numberOfAttempts,
+                                       name,
+                                       recordMeasurement,
+                                       null,
+                                       benchmark)
+            }
+            is BenchmarkEntryManual -> {
+                logger.log("Running manual benchmark $name")
+                val result = benchmark.lambda()
+                for ((i, durationNs) in result.durationsNs.withIndex()) {
```
LepilkinaElena: Are you sure that this new type still is needed? Only to get time/500?

performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
```diff
@@ -73,36 +146,28 @@ abstract class Launcher {
...
-                val scaledTime = time * 1.0 / autoEvaluatedNumberOfMeasureIteration
-                samples[k] = scaledTime
-                // Save benchmark object
-                benchmarkResults.add(BenchmarkResult("$prefix$name", BenchmarkResult.Status.PASSED,
-                        scaledTime / 1000, BenchmarkResult.Metric.EXECUTION_TIME, scaledTime / 1000,
-                        k + 1, numWarmIterations))
+            try {
+                runBenchmark(logger, numWarmIterations, numberOfAttempts, name, recordMeasurement, benchmark)
+            } catch (e: Throwable) {
+                printStderr("Failure while running benchmark $name: $e\n")
```
LepilkinaElena: It's also to add here BenchmarkResult with FAILED status.

performance/startup/src/main/kotlin-jvm/org/jetbrains/startup/Utils.kt
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.startup
+
+actual class Random actual constructor() {
```
LepilkinaElena: May be it's better to replace this class to `shared` folder

performance/startup/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+import org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType
+import org.jetbrains.kotlin.RunKotlinNativeTask
+
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+plugins {
+    id("benchmarking")
```
LepilkinaElena: There is also jvmRun in this plugin. It seems that now for this benchmark running process for jvm should fail with error on second run. It isn't critical, I think, it can be done later in separate review.

performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
projedi: I need these benchmarks not to run this code https://github.com/JetBrains/kotlin-native/pull/4018#diff-cb51810e983e4cc6a7fbd6f9e4e653b1R80 On the other hand, it can be solved by introducing a new runtime flag for the launcher to disable automatic iteration count calculation. What do you think?

performance/shared/src/main/kotlin/org/jetbrains/benchmarksLauncher/launcher.kt
projedi: My thought is that having an exception here is a fatal error just like `error("No matching benchmarks found")` above. I added a `try` block just to put `printStderr` in there to aid debugging if these happen.

LepilkinaElena: I think it's right to add here catching exceptions because we have one binary for several benchmarks and it's worth to run other even one fails. It's missed when it was written and after we didn't return. So I think that right way is to report failure and don't block running other benchmarks.

PR #4019, opened 2020-03-24
Store initializing singletons in a separate thread local small map

projedi:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2025,25 +2029,32 @@ OBJ_GETTER(initSharedInstance,
...
   while ((value = __sync_val_compare_and_swap(location, nullptr, initializing)) == initializing);
   if (value != nullptr) {
     // OK'ish, inited by someone else.
     RETURN_OBJ(value);
   }
   ObjHeader* object = AllocInstance(typeInfo, OBJ_RESULT);
-  UpdateHeapRef(localLocation, object);
+  auto insertIt = memoryState->initializingSingletons.insert({location, object});
+  RuntimeCheck(insertIt.second, "object cannot be assigned twice into initializingSingletons");
+  addHeapRef(object);
```
SvyatoslavScherbina: Is this actually required?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -451,6 +451,10 @@ struct MemoryState {
   uint64_t allocSinceLastGcThreshold;
 #endif // USE_GC
 
+  // This map is expected to be small, std::map consumes less memory than
+  // std::unordered_map and is just as efficient.
+  KStdOrderedMap<ObjHeader**, ObjHeader*> initializingSingletons;
```
SvyatoslavScherbina: Does using stack-like collection here has any effect on the performance?

backend.native/tests/runtime/basic/initializers6.kt
```diff
@@ -0,0 +1,75 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package runtime.basic.initializers6
+
+import kotlin.test.*
+
+import kotlin.native.concurrent.*
+
+val aWorkerId = 2
```
SvyatoslavScherbina: Not sure if this is always correct.

backend.native/tests/runtime/basic/initializers7.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package runtime.basic.initializers7
+
+import kotlin.test.*
+
+import kotlin.random.Random
+
+object A {
+    val a1 = Random.nextInt(100)
```
SvyatoslavScherbina: Using `Random` makes the test non-deterministic.

backend.native/tests/runtime/basic/initializers6.kt
```diff
@@ -0,0 +1,75 @@
...
+object A {
+    init {
+        // Must be called by aWorker only.
+        assertEquals(aWorkerId, Worker.current.id)
+        // Only allow b workers to run, when a worker has started initialization.
+        bWorkerUnlocker.increment()
+        // Only proceed with initialization, when all b workers have started executing.
+        while (aWorkerUnlocker.value < bWorkersCount) {}
+        // And now wait a bit, to increase probability of races.
+        Worker.current.park(1000 * 1000L)
```
SvyatoslavScherbina: 100 ms would be enough :)

runtime/src/main/cpp/Memory.cpp
projedi: I'm not sure, but what's protecting this reference (and all that it references) from being garbage collected while (potentially long) initialization is taking place?

backend.native/tests/runtime/basic/initializers7.kt
projedi: You mean if `a1` and `a2` become the same number and the test would fail to catch a potential bug? Fair enough, is there another way to dynamically initialize `a1` while ensuring compiler won't ever optimize it away?

SvyatoslavScherbina: I see.
Well, IIRC, the compiler doesn't optimize the dynamic initialization even if the initializer is literal.
This can likely be verified by more assertions (checking that values are zero in initial state).

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: The allocation [above](https://github.com/JetBrains/kotlin-native/pull/4019/files#diff-c7d26c58627c81c35c99320f067b2752R2046) actually creates a stack reference. It is alive at least until the return from the function.

runtime/src/main/cpp/Memory.cpp
projedi: It does, actually. It outperforms `std::map` on benchmarks from https://github.com/JetBrains/kotlin-native/pull/4018 And compared to the current implementation: it's 5% faster on a single initialization, and 10% slower on a nested initialization (with depth 500). I think, it's a win overall, since deeply nested initialization should be very rare.

Issue #3956, opened 2020-03-08
Kotlin 1.3.60 and 1.3.70: e: java.lang.IllegalStateException: org.jetbrains.kotlin.ir.descriptors.WrappedClassDescriptor


soywiz:
Reproduced:

This branch:
* https://github.com/korlibs/korge-hello-world/tree/bug/WrappedClassDescriptor
```
./gradlew runNativeDebug
```

```
e: Compilation failed: org.jetbrains.kotlin.ir.descriptors.WrappedClassDescriptor@564e1424

 * Source files:
 * Compiler version info: Konan: 1.3.70 / Kotlin: 1.3.70
 * Output kind: STATIC_CACHE

e: java.lang.IllegalStateException: org.jetbrains.kotlin.ir.descriptors.WrappedClassDescriptor@564e1424
        at org.jetbrains.kotlin.backend.konan.llvm.LlvmDeclarations.forClass(LlvmDeclarations.kt:47)
        at org.jetbrains.kotlin.backend.konan.llvm.ContextUtils$DefaultImpls.getTypeInfoPtr(ContextUtils.kt:192)
        at org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator.getTypeInfoPtr(CodeGenerator.kt:44)
        at org.jetbrains.kotlin.backend.konan.llvm.ContextUtils$DefaultImpls.getLlvmTypeInfoPtr(ContextUtils.kt:201)
        at org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator.getLlvmTypeInfoPtr(CodeGenerator.kt:44)
        at org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator.typeInfoValue(CodeGenerator.kt:56)
        at org.jetbrains.kotlin.backend.konan.llvm.IntrinsicGenerator.emitGetClassTypeInfo(IntrinsicGenerator.kt:428)
        at org.jetbrains.kotlin.backend.konan.llvm.IntrinsicGenerator.evaluateCall(IntrinsicGenerator.kt:235)
        at org.jetbrains.kotlin.backend.konan.llvm.IntrinsicGenerator.evaluateCall(IntrinsicGenerator.kt:193)
        at org.jetbrains.kotlin.backend.konan.llvm.CodeGeneratorVisitor.evaluateFunctionCall(IrToBitcode.kt:2132)
[...]

PR #4030, opened 2020-03-26
Local classes in inline bodies

homuroll:
Among other things, this fixes https://github.com/JetBrains/kotlin-native/issues/3956.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
@@ -3,6 +3,9 @@ package org.jetbrains.kotlin.backend.konan
 import org.jetbrains.kotlin.backend.common.*
 import org.jetbrains.kotlin.backend.common.lower.*
 import org.jetbrains.kotlin.backend.common.lower.inline.FunctionInlining
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesExtractionFromInlineFunctionsLowering
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesInInlineFunctionsLowering
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesInInlineLambdasLowering
 import org.jetbrains.kotlin.backend.common.lower.loops.ForLoopsLowering
 import org.jetbrains.kotlin.backend.common.phaser.*
 import org.jetbrains.kotlin.backend.konan.lower.*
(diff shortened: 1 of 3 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
```diff
@@ -86,20 +89,57 @@ internal val arrayConstructorPhase = makeKonanModuleLoweringPhase(
...
+        },
+        name = "Lateinit",
+        description = "Lateinit properties lowering"
+)
+
+internal val sharedVariablesPhase = makeKonanModuleLoweringPhase(
+        ::SharedVariablesLowering,
+        name = "SharedVariables",
+        description = "Shared variable lowering",
+        prerequisite = setOf(lateinitPhase)
```
SvyatoslavScherbina: Do I understand correctly that this dependency exists because of local `lateinit` variables? If shared variables are lowered first, then `LateinitUsageLowering` can't properly detect reads from "shared" `lateinit` variables, right?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
```diff
@@ -86,20 +89,57 @@ internal val arrayConstructorPhase = makeKonanModuleLoweringPhase(
...
+                }
+                LocalClassesExtractionFromInlineFunctionsLowering(context).run {
+                    input.files.forEach { lower(it) }
+                }
+                return input
+            }
+        },
+        name = "ExtractLocalClassesFromInlineBodies",
+        description = "Extraction of local classes from inline bodies",
+        prerequisite = setOf(sharedVariablesPhase),
```
SvyatoslavScherbina: Is this caused by "shared" variables captured by local classes lifted from inline lambdas?

homuroll: Yep.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
```diff
@@ -86,20 +89,57 @@ internal val arrayConstructorPhase = makeKonanModuleLoweringPhase(
...
+internal val extractLocalClassesFromInlineBodies = namedIrModulePhase(
+        lower = object : SameTypeCompilerPhase<Context, IrModuleFragment> {
+            override fun invoke(phaseConfig: PhaseConfig, phaserState: PhaserState<IrModuleFragment>, context: Context, input: IrModuleFragment): IrModuleFragment {
+                LocalClassesInInlineLambdasLowering(context).run {
+                    input.files.forEach { lower(it) }
+                }
+                LocalClassesInInlineFunctionsLowering(context).run {
+                    input.files.forEach { lower(it) }
+                }
+                LocalClassesExtractionFromInlineFunctionsLowering(context).run {
```
SvyatoslavScherbina: What exactly does it do?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
homuroll: Yeah, I have also considered an alternative - save `lateinit` modifier for shared variables, but it's worse, because we need to initialize them somewhere or replace simple code like `shared.element = value` with `shared = Ref(value)`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
homuroll: For local classes in inline functions not lambdas we split the lifting process onto two steps just like in the usual local declarations lowering: the first phase searches for captured variables and adds them to functions and constructors; the second phase just pulls up all local classes to an appropriate level.

PR #4031, opened 2020-03-26
Fix #3952

### LINKED ISSUE #3952: [Reverse C Interop] Unrelated packages with the same name collapsed into one
If you have two packages with the same leaf name, e.g. `knlibrary.sync` and `knlibrary.nested.sync`, they end up getting combined into one in the generated C header. Was able to reproduce on 1.3.61 and 1.3.70.

sync.kt:
```
package knlibrary.sync

class PlainSync {}
```

nested.sync.kt:
```
package knlibrary.nested.sync

class NestedSync {}
```

test.cpp:
```
#include <cstdio>
#define __attribute__(x) // MSVC doesn't support this
#include "knlibrary_api.h"

int main()
{
    knlibrary_ExportedSymbols* lib = knlibrary_symbols();
    auto knlibrary = lib->kotlin.root.knlibrary;
    
    // So far so good, NestedSync should be in knlibrary.nested.sync
    auto good = knlibrary.nested.sync.NestedSync.NestedSync();

    // Unexpected! PlainSync should be in knlibrary.sync
    auto bad = knlibrary.nested.sync.PlainSync.PlainSync();

    return 0;
}
```

build.cmd:
```
@echo off

call kotlinc-native -g sync.kt nested.sync.kt -produce dynamic -o knlibrary.dll

lib.exe /def:knlibrary.def /out:knlibrary.lib /machine:x64
cl.exe test.cpp knlibrary.lib
```

SvyatoslavScherbina:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -649,7 +649,7 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
     override fun visitPackageFragmentDescriptor(descriptor: PackageFragmentDescriptor, ignored: Void?): Boolean {
         val fqName = descriptor.fqName
         val name = if (fqName.isRoot) "root" else translateName(fqName.shortName().asString())
```
vvlevchenko, line 651: may be it'd be better to move name lower in lambda for `getOrPut`?

SvyatoslavScherbina: Done.

PR #4033, opened 2020-03-26
assertFailsWith with cause

ilya-g:
This is a native part of implementation of https://youtrack.jetbrains.com/issue/KT-37804 and https://youtrack.jetbrains.com/issue/KT-23514.
Common part is reviewed in https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4005

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,9 +20,9 @@ buildKotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds
 remoteRoot=konan_tests
 kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-dev-5111,branch:default:any,pinned:true/artifacts/content/maven
 kotlinVersion=1.4.0-dev-5111
-kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-dev-5111,branch:default:any,pinned:true/artifacts/content/maven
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -20,9 +20,9 @@ buildKotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds
 remoteRoot=konan_tests
 kotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-dev-4662,branch:default:any,pinned:true/artifacts/content/maven
 kotlinVersion=1.4.0-dev-4662
-kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-dev-4662,branch:default:any,pinned:true/artifacts/content/maven
-kotlinStdlibVersion=1.4.0-dev-4662
-kotlinStdlibTestsVersion=1.4.0-dev-4662
+kotlinStdlibRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-pub-2421,branch:default:any/artifacts/content/maven
+kotlinStdlibVersion=1.4.0-pub-2421
+kotlinStdlibTestsVersion=1.4.0-pub-2421
 testKotlinCompilerVersion=1.4.0-dev-4662
```
PavelPunegov: This should be also updated to the same version as compiler. This is a place where the  codegen/box test we get from.

vvlevchenko: why -pub-, but not -dev- build?

PR #4036, opened 2020-03-27
Update deprecations before 1.4

SvyatoslavScherbina:


runtime/src/main/kotlin/kotlin/native/concurrent/Worker.kt
```diff
@@ -54,6 +54,7 @@ public inline class Worker @PublishedApi internal constructor(val id: Int) {
          *
          * @param pointer value returned earlier by [Worker.asCPointer]
          */
+        @Deprecated("Use kotlinx.cinterop.StableRef instead", level = DeprecationLevel.WARNING)
```
olonho: Worked is an inline class over an Int, so intention of this API was to pass integer worker id as `COpaquePointer` without boxing. Stable ref to the box (which would be implicitly created)  is different operation, so I'd suggest either keep this operation as is, or avoid suggestion.

SvyatoslavScherbina: > Worked is an inline class over an Int

Exactly. And this blocks us from making `.name` work for terminated workers too, as I mentioned in the commit message.

> so intention of this API was to pass integer worker id as COpaquePointer without boxing

I'm not sure this operation is actually required.

> Stable ref to the box (which would be implicitly created) is different operation

Sure, but it should be enough for real use cases.

> keep this operation

Not an option, since this operation is harmful, at least for the reason described above.

> or avoid suggestion.

Not an option, we should provide an alternative for deprecated operations.

olonho: This will not change behavior of `name` in any sense, and as described in https://jetbrains.slack.com/archives/C0NKC62TV/p1584618031103200?thread_ts=1584613014.099300&cid=C0NKC62TV changing `name` behavior requires to change overall approach to the `Worker` API (make it object oriented and holding, not resource oriented). Worker API is modeled  after POSIX `pthread_t` and file/sockets descriptors, changing `*CPointer` API won't affect this design.

SvyatoslavScherbina: > This will not change behavior of name in any sense

Indeed. But AFAIK, this API is the only user-visible thing blocking from making `Worker` an object. 

> changing name behavior requires to change overall approach to the Worker API (make it object oriented and holding, not resource oriented). 

I don't think so. Making `Worker` an object (holding `id` and e.g. `name`) and maintaining a map fro `id` to `Worker` internally should fix the problem.

SvyatoslavScherbina: (Not sure if the map from `id` to `Worker` is required, it may be enough to have a worker-local variable for this).

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/StableRef.kt
```diff
@@ -16,7 +16,7 @@
 
 package kotlinx.cinterop
 
-@Deprecated("Use StableRef<T> instead", ReplaceWith("StableRef<T>"))
+@Deprecated("Use StableRef<T> instead", ReplaceWith("StableRef<T>"), DeprecationLevel.ERROR)
 typealias StableObjPtr = StableRef<*>
```
ilya-g, line 20: I'd rename StableRef to StableReference if there's a possibility

Dominaezzz: How about `StrongReference`? (Or `GlobalReference` as the JNI named it). I'm not sure what "Stable" in `StableRef` means.

SvyatoslavScherbina: Renaming is possible, but this is definitely out-of-scope for this PR.
Also it is not clear whether renaming makes sense here without redesigning.

PR #4037, opened 2020-03-27
Fix testObjCExport*

SvyatoslavScherbina:
Apparently compiler was using different copy of NativePlatformAnalyzerServices
until recently, so it didn't include new frontend checkers for `@Throws`


SvyatoslavScherbina:
See #4032 removing wrong copy of `NativePlatformAnalyzerServices`.

PR #4038, opened 2020-03-27
Fix ios_x64 platform libs cache for iOS 11 and 12

SvyatoslavScherbina:
Fixes https://youtrack.jetbrains.com/issue/KT-38045

### DIFF
--- a/platformLibs/src/platform/ios/CFNetwork.def
+++ b/platformLibs/src/platform/ios/CFNetwork.def
@@ -5,3 +5,7 @@ modules = CFNetwork
 
 compilerOpts = -framework CFNetwork
 linkerOpts = -framework CFNetwork
+
+excludedFunctions = kCFHTTPAuthenticationSchemeKerberos \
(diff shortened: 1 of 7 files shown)


benasher44:
Is there a workaround that can be applied locally for this one? We've upgraded locally, and now we find we can't test on iOS 12. Thanks!

SvyatoslavScherbina:
> Is there a workaround that can be applied locally for this one? We've upgraded locally, and now we find we can't test on iOS 12. Thanks!

Sure! Please disable compiler caches using Gradle project property `kotlin.native.cacheKind=none`.
Should be fixed in 1.4.

benasher44:
Thanks!

SvyatoslavScherbina:
Fixes https://youtrack.jetbrains.com/issue/KT-38045

SvyatoslavScherbina:
The fix is also backported to 1.3.7x branch: https://github.com/JetBrains/kotlin-native/pull/4071

PR #4043, opened 2020-03-31
[Interop][Linker] Skip declarations from cache dependencies v2

sbogolepov:
A bit more conservative version of https://github.com/JetBrains/kotlin-native/pull/4017 that doesn't use Lazy IR.


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/interop/IrProviderForCEnumAndCStructStubs.kt
```diff
@@ -60,15 +66,12 @@ internal class IrProviderForCEnumAndCStructStubs(
...
-                error("Module has already been set")
-            field = value
-            value.files += filesMap.values
-        }
+    /**
+     * The final output of this provider is a list of files which contain IR declarations
+     * of enums and structs that should be generated.
+     */
+    val outputFiles: List<IrFile>
+        get() = filesMap.filterKeys { moduleFilter(it.module) }.values.toList()
```
SvyatoslavScherbina: Why not filter these on call site?

sbogolepov: `moduleFilter` will have more uses inside `IrProviderForCEnumAndCStructStubs` later.

PR #4044, opened 2020-03-31
Implement WorkerBoundReference

projedi:


runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import kotlin.native.internal.*
+
+@SymbolName("Kotlin_SharedRef_createSharedRef")
+external private fun createSharedRef(value: Any): NativePtr
+
+/**
+ * A frozen shared reference to a Kotlin object
+ *
+ * Can be safely passed between workers, but can only be accessed on the worker it was created on
```
SvyatoslavScherbina, line 16: Unless the referred object is frozen too.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Note: Garbage collector currently cannot detect reference cycles between [SharedRef]s. Consider
```
SvyatoslavScherbina, line 18: Not exactly between `SharedRef`s, but any reference cycle which has at least one `SharedRef` involved.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+@Frozen
+@NoReorderFields
+@ExportTypeInfo("theSharedRefTypeInfo")
+public class SharedRef<out T : Any> private constructor(private var ptr: NativePtr) {
+
+    companion object {
+        /**
+         * Creates shared reference to given object
+         */
+        fun <T : Any> create(value: T) = SharedRef<T>(createSharedRef(value))
```
SvyatoslavScherbina, line 30: `SharedRef` should provide a constructor instead of this factory.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Returns the object this reference was [created][SharedRef.create] for.
+     */
+    @SymbolName("Kotlin_SharedRef_derefSharedRef")
+    external fun get(): T
```
SvyatoslavScherbina, line 37: Note that we have an option to make this a property (`.value`) with custom getter. This would be consistent with other classes in K/N (consider `AtomicReference`). But I'm not sure that property getter should throw an exception for illegal access.
A piece of advice from @ilya-g would be extremely useful here.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    companion object {
+        /**
+         * Creates shared reference to given object
+         */
+        fun <T : Any> create(value: T) = SharedRef<T>(createSharedRef(value))
+    }
+
+    /**
+     * Returns the object this reference was [created][SharedRef.create] for.
```
SvyatoslavScherbina, line 34: Should explicitly mention that it can fail with `IncorrectDereferenceException`.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun dispose() {
+        ref.value = null
+    }
+
+    /**
+     * Returns the object this reference was [created][DisposableSharedRef.create] for.
+     * @throws NullPointerException if [DisposableSharedRef.dispose] was called on this reference.
+     */
+    fun get(): T {
+        return ref.value?.get() ?: throw NullPointerException()
```
SvyatoslavScherbina, line 76: `IllegalStateException` maybe?

SvyatoslavScherbina: The exception should have proper message in any case.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -931,6 +931,9 @@ void freeAggregatingFrozenContainer(ContainerHeader* container) {
 void runDeallocationHooks(ContainerHeader* container) {
   ObjHeader* obj = reinterpret_cast<ObjHeader*>(container + 1);
   for (int index = 0; index < container->objectCount(); index++) {
+    if (obj->type_info() == theSharedRefTypeInfo) {
```
SvyatoslavScherbina, line 934: Does this have any effect on the performance for other classes?

runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -8,6 +8,19 @@
 #include "MemorySharedRefs.hpp"
 #include "Runtime.h"
 
+namespace {
```
SvyatoslavScherbina, line 11: Code supporting `SharedRef` doesn't belong to this file: it is a high-level user of `MemorySharedRefs.hpp`.
I believe it should be moved to a separate file then, e.g. `SharedRef.cpp`.

backend.native/tests/runtime/concurrent/shared0.kt
```diff
@@ -0,0 +1,301 @@
...
+    workerUnlocker.increment()
+
+    for (future in futures) {
+        future.result
+    }
+
+    for (worker in workers) {
+        worker.requestTermination().result
+    }
+}
```
SvyatoslavScherbina, line 301: No test for accessing `SharedRef` holding frozen object.
Object can be frozen both before and after `SharedRef` is created.

backend.native/tests/runtime/concurrent/shared1.kt
```diff
@@ -0,0 +1,508 @@
...
+    assertNull(localValueWeak.get())
+}
+
+@Test fun testDisposeAndCollect() {
+    val (localRef, localWeak, localValueWeak) = getWeaksAndAtomicReference(3)
+
+    localRef.value!!.dispose()
+    GC.collect()
+
+    assertNotNull(localWeak.get())
```
SvyatoslavScherbina, line 193: This is fragile. Maybe at least create explicit local `val` for `localRef.value!!`?

backend.native/tests/runtime/concurrent/shared1.kt
```diff
@@ -0,0 +1,508 @@
...
+    val semaphore: AtomicInt = AtomicInt(0)
+
+    val worker = Worker.start()
+
+    val (localWeak, localValueWeak, future) = disposeInWorker(worker, semaphore)
+    semaphore.increment()
+
+    future.result
+    GC.collect()
+    assertNotNull(localWeak.get())
```
SvyatoslavScherbina, line 284: Why is it so?

backend.native/tests/runtime/concurrent/shared1.kt
```diff
@@ -0,0 +1,508 @@
...
+    GC.collect()
+
+    // If these asserts fail, that means DisposableSharedRef managed to clean up cyclic garbage all by itself.
+    assertNotNull(weakB1.get())
+    assertNotNull(weakB2.get())
+}
+
+fun callDispose(ref: AtomicReference<DisposableSharedRef<B1>?>) {
+    ref.value!!.dispose()
+    ref.value = null
```
SvyatoslavScherbina, line 357: `ref.value = null` looks unnecessary and even harmful here.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
raniejade: Curious, what's the purpose/use-case for a `SharedRef`?

SvyatoslavScherbina: The purpose is to share reference without sharing contents. In particular, to have a reference from shared object to non-shared object. For example, when some methods of the shared object can be used from any execution context, while others must be called only from the particular worker and rely on objects of this worker.

backend.native/tests/runtime/concurrent/shared1.kt
projedi: Added a comment that hopefully explains the reasoning. However, variable naming is unfortunate, I'll try to improve it.

backend.native/tests/runtime/concurrent/shared1.kt
projedi: Left a comment explaining the reasoning, but will work on naming.

SvyatoslavScherbina: I still don't get it.
Is `localWeak.get()` a `DisposableSharedRef`? What exactly prevents it from removing by `GC.collect()` above?

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Note: Garbage collector currently cannot free any reference cycles with [SharedRef] or [DisposableSharedRef] in them.
+ * Consider using [DisposableSharedRef] to manually resolve cycles with explicit calls to [DisposableSharedRef.dispose]
+ */
+@Frozen
+@NoReorderFields
+@ExportTypeInfo("theSharedRefTypeInfo")
+public class SharedRef<out T : Any>(value: T) {
+
+    private var ptr = createSharedRef(value)
```
SvyatoslavScherbina, line 27: Why not `val`?

backend.native/tests/runtime/concurrent/shared0.kt
```diff
@@ -0,0 +1,429 @@
...
+        }
+        result
+    }
+
+    val value = future.result
+    assertEquals(0, value)
+    worker.requestTermination().result
+}
+
+val global3: SharedRef<A> = SharedRef(A(3).apply { freeze() })
```
SvyatoslavScherbina, line 50: > `A(3).apply { freeze() }`

`A(3).freeze()` should be the same.

backend.native/tests/runtime/concurrent/shared0.kt
```diff
@@ -0,0 +1,429 @@
...
+val global2: SharedRef<A> = SharedRef(A(3))
+
+@Test fun testGlobalDenyAccessOnWorker() {
+    assertEquals(3, global2.get().a)
+
+    val worker = Worker.start()
+    val future = worker.execute(TransferMode.SAFE, {}) {
+        var result = 0
+        assertFailsWith<IncorrectDereferenceException> {
+            result = global2.get().a
```
SvyatoslavScherbina, line 40: This actually checks that either `global2`, `.get()` or `.a` throws an exception.
It would make more sense to test that `.get()` throws.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,69 @@
...
+ * Can be safely passed between workers, but can only be accessed on the worker it was created on,
+ * unless the referred object is frozen too
+ *
+ * Note: Garbage collector currently cannot free any reference cycles with [SharedRef] or [DisposableSharedRef] in them.
+ * Consider using [DisposableSharedRef] to manually resolve cycles with explicit calls to [DisposableSharedRef.dispose]
+ */
+@Frozen
+@NoReorderFields
+@ExportTypeInfo("theSharedRefTypeInfo")
+public class SharedRef<out T : Any>(value: T) {
```
ilya-g, line 25: `...Ref` is inconsistent with other `...Reference` types, like `AtomicReference`, `WeakReference`

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,69 @@
...
+ * cycles with [SharedRef] or [DisposableSharedRef], consider using [SharedRef]
+ */
+@Frozen
+public class DisposableSharedRef<out T : Any>(value: T) {
+
+    private val ref: AtomicReference<SharedRef<T>?> = AtomicReference(SharedRef(value))
+
+    /**
+     * Free the reference. Any call to [DisposableSharedRef.get] after that will
+     * fail with [NullPointerException]
```
ilya-g, line 55: `with [NullPointerException]` — docs of `get` doesn't mention it.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlin.native.internal.*
+
+@SymbolName("Kotlin_SharedRef_createSharedRef")
+external private fun createSharedRef(value: Any): NativePtr
+
+/**
+ * A frozen shared reference to a Kotlin object
+ *
+ * Can be safely passed between workers, but can only be accessed on the worker it was created on,
+ * unless the referred object is frozen too
```
ilya-g, line 17: How is it supposed to be used from workers with this contract? Is there a real-world usage sample somewhere?

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
ilya-g: It meets two of three criteria when property should be preferred: https://kotlinlang.org/docs/reference/coding-conventions.html#functions-vs-properties

Sometimes, the first one ("Does not throw") can be dismissed, for example `Lazy.value` is a property that can throw whatever the lazy initializer throws. Whether or not it is acceptable depends on the use case.

runtime/src/main/cpp/Memory.cpp
projedi: Good thing you asked. I didn't think it would have a measurable impact since it runs during deallocation, it's a simple compare and jump, and `obj->type_info()` is accessed below regardless.

And then I ran `ring` benchmarks:
```
Performance details
=================
Benchmark                                         First score              Second score             Percent                  Ratio                    
------------------------------------------------------------------------------------------------------------------------------------------------------
Geometric mean                                    221.4273 ± 0.8246        220.2083 ± 0.8864        0.00 ± 0.00 %            1.01 ± 0.01              
------------------------------------------------------------------------------------------------------------------------------------------------------
Ring::Euler.problem9                              852.9136 ± 1.6820        574.3120 ± 1.8881        32.31 ± 0.44 %           1.49 ± 0.01              
Ring::Calls.interfaceMethodBimorphic              566.7952 ± 5.0681        394.2084 ± 0.8376        29.67 ± 1.25 %           1.44 ± 0.02              
Ring::Calls.openMethodTrimorphic                  1982.0209 ± 6.1022       1481.8918 ± 10.8697      24.45 ± 1.10 %           1.34 ± 0.01              
Ring::IntArray.countFilteredSomeLocal             22.6207 ± 0.0427         17.5005 ± 0.0401         22.31 ± 0.36 %           1.29 ± 0.01              
Ring::CompanionObject.invokeRegularFunction       0.0056 ± 0.0000          0.0046 ± 0.0000          16.12 ± 0.52 %           1.20 ± 0.01              
Ring::IntArray.countFilteredSomeManual            20.4322 ± 0.0561         17.5310 ± 0.0482         13.73 ± 0.47 %           1.17 ± 0.01              
Ring::String.stringBuilderConcat                  287.8454 ± 1.9528        255.7878 ± 2.8703        9.53 ± 2.01 %            1.13 ± 0.02              
Ring::String.stringBuilderConcatNullable          288.5585 ± 2.8659        256.7720 ± 2.3562        9.30 ± 1.78 %            1.12 ± 0.02              
Ring::ParameterNotNull.invokeTwoArgsWithNullCheck 0.0090 ± 0.0000          0.0081 ± 0.0000          9.08 ± 0.65 %            1.11 ± 0.01              
Ring::Euler.problem2                              6.8504 ± 0.0900          6.1866 ± 0.0220          8.16 ± 2.38 %            1.11 ± 0.02              
Ring::Euler.problem8                              339.8889 ± 1.4641        309.6184 ± 2.6217        7.74 ± 1.55 %            1.10 ± 0.01              
Ring::DefaultArgument.testFourOfFour              0.0051 ± 0.0000          0.0047 ± 0.0000          7.23 ± 1.63 %            1.09 ± 0.01              
Ring::ClassArray.countFilteredLocal               2497.6665 ± 20.3156      2293.5391 ± 6.3643       7.16 ± 1.50 %            1.09 ± 0.01              
Ring::Calls.interfaceMethodHexamorphic            2455.9505 ± 17.4846      2276.1167 ± 5.1275       6.45 ± 1.32 %            1.08 ± 0.01              
Ring::IntList.filterAndCount                      2617.9088 ± 10.9396      2481.8600 ± 17.6619      4.12 ± 1.35 %            1.05 ± 0.01              
Ring::ForLoops.arrayLoop                          7.5842 ± 0.0450          7.1792 ± 0.0610          3.96 ± 1.62 %            1.06 ± 0.02              
Ring::IntList.filterManual                        2595.6617 ± 2.9685       2488.8430 ± 4.4040       3.84 ± 0.34 %            1.04 ± 0.00              
Ring::OctoTest                                    3392.8933 ± 24.0985      3235.5541 ± 7.6086       3.73 ± 1.36 %            1.05 ± 0.01              
Ring::AbstractMethod.sortStringsWithComparator    227.7961 ± 0.9786        220.8034 ± 0.9333        2.24 ± 0.84 %            1.03 ± 0.01              
Ring::IntArray.copy                               1724.3959 ± 2.9719       1690.3470 ± 4.5611       1.54 ± 0.53 %            1.02 ± 0.00              
Ring::Calls.openMethodBimorphic                   394.3895 ± 1.7671        561.6845 ± 0.8763        -29.36 ± 0.63 %          0.70 ± 0.00              
Ring::ForLoops.charArrayLoop                      3.3796 ± 0.0171          4.6797 ± 0.0165          -27.16 ± 0.73 %          0.72 ± 0.01              
Ring::Calls.interfaceMethodTrimorphic             1498.6024 ± 4.4771       1981.0271 ± 5.7760       -23.90 ± 0.45 %          0.76 ± 0.00              
Ring::ParameterNotNull.invokeOneArgWithoutNullCheck0.0077 ± 0.0000          0.0099 ± 0.0000          -21.84 ± 0.59 %          0.78 ± 0.01              
Ring::Elvis.testElvis                             72.2298 ± 0.0781         85.2601 ± 0.2739         -14.92 ± 0.54 %          0.85 ± 0.00              
Ring::LocalObjects.localArray                     0.0856 ± 0.0006          0.1016 ± 0.0010          -14.35 ± 1.68 %          0.84 ± 0.01              
Ring::IntArray.countFilteredSome                  17.5423 ± 0.0493         20.3742 ± 0.0242         -13.55 ± 0.48 %          0.86 ± 0.00              
Ring::ClassStream.filterAndMapManual              2528.4038 ± 2.9074       2856.4001 ± 20.7141      -10.73 ± 1.29 %          0.89 ± 0.01              
Ring::IntBaseline.allocateArrayAndFill            26.7258 ± 0.1895         29.6828 ± 0.2427         -8.58 ± 1.48 %           0.90 ± 0.01              
Ring::WithIndicies.withIndiciesManual             2449.8847 ± 3.1263       2685.0674 ± 18.4686      -8.01 ± 1.26 %           0.91 ± 0.01              
Ring::ClassBaseline.consumeField                  58.4672 ± 0.2731         63.9413 ± 0.5942         -7.27 ± 1.71 %           0.91 ± 0.01              
Ring::Switch.testConstSwitch                      187.1804 ± 0.5415        202.5431 ± 0.6270        -7.03 ± 0.57 %           0.92 ± 0.01              
Ring::ParameterNotNull.invokeTwoArgsWithoutNullCheck0.0077 ± 0.0000          0.0082 ± 0.0001          -5.00 ± 2.14 %           0.94 ± 0.01              
Ring::Loop.arrayLoop                              66.5391 ± 0.5565         70.6445 ± 0.4824         -4.37 ± 1.59 %           0.94 ± 0.01              
Ring::Loop.arrayListLoop                          155.6977 ± 1.3134        164.9411 ± 0.9808        -4.24 ± 1.60 %           0.94 ± 0.01              
Ring::Loop.arrayIndexLoop                         96.0024 ± 0.4738         102.1985 ± 1.4875        -4.20 ± 2.75 %           0.94 ± 0.02              
Ring::Loop.arrayForeachLoop                       66.8288 ± 0.5850         70.4282 ± 0.4796         -3.62 ± 1.67 %           0.95 ± 0.01              
Ring::Loop.arrayListForeachLoop                   156.5447 ± 0.8745        164.1294 ± 1.3592        -3.29 ± 1.59 %           0.95 ± 0.01              
Ring::Switch.testVarSwitch                        190.1685 ± 0.5312        195.6584 ± 0.7369        -2.17 ± 0.73 %           0.97 ± 0.01              
Ring.compileTime                                  20548585.3580 ± 0.0000   20841423.6600 ± 0.0000   -1.41 ± 0.00 %           0.99 ± 0.00 
```
Some huge regressions and some huge improvements. This is all on my local machine, of course, but I reran the suite several times, and the results are stable. I'll be looking into it.

projedi: This is caused by `runDeallocationHooks` stopping to be automatically inlined and `DisposeSharedRef` being automatically inlined and also contributing with exception handlers around it. Put appropriate annotations to restore previous behavior. It's worth investigating whether we need to put more of `noexcept` annotations around and then revisiting this manual control of inlining.

backend.native/tests/runtime/concurrent/shared1.kt
projedi: Right, fixed it by holding onto `AtomicReference` of `DisposableSharedRef` on the main thread

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
SvyatoslavScherbina: For example, `(Disposable)SharedRef` is a replacement for `StableRef`, when the latter is (mis)used to share references to non-shared objects. See
https://github.com/Kotlin/kotlinx.coroutines/blob/b2ecddc0cc18dd46cb2e3513841177f2174ca113/kotlinx-coroutines-core/native/src/internal/Sharing.kt#L151

ilya-g: This doesn't make it more clear. 
I've noticed I'm not the first one who didn't get it, and you've already explained it once: https://github.com/JetBrains/kotlin-native/pull/4044#discussion_r402245428
Perhaps, it should be explained in the docs of the class.

ilya-g: IMO, the name `SharedReference` doesn't convey the concept behind it:
> The purpose is to share reference without sharing contents.

Take the following analogy: when I share a link to some document, everyone who has that link can access that document — this is what can be called a shared reference. That's not the case here, because only one worker can actually access the referenced object.

backend.native/tests/runtime/concurrent/shared0.kt
```diff
@@ -0,0 +1,454 @@
...
+val global2: SharedReference<A> = SharedReference(A(3))
+
+@Test
+fun testGlobalDenyAccessOnWorker() {
+    assertEquals(3, global2.value.a)
+
+    val worker = Worker.start()
+    val future = worker.execute(TransferMode.SAFE, {}) {
+        assertFailsWith<IncorrectDereferenceException> {
+            global2.value
```
SvyatoslavScherbina, line 41: Accessing to top-level fields in Kotlin/Native can fail with `IncorrectDereferenceException` too. So it makes a lot of sense to move `global2` out of this `assertFailsWith` too.

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
SvyatoslavScherbina: > when I share a link to some document, everyone who has that link can access that document — this is what can be called a shared reference. 

The new reference implemented in this PR is more like a link to a private document which you can save to some public document: if anybody but you follows this link, he/she gets "no access exception".

Do you have better naming suggestion? Maybe @elizarov has some ideas?

runtime/src/main/kotlin/kotlin/native/concurrent/SharedReference.kt
```diff
@@ -0,0 +1,71 @@
...
+@Frozen
+@NoReorderFields
+@ExportTypeInfo("theSharedReferenceTypeInfo")
+public class SharedReference<out T : Any>(value: T) {
+
+    private val ptr = createSharedRef(value)
+
+    /**
+     * The referenced value.
+     * @throws IncorrectDereferenceException if referred object is not frozen and this is called from a different worker, than [SharedReference] was created on
```
SvyatoslavScherbina, line 34: If the access fails with exception, it is extremely useful to report some additional information, for example the name of the worker owning this reference and the name of current worker. (Information about the object should still be included).

Supporting this ad hoc for `SharedReference` (instead of `KRefSharedHolder`) is ok for now.
Note that `Worker.name` doesn't work for terminated workers, so the name of the owner should probably be stored to the reference when it is created.

SvyatoslavScherbina: (Making exceptions more verbose here can also be postponed to next PR).

runtime/src/main/kotlin/kotlin/native/concurrent/SharedRef.kt
elizarov: Possible names:
* `Moniker`
* `IdentityReference`
* `OpaqueReference`
* `ForeignReference`
* `ShieldedReference`
* `InsulatedReference`
* `SafeReference`

ilya-g: - `ExclusiveReference`
- `WorkerBoundReference`

elizarov: More:
* `WorkerConfinedReference`
* `WorkerLocalReference`

I like the names that start with `Worker` as they better highlight cases when these classes should be used (that is, when you use workers) and it would be even better if/when they expose `val worker: Worker` property.

ilya-g: `WorkerConfinedReference` seems fine to me.

SvyatoslavScherbina: Problem is: this name doesn't reflect the fact that this reference itself is shareable.

projedi: I'll settle on `WorkerBoundReference`, if that's alright.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    private val ptr = createWorkerBoundReference(value)
+
+    /**
+     * The referenced value.
+     * @throws IncorrectDereferenceException if referred object is not frozen and this is called from a different worker, than [WorkerBoundReference] was created on
+     */
+    val value: T
+        get() = @Suppress("UNCHECKED_CAST") (derefWorkerBoundReference(ptr) as T)
+}
```
elizarov, line 38: Please, consider adding `val worker: Worker` property.

SvyatoslavScherbina: We've considered this and decided to postpone.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -0,0 +1,71 @@
...
+@Frozen
+@NoReorderFields
+@ExportTypeInfo("theWorkerBoundReferenceTypeInfo")
+public class WorkerBoundReference<out T : Any>(value: T) {
+
+    private val ptr = createWorkerBoundReference(value)
+
+    /**
+     * The referenced value.
+     * @throws IncorrectDereferenceException if referred object is not frozen and this is called from a different worker, than [WorkerBoundReference] was created on
```
SvyatoslavScherbina, line 34: `if referred object is not frozen and current worker is different from the one created [this]`
?

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -0,0 +1,71 @@
...
+@SymbolName("Kotlin_WorkerBoundReference_create")
+external private fun createWorkerBoundReference(value: Any): NativePtr
+
+@SymbolName("Kotlin_WorkerBoundReference_deref")
+external private fun derefWorkerBoundReference(ref: NativePtr): Any
+
+/**
+ * A frozen shared reference to a Kotlin object
+ *
+ * Can be safely passed between workers, but can only be accessed on the worker it was created on,
```
SvyatoslavScherbina, line 19: > but can only be accessed on the worker

"It accessed" is ambiguous. Maybe explicitly mention `[value]` or "referred object", or use "dereference" instead of "access"?

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -0,0 +1,71 @@
...
+@ExportTypeInfo("theWorkerBoundReferenceTypeInfo")
+public class WorkerBoundReference<out T : Any>(value: T) {
+
+    private val ptr = createWorkerBoundReference(value)
+
+    /**
+     * The referenced value.
+     * @throws IncorrectDereferenceException if referred object is not frozen and this is called from a different worker, than [WorkerBoundReference] was created on
+     */
+    val value: T
```
ilya-g, line 36: Do we need something like `valueOrNull` which doesn't throw in other workers, but returns null instead?
cc @elizarov

elizarov: You can take a look at my use-cases here: https://github.com/Kotlin/kotlinx.coroutines/blob/native-mt/kotlinx-coroutines-core/native/src/internal/Sharing.kt#L199

As you can see, the most important additional thing I need is a reference to the `worker`.

elizarov: However, `valueOrNull` would be occasionally useful, too. See usages of this function, for example: https://github.com/Kotlin/kotlinx.coroutines/blob/native-mt/kotlinx-coroutines-core/native/src/internal/Sharing.kt#L159

runtime/src/main/kotlin/kotlin/native/concurrent/SharedReference.kt
projedi: Implemented in https://github.com/JetBrains/kotlin-native/pull/4115

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: Implemented in https://github.com/JetBrains/kotlin-native/pull/4115

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: Implemented in https://github.com/JetBrains/kotlin-native/pull/4135

PR #4045, opened 2020-03-31
[runtime][source info][mac] more precision source info fetching

vvlevchenko:


runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -92,28 +106,38 @@ bool TryInitializeCoreSymbolication() {
...
+              && address < range.location + range.length) {
+            RuntimeAssert(result.fileName == nullptr, "shouldn't be more than one value per address");
+            const char* fileName = CSSourceInfoGetPath(ref);
+            if (fileName != nullptr) {
+              result.fileName = fileName;
+              result.lineNumber = lineNumber;
+              result.column = CSSourceInfoGetColumn(ref);
+            }
+       }
+       return 1;
```
SvyatoslavScherbina, line 138: What does `1` mean here?

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -92,28 +106,38 @@ bool TryInitializeCoreSymbolication() {
...
+      return result;
+
+    CSSymbolForeachSourceInfo(symbol,
+      ^(CSSourceInfoRef ref) {
+          uint32_t lineNumber = CSSourceInfoGetLineNumber(ref);
+          CSRange range = CSSourceInfoGetRange(ref);
+          if (lineNumber != 0
+              && address >= range.location
+              && address < range.location + range.length) {
+            RuntimeAssert(result.fileName == nullptr, "shouldn't be more than one value per address");
```
SvyatoslavScherbina, line 130: What does guarantee this?
When `CSSymbolRef` has multiple `CSSourceInfoRef`s, what do they look like?

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: I haven't found any examples or documentation how to use it properly, so just guessed value.

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: I've seen several SCSourceInfoRefs per CSSymbolRef on inlines. Current source info selection based on range, but we I haven't got enough information about CoreSymbolisation if it enough.

runtime/src/debug/cpp/SourceInfo.cpp
SvyatoslavScherbina: Googling this function reveals two usages, both returning `0`
`¯\_(ツ)_/¯`

runtime/src/debug/cpp/SourceInfo.cpp
SvyatoslavScherbina: Can we rely on the order of `SCSourceInfoRef`s here?

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: ok

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: Looks so, but it also iterates source info of closing scopes too.

SvyatoslavScherbina: I don't get you.
Can we always use first or last `CSSourceInfoRef`?

vvlevchenko: no, more likely [length/2]

SvyatoslavScherbina: I see.
Let's stop reverse engineering for now then :)
Current heuristic is probably ok (although I have no idea what it means), but let's remove the assertion (since we don't now if the heuristic checks are enough).

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -64,9 +64,18 @@ uint32_t (*CSSourceInfoGetLineNumber)(CSSourceInfoRef info);
 
 uint32_t (*CSSourceInfoGetColumn)(CSSourceInfoRef info);
 
+void* (*CSSymbolOwnerGetBaseAddress)(CSSymbolOwnerRef info);
```
SvyatoslavScherbina, line 67: Is it used?

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: ok

runtime/src/debug/cpp/SourceInfo.cpp
vvlevchenko: no, I've used it for  debugging

PR #4049, opened 2020-04-01
[IR] Run lowerings before inlining on the entire IR

homuroll:
Since inlining is a cross-module operation (we need to know the bodies of
inline functions), all lowerings that are supposed to run before the inliner,
need to be run on the entire IR (granted, they can be run only on the inline functions,
but, for simplicity, just run them on the entire IR)

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
@@ -317,17 +317,23 @@ internal val linkerPhase = konanUnitPhase(
         description = "Linker"
 )
 
-internal val allLoweringsPhase = namedIrModulePhase(
-        name = "IrLowering",
-        description = "IR Lowering",
+internal val preInlinePhase = namedIrModulePhase(
+        name = "IrPreInlineLowering",
+        description = "IR lowerings before inlining",
         lower = removeExpectDeclarationsPhase then
                 stripTypeAliasDeclarationsPhase then
                 lowerBeforeInlinePhase then
                 arrayConstructorPhase then
                 lateinitPhase then
                 sharedVariablesPhase then
-                extractLocalClassesFromInlineBodies then
-                inlinePhase then
+                extractLocalClassesFromInlineBodies,
+        actions = setOf(defaultDumper, ::moduleValidationCallback)
+)
+
+internal val allLoweringsPhase = namedIrModulePhase(
+        name = "IrLowering",
+        description = "IR Lowering",
+        lower = inlinePhase then
                 provisionalFunctionExpressionPhase then
                 lowerAfterInlinePhase then
                 performByIrFile(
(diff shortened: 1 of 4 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -317,17 +317,23 @@ internal val linkerPhase = konanUnitPhase(
         description = "Linker"
 )
 
-internal val allLoweringsPhase = namedIrModulePhase(
-        name = "IrLowering",
-        description = "IR Lowering",
+internal val preInlinePhase = namedIrModulePhase(
+        name = "IrPreInlineLowering",
+        description = "IR lowerings before inlining",
```
SvyatoslavScherbina, line 322: Did we have similar issues with other lowerings from the sequence below?

backend.native/tests/build.gradle
```diff
@@ -3301,6 +3301,12 @@ linkTest("inline_defaultArgs_linkTest") {
     lib = "codegen/inline/defaultArgs_linkTest_lib.kt"
 }
 
+linkTest("inline_sharedVar_linkTest") {
```
SvyatoslavScherbina, line 3304: Does this test fail on JS? If not, how does JS backend handle this? Using permanent IR approach?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: Correct me if I wrong but the inline lowering is the only lowering which requires function bodies from other modules, other only require headers and for headers there are some maps in `context` to not be dependent on the order of modules.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -450,6 +485,8 @@ private val backendCodegen = namedUnitPhase(
         name = "Backend codegen",
         description = "Backend code generation",
         lower = takeFromContext<Context, Unit, IrModuleFragment> { it.irModule!! } then
+                preInlinePhase then
+                dependenciesPreInlinePhase then
```
SvyatoslavScherbina, line 489: Does this approach work with caches? When `inline` function comes from cached library.

backend.native/tests/build.gradle
homuroll: They run all lowerings on the entire world.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: I'll check.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
SvyatoslavScherbina: I mean that `lateinitPhase`, for example, is probably not very different from `sharedVariablesPhase`.
What happens if  `inline` function from other module has `lateinit var`?

backend.native/tests/build.gradle
SvyatoslavScherbina: What about incremental compilation?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: I don't get it. We run all lowerings that are run before the inline phase for entire IR, `lateinitPhase` included.

SvyatoslavScherbina: I mean _before your fix_.

backend.native/tests/build.gradle
homuroll: For IC it is ensured that before a particular lowering is about to run all previous lowerings have been run.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: Ah, I see, you're implying to try to find similar bugs and add tests on'em?

PR #4054, opened 2020-04-02
Report unsupported fields of Companion of ObjC subclass

knebekaizer:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/KotlinObjCClassInfoGenerator.kt
```diff
@@ -27,10 +27,17 @@ internal class KotlinObjCClassInfoGenerator(override val context: Context) : Con
...
-        val companionObject = irClass.declarations.filterIsInstance<IrClass>().atMostOne { it.isCompanion  }
+        val companionObject = irClass.companionObject()
         val classMethods = companionObject?.generateMethodDescs().orEmpty()
 
-        val superclassName = irClass.getSuperClassNotAny()!!.let {
+        val superClass = irClass.getSuperClassNotAny()!!
+        if (superClass.isObjCClass() && companionObject != null) {
+            val companionBodySize = LLVMStoreSizeOfType(llvmTargetData, context.llvmDeclarations.forClass(companionObject).bodyType).toInt()
+            if (companionBodySize > LLVMStoreSizeOfType(llvmTargetData, int8TypePtr).toInt())
+                context.reportCompilationError("Fields are not supported for Companion of subclass of ObjC type")
```
SvyatoslavScherbina, line 37: We have bunch of compilation errors reported from `InteropLowering.kt`.
I guess this check should be moved there too.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/KotlinObjCClassInfoGenerator.kt
```diff
@@ -27,10 +27,17 @@ internal class KotlinObjCClassInfoGenerator(override val context: Context) : Con
...
         val instanceMethods = generateInstanceMethodDescs(irClass)
 
-        val companionObject = irClass.declarations.filterIsInstance<IrClass>().atMostOne { it.isCompanion  }
+        val companionObject = irClass.companionObject()
         val classMethods = companionObject?.generateMethodDescs().orEmpty()
 
-        val superclassName = irClass.getSuperClassNotAny()!!.let {
+        val superClass = irClass.getSuperClassNotAny()!!
+        if (superClass.isObjCClass() && companionObject != null) {
+            val companionBodySize = LLVMStoreSizeOfType(llvmTargetData, context.llvmDeclarations.forClass(companionObject).bodyType).toInt()
```
SvyatoslavScherbina, line 35: It would be more straightforward and reliable to check presence of `IrField`s in the class instead.

knebekaizer: I tried both approaches, they have the same effect of course. My choice was based on the same "straightforward" rationale: I want to check for non-trivial body which we don't initialize correctly, and size (i.e. non-initialized memory slots) is exactly what differentiate it from trivial one. Also, checking the size is simpler, it works for backing fields, etc. Saying that, I would agree that testing against pointer size is not exactly "trivial body test" and may be fragile. So I have no clear preferences.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -482,6 +491,16 @@ private class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfor
             )
         }
 
+        irClass.companionObject()?.let {
+            if (it.hasFields() ||
+                    it.getSuperClassNotAny()?.hasFields() ?: false) {
```
SvyatoslavScherbina, line 496: IIRC, super classes can't have fields here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -482,6 +491,16 @@ private class InteropLoweringPart1(val context: Context) : BaseInteropIrTransfor
...
+                context.reportCompilationError(
+                        "Fields are not supported for Companion of subclass of ObjC type",
+                        currentFile, irClass
```
SvyatoslavScherbina, line 499: Reporting the location of particular field (property?) would significantly improve user experience.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/KotlinObjCClassInfoGenerator.kt
SvyatoslavScherbina: Thank you for the explanation.
We have plans to migrate to frontend all these compilation errors reported by backend. So adding this check to InteropLowering will simplify the migration for the check.

Checking this on IR level instead of "body-size" level makes it more straightforward to report location of particular field instead of class (see https://github.com/JetBrains/kotlin-native/pull/4054#discussion_r402954213).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
knebekaizer: My testcase:
```
open class Dummy {
    val Z = 42
}
class mySss:NSData() {
    companion object : Dummy() {
        fun textLength() = Z
    }
    fun myComp() = Companion
}
```
passed the first check and failed at runtime.

SvyatoslavScherbina: I see.
I presumed that `companion object` should have super types that are meta-types for super types of the containing class (e.g. `NSDataMeta`), but we don't have checks for this either.
Having other super types in such companions is not supported correctly too.
Instance of the companion in your testcase don't conform to `Dummy` on runtime.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
knebekaizer: Let's consider it as follow-up enhancement

PR #4056, opened 2020-04-03
[StubIR][Metadata] fix for ObjCBlockVar

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -560,4 +565,24 @@ private class MappingExtensions(
...
     private val TypeParameterStub.id: Int
         get() = typeParametersInterner.intern(this)
+}
+
+private enum class TypeAliasExpansionMode {
+    /**
+     * Substitution of type alias with fully expanded type.
+     * Typical usage: expanded type of typealias.
+     */
+    ERASE,
```
SvyatoslavScherbina, line 575: Do I understand correctly that this is the only new mode?
`NONE` is equal to `shouldExpandTypeAliases = false`, while `EXPANSION` is equal to `shouldExpandTypeAliases  = true`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -451,39 +451,44 @@ private class MappingExtensions(
...
-                km.arguments += typeArguments
-            }
-            if (shouldExpandTypeAliases) {
-                // Abbreviated and expanded types have the same nullability.
-                KmType(flags).also { km ->
-                    km.abbreviatedType = abbreviatedType
-                    val kmUnderlyingType = underlyingType.map(true)
-                    km.arguments += kmUnderlyingType.arguments
-                    km.classifier = kmUnderlyingType.classifier
+    fun StubType.map(mode: TypeAliasExpansionMode = TypeAliasExpansionMode.EXPANSION): KmType = when (this) {
```
SvyatoslavScherbina, line 454: Seems correct from my point of view, but I'm not an expert in metadata.
Have you consulted anyone who is?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: Yep.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: No, good idea.

PR #4066, opened 2020-04-06
[IR] Run lowerings before the inliner on all inline functions

homuroll:
Since inlining is a cross-module operation (we need to know the bodies of
inline functions), all lowerings that are supposed to run before the inliner,
need to be run on all inline functions.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
@@ -66,6 +66,8 @@ internal class SpecialDeclarationsFactory(val context: Context) {
     private val loweredEnums = mutableMapOf<IrClass, LoweredEnum>()
     private val ordinals = mutableMapOf<ClassDescriptor, Map<ClassDescriptor, Int>>()
 
+    val loweredInlineFunctions = mutableSetOf<IrFunction>()
+
     object DECLARATION_ORIGIN_FIELD_FOR_OUTER_THIS :
             IrDeclarationOriginImpl("FIELD_FOR_OUTER_THIS")
 
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
@@ -25,11 +25,7 @@ import org.jetbrains.kotlin.descriptors.konan.isNativeStdlib
 import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
 import org.jetbrains.kotlin.ir.symbols.IrSymbol
-import org.jetbrains.kotlin.ir.util.DeclarationStubGenerator
-import org.jetbrains.kotlin.ir.util.SymbolTable
-import org.jetbrains.kotlin.ir.util.addChild
-import org.jetbrains.kotlin.ir.util.addFile
-import org.jetbrains.kotlin.ir.util.file
+import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.acceptVoid
 import org.jetbrains.kotlin.konan.target.CompilerOutputKind
 import org.jetbrains.kotlin.psi2ir.Psi2IrConfiguration
@@ -320,6 +316,7 @@ internal val linkerPhase = konanUnitPhase(
 internal val allLoweringsPhase = namedIrModulePhase(
         name = "IrLowering",
         description = "IR Lowering",
+        // TODO: The lowerings before inlinePhase should be aligned with [NativeInlineFunctionResolver.kt]
         lower = removeExpectDeclarationsPhase then
                 stripTypeAliasDeclarationsPhase then
                 lowerBeforeInlinePhase then
@@ -392,6 +389,7 @@ internal val dependenciesLowerPhase = SameTypeNamedPhaseWrapper(
                                     ?: return@forEach
                             input.files += libModule.files
                         }
+
                 input.files += files
 
                 return input
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.kotlin.backend.konan.lower
+
+import org.jetbrains.kotlin.backend.common.DeclarationTransformer
+import org.jetbrains.kotlin.backend.common.lower.*
+import org.jetbrains.kotlin.backend.common.lower.inline.DefaultInlineFunctionResolver
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesExtractionFromInlineFunctionsLowering
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesInInlineFunctionsLowering
+import org.jetbrains.kotlin.backend.common.lower.inline.LocalClassesInInlineLambdasLowering
+import org.jetbrains.kotlin.backend.common.runPostfix
+import org.jetbrains.kotlin.backend.konan.Context
+import org.jetbrains.kotlin.ir.declarations.IrFunction
+import org.jetbrains.kotlin.ir.symbols.IrFunctionSymbol
+import org.jetbrains.kotlin.ir.util.dump
+
+// TODO: This is a bit hacky. Think about adopting persistent IR ideas.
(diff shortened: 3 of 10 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+                    break
+                prevParent = parent as IrDeclaration
+            }
+
+            val originalFile = prevParent.parent as IrFile
+            val irFile = context.irFilesForInlineFunctions.getOrPut(originalFile) {
+                IrFileImpl(
+                        originalFile.fileEntry,
+                        EmptyPackageFragmentDescriptor(context.moduleDescriptor,
+                                FqName.topLevel(Name.special("<inline_functions_from_cached_libraries>")))
```
SvyatoslavScherbina, line 50: Maybe reuse the same package fq name?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+            }
+            irFile.addChild(prevParent)
+        }
+
+        PreInlineLowering(context).lower(body, function)
+
+        ArrayConstructorLowering(context).lower(body, function)
+
+        NullableFieldsForLateinitCreationLowering(context)
+                .runPostfix(true).transformFlat(function)
```
SvyatoslavScherbina, line 61: Why does it ignore `transformFlat` result?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+        NullableFieldsDeclarationLowering(context)
+                .runPostfix(true).transformFlat(function)
+        LateinitUsageLowering(context).lower(body, function)
+
+        SharedVariablesLowering(context).lower(body, function)
+
+        LocalClassesInInlineLambdasLowering(context).lower(body, function)
+        LocalClassesInInlineFunctionsLowering(context).lower(body, function)
+        LocalClassesExtractionFromInlineFunctionsLowering(context).lower(body, function)
```
SvyatoslavScherbina, line 70: Can we reuse pipeline from TopLevelPhases.kt?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
homuroll: For simplicity - the lowering doesn't change functions.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
homuroll: I suppose.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
homuroll: How? They are supposed to be applied to either `IrModule` or `IrFile`. We can probably create a number of additional phases to be applied to `IrFunction`, but it will still be a copy/paste, so don't see a point in it. And by the way it is a hack, we should later consider using PIR for that, so may be it is not that bad.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -384,6 +380,14 @@ internal val dependenciesLowerPhase = SameTypeNamedPhaseWrapper(
                             input.files.clear()
                         }
 
+                // Lower all referenced inline functions from cached libraries.
+                if (context.irFilesForInlineFunctions.isNotEmpty()) {
+                    input.files += context.irFilesForInlineFunctions.values
+                    allLoweringsPhase.invoke(phaseConfig, phaserState, context, input)
```
SvyatoslavScherbina, line 386: What happens if it produces more `irFilesForInlineFunctions` or adds declarations to existing ones?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
SvyatoslavScherbina: Then we should check this fact explicitly here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
SvyatoslavScherbina: We should ensure this hack remains synchronized with the main pipeline. Alternative solution would be to add some comments to both copies.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -47,6 +47,7 @@ import org.jetbrains.kotlin.backend.common.ir.copyTo
 import org.jetbrains.kotlin.backend.common.ir.copyToWithoutSuperTypes
 import org.jetbrains.kotlin.backend.konan.objcexport.ObjCExport
 import org.jetbrains.kotlin.backend.konan.llvm.coverage.CoverageManager
+import org.jetbrains.kotlin.ir.declarations.impl.IrFileImpl
```
SvyatoslavScherbina, line 50: Unused.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
```diff
@@ -0,0 +1,58 @@
...
+        context.specialDeclarationsFactory.loweredInlineFunctions.add(function)
+
+        PreInlineLowering(context).lower(body, function)
+
+        ArrayConstructorLowering(context).lower(body, function)
+
+        NullableFieldsForLateinitCreationLowering(context)
+                .runPostfix(true).transformFlat(function)
+        NullableFieldsDeclarationLowering(context)
+                .runPostfix(true).transformFlat(function)
```
SvyatoslavScherbina, line 43: I noticed the difference between `runPostfix(true)` here and `runPostfix(false)` when applying these lowerings in main pipeline.
Apparently we have a bug in the latter case:
```kotlin
fun main() {
    val o = object : Any() {
        lateinit var x: Any
    }

    println(o.x)
}
```
Prints `null` on 1.4-M1 and master, fails with exception on 1.3.71.

SvyatoslavScherbina: Could you please also check other lowerings for such issues?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
SvyatoslavScherbina: (this is still required)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
homuroll: Other possible candidates are inheritors of `DeclarationTransformer`. We use `StripTypeAliasLowering` - for it doesn't matter since type aliases can only be on top level; `NullableFieldsForLateinitCreationLowering` and `NullableFieldsDeclarationLowering` - both need fixing; and the last are `DefaultArgumentStubGenerator` and `DefaultParameterCleaner` - they are run after the local declarations lowering, and therefore it is safe to skip local declarations at that point.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/NativeInlineFunctionResolver.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        PreInlineLowering(context).lower(body, function)
+
+        ArrayConstructorLowering(context).lower(body, function)
+
+        assert(NullableFieldsForLateinitCreationLowering(context)
+                .runPostfix(true).transformFlat(function) == null) {
+            "Unexpected transformation of function ${function.dump()}"
+        }
+        assert(NullableFieldsDeclarationLowering(context)
```
SvyatoslavScherbina, line 38: Assertions can be disabled.

PR #4069, opened 2020-04-07
Support clang-format in runtime

projedi:
Applying (with some manual tweaks and disabling formatting on some blocks) clang-format with this config produces something like this:
https://github.com/JetBrains/kotlin-native/commit/7556973210ec98d13a5439a56711ccb1b16d0c45


.clang-format
```diff
@@ -0,0 +1,110 @@
+---
```
projedi, line 1: Differences from Chromium style:
* Set `AllowAllParametersOfDeclarationOnNextLine: true`
* Set `BinPackParameters: true`
* Removed `ForEachMacros`
* Tweaked `IncludeCategories` and `IncludeIsMainRegex`
* Removed `RawStringFormats`
* Set `Standard: Cpp11`
* Removed `StatementMacros`

projedi, line 1: Documentation on all the options can be found here: https://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html

.clang-format
```diff
@@ -0,0 +1,110 @@
...
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
```
projedi, line 12: With `true`: https://github.com/JetBrains/kotlin-native/commit/a42e18551cd5f84189551519719f60d2ad8dea04
With `Inline` (diff from `true`): https://github.com/JetBrains/kotlin-native/commit/7ed705072a679d5c13278cdf6c05ae1bc10b4805

build.gradle
```diff
@@ -738,3 +739,11 @@ task pusher(type: KotlinBuildPusher){
     overrideConfigurationId = System.getenv("TEAMCITY_OVERRIDE_NATIVE_ID") ?: "Kotlin_KotlinDev_DeployMavenArtifacts_OverrideNative"
     token = project.findProperty("teamcityBearToken") ?: System.getenv("TEAMCITY_BEAR_TOKEN")
 }
+
+task format {
+    dependsOn ":runtime:format"
+}
+
+task formatCheck {
```
projedi, line 747: This should be run during a precommit check as a part of some existing build or as a brand new build (which can be run on any platform, since there isn't anything platform-specific in this style check)

runtime/build.gradle
```diff
@@ -101,3 +103,88 @@ task generateJsMath {
...
+    include '**/*.c'
+    include '**/*.cpp'
+    include '**/*.h'
+    include '**/*.hpp'
+    include '**/*.m'
+    include '**/*.mm'
+
+    exclude 'cpp/utf8.h' // part of external library
+
+    // TODO: Reformat all these files.
```
projedi, line 117: I think it's better to start gradually fixing these files instead of doing it in a single PR

.clang-format
projedi: I like the `Inline` because it basically means that simple functions defined in the header (usually doing simple arithmetics or just forwarding the call) and simple class methods (usually getters and setters) get single-lined, and everything else gets a proper multi-line body.

There's also an option `false` which will make every function definition multi-line.

.clang-format
SvyatoslavScherbina: It makes sense to have two commits for this file: one adding original Chromium style, next one applying all the tweaks.

runtime/build.gradle
projedi: Put `// clang-format off` to every existing file's header.

build.gradle
projedi: @PavelPunegov which way is better in your opinion: make this check a part of an existing build, or create another build?

PavelPunegov: I think it could be done as a part of sources checkout job. 
I see two scenarios here:
1. We could implement some reporting that will be collected and shown. Some kind of warnings.
2. Fail on checkout and fail the whole job chain line. This one enforces usage of format rules, but may add some undesired failures where the code is not formatted yet or under the discussion whether formatter correct or should be changed, and the code itself doesn't build and tested.

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecClangFormat.kt
```diff
@@ -0,0 +1,43 @@
...
+        }.assertNormalExitValue()
+        val diffResult = project.exec {
+            it.commandLine = listOf("diff", "-u", target, "-")
+            it.standardInput = ByteArrayInputStream(outputStream.toByteArray())
+            it.isIgnoreExitValue = true
+        }
+        if (diffResult.exitValue == 0) {
+            return
+        }
+        throw GradleException("$target is not formatted")
```
PavelPunegov, line 41: I'd like to add the diff result as well

build.gradle
projedi: I don't think warnings will do the job: one can easily forget to check them. Is there no simple way to fail a build, while allowing the build chain to continue?
I do agree that failing the entire chain because of formatting is at the very least annoying.

SvyatoslavScherbina: AFAIK, adding this check as additional "Precommit" configuration dependency should do the job.

projedi:
Alternatively I tried to match [Kotlin style](https://kotlinlang.org/docs/reference/coding-conventions.html) as much as possible: https://github.com/JetBrains/kotlin-native/blob/b91442331d95200ef406b1d8e9ab1a0ffdeb063a/.clang-format This is how it looks on the current code: https://github.com/JetBrains/kotlin-native/commit/b83b5da0366ea3c48188965492a056ecd91ba247 (I didn't put any `// clang-format off` around arrays, so ignore them)

projedi:
Was superseded by https://github.com/JetBrains/kotlin-native/pull/4254

PR #4074, opened 2020-04-08
[Tools] Support basic navigation at runtime source in Clion

knebekaizer:


sbogolepov:
cc @SvyatoslavScherbina, @projedi

PR #4082, opened 2020-04-09
Hint for users about disabling caches

LepilkinaElena:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -133,7 +133,13 @@ internal class Linker(val context: Context) {
                 it.execute()
             }
         } catch (e: KonanExternalToolFailure) {
-            context.reportCompilationError("${e.toolName} invocation reported errors\n${e.message}")
+            val extraUserInfo =
+                    if (caches.static.isNotEmpty() && context.config.produce != CompilerOutputKind.STATIC_CACHE)
```
SvyatoslavScherbina, line 137: Technically, if producing static cache, disabling caches would help too.
So I guess it makes sense to remove `context.config.produce != CompilerOutputKind.STATIC_CACHE` check.

Also we could add `|| caches.dynamic.isNotEmpty()`.

PR #4085, opened 2020-04-10
Enable light debug information

LepilkinaElena:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -152,11 +152,13 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var frameworkImportHeaders: Array<String>? = null
 
-    // FIXME: same name.
     @Argument(
-            value = "-Xg0",
+            value = "-Xadd-light-debug",
             valueDescription = "{disable|enable}",
-            description = "Add light debug information"
+            deprecatedName = "-Xg0",
```
SvyatoslavScherbina, line 158: The idea is to avoid using `-Xg0` at all. Otherwise existing projects adding `-Xg0` (without value) will fail to compile.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -152,11 +152,13 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var frameworkImportHeaders: Array<String>? = null
 
-    // FIXME: same name.
     @Argument(
-            value = "-Xg0",
+            value = "-Xadd-light-debug",
```
SvyatoslavScherbina, line 156: Parser of this option mentions `-Xg0` too.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -152,11 +152,13 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
...
             valueDescription = "{disable|enable}",
-            description = "Add light debug information"
+            deprecatedName = "-Xg0",
+            description = "Add light debug information for optimized builds. This option is skipped in debug builds.\n" +
+                    "It's enabled by default on Darwin platforms where collected debug information is stored in .dSYM file.\n" +
+                    "Current option is disabled by default on other platforms."
```
SvyatoslavScherbina, line 161: Currently

PR #4088, opened 2020-04-11
Make static cache symbols hidden

SvyatoslavScherbina:
Enables DCE for cached code when linking dynamic library too

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OptimizationPipeline.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OptimizationPipeline.kt
@@ -5,6 +5,8 @@ import kotlinx.cinterop.memScoped
 import kotlinx.cinterop.ptr
 import kotlinx.cinterop.value
 import llvm.*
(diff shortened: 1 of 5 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/OptimizationPipeline.kt
```diff
@@ -175,6 +177,11 @@ internal fun runLlvmOptimizationPipeline(context: Context) {
             // Since we are in a "closed world" internalization can be safely used
             // to reduce size of a bitcode with global dce.
             LLVMAddInternalizePass(modulePasses, 0)
+        } else if (context.config.produce == CompilerOutputKind.STATIC_CACHE) {
+            // Hidden visibility makes symbols internal when linking the binary.
+            // When producing dynamic library, this enables DCE similar to internalize but later:
```
sbogolepov, line 182: Maybe "..., this enables stripping unused symbols from binary with `-dead_strip` flag"?

SvyatoslavScherbina: Makes sense.

PR #4091, opened 2020-04-13
Adding CompilationDatabase generation

projedi:


build-tools/src/main/kotlin/org/jetbrains/kotlin/CompileToBitcode.kt
```diff
@@ -46,11 +46,45 @@ open class CompileToBitcode @Inject constructor(@InputDirectory val srcRoot: Fil
 
     private val targetDir by lazy { File(project.buildDir, target) }
 
-    private val objDir by lazy { File(targetDir, folderName) }
+    val objDir by lazy { File(targetDir, folderName) }
```
projedi, line 49: I don't like exposing internals here and in `ExecClang`, so that we can scavenge them and assemble a compilation database. But I don't have any better ideas just yet.

build.gradle
```diff
@@ -739,3 +741,10 @@ task pusher(type: KotlinBuildPusher){
     overrideConfigurationId = System.getenv("TEAMCITY_OVERRIDE_NATIVE_ID") ?: "Kotlin_KotlinDev_DeployMavenArtifacts_OverrideNative"
     token = project.findProperty("teamcityBearToken") ?: System.getenv("TEAMCITY_BEAR_TOKEN")
 }
+
+CompilationDatabaseKt.mergeCompilationDatabases(project, "compdb", [
```
projedi, line 745: It might be useful to make this task a depedency of some frequently executed task, so that compilation database would have a chance to regenerate on a regular basis. The job itself is rather inexpensive compared to compilation, so having it as a dependency won't have a noticable impact.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CompilationDatabase.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import org.jetbrains.kotlin.konan.target.HostManager
+
+@Serializable
```
SvyatoslavScherbina, line 24: IIRC, we had strong opinion against using `kotlinx.serialization` in our build scripts.
@PavelPunegov probably knows more.

PavelPunegov: Well, we still use them, but I think it's better to avoid to do that. It's ok for now, as I'm going to rewrite that and have started actually.

projedi: What's the preferred way to deal JSON then?

PavelPunegov: There is a https://github.com/JetBrains/kotlin-native/tree/master/tools/benchmarks/shared/src/main/kotlin/org/jetbrains/report/json
but it requires to be moved into `build-tools`. And that I was doing but haven't finished yet.

PR #4096, opened 2020-04-14
[devirtualization] Removed redundant code with associated objects + test

homuroll:
It turned out that it is not needed to explicitly add all associated objects to instantiating classes
since their constructors have been already added to the root set and all constructors have additional parameter (<this>) and its type will be added to instantiating classes since it is final for all objects.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt
@@ -97,7 +97,9 @@ internal object Devirtualization {
                 moduleDFG.symbolTable.functionMap.values.filter { it.explicitlyExported } +
                 externalModulesDFG.functionDFGs.keys.filter { it.explicitlyExported }
 
-        // Conservatively assume each associated object could be instantiated.
+        // Conservatively assume each associated object could be called.
+        // Note: for constructors there is additional parameter (<this>) and its type will be added
+        // to instantiating classes since all objects are final types.
         val associatedObjects = mutableListOf<DataFlowIR.FunctionSymbol>()
         context.irModule!!.acceptChildrenVoid(object: IrElementVisitorVoid {
             override fun visitElement(element: IrElement) {
@@ -898,6 +900,8 @@ internal object Devirtualization {
                 }
                 if (entryPoint == null) {
                     // For library assume all public non-abstract classes could be instantiated.
+                    // Note: for constructors there is additional parameter (<this>) and for associated objects
+                    // its type will be added to instantiating classes since all objects are final types.
                     symbolTable.classMap.values
                             .filterIsInstance<DataFlowIR.Type.Public>()
                             .filter { !it.isAbstract }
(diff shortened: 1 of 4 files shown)


backend.native/tests/objcexport/values.swift
```diff
@@ -1278,6 +1278,9 @@ func testGH3825() throws {
     try assertEquals(actual: count, expected: 3)
 }
 
+func testKt35940() throws {
```
SvyatoslavScherbina, line 1281: Please move this test to a separate file.
See `backend.native/tests/objcexport/gh4002.*` as an example.

backend.native/tests/objcexport/values.kt
```diff
@@ -1012,3 +1011,30 @@ fun mutULong2Long(): MutableMap<ULong, Long> = mutableMapOf(Pair(0x8000_0000_000
...
+    override fun bar(c: CKt35940) {
+        c.list = mutableListOf("zzz")
+    }
+}
+
+@Associated_Kt35940(IKt35940Impl::class)
+private class IKt35940ImplHolder
+
+@OptIn(ExperimentalAssociatedObjects::class)
+fun testKt35940(): String {
```
SvyatoslavScherbina, line 1035: What exactly does it test?
I tried removing the code for associated objects from `Devirtualization`, but the test still passes with `-opt`.

homuroll: Hmm, it's been copied from the test `associatedObjects1`, may be I've copied wrong part.

homuroll: Weird, if I move the test to a separate file and compile it to a `kexe`, then it fails as expected.

SvyatoslavScherbina: This probably means that the test is not self-contained enough. IIRC, it checks that `ArrayList` is in instantiating classes.

homuroll: Actually, the call site hasn't been devirtualized, I'll dig deeper to find out why.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt
```diff
@@ -97,7 +97,9 @@ internal object Devirtualization {
                 moduleDFG.symbolTable.functionMap.values.filter { it.explicitlyExported } +
                 externalModulesDFG.functionDFGs.keys.filter { it.explicitlyExported }
 
-        // Conservatively assume each associated object could be instantiated.
+        // Conservatively assume each associated object could be called.
+        // Note: for constructors there is additional parameter (<this>) and its type will be added
```
SvyatoslavScherbina, line 101: Looks fragile.
Please add a comment about this expectation to the code that adds final types to instantiating classes.

backend.native/tests/objcexport/values.kt
```diff
@@ -13,8 +13,7 @@ import kotlin.native.concurrent.isFrozen
 import kotlin.native.internal.ObjCErrorException
 import kotlin.native.ref.WeakReference
 import kotlin.properties.ReadWriteProperty
-import kotlin.reflect.KClass
-import kotlin.reflect.KProperty
+import kotlin.reflect.*
```
SvyatoslavScherbina, line 16: Please avoid committing redundant changes in `values.kt` and `values.swift`.

PR #4100, opened 2020-04-14
Rewrite :shared:buildSrc to Java to avoid dependency to kotlin stdlib

goodwinnk:
Composite build hangs on some version combination after start using 1.4.255-SNAPSHOT in
Kotlin build. This is probably caused by kotlin-stdlib substitution in buildSrc build.


goodwinnk:
Merged manually.

PR #4105, opened 2020-04-15
Fix race in cyclic collector

SvyatoslavScherbina:
Prevent cyclic collector from observing newly allocated atomic root with RC = 0.

### DIFF
--- a/runtime/src/main/cpp/Memory.cpp
+++ b/runtime/src/main/cpp/Memory.cpp
@@ -1940,11 +1940,6 @@ OBJ_GETTER(allocInstance, const TypeInfo* type_info) {
 #endif  // USE_GC
   auto container = ObjectContainer(state, type_info);
   ObjHeader* obj = container.GetPlace();
-#if USE_CYCLIC_GC
-  if ((obj->type_info()->flags_ & TF_LEAK_DETECTOR_CANDIDATE) != 0) {
-    cyclicAddAtomicRoot(obj);
-  }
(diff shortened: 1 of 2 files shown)


runtime/src/main/cpp/Memory.cpp
```diff
@@ -1940,18 +1940,21 @@ OBJ_GETTER(allocInstance, const TypeInfo* type_info) {
...
     rememberNewContainer(container.header());
   } else {
     makeShareable(container.header());
   }
 #endif  // USE_GC
+#if USE_CYCLIC_GC
+  if ((obj->type_info()->flags_ & TF_LEAK_DETECTOR_CANDIDATE) != 0) {
+    // Note: this should be performed after [rememberNewContainer] (above).
+    // Otherwise cyclic collector can observe this atomic root with RC = 0,
+    // thus consider it garbage and then zero it after initialization.
```
projedi, line 1954: Is there a way to write a test that triggers this race in a more direct way than `worker_bound_reference1`? Maybe creating a ton of `AtomicReference`s in a loop in several workers while trying to trigger cyclic collection amidst all of it?

projedi: I gave it a shot: https://github.com/JetBrains/kotlin-native/pull/4112/files Unfortunately it doesn't reproduce locally without `usleep` after `cyclicAddAtomicRoot`. Let's see if it's enough for the build agents, and then decide whether the test is useful at all.

PR #4108, opened 2020-04-16
[runtime] Terminate on IncorrectDereferenceException from ObjC weak p…

knebekaizer:
…roperty

Root cause: ObjC runtime gets a os_unfair_lock when accessing weak property.
If exception is thrown from _tryRetainImp this lock left unlocked,
so it will hang on `__ulock_wait` on the next access.
Interim workaround: terminate instead of exception propagating.

### DIFF
--- a/runtime/src/main/cpp/ObjCExportErrors.mm
+++ b/runtime/src/main/cpp/ObjCExportErrors.mm
@@ -46,7 +46,7 @@ extern "C" RUNTIME_NORETURN void Kotlin_ObjCExport_trapOnUndeclaredException(KRe
 static char kotlinExceptionOriginChar;
 
 static bool isExceptionOfType(KRef exception, const TypeInfo** types) {
-  for (int i = 0; types[i] != nullptr; ++i) {
+  if (types) for (int i = 0; types[i] != nullptr; ++i) {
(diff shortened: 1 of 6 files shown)


runtime/src/main/cpp/ObjCExportErrors.mm
```diff
@@ -46,7 +46,7 @@ static void printlnMessage(const char* message) {
 static char kotlinExceptionOriginChar;
 
 static bool isExceptionOfType(KRef exception, const TypeInfo** types) {
-  for (int i = 0; types[i] != nullptr; ++i) {
+  if (types) for (int i = 0; types[i] != nullptr; ++i) {
```
SvyatoslavScherbina, line 49: What is this patch for?
IIRC, `types` is guaranteed to be non-null.

knebekaizer: `types` comes from another module and has never been checked, so it is _supposed_ to be not null, not _guaranteed_. Using `gsl::not_null` would be better solution indeed (let's consider).  The patch arose as a result of my attempt (abandoned now) to implement "optional" behavior (similar to `std::optional`).

knebekaizer:
Reworked as #4126

PR #4109, opened 2020-04-16
[runtime] Fix snprintf usage

knebekaizer:
Correct usage of return value.

### DIFF
--- a/runtime/src/main/cpp/KString.cpp
+++ b/runtime/src/main/cpp/KString.cpp
@@ -919,13 +919,14 @@ KInt Kotlin_StringBuilder_insertInt(KRef builder, KInt position, KInt value) {
   RuntimeAssert(toArray->count_ >= 11 + position, "must be true");
   char cstring[12];
   auto length = konan::snprintf(cstring, sizeof(cstring), "%d", value);
+  RuntimeAssert(length >= 0, "This should never happen"); // may be overkill
+  RuntimeAssert(length < sizeof(cstring), "Unexpectedly large value"); // Can't be, but this is what sNprintf for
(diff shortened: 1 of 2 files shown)


runtime/src/main/cpp/KString.cpp
```diff
@@ -919,13 +919,13 @@ KInt Kotlin_StringBuilder_insertInt(KRef builder, KInt position, KInt value) {
   RuntimeAssert(toArray->count_ >= 11 + position, "must be true");
   char cstring[12];
   auto length = konan::snprintf(cstring, sizeof(cstring), "%d", value);
+  RuntimeAssert(length < sizeof(cstring), "Unexpectedly large value"); // Can't be, but this is what sNprintf for
```
projedi, line 922: Maybe also check `lengh > 0`?

runtime/src/main/cpp/Porting.cpp
```diff
@@ -110,16 +110,19 @@ void consolePrintf(const char* format, ...) {
   char buffer[1024];
   va_list args;
   va_start(args, format);
-  int rv = vsnprintf_impl(buffer, sizeof(buffer) - 1, format, args);
+  int rv = vsnprintf_impl(buffer, sizeof(buffer), format, args);
+  if (rv > sizeof(buffer) - 1) rv = sizeof(buffer) - 1;  // TODO: Consider realloc or report truncating.
```
projedi, line 114: Also needs a check for `rv < 0`. Although, I'm not sure what to do in this case. `RuntimeAssert` itself calls `konan::snprintf`, and if `snprintf` failed because it failed to allocate memory…

PR #4112, opened 2020-04-16
Fix cyclic race repro

projedi:


projedi:
On commit c8a418b [Windows Sanity](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinNative_Development_KotlinNativeWindowsSanity/70712637?) and [Linux Sanity](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinNative_Development_KotlinNativeLinuxSanity/70712633?) did successfully reproduce the problem. On a second try (on 7aa8111) no luck.

So, there's a signal from this test, but it's a weak one. It might not catch the offending patch, we'll have to monitor for flakiness. @SvyatoslavScherbina what do you think?

SvyatoslavScherbina:
Please refer the fix from the commit message (to simplify investigating the test when it fail).

PR #4115, opened 2020-04-17
Improve WorkerBoundReference exceptions and add valueOrNull

projedi:


runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -68,4 +82,10 @@ public class DisposableWorkerBoundReference<out T : Any>(value: T) {
      */
     val value: T
         get() = ref.value?.value ?: throw IllegalStateException("illegal attempt to dereference disposed $this")
+
+    /**
+     * The referenced value or null if [this] is disposed, or if referred object is not frozen and current worker is different from the one created [this]
+     */
+    val valueOrNull: T?
```
projedi, line 89: @elizarov do you need a middleground `valueOrNull` that fails with exception if `this` is disposed, but returns `null` if called on the wrong worker?

elizarov: Yes. It is not that much useful without `worker` reference, though, since I'll have to store a worker reference myself anyway, but it will have me from having to use `currentWorker` and compare myself.

runtime/src/main/kotlin/kotlin/Any.kt
```diff
@@ -41,10 +42,9 @@ public open class Any {
...
     public open fun toString(): String {
         val kClass = this::class
-        val className = kClass.qualifiedName ?: kClass.simpleName ?: "<object>"
         // TODO: consider using [identityHashCode].
-        val unsignedHashCode = this.hashCode().toLong() and 0xffffffffL
-        val hashCodeStr = unsignedHashCode.toString(16)
-        return "$className@$hashCodeStr"
+        val identity = this.hashCode()
+        // Use debug description as the default string representation.
+        return debugDescription(kClass, identity)
```
SvyatoslavScherbina, line 48: `Any.toString()` is not exactly the same (abstraction) as debug description.
I would prefer this patch to be reverted, code duplication is intentional here.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
SvyatoslavScherbina: > do you need a middleground valueOrNull that fails with exception if this is disposed, but returns null if called on the wrong worker?

I'm not sure that current behaviour is required. So this "middleground" should probably be the only variant.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -6,12 +6,16 @@
 package kotlin.native.concurrent
 
 import kotlin.native.internal.*
+import kotlin.native.identityHashCode
```
SvyatoslavScherbina, line 9: Seems unused.

runtime/src/main/cpp/Utils.h
```diff
@@ -53,3 +55,14 @@ class LockGuard {
   LockGuard(const LockGuard&) = delete;
   LockGuard& operator=(const LockGuard&) = delete;
 };
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Returns a string describing object at `address` of type `typeInfo`.
+OBJ_GETTER(DescribeObjectForDebugging, KConstNativePtr typeInfo, KConstNativePtr address);
```
SvyatoslavScherbina, line 64: This file is not the best place for such a high-level function.
AFAIK, there is only one usage of this function in C++ code, so it is ok to move this declaration to the same file.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: >  It is not that much useful without worker reference

I'll be adding `worker` getter in a later PR.

> I'm not sure that current behaviour is required. So this "middleground" should probably be the only variant.

@elizarov what do you think?

runtime/src/main/kotlin/kotlin/Any.kt
projedi: Okay, I just thought that the default `toString()` implementation is exactly the debug description of an object.

SvyatoslavScherbina: Just a coincidence :)

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: There's also an issue with naming. It feels strange to me, if `valueOrNull` starts throwing exceptions. Maybe `valueOrNullIfSet`? @ilya-g what do you think?

ilya-g: First about throwing ISE from this property:
- if using WBR after disposal is expected to be a normal case, then I'd prefer this property not to throw.
- if that's an exceptional case, then it's ok to throw an exception.
Then about naming:
- if that's an exceptional case so that the property should not be accessed in a normal code, then it's ok to leave the name `valueOrNull`
- only if that is a normal case (to access the property after disposal), and you insist on throwing exception in this case, then the property should be named differently.

ilya-g:
Please improve doc comments in WorkerBoundReference.kt: end sentences with a full stop; use third person verbs for describing actions like `dispose`.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
elizarov: The whole `DisposableWBR` does not look very useful to me anyway. It does not need any integration with GC or anything. For convenience and easy-to-troubleshoot exceptions I'll have to write it myself anyway. I have a radical suggestion to drop `DisposableWBR` variant completely which completely avoid the needless design discussion on its behavior.

projedi: Okay, let's remove it then.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -11,61 +11,42 @@ import kotlin.native.internal.*
...
- * Call [dispose] manually to resolve cycles
- *
- * Note: This class has more expensive [value] getter than [WorkerBoundReference]. If you don't have reference
- * cycles with [WorkerBoundReference] or [DisposableWorkerBoundReference], consider using [WorkerBoundReference]
- */
-@Frozen
-public class DisposableWorkerBoundReference<out T : Any>(value: T) {
-
-    private val ref: AtomicReference<WorkerBoundReference<T>?> = AtomicReference(WorkerBoundReference(value))
+        get() = valueOrNull ?: throw IncorrectDereferenceException("illegal attempt to access non-shared $valueDescription bound to `$ownerName` from `${Worker.current.name}`")
```
SvyatoslavScherbina, line 45: The exception message can be tested too.

SvyatoslavScherbina: (Single test would be enough for now).

backend.native/tests/runtime/concurrent/worker_bound_reference0.kt
```diff
@@ -568,3 +568,21 @@ fun concurrentAccess() {
...
+
+@Test
+fun testExceptionMessage() {
+    val worker = Worker.start()
+    val future = worker.execute(TransferMode.SAFE, {}) {
+        WorkerBoundReference(A(3))
+    }
+    val value = future.result
+
+    val valueDescription = A(3).toString()
```
SvyatoslavScherbina, line 580: How can this be equal to `value` description? I guess hashes can differ.

PR #4116, opened 2020-04-17
[runtime][source info][mac] return "inline at" position for stack trace

vvlevchenko:


runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -22,6 +22,17 @@
 #include <limits.h>
 #include <stdint.h>
 #include <unistd.h>
+#include <string.h>
+
+#define TRACE_SYMBOLICATION 1
```
SvyatoslavScherbina, line 27: Should probably be `0`.

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -22,6 +22,17 @@
...
+#if TRACE_SYMBOLICATION
+#include <stdio.h>
+#define SYM_LOG(...) fprintf(stderr, __VA_ARGS__)
+#define SYM_DUMP(p) CSShow((p))
+#else
+#define SYM_LOG(...)
+#define SYM_DUMP(p) CSShow((p))
```
SvyatoslavScherbina, line 34: Should be empty.

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -114,22 +143,61 @@ extern "C" struct SourceInfo Kotlin_getSourceInfo(void* addr) {
...
+     * ASSUMPTION: we assume that the _first_ and the _last_ source infos should belong to real function(symbol) the rest might belong to
+     * inlined functions.
+     */
+    CSSymbolForeachSourceInfo(symbol,
+          ^(CSSourceInfoRef ref) {
+           uint32_t lineNumber = CSSourceInfoGetLineNumber(ref);
+           if (lineNumber == 0)
+             return 0;
+           if (limits.start == -1) {
+              limits.start = lineNumber;
```
SvyatoslavScherbina, line 159: Indent has 3 spaces.

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -114,22 +143,61 @@ extern "C" struct SourceInfo Kotlin_getSourceInfo(void* addr) {
...
+              limits.fileName = CSSourceInfoGetPath(ref);
+           } else
+              limits.end = lineNumber;
```
SvyatoslavScherbina, line 162: No `{}` for `else` branch, while `if` has `{}`.

backend.native/tests/runtime/exceptions/kt-37572.kt
```diff
@@ -0,0 +1,41 @@
...
+        null,
+        "kt-37572.kt" to 25,
+        "kt-37572.kt" to 16,
+        "kt-37572.kt" to 6,
+        "kt-37572.kt" to 4)
+
+internal fun checkFrame(value:String) {
+    val (pos, file, line) = regex.find(value)!!.destructured
+    goldValues[pos.toInt()]?.let{
+        assertTrue(file == it.first && line.toInt() == it.second, "expected ${it.first}:${it.second} got $file:$line")
```
SvyatoslavScherbina, line 39: Why not just have two `assertEquals` instead?
`assertEquals` has proper message by default.

backend.native/tests/runtime/exceptions/kt-37572.kt
```diff
@@ -0,0 +1,42 @@
...
+        null,
+        "kt-37572.kt" to 25,
+        "kt-37572.kt" to 16,
+        "kt-37572.kt" to 6,
+        "kt-37572.kt" to 4)
+
+internal fun checkFrame(value:String) {
+    val (pos, file, line) = regex.find(value)!!.destructured
+    goldValues[pos.toInt()]?.let{
+        assertEquals(file, it.first)
```
SvyatoslavScherbina, line 39: Arguments should probably be flipped: expected value must be the first.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1703,17 +1700,17 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
 
         override fun returnableBlockScope(): CodeContext? = this
 
-        override fun location(line: Int, column: Int): LocationInfo? {
+        override fun location(offset: Int): LocationInfo? {
             return if (returnableBlock.inlineFunctionSymbol != null) {
                 val diScope = functionScope ?: return null
                 val outerFileEntry = outerFileEntry()
-                val inlinedAt = outerContext.location(outerFileEntry.line(returnableBlock.startOffset), outerFileEntry.column(returnableBlock.startOffset))
+                val inlinedAt = outerContext.location(outerFileEntry.line(returnableBlock.startOffset))
```
SvyatoslavScherbina, line 1707: Is this correct?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -684,7 +680,8 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
             declaration?.scope() ?: llvmFunction!!.scope(0, subroutineType(context, codegen.llvmTargetData, listOf(context.irBuiltIns.intType)))
         }
 
-        override fun location(line: Int, column: Int) = scope?.let { LocationInfo(it, line, column) }
+        private val fileScope = (fileScope() as? FileScope)
+        override fun location(offset: Int) = scope?.let { scope -> fileScope?.let{LocationInfo(scope, it.file.fileEntry.line(offset), it.file.fileEntry.column(offset)) } }
```
SvyatoslavScherbina, line 684: What exactly does this fix?

SvyatoslavScherbina: I mean that probably we've got wrong file scope when calling `line`/`column` at call site. Could you point me at the specific code fragment that had wrong file scope?

vvlevchenko: no we have it right for most of the block (it's returnable block), but when we calculate location for variable (captured by inliner) we could be in the middle of returnable block and file isn't actual for variable belonging to call site.    this exactly the place we used wrong place to calculate line/column position

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
vvlevchenko: yep, we calculate real line/column position using actual file

SvyatoslavScherbina: `location` expects `offset` but gets line number.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
SvyatoslavScherbina: That's what I meant :)
Is it code context from `switchSymbolizationContextTo` that was affected by the fix?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
vvlevchenko: Oh, that definitely shouldn't be in this way

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
vvlevchenko: yes

PR #4118, opened 2020-04-18
Add "use" extension function to StableRef

serebit:


Dominaezzz:
How about `consume()`?

ilya-g:
Could you provide some background on the proposed method? In what scenarios it would be useful? How common are they?

serebit:
> Could you provide some background on the proposed method? In what scenarios it would be useful? How common are they?

I... completely overlooked the fact that the data that `StableRef`s hold can't be used after they are disposed. With that in mind, I have a function to replace this with that solves the same problem in a different way.

Dominaezzz:
What do you mean? The data can be used, it just won't be accessible via the `StableRef` anymore.

serebit:
> What do you mean? The data can be used, it just won't be accessible via the `StableRef` anymore.

I wasn't sure, as the docs are ambiguous on whether or not the data itself becomes inaccessible. If you're positive that that's the case, then I'm in favor of `consume` over `use`.

Dominaezzz:
Yes, I'm positive. `StableRef` is more/less a glorified reference count.

serebit:
Hmm. If that's true, then disposing of the stable ref immediately after `get`ting from it could *potentially* be prone to using an object after it's been GCed...

SvyatoslavScherbina:
> Hmm. If that's true, then disposing of the stable ref immediately after `get`ting from it could *potentially* be prone to using an object after it's been GCed...

I don't think so. Could you provide an example?

ilya-g:
>> Could you provide some background on the proposed method? In what scenarios it would be useful? How common are they?

 > I... completely overlooked the fact that the data that StableRefs hold can't be used after they are disposed. With that in mind, I have a function to replace this with that solves the same problem in a different way.

Sorry, that doesn't answer the questions I asked. What I asked for is to provide the motivation behind this proposal.

serebit:
My apologies @ilya-g and @SvyatoslavScherbina, I got confused as I wasn't sure if I fully understood how StableRefs work. I'm not familiar *at all* with Kotlin/Native internals, so I'll defer to you on any technical concerns with this proposal.

My motivation is the existence of single-use callbacks that are guaranteed to only be run once. With these callbacks, the StableRef can be disposed immediately after obtaining its held value if this single-use callback is the only callback in which it will be used. I believe this use case is common enough that a convenience function should be included in the standard library.

serebit:
Decided to close this, as I don't really see the value in it anymore.

SvyatoslavScherbina:
The case you described ("StableRef can be disposed immediately after obtaining its held value") should probably be addressed with something like `.getAndDispose()`, which gets the reference from `StableRef` and then disposes the latter.

PR #4120, opened 2020-04-20
Common Throws annotation

ilya-g:
Common part review: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4078

### DIFF
--- a/runtime/src/main/kotlin/kotlin/Annotations.kt
+++ b/runtime/src/main/kotlin/kotlin/Annotations.kt
@@ -8,6 +8,7 @@ package kotlin
 import kotlin.annotation.AnnotationRetention.BINARY
 import kotlin.annotation.AnnotationRetention.SOURCE
 import kotlin.annotation.AnnotationTarget.*
+import kotlin.reflect.KClass
 
 /**
  * Marks the annotated declaration as deprecated.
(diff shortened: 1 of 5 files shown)


runtime/src/main/kotlin/kotlin/Annotations.kt
```diff
@@ -148,4 +149,23 @@ public annotation class DslMarker
...
+ * which is an instance of one of the [exceptionClasses] or their subclasses,
+ * it is propagated as `NSError`. Other Kotlin exceptions reaching Swift/Objective-C
+ * are considered unhandled and cause program termination.
+ *
+ * @property exceptionClasses the list of checked exception classes that may be thrown by the function.
+ */
+@SinceKotlin("1.4")
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER, AnnotationTarget.CONSTRUCTOR)
+@Retention(AnnotationRetention.BINARY)
+public actual annotation class Throws(actual vararg val exceptionClasses: KClass<out Throwable>)
```
ilya-g, line 171: The fully qualified name has been changed in the result, but I don't think it's worth hassle to provide a migration for that.

runtime/src/main/kotlin/kotlin/Annotations.kt
```diff
@@ -148,4 +149,23 @@ public annotation class DslMarker
...
+ *
+ * When Kotlin function called from Swift/Objective-C code throws an exception
+ * which is an instance of one of the [exceptionClasses] or their subclasses,
+ * it is propagated as `NSError`. Other Kotlin exceptions reaching Swift/Objective-C
+ * are considered unhandled and cause program termination.
+ *
+ * @property exceptionClasses the list of checked exception classes that may be thrown by the function.
+ */
+@SinceKotlin("1.4")
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER, AnnotationTarget.CONSTRUCTOR)
```
ilya-g, line 169: I've extended targets list to match JVM annotation. Hope that wouldn't be a problem.

SvyatoslavScherbina: I'm not sure backend is ready for this.
I guess `@Throws` should be forbidden on property accessors for native in frontend.

runtime/src/main/kotlin/kotlin/Annotations.kt
SvyatoslavScherbina: Can we provide `typealias Throws` in `kotlin.native` package?

ilya-g: This would likely have the same ambiguity consequences as having `kotlin.Throws` and `kotlin.jvm.Throws` in default-imported packages in Kotlin/JVM. That's what I call a hassle.

SvyatoslavScherbina: I get this.
I mean that you probably have a solution for this ambiguity for JVM. How hard would it be to apply the same solution to Native?

runtime/src/main/kotlin/kotlin/Annotations.kt
ilya-g: Reverted accessor targets for K/N.

PR #4125, opened 2020-04-21
[debug] Forbade getting llvm for external non-exported functions + test

homuroll:
This could happen for inline lambdas

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -3340,6 +3340,11 @@ linkTest("inline_lateinitProperty_linkTest") {
     lib = "codegen/inline/lateinitProperty_linkTest_lib.kt"
 }
 
+linkTest("inline_inlineCtor_linkTest") {
+    source = "codegen/inline/inlineCtor_linkTest_main.kt"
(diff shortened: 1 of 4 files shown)


backend.native/tests/build.gradle
```diff
@@ -3340,6 +3340,11 @@ linkTest("inline_lateinitProperty_linkTest") {
     lib = "codegen/inline/lateinitProperty_linkTest_lib.kt"
 }
 
+linkTest("inline_inlineCtor_linkTest") {
```
SvyatoslavScherbina, line 3343: I can't make this test fail when the fix is reverted. Can you?

SvyatoslavScherbina: (It looks like I've got tricked by Gradle, let me recheck)

homuroll: Hmm, did you compile with caches on?

SvyatoslavScherbina: Well, we don't apply caches for link tests at the moment. I believe we have ticket for this.
Let's keep it as is for now.

PR #4126, opened 2020-04-21
[runtime] Terminate on IncorrectDereferenceException from ObjC weak p…

knebekaizer:
…roperty

Issue: ObjC runtime gets a os_unfair_lock when accessing weak property.
If exception is thrown from _tryRetainImp this lock left unlocked,
so it will hang on `__ulock_wait` on the next access.
Interim workaround: terminate instead of exception propagating.

However TerminateWithUnhandledException causes stacktrace dump to fail
in debug mode because of CSSymbolOwnerGetSymbolWithAddress failed on
the same recursive os_unfair_lock attempt.

### DIFF
--- a/runtime/src/main/cpp/ObjCInterop.mm
+++ b/runtime/src/main/cpp/ObjCInterop.mm
@@ -16,6 +16,9 @@
 
 #if KONAN_OBJC_INTEROP
 
+#import <Foundation/NSException.h>
+#import <objc/objc-exception.h>
+
 #include <objc/objc.h>
 #include <objc/runtime.h>
 #include <objc/message.h>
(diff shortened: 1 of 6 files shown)


runtime/src/main/cpp/ObjCInterop.mm
```diff
@@ -28,6 +30,7 @@
 #include "Natives.h"
 #include "ObjCInterop.h"
 #include "ObjCExportPrivate.h"
+//#include "ObjCExportErrors.h"
```
SvyatoslavScherbina, line 33: Kinda unused.

runtime/src/main/cpp/ObjCInterop.mm
```diff
@@ -95,7 +98,15 @@ BOOL _tryRetainImp(id self, SEL _cmd) {
   // this is a regression for instances of Kotlin subclasses of Obj-C classes:
   // loading a reference to such an object from Obj-C weak reference now fails on "wrong" thread
   // unless the object is frozen.
-  return getBackRef(self)->tryAddRef();
+  try {
+    return getBackRef(self)->tryAddRef();
+  }
+  catch (ExceptionObjHolder& e) {
+    // TODO: check for IncorrectDereferenceException and possible weak property access
+    [NSException raise:NSGenericException
```
SvyatoslavScherbina, line 106: Should also terminate, like
```objc
@try {
    [NSException raise:...];
} @catch (...) {
    objc_terminate();
}
```
Ugly but somewhat necessary `¯\_(ツ)_/¯`.

Termination would be required for other operations too, as we discussed offline.

PR #4127, opened 2020-04-22
[IR] Fixed IR_VALIDATION warning

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/KonanDefaultParameterInjector.kt
```diff
@@ -22,7 +23,11 @@ internal class KonanDefaultParameterInjector(private val konanContext: KonanBack
             null -> IrConstImpl.constNull(startOffset, endOffset, context.irBuiltIns.nothingNType)
             PrimitiveBinaryType.BOOLEAN -> IrConstImpl.boolean(startOffset, endOffset, type, false)
             PrimitiveBinaryType.BYTE -> IrConstImpl.byte(startOffset, endOffset, type, 0)
-            PrimitiveBinaryType.SHORT -> IrConstImpl.short(startOffset, endOffset, type, 0)
+            PrimitiveBinaryType.SHORT -> when (type.getInlinedClassNative()) {
+                context.irBuiltIns.short -> IrConstImpl.short(startOffset, endOffset, type, 0)
+                context.irBuiltIns.char -> IrConstImpl.char(startOffset, endOffset, type, 0.toChar())
+                else -> error("Unexpected type with PrimitiveBinaryType.SHORT: ${type.render()}")
```
SvyatoslavScherbina, line 29: This is fragile. What happens if type is `inline class Foo(val value: Short)`?

AFAIK, current implementation uses `Foo?` parameter type in this case, but this may change.
Please consider using `unwrapToPrimitiveOrReference` (suitable for the entire `when` too).

homuroll: Done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/KonanDefaultParameterInjector.kt
```diff
@@ -18,18 +18,27 @@ internal class KonanDefaultParameterInjector(private val konanContext: KonanBack
...
-            PrimitiveBinaryType.DOUBLE -> IrConstImpl.double(startOffset, endOffset, type, 0.0)
-            PrimitiveBinaryType.POINTER -> irCall(startOffset, endOffset, symbols.getNativeNullPtr.owner, emptyList())
-            PrimitiveBinaryType.VECTOR128 -> TODO()
-        }
+        val nullConstOfEquivalentType = type.unwrapToPrimitiveOrReference(
+                eachInlinedClass = { _, _ -> },
+                ifReference = { _ -> IrConstImpl.constNull(startOffset, endOffset, context.irBuiltIns.nothingNType) },
+                ifPrimitive = { primitiveType, nullable ->
+                    if (nullable)
+                        IrConstImpl.constNull(startOffset, endOffset, context.irBuiltIns.nothingNType)
```
SvyatoslavScherbina, line 26: I'm not sure this is correct. Consider `type = Foo` where
```
inline class Foo(val value: NonNullNativePtr?)
```
Then `null` (of `Nothing?` type) doesn't conform to `Foo`.
Previous variant is probably wrong too, and I suspect this code is unreachable.

SvyatoslavScherbina: (Discussed offline).

PR #4128, opened 2020-04-22
[CInterop] Fix keyword mangling in metadata mode

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
```diff
@@ -391,7 +391,7 @@ class StubIrTextEmitter(
     }
 
     private fun renderClassifierDeclaration(classifier: Classifier): String =
-            kotlinFile.declare(classifier)
+            kotlinFile.declare(classifier).asSimpleName()
```
SvyatoslavScherbina, line 394: Are there other usages of `declare`?

backend.native/tests/interop/basics/mangling_keywords.kt
```diff
@@ -0,0 +1,35 @@
+import kotlin.test.*
+import mangling_keywords.*
+
+fun main() {
+    // Check that all Kotlin keywords are imported and mangled.
+    assertEquals(`as`, "as")
```
SvyatoslavScherbina, line 6: Arguments should probably be swapped.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrTextEmitter.kt
sbogolepov: No. I'll try to gradually refactor `KotlinFile` later.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
```diff
@@ -320,7 +320,7 @@ abstract class KotlinFile(
                 // TODO: using original global name should be preferred to importing the clashed name.
             } else {
                 declaredProperties.add(name)
-                name.asSimpleName()
+                name
```
knebekaizer, line 323: Have you ever consider different types for MangledName and non-mangled? Even if they have the same text representation (for most of cases I guess) they still have different usage areas.

sbogolepov: Yeah, it's may be a good idea to use `inline class` for such thing.

PR #4136, opened 2020-04-24
[kotlin compiler][update][build] 1.4.0-dev-7380

vvlevchenko:
fixes the issue with communication of daemon and gradle plugin

### DIFF
--- a/build-tools/src/main/kotlin/org/jetbrains/kotlin/Reporter.kt
+++ b/build-tools/src/main/kotlin/org/jetbrains/kotlin/Reporter.kt
@@ -6,7 +6,6 @@
 package org.jetbrains.kotlin
 
 import com.ullink.slack.simpleslackapi.impl.SlackSessionFactory
-import kotlinx.serialization.ImplicitReflectionSerializer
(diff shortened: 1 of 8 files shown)


build-tools/src/main/kotlin/org/jetbrains/kotlin/Reporter.kt
```diff
@@ -74,12 +72,14 @@ open class Reporter : DefaultTask() {
...
-        sendTextToSlack(report)
+        if (doSlackSending())
+            sendTextToSlack(report)
     }
-
-
 }
 
+private fun DefaultTask.doSlackSending() = !project.hasProperty("build.reporter.noSlack")
+        || project.property("build.reporter.noSlack").toString().toBoolean()
```
SvyatoslavScherbina, line 81: So if `-Pbuild.reporter.noSlack=true`, the script does send Slack report?

vvlevchenko: oops

PR #4144, opened 2020-04-28
Add fir-serialization to libraries with ignored collisions

mglukhikh:


udalov:
BTW is it possible to invent a more robust solution to prevent hardcoding all compiler modules here and break the build each time a new module is added?

PR #4145, opened 2020-04-28
Avoid duplicated selectors when Kotlin class inherits from muliple ba…

knebekaizer:
…ses (protocols) [KT-38234]

Issue: RuntimeAssert in AddMethods (ObjCInterop.mm) when kotlin class overrides the same
selector from multiple bases, e.g. from the base objc class and a protocol(s).

### DIFF
--- a/backend.native/tests/interop/objc/tests/KT38234_override.h
+++ b/backend.native/tests/interop/objc/tests/KT38234_override.h
@@ -1,0 +1,9 @@
+#import <Foundation/NSObject.h>
+
+@protocol P1
+-(int)foo;
+@end;
+
+@interface Base : NSObject <P1>
+-(int)callFoo;
+@end;
--- a/backend.native/tests/interop/objc/tests/KT38234_override.h
+++ b/backend.native/tests/interop/objc/tests/KT38234_override.h
@@ -1,9 +1,9 @@
 #import <Foundation/NSObject.h>
 
-@protocol P1
+@protocol KT38234_P1
 -(int)foo;
 @end;
 
-@interface Base : NSObject <P1>
+@interface KT38234_Base : NSObject <KT38234_P1>
 -(int)callFoo;
 @end;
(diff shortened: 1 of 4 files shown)


sbogolepov:
Don't forget to add a test case.

knebekaizer:
Test added

backend.native/tests/interop/objc/tests/KT38234_override.h
```diff
@@ -0,0 +1,9 @@
+#import <Foundation/NSObject.h>
+
+@protocol P1
```
SvyatoslavScherbina, line 3: Currently all tests in this directory are compiled together, so using more unique names (like in [original reproducer](https://github.com/JetBrains/kotlin-native/compare/kt-38234#diff-62b02828c319a01dde4c977ea7a7cab2R3)) is preferable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/KotlinObjCClassInfoGenerator.kt
```diff
@@ -117,19 +117,21 @@ internal class KotlinObjCClassInfoGenerator(override val context: Context) : Con
...
-                                return@mapNotNull null
-
-                ObjCMethodDesc(
-                        annotation.getAnnotationStringValue("selector"),
-                        annotation.getAnnotationStringValue("encoding"),
-                        it.llvmFunction
-                )
-            }
+    private fun IrClass.generateImpMethodDescs(): List<ObjCMethodDesc> {
+        val noDups = mutableMapOf<String, ObjCMethodDesc>()  // remove duplicates [KT-38234]
```
SvyatoslavScherbina, line 121: Note that the same can probably be done by keeping most of the code here intact, by adding `.distinctBy { it.selector }` at the end.

PR #4147, opened 2020-04-29
[CInterop] Fix global property mangling

sbogolepov:
Do not mangle properties with the same name but with different receivers.

### DIFF
--- a/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
+++ b/Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
@@ -33,7 +33,7 @@ interface KotlinScope {
      * @return the string to be used as a name in the declaration of the property in current scope,
      * or `null` if the property with given name can't be declared.
      */
-    fun declareProperty(name: String): String?
+    fun declareProperty(receiver: String?, name: String): String?
 
     val mappingBridgeGenerator: MappingBridgeGenerator
 }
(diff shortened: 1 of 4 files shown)


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
```diff
@@ -334,6 +336,17 @@ abstract class KotlinFile(
 
 }
 
+internal fun getTopLevelPropertyDeclarationName(scope: KotlinScope, property: PropertyStub): String {
+    val receiverName = when (val type = property.receiverType) {
+        null -> null
+        is ClassifierStubType -> type.classifier.getRelativeFqName()
```
SvyatoslavScherbina, line 342: Why relative?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/KotlinCodeModel.kt
```diff
@@ -339,9 +339,9 @@ abstract class KotlinFile(
 internal fun getTopLevelPropertyDeclarationName(scope: KotlinScope, property: PropertyStub): String {
     val receiverName = when (val type = property.receiverType) {
         null -> null
-        is ClassifierStubType -> type.classifier.getRelativeFqName()
-        is AbbreviatedType -> type.abbreviatedClassifier.getRelativeFqName()
-        is FunctionalType -> type.classifier.getRelativeFqName()
+        is ClassifierStubType -> type.classifier.fqName
+        is AbbreviatedType -> type.abbreviatedClassifier.fqName
```
SvyatoslavScherbina, line 343: This doesn't seem consistent with Kotlin overloading rules.
Maybe use `.underlyingType`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrExtensions.kt
```diff
@@ -86,4 +86,15 @@ fun ConstantStub.determineConstantAnnotationClassifier(): Classifier = when (thi
...
         else -> error("Real constant with unexpected size of $size.")
     }
-}.let { Classifier.topLevel(cinteropInternalPackage, "ConstantValue").nested(it) }
\ No newline at end of file
+}.let { Classifier.topLevel(cinteropInternalPackage, "ConstantValue").nested(it) }
+
+/**
+ * Returns the original name of the given type.
+ */
+val StubType.underlyingTypeFqName: String
```
knebekaizer, line 94: Should it be the abstract property of StubType instead of extension? Dynamic type check `when(this) is ...` smells like an antipattern, isn't it?

sbogolepov: It is easy to replace extension with abstract method if we see that this method affects performance. Until then I prefer to keep classes as simple as possible.

PR #4150, opened 2020-04-29
[GC] Added new threshold for collecting cycles in GC (KT-19076)

LepilkinaElena:
Fix for https://youtrack.jetbrains.com/issue/KT-19076

### DIFF
--- a/runtime/src/main/cpp/Memory.cpp
+++ b/runtime/src/main/cpp/Memory.cpp
@@ -99,6 +99,8 @@ constexpr size_t kMaxGcAllocThreshold = 8 * 1024 * 1024;
 // If GC base work to collection cycles time ratio is less this value,
 // increase GC threshold for cycles collection.
 constexpr double kGcCollectCyclesLoadRatio = 0.05;
+// Minimum time of cycles collection to change thresholds.
+constexpr size_t kGcCollectCyclesMinimumDuration = 300;
 
 #endif  // USE_GC
 
@@ -1674,9 +1676,10 @@ void garbageCollect(MemoryState* state, bool force) {
     }
   }
 
-  if (state->gcErgonomics && collectCyclesDuration > 0 &&
+  if (state->gcErgonomics && collectCyclesDuration > kGcCollectCyclesMinimumDuration &&
     double(processFinalizerQueueDuration) / collectCyclesDuration < kGcCollectCyclesLoadRatio) {
-     increaseGcCollectCyclesThreshold(state);
+    increaseGcCollectCyclesThreshold(state);
+    GC_LOG("Adjusting GC collecting cycles threshold to %lld\n", state->gcCollectCyclesThreshold);
   }
 
   state->gcInProgress = false;
--- a/runtime/src/main/cpp/Memory.cpp
+++ b/runtime/src/main/cpp/Memory.cpp
@@ -100,7 +100,7 @@ constexpr size_t kMaxGcAllocThreshold = 8 * 1024 * 1024;
 // increase GC threshold for cycles collection.
 constexpr double kGcCollectCyclesLoadRatio = 0.05;
 // Minimum time of cycles collection to change thresholds.
-constexpr size_t kGcCollectCyclesMinimumDuration = 300;
+constexpr size_t kGcCollectCyclesMinimumDuration = 200;
 
 #endif  // USE_GC
 
@@ -2303,9 +2303,10 @@ void startGC() {
 
 void setGCThreshold(KInt value) {
   GC_LOG("setGCThreshold %d\n", value)
-  if (value > 0) {
-    initGcThreshold(memoryState, value);
+  if (value <= 0) {
+    ThrowIllegalArgumentException();
   }
+  initGcThreshold(memoryState, value);
 }
 
 KInt getGCThreshold() {
@@ -2315,9 +2316,10 @@ KInt getGCThreshold() {
 
 void setGCCollectCyclesThreshold(KInt value) {
   GC_LOG("setGCCollectCyclesThreshold %d\n", value)
-  if (value > 0) {
-    initGcCollectCyclesThreshold(memoryState, value);
+  if (value <= 0) {
+    ThrowIllegalArgumentException();
   }
+  initGcCollectCyclesThreshold(memoryState, value);
 }
 
 KInt getGCCollectCyclesThreshold() {
@@ -2327,9 +2329,11 @@ KInt getGCCollectCyclesThreshold() {
 
 void setGCThresholdAllocations(KLong value) {
   GC_LOG("setGCThresholdAllocations %lld\n", value)
-  if (value > 0) {
-    memoryState->allocSinceLastGcThreshold = value;
+  if (value <= 0) {
+    ThrowIllegalArgumentException();
   }
+
+  memoryState->allocSinceLastGcThreshold = value;
 }
 
 KLong getGCThresholdAllocations() {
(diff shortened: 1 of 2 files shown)


runtime/src/main/cpp/Memory.cpp
```diff
@@ -87,11 +89,19 @@ constexpr double kGcToComputeRatioThreshold = 0.5;
...
+constexpr size_t kMaxErgonomicToFreeSizeThreshold = 8 * 1024 * 1024;
 // How many elements in finalizer queue allowed before cleaning it up.
 constexpr size_t kFinalizerQueueThreshold = 32;
 // If allocated that much memory since last GC - force new GC.
 constexpr size_t kMaxGcAllocThreshold = 8 * 1024 * 1024;
+// If GC base work to collection cycles time ratio is less this value,
+// increase GC threshold for cycles collection.
+constexpr double kGcCollectCyclesLoadRatio = 0.05;
+// Minimum time of cycles collection to change thresholds.
+constexpr size_t kGcCollectCyclesMinimumDuration = 100;
```
LepilkinaElena, line 103: I'm waiting all results of experiments for this value on TC.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2254,6 +2313,18 @@ KInt getGCThreshold() {
   return memoryState->gcThreshold;
 }
 
+void setGCCollectCyclesThreshold(KInt value) {
+  GC_LOG("setGCCollectCyclesThreshold %d\n", value)
+  if (value > 0) {
```
projedi, line 2318: I think it's better to throw `IllegalArgumentException` here (when it's called from kotlin user code; when it gets called from runtime on it's own behalf, it should `RuntimeAssert`)

runtime/src/main/cpp/Memory.cpp
```diff
@@ -87,11 +89,19 @@ constexpr double kGcToComputeRatioThreshold = 0.5;
...
 // Threshold of size for toFree set, triggering actual cycle collector.
-constexpr size_t kMaxToFreeSize = 8 * 1024;
+constexpr size_t kMaxToFreeSizeThreshold = 8 * 1024;
+// Never exceed this value when increasing size for toFree set, triggering actual cycle collector.
+constexpr size_t kMaxErgonomicToFreeSizeThreshold = 8 * 1024 * 1024;
 // How many elements in finalizer queue allowed before cleaning it up.
 constexpr size_t kFinalizerQueueThreshold = 32;
 // If allocated that much memory since last GC - force new GC.
 constexpr size_t kMaxGcAllocThreshold = 8 * 1024 * 1024;
+// If GC base work to collection cycles time ratio is greater this value,
```
SvyatoslavScherbina, line 99: > If GC base work to collection cycles time ratio

Seems wrong.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2953,6 +3035,20 @@ KInt Kotlin_native_internal_GC_getThreshold(KRef) {
 #endif
 }
 
+void Kotlin_native_internal_GC_setCollectCyclesThreshold(KRef, KLong value) {
+#if USE_GC
+  setGCCollectCyclesThreshold(value);
```
SvyatoslavScherbina, line 3040: Type mismatch here: `KLong` gets silently truncated to `KInt`.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1604,38 +1629,77 @@ void garbageCollect(MemoryState* state, bool force) {
...
+  auto processFinalizerQueueDuration = konan::getTimeMicros() - processFinalizerQueueStartTime;
+  GC_LOG("||| GC: processFinalizerQueueDuration %lld\n", processFinalizerQueueDuration);
+#endif
 
-  if (force || state->toFree->size() > kMaxToFreeSize) {
+  int64_t collectCyclesDuration = 0;
+  int64_t cyclicGcEndTime = 0;
+  if (force || state->toFree->size() > state->gcCollectCyclesThreshold) {
     while (state->toFree->size() > 0) {
+      auto collectCyclesStartTime = konan::getTimeMicros();
```
projedi, line 1668: I'm wondering: what if we include `processFinalizerQueue` execution time from below into `collectCycles` time? This way we can wrap the loop in `cyclicGcStartTime = ...; while(...) {...}; cyclicGcEndTime = ...;` and adjust the `if` below `double(cyclicGcEndTime - cyclicGcStartTime) / (cyclicGcStartTime - state->lastCyclicGcTimestamp + 1)`. Rationale behind this: `processFinalizerQueue` work is done precisely because we were able to resolve some cycles (or am I wrong here?). What do you think?

runtime/src/main/cpp/Memory.cpp
LepilkinaElena: // If the ratio of GC collection cycles time to program execution time is greater this value,
// increase GC threshold for cycles collection.

runtime/src/main/cpp/Memory.cpp
LepilkinaElena: I think it would be fine for current approach, if there is a lot of garbage then  just freeing takes quite a big amount of time. But usually there is small amount of cyclic garbage. I tried this to use less calls for measuring time and check on benchmarks.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1604,37 +1629,75 @@ void garbageCollect(MemoryState* state, bool force) {
...
+        GC_LOG("||| GC: processFinalizerQueueDuration = %lld\n", processFinalizerQueueDuration);
+      #endif
+    }
+    auto cyclicGcEndTime = konan::getTimeMicros();
+    #if PROFILE_GC
+      GC_LOG("||| GC: collectCyclesDuration = %lld\n", cyclicGcEndTime - cyclicGcStartTime);
+    #endif
+    auto cyclicGcDuration = cyclicGcEndTime - cyclicGcStartTime;
+    if (state->gcErgonomics && cyclicGcDuration > kGcCollectCyclesMinimumDuration &&
+        double(cyclicGcDuration) / (gcStartTime - state->lastCyclicGcTimestamp + 1) > kGcCollectCyclesLoadRatio) {
```
projedi, line 1683: `s/gcStartTime/cyclicGcStartTime/`?

LepilkinaElena: I amn't sure  because we wanted time of program execution, in this case we get time of program execution + part of GC. But at the same time we don't subtract time of other GC without cyclic collection. Ok, I'll change for consistency.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: IMHO, measuring only `collectCycles` seems more precise, because processing finalizer queue is always a "useful work". If cycle collection takes 1 ms while processing finalizer queue takes 1000 ms, the latter shouldn't be the reason to perform the former less often.
On the other hand, T(cycle collection) > kT(process finalizer queue). So this example is not exactly possible.

Considering the fact that measuring processing finalizer queue makes the code simpler, I guess it is ok to keep it that way.

runtime/src/main/cpp/Memory.cpp
SvyatoslavScherbina: lgtm.

runtime/src/main/cpp/Memory.cpp
LepilkinaElena: I agree that it's more precise, but impact of processing finalizer queue is quite small. Nevertheless I checked on benchmarks, nothing critical because of this change.

PR #4152, opened 2020-04-30
Align docs with USE_CYCLIC_GC=0 (ed36f02c)

SvyatoslavScherbina:


runtime/src/main/kotlin/kotlin/native/concurrent/Atomics.kt
```diff
@@ -216,7 +216,7 @@ private fun debugString(value: Any?): String {
 /**
  * An atomic reference to a frozen Kotlin object. Can be used in concurrent scenarious
  * but frequently shall be of nullable type and be zeroed out once no longer needed.
- * Asynchronous cycle collector takes care of cyclic references of that kind.
+ * Otherwise memory leak could happen if atomic reference is a part of a reference cycle.
```
ilya-g, line 219:
```suggestion
 * Otherwise memory leak could happen if the atomic reference is a part of a reference cycle.
```

Issue #2316, opened 2018-11-13
K/N debugger: no way to access `this` in debugger


ddolovov:
**Problem:** This is no way to access `this` in Kotlin/Native debugger would it be LLDB console or CLion expression evaluator window.
<img src="http://drive.google.com/uc?export=view&id=1DCsXKyC1SkDPC72m2UO0c_YJat3zSFu_">
<img src="https://drive.google.com/uc?export=view&id=1F0gwpHjCKBRpsrqAjx-nJagL1nkntImv">

**Expected:** It should be possible to use `this` (or close analogue) in LLDB console and in CLion expression evaluator.

vvlevchenko:
please take a look at #2323, it's workaround over limitation of c-expression evaluator. Perhaps it could be used in CLion, for a period of lack of kotlin expression evaluator.

ddolovov:
I would like to reopen this issue because I believe we do need possibility to evaluate expressions with `this` keyword in K/N debugger. As for K/N support in IDE this even becomes critical (other flavors of Kotlin allow using `this` in "Expression Evaluator" window).

SvyatoslavScherbina:
@vvlevchenko @ddolovov is this still the case?

ddolovov:
Yes, it is. There is still no way to access `this` in IDE, e.g. in "Evaluate Expression"

vvlevchenko:
```

SvyatoslavScherbina:
I see.
> as a workaround it's possible to rename: <this> -> _this

Let's use this approach.

vvlevchenko:
the issue's been fixed in agreed way.

PR #4153, opened 2020-04-30
[codegen][debug-info] generate `_this` name for <this> special variable

vvlevchenko:


vvlevchenko:
fix for #2316

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2489,6 +2490,24 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
     }
 }
 
+private fun IrValueParameter.nameConversion() = descriptor.name.nameConversion()
```
SvyatoslavScherbina, line 2493: `nameConversion` function name here and below is too ambiguous for this context.
Maybe `nameForDebugInfo` or `debugInfoName`?

ddolovov:
LGTM

PR #4154, opened 2020-04-30
Provide an -Xmanifest-targets flag, allowing to specify manifest targets explicitly

dsavvinov:
This is needed for https://youtrack.jetbrains.com/issue/KT-38658. This commit only exposes a CLI-argument,
which later will be used by Gradle tooling


dsavvinov:
Respective review with the usage of API

dsavvinov:
Turns out there's a `-manifest` argument, which provides exactly the needed functionality. We'll stick to using it. Thanks everyone!

PR #4157, opened 2020-04-30
Decommonize collection builders

qurbonzoda:


runtime/src/main/kotlin/kotlin/collections/ArrayList.kt
```diff
@@ -9,52 +9,45 @@ actual class ArrayList<E> private constructor(
...
+    }
+
+    override actual val size: Int
         get() = length
 
     override actual fun isEmpty(): Boolean = length == 0
 
     override actual fun get(index: Int): E {
-        checkIndex(index)
+        AbstractList.checkElementIndex(index, length)
```
ilya-g, line 39: Let's make a copy of `checkElementIndex` and other checks from `AbstractList` companion as member functions.

runtime/src/main/kotlin/kotlin/collections/ArrayList.kt
```diff
@@ -309,7 +309,8 @@ actual class ArrayList<E> private constructor(
         }
 
         override fun set(element: E) {
-            list.checkIndex(lastIndex)
+            list.checkIsMutable()
+            check(lastIndex != -1) { "Call next() or previous() before replacing element from the iterator." }
```
ilya-g, line 313: Is it possible to get `IOOBE` even after this check?

runtime/src/main/kotlin/kotlin/collections/Collections.kt
```diff
@@ -41,6 +43,23 @@ public interface MutableIterable<out T> : Iterable<T> {
 }
 
 
+@PublishedApi
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+@kotlin.internal.InlineOnly
+internal actual inline fun <E> buildListInternal(@BuilderInference builderAction: MutableList<E>.() -> Unit): List<E> {
```
ilya-g, line 50: Since these functions are internal they likely do not need `@BuilderInference` annotation.

runtime/src/main/kotlin/kotlin/collections/ArrayList.kt
qurbonzoda: `lastIndex` is changed to non `-1` only in `previous()` and `next()` operations. Hence if the list is not modified bypassing this iterator, this check should be enough.

ilya-g: Yes, I mean the situation when the list is modified after moving the iterator. Is it possible, for example, to write the value beyond the last element of the sublist and go unnoticed?

runtime/src/main/kotlin/kotlin/collections/Collections.kt
qurbonzoda: Fixed

backend.native/tests/runtime/collections/hash_map0.kt
```diff
@@ -184,9 +182,9 @@ fun testPutEntry() {
     assertTrue(m.entries.contains(e))
     assertTrue(m.entries.remove(e))
     assertTrue(mapOf("b" to "2", "c" to "3") == m)
-    assertTrue(m.entries.add(e))
+    assertEquals(null, m.put(e.key, e.value))
```
qurbonzoda, line 185: `entries` property shouldn't support inserting entries

backend.native/tests/runtime/collections/hash_map0.kt
```diff
@@ -155,7 +155,6 @@ fun testEquals() {
     assertFalse(m == mapOf("a" to "1", "b" to "2", "c" to "5"))
     assertFalse(m == mapOf("a" to "1", "b" to "2"))
     assertEquals(m.keys, expected.keys)
-    assertEquals(m.values, expected.values)
```
qurbonzoda, line 158: After changes in this commit `values` collection inherits `Object.equals()` and `Object.hashCode()` methods. 
Such behaviour is consistent with java `LinkedHashMap`

runtime/src/main/kotlin/kotlin/collections/ArrayList.kt
qurbonzoda: I believe it's not possible if the list is modified via this iterator, because iterator modifies only with `add` or `remove` operations, and they both set `lastIndex` to `-1`. 
IIRC if the list/sublist is modified bypassing its already created iterator, the iterators behaviour becomes undefined.

qurbonzoda: Index out of bounds is also checked now

backend.native/tests/runtime/collections/hash_map0.kt
ilya-g: Can be replaced with `m.values.toList()` and `expected.values.toList()`

qurbonzoda: done

PR #4161, opened 2020-05-06
(~)[runtime] Use gsl-lite span

knebekaizer:
Syntax sugar, does not affect compiled code

### DIFF
--- a/runtime/src/main/cpp/TypeInfo.h
+++ b/runtime/src/main/cpp/TypeInfo.h
@@ -22,6 +22,8 @@
 #include "Common.h"
 #include "Names.h"
 
+#include "gsl-lite.h"
+
 #if KONAN_TYPE_INFO_HAS_WRITABLE_PART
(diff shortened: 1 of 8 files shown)


runtime/src/main/cpp/TypeInfo.h
```diff
@@ -109,16 +111,23 @@ struct TypeInfo {
     int32_t instanceSize_;
     // Must be pointer to Any for array classes, and null for Any.
     const TypeInfo* superType_;
+
     // All object reference fields inside this object.
     const int32_t* objOffsets_;
     // Count of object reference fields inside this object.
     // 1 for kotlin.Array to mark it as non-leaf.
     int32_t objOffsetsCount_;
+	auto fieldOffsets() const { return gsl::span<const int32_t>(objOffsets_, (size_t)objOffsetsCount_); }
```
SvyatoslavScherbina, line 120: Something wrong with indents.

PR #4165, opened 2020-05-12
Add test for [KT-37067]

knebekaizer:
* Issue is already fixed by metadata-based cinterop
* Fix sample file naming for case-sensitive FS


SvyatoslavScherbina:
Please fix relevant annotation declarations (`CCall.ReturnsRetained` etc) too, currently these declarations are applicable only to functions but not to property accessors, according to the language rules.

SvyatoslavScherbina (approved):

SvyatoslavScherbina:
I've rerun the tests on rebased version. Everything is ok, merging.

PR #4166, opened 2020-05-12
Do not generate getters for global singletons

projedi:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -901,31 +901,44 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
-        // If object is imported - access it via getter function.
-        if (isExternal(irClass)) {
-            val valueGetterName = irClass.objectInstanceGetterSymbolName
-            val valueGetterFunction = LLVMGetNamedFunction(context.llvmModule, valueGetterName) ?:
-                LLVMAddFunction(context.llvmModule, valueGetterName,
-                        functionType(kObjHeaderPtr, false, kObjHeaderPtrPtr))
+        val storageKind = irClass.storageKind(context)
+
+        // If thread local object is imported - access it via getter function.
+        if (isExternal(irClass) && storageKind == ObjectStorageKind.THREAD_LOCAL) {
```
SvyatoslavScherbina, line 907: It makes sense to expose a thread-local singleton by generating getter for the instance storage instead of getter for the instance itself. Would this make the compiler code simpler by unifying thread-local and global singletons?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt
```diff
@@ -85,7 +85,16 @@ internal val IrClass.writableTypeInfoSymbolName: String
         return "ktypew:" + this.fqNameForIrSerialization.toString()
     }
 
-internal val IrClass.objectInstanceGetterSymbolName: String
+internal val IrClass.globalObjectInstanceSymbolName: String
```
SvyatoslavScherbina, line 88: It's not an "instance" symbol name, but a name for "instance storage". Same below.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -901,31 +901,48 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+            val singleton = context.llvmDeclarations.forSingleton(irClass)
+            val instanceAddress = singleton.instanceStorage
+            instanceAddress.getAddress(this)
         }
 
-        val storageKind = irClass.storageKind(context)
         when (storageKind) {
-            ObjectStorageKind.SHARED -> context.llvm.sharedObjects += irClass
+            ObjectStorageKind.SHARED -> context.llvm.sharedObjects += objectPtr
             ObjectStorageKind.THREAD_LOCAL -> context.llvm.objects += irClass
```
SvyatoslavScherbina, line 939: We need to recheck if `objects` set is still required.
Motivation: the usage before this PR was kinda a hack to decide on (de)initializers generation, this hack probably don't work as expected anymore.

PR #4169, opened 2020-05-12
Update GRADLE_PLUGIN.md

belovrv:


GRADLE_PLUGIN.md
```diff
@@ -61,7 +61,7 @@ kotlin {
     println(targets.windows)
     
     // You also can configure all native targets in a single block.
-    targets.withType(KotlinNativeTarget) {
+    targets.withType(KotlinNativeTarget::class.java) {
```
SvyatoslavScherbina, line 64: This code snippet begins with &#96;&#96;&#96;groovy.
Is `KotlinNativeTarget::class.java` valid for Groovy?

ilmat192: You're right, Groovy doesn't allow this syntax. But this snippet still needs to be updated because we must write an FQ class name here or add a corresponding import (see the suggestion).
```suggestion
    targets.withType(org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget) {
```

SvyatoslavScherbina: This script starts with 
```groovy
import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
```
Isn't it enough?

ilmat192: My bad. This import is enough, so closing the PR.

PR #4171, opened 2020-05-13
Remove unused kotlinx-serialization-runtime dependency

ilya-g:


ilya-g:
There's also this leftover: https://github.com/JetBrains/kotlin-native/blob/master/build-tools/settings.gradle.kts#L19

PR #4178, opened 2020-05-19
[Interop] Fixes for struct accessors

sbogolepov:
- Fix `structVar.field = TODO()`
- Fix `enumVar.value = TODO()`
- Add support for fields of Objective-C types
- Tests for these cases


backend.native/tests/interop/objc/tests/structWithNSObject.h
```diff
@@ -0,0 +1,15 @@
...
+    NSString* nsString;
+    NSObject* object;
+    int (^block)(void);
+
+    NSArray* array;
+    NSMutableArray* mutableArray;
+
+    NSSet* set;
+
+    NSDictionary* dictionary;
```
SvyatoslavScherbina, line 13: `id`-typed field is missing.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -123,10 +133,12 @@ private fun InteropCallContext.writeValueToMemory(
 
 private fun InteropCallContext.determineInMemoryType(type: IrType): IrType {
     val classifier = type.classOrNull!!
-    return if (classifier in symbols.unsignedIntegerClasses) {
-        symbols.unsignedToSignedOfSameBitWidth.getValue(classifier).owner.defaultType
-    } else {
-        type
+    return when {
+        // enum may have an unsigned base type so we perform a recursive call.
+        type.isCEnumType() -> determineInMemoryType(type.getCEnumPrimitiveType())
```
SvyatoslavScherbina, line 138: Is this code reachable? Could you provide an example?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -56,6 +57,14 @@ private class InteropCallContext(
 
     fun IrType.isStoredInMemoryDirectly(): Boolean =
             isPrimitiveType() || isUnsigned() || isVector()
+
+    fun IrType.isSupportedReference(): Boolean = isObjCObjectType()
+            || isNullableString()
```
SvyatoslavScherbina, line 62: Why nullable?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -56,6 +57,14 @@ private class InteropCallContext(
...
+            || classOrNull == symbols.list
+            || classOrNull == symbols.mutableList
+            || classOrNull == symbols.set
+            || classOrNull == symbols.map
+            || isFunction()
```
SvyatoslavScherbina, line 67: The entire logic of determining and handling types in this file becomes similar to what `CBridgeGen` does. Is it possible to reuse some of the code?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -151,6 +164,7 @@ private fun InteropCallContext.castPrimitiveIfNeeded(
 }
 
 private fun InteropCallContext.convertEnumToIntegral(enumValue: IrExpression): IrExpression {
+    if (enumValue.type.classOrNull == symbols.nothing) return enumValue
```
SvyatoslavScherbina, line 167: This is fragile:
```c
enum E {
    ZERO
};

struct S {
    enum E e;
};
```

```kotlin
fun main() {
    foo(E.ZERO)
}

fun <T : E> foo(e: T) {
    val s = nativeHeap.alloc<S>()
    s.e = e
}
```
-> NPE below.

The compiler should always rely on the type of (setter) parameter, not the type of argument.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
sbogolepov: Because we generate field of nullable type for `NSString*`. 
Oh, I see that there is a `IrType.isStringClassType()` extension which is more flexible.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
sbogolepov: It is reachable via `writeEnumValueToMemory`. I see that it is simpler to pass  `IrType.getCEnumPrimitiveType()` directly to `writeValueToMemory`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
sbogolepov: 6d13354955feb7f34a091ba8f9ef36b296aef685

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
SvyatoslavScherbina: > Because we generate field of nullable type for `NSString*`.

Which is probably not true if the field is declared as non-nullable:
```obj-c
NSString* _Nonnull s;
```

sbogolepov: Yeah, right. Anyway, this code is superseded by 6d13354.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
```diff
@@ -121,20 +88,12 @@ private fun InteropCallContext.writeValueToMemory(
...
-        type
-    }
-}
-
 private fun InteropCallContext.castPrimitiveIfNeeded(
         value: IrExpression,
         targetType: IrType
 ): IrExpression {
     val valueClass = value.type.classOrNull!!
+    if (valueClass === symbols.nothing) return value
```
SvyatoslavScherbina, line 96: Same here:
```c
struct S {
    int x;
};
```

```kotlin
fun main() {
    foo(42)
}

fun <T : Int> foo(x: T) {
    val s = nativeHeap.alloc<S>()
    s.x = x
}
```

-> NPE above.

Please ensure you don't rely on type of IR expressions in this file.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropCallConvertors.kt
SvyatoslavScherbina: This case should be checked by tests too (same for https://github.com/JetBrains/kotlin-native/pull/4178#discussion_r427171368 and https://github.com/JetBrains/kotlin-native/pull/4178#discussion_r427854784).

PR #4181, opened 2020-05-20
Add kotlin.coroutines.cancellation.CancellationException to stdlib

SvyatoslavScherbina:
Note: first commit just copies the declaration from https://github.com/Kotlin/kotlinx.coroutines/blob/583ec6e862fb70c7fc0232af948a028bab3b20de/kotlinx-coroutines-core/native/src/Exceptions.kt#L13

### DIFF
--- a/runtime/src/main/kotlin/kotlin/coroutines/CancellationException.kt
+++ b/runtime/src/main/kotlin/kotlin/coroutines/CancellationException.kt
@@ -6,14 +6,12 @@
 package kotlin.coroutines.cancellation
 
 /**
- * Thrown by cancellable suspending functions if the [Job] of the coroutine is cancelled while it is suspending.
+ * Thrown by cancellable suspending functions if the coroutine is cancelled while it is suspending.
  * It indicates _normal_ cancellation of a coroutine.
- * **It is not printed to console/log by default uncaught exception handler**.
- * (see [CoroutineExceptionHandler]).
  */
-public actual open class CancellationException(
+public open class CancellationException(
         message: String?,
         cause: Throwable?
 ) : IllegalStateException(message, cause) {
-    public actual constructor(message: String?) : this(message, null)
+    public constructor(message: String?) : this(message, null)
 }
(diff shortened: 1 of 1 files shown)


runtime/src/main/kotlin/kotlin/coroutines/CancellationException.kt
```diff
@@ -0,0 +1,17 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.coroutines.cancellation
+
+/**
+ * Thrown by cancellable suspending functions if the coroutine is cancelled while it is suspending.
+ * It indicates _normal_ cancellation of a coroutine.
+ */
+public open class CancellationException(
```
ilya-g, line 12: For consistency please provide the following set of constructors:
https://github.com/JetBrains/kotlin/blob/b06780988df1e9cf207d3b663017138aff0e665e/libraries/stdlib/common/src/kotlin/ExceptionsH.kt#L118-L121

runtime/src/main/kotlin/kotlin/coroutines/CancellationException.kt
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.coroutines.cancellation
+
+/**
+ * Thrown by cancellable suspending functions if the coroutine is cancelled while it is suspending.
```
ilya-g, line 9: while it is suspended?

runtime/src/main/kotlin/kotlin/coroutines/CancellationException.kt
```diff
@@ -0,0 +1,17 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.coroutines.cancellation
+
+/**
+ * Thrown by cancellable suspending functions if the coroutine is cancelled while it is suspended.
+ * It indicates _normal_ cancellation of a coroutine.
+ */
+public open class CancellationException : IllegalStateException {
```
ilya-g, line 12: Also makes sense to annotate with `@SinceKotlin("1.4")`

SvyatoslavScherbina: Agree. Done.

PR #4183, opened 2020-05-21
Unicode support in Windows console: reading in UTF-16 and converting to UTF-8

PavelPunegov:


PavelPunegov:
This is still work in progress, as this is only a proof that it works in Console, while it may not work for UTF-8 input

PavelPunegov:
This is a fix for the https://youtrack.jetbrains.com/issue/KT-38736

PR #4184, opened 2020-05-21
Do not use Kotlin exceptions from MemorySharedRefs in objc context

projedi:


runtime/src/objc/cpp/ObjCExportClasses.mm
```diff
@@ -36,6 +36,9 @@ -(BOOL)_tryRetain;
 
 static void injectToRuntime();
 
+// Note: `KotlinBase`'s `toKotlin`, `retain` and `_tryRetain` methods will terminate if
```
SvyatoslavScherbina, line 39: Not entirely true for `retain`: it will fail only if back ref's `refCount` is zero, which actually shouldn't happen.

SvyatoslavScherbina: Same with `BackRefFromAssociatedObject::addRef`.

runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -126,15 +138,33 @@ void BackRefFromAssociatedObject::releaseRef() {
 }
 
 ObjHeader* BackRefFromAssociatedObject::ref() const {
-  ensureRefAccessible();
+  return konan::tryOrTerminate([this]() { return refOrThrow(); });
```
knebekaizer, line 141: This is better than before, but still suffers from bloating / duplication. Consider templating of entire call tree: ref, addRef, tryAddRef etc. See https://github.com/JetBrains/kotlin-native/tree/vi/ref-shared-holder-exceptions-templating for example

runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -102,12 +110,16 @@ void BackRefFromAssociatedObject::addRef() {
...
+
+bool BackRefFromAssociatedObject::tryAddRefOrThrow() {
   // Suboptimal but simple:
-  this->ensureRefAccessible();
+  this->ensureRefAccessibleOrThrow();
   ObjHeader* obj = this->obj_;
 
   if (!TryAddHeapRef(obj)) return false;
-  this->addRef();
+  this->addRefOrThrow();
```
knebekaizer, line 122: Ypu may need RAII here to balance TryAddHeapRef. However, looks like it can't throw actuall after `ensureRefAccessibleOrThrow` above. Anyway, consider to clarify the point with comment or RAII or (probably better) try to avoid duplicated check in ensureRefAccessibleOrThrow an then in addRefOrThrow

runtime/src/main/cpp/Porting.h
```diff
@@ -94,6 +99,23 @@ do {                          \
 } while(0)
 #endif
 
+template <typename F>
+// TODO: Use invoke_result without && when compiled with C++17
+ALWAYS_INLINE typename std::result_of<F&&()>::type tryOrTerminate(F&& tryClause) {
```
SvyatoslavScherbina, line 104: It might make more sense to put this function to `Exceptions.h`.
Motivation -- the body relies on at least two implementation details of that module:
1. Kotlin exceptions are C++ exceptions.
2. `std::terminate` prints stack trace of "current" Kotlin exception.

runtime/src/main/cpp/Porting.h
```diff
@@ -94,6 +99,23 @@ do {                          \
...
+#if KONAN_NO_EXCEPTIONS
+  // This will terminate the program if exception is thrown.
+  return tryClause();
+#else
+  try {
+    return tryClause();
+  } catch (...) {
```
SvyatoslavScherbina, line 111: Does this additional `catch` impose performance penalty?

runtime/src/main/cpp/MemorySharedRefs.cpp
projedi: Good idea. I replaced this with `RuntimeAssert` + (crashing) `addRef` for now, but it's still suboptimal. Better yet is to use a variant of `addRef` that doesn't have any checks at all. I'm still figuring out a nicer way of doing that, considering your comment above

runtime/src/main/cpp/Porting.h
projedi: The same goes for the `TRY_CATCH` defined above I suppose?

runtime/src/main/cpp/Porting.h
projedi: No measurable difference according to `objcinterop` benchmarks. However, I do worry about having an extra `try-catch` on a happy path.

runtime/src/main/cpp/Porting.h
SvyatoslavScherbina: I'm not sure about this: `TRY_CATCH` is probably used only for C++ exceptions.

runtime/src/main/cpp/Porting.h
SvyatoslavScherbina: > having an extra `try-catch` on a happy path.

Exactly! That's what I meant.

runtime/src/main/cpp/Porting.h
knebekaizer: I suggested above templated variant which is less bloated and does not imply extra try/catch, please consider.

runtime/src/main/cpp/Porting.h
```diff
@@ -94,6 +99,23 @@ do {                          \
 } while(0)
 #endif
 
+template <typename F>
+// TODO: Use invoke_result without && when compiled with C++17
```
knebekaizer, line 103: Can't catch this comment. For C++14 and above a simple `auto` return would be sufficient. If you mean "stay with c++11 but be ready for 14-17" you may use a form `auto + trailing return type` and say "TODO: remove trailing for c++14"
```
auto tryOrTerminate(F&& tryClause) -> typename std::result_of<F&&()>::type {
```
As for me, `auto` seems more readable, but it's only my personal preference.

projedi: To be honest, I don't like type inference in function type signatures. I use `auto` for local variables only. But then again, this is just my personal preference.

As for what I meant: `result_of` is deprecated in c++17 and removed in c++20, having been replaced with `invoke_result`.

runtime/src/main/cpp/Porting.h
projedi: > Clang and gcc follow zero-cost Itanium ABI

What about ABI on ARM? Does it do a similar zero-cost handling?

runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -47,21 +83,20 @@ void KRefSharedHolder::init(ObjHeader* obj) {
...
     return nullptr;
   }
+
   AdoptReferenceFromSharedVariable(obj_);
   return obj_;
 }
 
+template ObjHeader* KRefSharedHolder::ref<ErrorHandlingPolicy::kReturnDefault>() const;
+template ObjHeader* KRefSharedHolder::ref<ErrorHandlingPolicy::kThrowException>() const;
+template ObjHeader* KRefSharedHolder::ref<ErrorHandlingPolicy::kTerminate>() const;
```
knebekaizer, line 98: Is this explicit instantiation necessary?  See more below.

runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -33,6 +33,42 @@ RUNTIME_NORETURN inline void throwIllegalSharingException(ObjHeader* object) {
...
+template <ErrorHandlingPolicy errorHandlingPolicy>
+bool ensureRefAccessible(ObjHeader* object, ForeignRefContext context) {
+  if (errorHandlingPolicy == ErrorHandlingPolicy::kIgnore)
+    return true;
+
+  if (isForeignRefAccessible(object, context)) {
+    return true;
+  }
+
+  switch (errorHandlingPolicy) {
```
knebekaizer, line 60: Consider specialization instead of switch for template parameter:
```
template <ErrorHandlingPolicy>
bool ensureRefAccessible(ObjHeader* object, ForeignRefManager* context) {
    return isForeignRefAccessible(object, context);  // default behaviour
}
// specializations:
template <>
bool ensureRefAccessible<ErrorHandlingPolicy::kIgnore>(ObjHeader* object, ForeignRefManager* context) {
    return true;
}
template <>
bool ensureRefAccessible<ErrorHandlingPolicy::kThrowException>(ObjHeader* object, ForeignRefManager* context) {
    if (isForeignRefAccessible(object, context))) return true;
    throwIllegalSharingException(object);
}
// ...
```
Looks simpler and probably more idiomatic.

runtime/src/main/cpp/MemorySharedRefs.hpp
```diff
@@ -10,14 +10,32 @@
...
  public:
   void initLocal(ObjHeader* obj);
 
   void init(ObjHeader* obj);
 
+  // Error if called from the wrong worker with non-frozen obj_.
+  template <ErrorHandlingPolicy errorHandlingPolicy>
   ObjHeader* ref() const;
-  ObjHeader* refOrNull() const;
+  ObjHeader* refOrTerminate() const {
```
knebekaizer, line 30: What is the value of those wrappers? It is essentially a parameter value hardcoded to a function name. Looks like anti-pattern, isn't it? There are more below.

runtime/src/main/cpp/MemorySharedRefs.hpp
```diff
@@ -37,13 +53,40 @@ class BackRefFromAssociatedObject {
...
+  bool tryAddRefOrThrow() {
+    return tryAddRef<ErrorHandlingPolicy::kThrowException>();
+  }
 
   void releaseRef();
 
+  // Error if called from the wrong worker with non-frozen obj_.
+  template <ErrorHandlingPolicy errorHandlingPolicy>
   ObjHeader* ref() const;
+  ObjHeader* refOrTerminate() const {
```
knebekaizer, line 81: Again, new functions where parameterized one already exist. Just extra layer and mind bloating.
Is  `refOrThrow()` simpler than `ref<kThrow>()`? You may consider shorter names for parameter though.

runtime/src/main/cpp/MemorySharedRefs.cpp
projedi: Yep. Since the definition of `KRefSharedHolder::ref` is in this file, compilation of this file must produce all instantiations that might be used. Explicit instantiation makes sure this happens.

It is strange, however, that these instantiations are not generated by `KRefSharedHolder::refOr…` methods (possibly, because they are `inline`). But we're removing them anyway, so explicit instantiations are mandatory.

runtime/src/main/cpp/MemorySharedRefs.hpp
projedi: I personally find them more readable. But I'm not overly attached to them, so they can go once enum is renamed to be a tad shorter.

PR #4186, opened 2020-05-22
Correct docs of ThreadLocal and SharedImmutable annotations

ilya-g:
- clarify that they have effect only in K/N
- correct possible application targets of SharedImmutable

Relates to [KT-36245](https://youtrack.jetbrains.com/issue/KT-36245)


SvyatoslavScherbina:
Thanks!

PR #4187, opened 2020-05-24
Fix compilation problem by gradle version

1552980358:
Fix `Could not initialize class org.codehaus.groovy.runtime.InvokerHelper` on which JDK version >= 14


ilmat192:
@1552980358, We've update Gradle version to 6.5.1: https://github.com/JetBrains/kotlin-native/commit/9a4e8b83b2f7735f6c5d933c1c6dde53e249524f.

PR #4200, opened 2020-05-28
[c-interop][reverse][kt-36878] unifiyng generation of type handlers inlined and not inlined types

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -797,20 +798,14 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
...
         val set = mutableSetOf<ClassDescriptor>()
         defineUsedTypesImpl(scope, set)
-        // Add nullable primitives.
-        predefinedTypes.forEach {
-            val nullableIt = it.makeNullable()
+        val types = predefinedTypes + set
+                .map {  if (it.isInline) it.underlyingRepresentation()!!.type else it.defaultType }
+                .filter{isMappedToReference(it)}
+        val nullableTypes = types.map { it.makeNullable() }.toSet()
+        nullableTypes.forEach {
```
SvyatoslavScherbina, line 805: Does this work for 
```kotlin
inline class Foo(val value: Any?)
inline class Bar(val value: Foo)

fun takeNullableBar(nullableBar: Bar?) {}
```
?

vvlevchenko: I've added support inline and recursive inline classes for arguments and return types.

SvyatoslavScherbina: Please add the code above to the test suite.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -297,12 +296,12 @@ private class ExportedElement(val kind: ElementKind,
             original is ConstructorDescriptor ->
                 SignatureElement(uniqueName(original, shortName), original.constructedClass.defaultType)
             else ->
-                SignatureElement(uniqueName(original, shortName), original.returnType!!)
+                SignatureElement(uniqueName(original, shortName), original.returnType!!.noInlineUnderlyingRepresentation() ?: original.returnType!!)
```
SvyatoslavScherbina, line 299: What is the purpose of this patch? Does it change how types are mapped?

SvyatoslavScherbina: I see.
New mapping behaviour seems incorrect. For `takeNullableBar` from [the motivational test case](https://github.com/JetBrains/kotlin-native/pull/4200/commits/ac0a57c3f7df6cc3d7834bad2ade56ae5807d701#r434619545) it should produce `Bar` as before, not `Any`.

Mapping was ok, it is generation of used types that was broken.

vvlevchenko: Hm, I'm totally confused now. current (master) behaviour is: avoid type generation for inline classes. and my patch prevent appearing "unwanted" types in return and as parameter. Or I incorrectly understand the idea?

SvyatoslavScherbina:
To be moved to https://github.com/JetBrains/kotlin.

PR #4201, opened 2020-05-28
Lift experimental from string builder

qurbonzoda:


runtime/src/main/kotlin/kotlin/text/StringBuilder.kt
```diff
@@ -887,10 +884,14 @@ public inline fun StringBuilder.setLength(l: Int) = this.setLength(newLength = l
  * @throws IndexOutOfBoundsException or [IllegalArgumentException] when [start] or [end] is out of range of the [csq] character sequence indices or when `start > end`.
  * @throws IndexOutOfBoundsException if [index] is less than zero or greater than the length of this string builder.
  */
-@OptIn(ExperimentalStdlibApi::class)
+@Deprecated(
+        "Use insertRange(index: Int, csq: CharSequence?, start: Int, end: Int) instead",
+        ReplaceWith("insertRange(index, csq, start, end)"),
```
ilya-g, line 889: Correct the deprecation message and replacement:
```suggestion
        "Use insertRange(index: Int, csq: CharSequence, start: Int, end: Int) instead",
        ReplaceWith("insertRange(index, csq ?: "null", start, end)"),
```

PR #4207, opened 2020-06-02
Advance buildKotlinVersion to 1.4.0-dev-9619

ilya-g:
Use it to build shared tools.
Earlier, kotlinVersion was used, which caused shared and main project being built with different versions of the compiler.

The new version is required because of new `flatMap` overloads in stdlib, which can be resolved only with a compiler that supports such overloads.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,8 +15,8 @@
 #
 
 # A version of the Kotlin compiler that is used to build Kotlin/Native.
-buildKotlinVersion=1.4.0-dev-8620
-buildKotlinCompilerRepo=https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:Kotlin_KotlinPublic_Compiler),number:1.4.0-dev-8620,branch:default:any,pinned:true/artifacts/content/maven
+buildKotlinVersion=1.4.0-dev-9619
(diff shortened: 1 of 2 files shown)


vvlevchenko:
It's exactly the idea to compile with one version of the kotlin (build), but use the newest kotlin plugin to get up-to-date plugin's infrastructure (e.g. host manager). @ilmat192 please correct me if wrong.

ilya-g:
Do we need the newest kotlin plugin to build `shared`?

ilmat192:
> It's exactly the idea to compile with one version of the kotlin (build), but use the newest kotlin plugin to get up-to-date plugin's infrastructure (e.g. host manager).

Yes, generally this is the case. Only small correction: the host manager is not a part of plugin's infrastructure, it's a part of K/N target infrastructure. We need info about targets in our buildscripts, so we have to use the newest plugin.

> Do we need the newest kotlin plugin to build shared?

AFAIK no, `shared` can be built with an older plugin. But I think that it's better to keep the same plugin version in all K/N modules. 

Can't we just override the compiler version like we do for the main module:
```kotlin
dependencies {
    kotlinCompilerClasspath("org.jetbrains.kotlin:kotlin-compiler-embeddable:$buildKotlinVersion")
}
```
?

We had such overriding in `shared` earlier but it was lost with switching to Kotlin DSL.

ilmat192:
Discussed offline, decided to merge it as is and then add compiler overriding in a separate PR.

PR #4211, opened 2020-06-03
[IR][codegen] Moved objc allocs from codegen to InteropLowering

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1395,20 +1395,12 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-                    )
-                } else {
-                    val dstTypeInfo = functionGenerationContext.bitcast(kInt8Ptr, codegen.typeInfoValue(dstClass))
-                    callDirect(
-                            context.ir.symbols.throwClassCastException.owner,
-                            listOf(srcArg, dstTypeInfo),
-                            Lifetime.GLOBAL
-                    )
-                }
+                val dstTypeInfo = functionGenerationContext.bitcast(kInt8Ptr, codegen.typeInfoValue(dstClass))
```
knebekaizer, line 1398: `functionGenerationContext.` seems redundant here

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/CallGraphBuilder.kt
```diff
@@ -111,31 +109,6 @@ internal class CallGraphBuilder(val context: Context,
     private inline fun DataFlowIR.FunctionBody.forEachCallSite(block: (DataFlowIR.Node.Call) -> Unit) =
             nodes.forEach { node ->
                 when (node) {
-                    // TODO: OBJC-CONSTRUCTOR-CALL
-                    is DataFlowIR.Node.NewObject -> {
```
knebekaizer, line 115: Is `NewObject` still possible here? Otherwise consider assertion

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
homuroll: This was before my PR also, but, sure, I'll fix it.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/CallGraphBuilder.kt
homuroll: `NewObject` is a subtype of `Call`, before the PR it required special handling, now doesn't.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -1153,6 +1186,54 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
         }
     }
 
+    override fun visitTypeOperator(expression: IrTypeOperatorCall): IrExpression {
```
SvyatoslavScherbina, line 1189: What is handling Obj-C type checks here required for?

homuroll: To let analyses see some of the underlying magic. For instance, for escape analysis it's not clear why `is` operator might force argument to escape.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
```diff
@@ -479,6 +467,89 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
...
+                irClass.llvmSymbolOrigin,
+                independent = true // Protocol is header-only declaration.
+        )
+
+        return call(protocolGetter, emptyList())
+    }
+
+    private fun emitObjCInstanceOf(callSite: IrCall, args: List<LLVMValueRef>): LLVMValueRef {
+        val dstClass = callSite.getTypeArgument(0)?.getClass()
+                ?: error("Expected type argument")
```
SvyatoslavScherbina, line 496:
```kotlin
fun <T : kotlinx.cinterop.ObjCObject> foo(x: Any) {
    x as T
}
```

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -1153,6 +1186,54 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
...
+                val value = irTemporary(expression.argument)
+                +irObjCInstanceOf(value, expression.typeOperand)
+            }
+
+            IrTypeOperator.NOT_INSTANCEOF -> builder.at(expression).irBlock {
+                val value = irTemporary(expression.argument)
+                +irNot(irObjCInstanceOf(value, expression.typeOperand))
+            }
+
+            else -> expression
```
SvyatoslavScherbina, line 1222:
```kotlin
fun foo(x: Any) {
    x as? kotlinx.cinterop.ObjCObject
}
```

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
SvyatoslavScherbina: Previous implementation for Obj-C type checks was designed with avoiding code duplication in mind. See comments with reproducers for related bugs in new implementation.
So it might be easier to let analyses see the underlying magic.

homuroll: I thought that the more we can do in lowerings and not in codegen the better.

SvyatoslavScherbina: It depends. In this particular case it might be true, but avoiding code duplication (to support type operand erasure, safe casts etc.) is important too. So, it would be great to achieve both.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
SvyatoslavScherbina: Also please add some info to error message, to make the error debuggable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1322,23 +1322,25 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
+        IrTypeOperator.IMPLICIT_COERCION_TO_UNIT -> {
+            evaluateExpression(value.argument)
+            functionGenerationContext.theUnitInstanceRef.llvm
         }
+
+        IrTypeOperator.CAST,
+        IrTypeOperator.SAFE_CAST,
+        IrTypeOperator.INSTANCEOF,
+        IrTypeOperator.NOT_INSTANCEOF ->
+            error("Should've been lowered: ${value.dump()} ${functionGenerationContext.irFunction?.render()}")
```
SvyatoslavScherbina, line 1337: You can report the exact source location here too, see https://github.com/JetBrains/kotlin-native/blob/5cb1de5d84392f6d1bb836f299fc6fbc8bbe962d/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Reporting.kt#L34

runtime/src/main/kotlin/kotlin/native/internal/KClassImpl.kt
```diff
@@ -87,3 +87,6 @@ private external fun getRelativeName(typeInfo: NativePtr): String?
 
 @SymbolName("Kotlin_TypeInfo_isInstance")
 private external fun isInstance(obj: Any, typeInfo: NativePtr): Boolean
+
+@TypedIntrinsic(IntrinsicType.INSTANCE_OF)
+internal external inline fun <reified T : Any> instanceOf(obj: Any): Boolean
```
SvyatoslavScherbina, line 92: Placing this intrinsic to `Intrinsics.kt` would probably make more sense.
Also please add `@PublishedApi`, we follow convention of using it to indicate that `internal` declaration can be used from foreign module (which is the case here).

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/ObjectiveCImpl.kt
```diff
@@ -132,6 +132,10 @@ internal annotation class ObjCMethodImp(val selector: String, val encoding: Stri
 @TypedIntrinsic(IntrinsicType.OBJC_GET_SELECTOR)
 internal external fun objCGetSelector(selector: String): COpaquePointer
 
+@PublishedApi
+@TypedIntrinsic(IntrinsicType.OBJC_INSTANCE_OF)
+internal external inline fun <reified T : ObjCObjectBase> objCInstanceOf(obj: Any): Boolean
```
SvyatoslavScherbina, line 137:
```suggestion
internal external inline fun <reified T : ObjCObject> objCInstanceOf(obj: Any): Boolean
```
(would probably be more correct here)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TypeOperatorLowering.kt
```diff
@@ -75,6 +63,34 @@ private class TypeOperatorTransformer(val context: CommonBackendContext, val fun
         }
     }
 
+    fun IrBuilderWithScope.irCast(argument: IrExpression, type: IrType, argumentIsNotNull: Boolean = false) = irBlock {
+        val value = irTemporary(argument)
+        +irIfThenElse(
+                type,
+                condition = irInstanceOf(value, type, argumentIsNotNull),
+                thenPart = irImplicitCast(irGet(value), type),
+                elsePart = irCall(symbols.ThrowTypeCastException)
```
SvyatoslavScherbina, line 72: If I understand correctly, previous implementation of casts was throwing an exception with proper message in some cases. This implementation doesn't seem to do so.

If we don't have a test for the message, please add it.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TypeOperatorLowering.kt
```diff
@@ -75,6 +63,34 @@ private class TypeOperatorTransformer(val context: CommonBackendContext, val fun
         }
     }
 
+    fun IrBuilderWithScope.irCast(argument: IrExpression, type: IrType, argumentIsNotNull: Boolean = false) = irBlock {
+        val value = irTemporary(argument)
+        +irIfThenElse(
+                type,
+                condition = irInstanceOf(value, type, argumentIsNotNull),
+                thenPart = irImplicitCast(irGet(value), type),
```
SvyatoslavScherbina, line 71: Note that implicit cast is handled by autoboxing through `Any?` type, it can box argument and unbox result. Please make sure this new cast implementation doesn't affect performance.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TypeOperatorLowering.kt
```diff
@@ -96,43 +112,52 @@ private class TypeOperatorTransformer(val context: CommonBackendContext, val fun
...
                 }
             }
 
-            typeOperand.isSimpleTypeWithQuestionMark -> builder.irAs(expression.argument, typeOperand.makeNotNull())
+            typeOperand.isSimpleTypeWithQuestionMark -> builder.irCast(expression.argument, typeOperand.makeNotNull())
 
-            typeOperand == expression.typeOperand -> expression
-
-            else -> builder.irAs(expression.argument, typeOperand)
+            else -> builder.irCast(expression.argument, typeOperand)
```
SvyatoslavScherbina, line 126: Can we add `argumentIsNotNull = true` here and/or above?

PR #4213, opened 2020-06-03
Test for illegal sharing with objc weaks

projedi:
Tests #4126


backend.native/tests/interop/objc/illegal_sharing_with_weak/objclib.m
```diff
@@ -0,0 +1,11 @@
...
+
+static NSObject* __weak globalObject = nil;
+
+void setObject(NSObject* obj) {
+  globalObject = obj;
+}
+
+bool isObjectAliveShouldCrash() {
+  return globalObject != nil;
+}
```
SvyatoslavScherbina, line 11: It might be enough to put these declarations to `.h`, making building `.dylib` unnecessary.

backend.native/tests/build.gradle
```diff
@@ -3880,6 +3884,37 @@ if (PlatformInfo.isAppleTarget(project)) {
...
+                args "-lobjc", '-fobjc-arc'
+                args '-fPIC', '-shared', '-o', "$buildDir/libobjc_illegal_sharing_with_weak.dylib"
+                // Enable ARC optimizations to prevent some objects from leaking in Obj-C code due to exceptions:
+                args '-O2'
+            }
+            if (project.target instanceof KonanTarget.IOS_X64) {
+                UtilsKt.codesign(project, "$buildDir/libobjc_illegal_sharing_with_weak.dylib")
+            }
+
+            copy {
```
SvyatoslavScherbina, line 3912: Doesn't seem required.

backend.native/tests/build.gradle
```diff
@@ -3880,6 +3884,37 @@ if (PlatformInfo.isAppleTarget(project)) {
...
+
+    interopTest("interop_objc_illegal_sharing_with_weak") {
+        source = "interop/objc/illegal_sharing_with_weak/main.kt"
+        interop = 'objc_illegal_sharing_with_weak'
+
+        expectedExitStatusChecker = { it != 0 }
+        outputChecker = {
+            it.startsWith("Before") &&                   // Should crash after "Before"
+                !it.contains("After") &&                 // But before "After"
+                it.contains("isObjectAliveShouldCrash")  // And should contain stack trace.
```
SvyatoslavScherbina, line 3896: Btw, does this work with `-opt`?

runtime/src/main/cpp/ObjCInterop.mm
```diff
@@ -102,17 +102,14 @@ BOOL _tryRetainImp(id self, SEL _cmd) {
...
-                  format:@"Possible illegal attempt to access weak property from non-owning thread"];
-    } @catch (...) {
-      objc_terminate();
-    }
+    // Cannot use SourceInfo here, because CoreSymbolication framework (CSSymbolOwnerGetSymbolWithAddress)
+    // fails at recursive retain lock. Similarly, cannot use objc exception here, because its unhandled
+    // exception handler might fail at recursive retain lock too.
+    // TODO: Refactor to be more explicit. Instead of relying on an unhandled exception termination
+    // (and effectively setting a global to alter its behavior), just call an appropriate termination
+    // function by hand.
```
SvyatoslavScherbina, line 110: ➕

backend.native/tests/interop/objc/illegal_sharing_with_weak/objclib.m
projedi: That simplified things indeed. Thanks!

backend.native/tests/build.gradle
projedi: Marked the function as `noinline`. Now it definitely should.

runtime/src/main/cpp/Exceptions.h
```diff
@@ -67,4 +67,23 @@ void PrintThrowable(KRef);
...
+  ScopedDisallowSourceInfo(ScopedDisallowSourceInfo&&) = delete;
+  ScopedDisallowSourceInfo& operator=(const ScopedDisallowSourceInfo&) = delete;
+  ScopedDisallowSourceInfo& operator=(ScopedDisallowSourceInfo&&) = delete;
+
+  static bool IsActive();
+
+ private:
+  // Making a counter instead of a boolean, if this lock ever gets nested.
+  // Shouldn't have any significant performance penalty.
+  static THREAD_LOCAL_VARIABLE int_fast8_t activeCount;
```
SvyatoslavScherbina, line 86: Btw, what is the motivation behind using `int_fast8_t`? Is it an optimization?

projedi: > Is it an optimization?

Essentially, yes. I wanted to use the fastest integer available and found the `int_fast*_t` family, which does exactly that. I don't think there's any harm in using those (except an element of surprise on the first encounter).

Also, with [Apple libc](https://opensource.apple.com/source/Libc/Libc-1353.41.1/include/stdint.h.auto.html) they seem to map to a corresponding `int*_t`, but on [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/stdint.h;h=2df07e485b402f9328c51e6cd0cabb0cf5afd6c1;hb=HEAD#l55) they do depend on a machine word size.

SvyatoslavScherbina: I see. Well, I don't think this is worth optimizing :)

projedi: I agree. The cost of doing so isn't very big, though.

knebekaizer: If you only need the fastest counter then plain `int` may be faster (and more universal) than any int8 variant.

projedi: Technically, `int_fastx_t` is the fastest integer of size at least `x`. So, it may as well be `int` even for 8 byte variant. And on [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/stdint.h;h=2df07e485b402f9328c51e6cd0cabb0cf5afd6c1;hb=HEAD#l55) that's pretty much what's happening for 16 and 32 bit variants. For 8 bit, however, it uses `signed char` and I have no idea why.

projedi:
> The intention (my guess, that's it) is to modify behavior of terminate handler, or (narrowing) just `Kotlin_getSourceInfo` which may be called from terminate handler. Should we express the intention clearly, in a direct way?

> However, the best way may be to create custom terminate function instead of `std::terminate`

This is precisely what I want to end up with. I left a [TODO](https://github.com/JetBrains/kotlin-native/pull/4213/files/9000bf892a0723eafbfcaf31c9a4783386b3e88e#diff-a43175483524cdd8f2def29406309f78R108) about it. I just wanted to do it after this PR (and possibly https://github.com/JetBrains/kotlin-native/pull/4184 ) are merged.

> The name is ScopedXXX but destructor seems to be never called, as the control flow does not leave syntax scope after terminate

In the original version I had a function that globally disabled SourceInfo, but ultimately decided against it, because the current version of the hack allows a quickfix to be put anywhere until we refactor the code to make it unneeded.

knebekaizer:
> In the original version I had a function that globally disabled SourceInfo, but ultimately decided against it, because the current version of the hack allows a quickfix to be put anywhere until we refactor the code to make it unneeded.

Yes, and this is the concern: you still globally disable GetSourceInfo, but this fact is hidden in a 17-lines class. At the call site it looks like local thread object, but in fact it is not. Please make the intention and behavior clear.
* Destructor is never called, so `Scoped...` name is mispointing
* You don't need a counter because it's never decreased. Essentially it is a flag related to the _caller context_ (it means:  I gonna terminate but it is unsafe to use SourceInfo in the current context). This is a _caller_ context, so that no callee may modify it. The flag is set once and never cleared. It may be an empty singleton object or "set only" global flag variable.  As a bonus, the same singleton flag may be used to block any possible concurrent termination flow and prevent interleaving.

SvyatoslavScherbina:
> At the call site it looks like local thread object, but in fact it is not.

Why so? Could you clarify?

runtime/src/main/cpp/Exceptions.cpp
```diff
@@ -103,6 +103,12 @@ _Unwind_Reason_Code unwindCallback(
 }
 #endif
 
+SourceInfo GetSourceInfo(KConstRef stackTrace, int index) {
```
knebekaizer, line 106: Prefer `lowCamelCase` for private function. `UpperCamelCase` to share C name with kotlin (aka internal interop)

PR #4218, opened 2020-06-04
Update docs for `@Throws suspend fun`

SvyatoslavScherbina:


OBJC_INTEROP.md
```diff
@@ -133,15 +133,20 @@ Swift has only checked errors. So if Swift or Objective-C code calls a Kotlin me
...
-and as `throws` methods in Swift.
+and as `throws` methods in Swift. Representations for `suspend` functions always have
+`NSError*`/`Error` parameter in completion handler.
 
 When Kotlin function called from Swift/Objective-C code throws an exception
 which is an instance of one of the `@Throws`-specified classes or their subclasses,
 it is propagated as `NSError`. Other Kotlin exceptions reaching Swift/Objective-C
 are considered unhandled and cause program termination.
 
+In the absence of `@Throws`, `suspend` functions propagate only
```
p7nov, line 146:
```suggestion
`suspend` functions without `@Throws` propagate only
```

SvyatoslavScherbina: Do you find this more readable?

p7nov: Yes, especially for non-native speakers. **Function without .. annotation** contains only words that any developer knows; **in the absence** sounds heavier and may not be familiar to them.

SvyatoslavScherbina: I see. Ok.

Should we use the same simplified wording for the annotation [comment](https://github.com/JetBrains/kotlin-native/pull/4218/files#diff-aecbddbc81c0e1ad924e886c998bcb0fR167-R170)? cc @ilya-g

PR #4219, opened 2020-06-04
[Interop] Use m128 to declare simd type for Windows target and MSVC

knebekaizer:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -848,7 +848,17 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
...
-        output("typedef float __attribute__ ((__vector_size__ (16))) ${prefix}_KVector128;")
+
+        if (context.config.target.family == Family.MINGW) {
+            output("#ifndef _MSC_VER")
+            output("typedef float __attribute__ ((__vector_size__ (16))) ${prefix}_KVector128;")
+            output("#else")
+            output("typedef __m128 ${prefix}_KVector128;")
+            output("#endif")
+        } else {
+            output("typedef float __attribute__ ((__vector_size__ (16))) ${prefix}_KVector128;")
```
SvyatoslavScherbina, line 859: Maybe create a variable with this string to avoid code duplication?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
```diff
@@ -848,7 +848,17 @@ internal class CAdapterGenerator(val context: Context) : DeclarationDescriptorVi
         output("typedef unsigned long long ${prefix}_KULong;")
         output("typedef float              ${prefix}_KFloat;")
         output("typedef double             ${prefix}_KDouble;")
-        output("typedef float __attribute__ ((__vector_size__ (16))) ${prefix}_KVector128;")
+
+        if (context.config.target.family == Family.MINGW) {
+            output("#ifndef _MSC_VER")
+            output("typedef float __attribute__ ((__vector_size__ (16))) ${prefix}_KVector128;")
+            output("#else")
+            output("typedef __m128 ${prefix}_KVector128;")
```
SvyatoslavScherbina, line 856: Does using `__m128` type require a header to be included?
According to the documentation, it should be `xmmintrin.h`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt
knebekaizer: @SvyatoslavScherbina  I thought on that too but stalled at the question: where to put it to share with both cinterop and kotlinc. Any ideas?

SvyatoslavScherbina: I mean code duplication with the line above: https://github.com/JetBrains/kotlin-native/blob/0ef167b50481b31bdf7f52aa7eabb716485737f0/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CAdapterGenerator.kt#L854

PR #4222, opened 2020-06-04
Add metadata comparison tests

PavelPunegov:


backend.native/tests/build.gradle
```diff
@@ -240,6 +240,7 @@ run {
     dependsOn(tasksOf(KonanTest))
     // Add framework tests
     dependsOn(tasksOf(FrameworkTest))
+    dependsOn(tasksOf(MetadataComparisonTest))
```
SvyatoslavScherbina, line 243: We might need to add this test to sanity.
Also there are probably other test kinds that don't run with `run` or `sanity`. For example, `CoverageTest`.

PavelPunegov: I'm looking through all our tests to find out what else isn't run in a `run {}` and in `sanity {}` but could have been.

SvyatoslavScherbina:
@PavelPunegov Do you plan to [include](https://github.com/JetBrains/kotlin-native/pull/4222#discussion_r435343037) coverage (and maybe other) tests with a separate PR?

PavelPunegov:
> @PavelPunegov Do you plan to [include](https://github.com/JetBrains/kotlin-native/pull/4222#discussion_r435343037) coverage (and maybe other) tests with a separate PR?

No, I've just find out that you have merged your PR. I am going to add it here then.

PR #4241, opened 2020-06-18
KT-39670: -Xdebug-prefix-map for path remapping

mpetrov:
This introduces a new `-Xdebug-prefix-map` flag. When compiling files
with directories prefixed by the map key, the prefix will be changed to
the map value. This allows for reproducable builds by replacing
absolute paths with relative ones.

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
@@ -37,11 +37,13 @@ internal val contextLLVMSetupPhase = makeKonanModuleOpPhase(
 
             // we don't split path to filename and directory to provide enough level uniquely for dsymutil to avoid symbol
             // clashing, which happens on linking with libraries produced from intercepting sources.
+            val filePath = context.config.outputFile.toFileAndFolder(context).path()
+
             context.debugInfo.compilationUnit = if (context.shouldContainLocationDebugInfo()) DICreateCompilationUnit(
                     builder = context.debugInfo.builder,
                     lang = DWARF.language(context.config),
-                    File = File(context.config.outputFile).absolutePath,
-                    dir = "-",
+                    File = filePath,
+                    dir = "",
                     producer = DWARF.producer,
                     isOptimized = 0,
                     flags = "",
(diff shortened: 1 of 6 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BitcodePhases.kt
```diff
@@ -37,11 +37,13 @@ internal val contextLLVMSetupPhase = makeKonanModuleOpPhase(
...
             // clashing, which happens on linking with libraries produced from intercepting sources.
+            val filePath = context.config.outputFile.toFileAndFolder(context).path()
+
             context.debugInfo.compilationUnit = if (context.shouldContainLocationDebugInfo()) DICreateCompilationUnit(
                     builder = context.debugInfo.builder,
                     lang = DWARF.language(context.config),
-                    File = File(context.config.outputFile).absolutePath,
-                    dir = "-",
+                    File = filePath,
+                    dir = "",
```
vvlevchenko, line 46: It'd suggest to drop prefixing on CompilationUnit generation. path stored in CU isn't involved in source location resolving directly, but impact on identification of CU. it's like path of object file (.o), additionally it can hint in issue investigations e.g. where file were produced and so on.

mpetrov: Thanks for looking at this, I honestly don't have too much insight into the compilation unit naming.
To make sure I'm parsing your comment right: is the suggestion to revert these two lines and go back to using `absolutePath`?

If so my concern would be that the build would not be reproducible. e.g. If you were to build on teamcity you'd get something like `/Users/teamcity/buildAgent/work/4d622a065c544371/compilationUnit.o` as the file name. If I were to run the same build with identical inputs on a different build agent instance I'd get a change in path, and checksums.

Happy to also split this off into a follow-up PR if you'd like to keep the scope of this change small. Finding sources was the most immediate concern, but we do use bazel so reproducibility is something I'd like to have handled.

vvlevchenko: > If so my concern would be that the build would not be reproducible. e.g. If you were to build on teamcity you'd get something like ...

vvlevchenko: > If so my concern would be that the build would not be reproducible. e.g. If you were to build on teamcity you'd get something like ...

Ok, this make sense. I'll add tests for this option and I haven't objections to apply this PR

vvlevchenko:
applied together with tests in #4261. thank you for contribution.

PR #4242, opened 2020-06-18
Ignored caches for optimized compilation

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/CacheSupport.kt
```diff
@@ -49,8 +50,13 @@ class CacheSupport(
             library to cachePath
         }
 
+        val optimized = configuration.getBoolean(KonanConfigKeys.OPTIMIZATION)
+        if (optimized && (explicitCacheFiles.isNotEmpty() || implicitCacheDirectories.isNotEmpty()))
+            configuration.report(CompilerMessageSeverity.WARNING, "Cached libraries will not be used for optimized compilation")
+
         CachedLibraries(
                 target = target,
+                optimized = optimized,
```
SvyatoslavScherbina, line 59: Why not just pass empty `explicitCacheFiles` and `implicitCacheDirectories` if `optimized`?
`CachedLibraries` doesn't seem the right place to make it depend on `optimized`.

PR #4256, opened 2020-06-25
Sync docs and behavior of decapitalize and capitalize

ilya-g:


ilmat192:
Can't this patch be merged to `master` (and `1.4.0`/`1.4-M3`) without a compiler update?

ilya-g:
Yes, it can be merged to the other branches as well without a compiler update.
I just prefer putting commits in `kotlin-mirror/master` branch when they are a part of a change that affects both `kotlin` and `kotlin-native` repos.

PR #4257, opened 2020-06-25
Add changelog for 1.4-M3

ilmat192:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# v1.4.0-M3 (Jun 2020)
+  * Tune GC to improve execution time performance ([`KT-19076`](https://youtrack.jetbrains.com/issue/KT-19076)).
+  * Support @Throws annotation ([`KT-34848`](https://youtrack.jetbrains.com/issue/KT-34848))
```
SvyatoslavScherbina, line 3: K/N has already supported `@Throws` annotation. KT-34848 adds `annotation class Throws` to common. I'm not sure this belongs to Native changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+  * Fix support of variadic block types in ObjC interop ([`KT-36766`](https://youtrack.jetbrains.com/issue/KT-36766))
```
SvyatoslavScherbina, line 4: It's nothing like "fix support", just "handle": variadic block types are still unsupported, but at least don't cause `cinterop` crash.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+  * Support `suspend fun` in ObjC/Swift interop.
```
SvyatoslavScherbina, line 5: 1. This support was introduced in 1.4-M2.
2. The wording is kinda too wide: it can be misread as "now async Obj-C methods are imported as Kotlin `suspend fun`", which is incorrect.

CHANGELOG.md
ilmat192: Ok, removed.

CHANGELOG.md
ilmat192: Fixed

CHANGELOG.md
ilmat192: Ok, I've removed it from the M3 changelog. @PavelPunegov, please take into account the second point when you'll work on the M2 changelog.

PR #4258, opened 2020-06-25
Bridges to nothing

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -96,22 +94,36 @@ internal fun IrFunction.hasReferenceAt(index: Int): Boolean {
...
-internal enum class BridgeDirection {
-    NOT_NEEDED,
-    FROM_VALUE_TYPE,
-    TO_VALUE_TYPE
+internal sealed class BridgeDirection {
+    object NOT_NEEDED : BridgeDirection()
+    object FROM_VALUE_TYPE : BridgeDirection()
+    object TO_VALUE_TYPE : BridgeDirection()
+    object TO_NULL : BridgeDirection()
+    data class TO_NOTHING(val fromBinaryType: PrimitiveBinaryType?) : BridgeDirection()
```
SvyatoslavScherbina, line 108: Do we really need all these kinds?
We already have generic machinery that supports arbitrary upcasting and downcasting.
So why is it required to distinguish all these cases in bridge generator, isn't it enough to have only "needed" and "not needed"?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -112,20 +113,31 @@ internal class SpecialDeclarationsFactory(val context: Context) {
...
     }
 
-    private fun createBridge(function: IrSimpleFunction,
-                             bridgeDirections: BridgeDirections) = WrappedSimpleFunctionDescriptor().let { descriptor ->
+    private fun createBridge(key: BridgeKey) = WrappedSimpleFunctionDescriptor().let { descriptor ->
+        val (function, bridgeDirections) = key
         val startOffset = function.startOffset
         val endOffset = function.endOffset
-        val returnType = when (bridgeDirections.array[0]) {
+        val returnType = when (val returnBridgeDirection = bridgeDirections.array[0]) {
```
SvyatoslavScherbina, line 124: How is this essentially different from `function.returnType`? Could you provide and example?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
homuroll: The original idea was that a bridge could be described by a pair `<function, bridgeDirections>`, and that one bridge could be used to override more than one function. Note that `bridgeDirections` is used as a key in the vtable generator, so I tried not to touch that. We can discuss the refactoring of the whole bridge/vtable system but I think it is a separate thing and IMO not for 1.4-M3.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
homuroll: Here `function` is not the function being overridden, but the target to call. So exactly for targets returning `Nothing` or `Nothing?` return type is different.

SvyatoslavScherbina: Ah, I see. Then:
How is this essentially different from `overriddenFunction.returnType`? Could you provide an example?
Do I understand correctly that `overriddenFunction` is not currently available here, and this is the only blocker?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
SvyatoslavScherbina: > and IMO not for 1.4-M3.

This entire PR is not for 1.4-M3 or 1.4 anyway.
I find this classification of bridge directions too cumbersome, I believe it is possible to make it much simpler without refactoring the whole system.

homuroll: Ok, if it is not intended to be fixed in 1.4, then, sure, I'll try to simplify things a bit.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
homuroll: Yes, I think, `overriddenFunction.returnType` will do, I guess, it is not available here for the same reason - I wanted a bridge to be built only by its target and directions.

SvyatoslavScherbina: Of course.
But it might be possible to include the type (or information to construct it) to `BridgeDirection`.
This might involve using result of https://github.com/JetBrains/kotlin-native/blob/7a24e9ea0ecb7a8f758ae13fa8f3186700c7bfe9/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt#L27 for example.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -78,55 +74,93 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
+    }
+}
+
+internal fun IrType.isNullableNothing() =
+        isNullable() && classifierOrNull?.isClassWithFqName(KotlinBuiltIns.FQ_NAMES.nothing) == true
+
+private fun IrFunction.bridgeDirectionToAt(overriddenFunction: IrFunction, index: Int): BridgeDirection {
+    val irClass = overriddenFunction.typeAt(index)?.erasure()
+    return when {
+        index == 0 && returnType.isNothing() && !overriddenFunction.returnType.isNothing() ->
```
SvyatoslavScherbina, line 127: Is this correct for `returnType == Nothing, overriddenFunction.returnType == Any`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -78,55 +74,93 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
+
+        hasValueTypeAt(index) && overriddenFunction.hasReferenceAt(index) ->
+            // Erase to [Any?].
+            BridgeDirection(null, BridgeDirectionKind.BOX)
+
+        hasReferenceAt(index) && overriddenFunction.hasValueTypeAt(index) ->
+            BridgeDirection(irClass, BridgeDirectionKind.UNBOX)
+
+        else -> BridgeDirection.NONE
+    }
```
SvyatoslavScherbina, line 142: Why not reuse the logic from `getTypeConversion`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
```diff
@@ -249,7 +249,13 @@ private fun Context.buildBridge(startOffset: Int, endOffset: Int,
             }
         }
 
-        +irReturn(delegatingCall)
+        if (bridge.returnType.isNullableNothing()
```
SvyatoslavScherbina, line 252: What is this required for? Does `+irReturn(delegatingCall)` work in this case?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -78,55 +74,93 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
+}
+
+internal data class BridgeDirection(val irClass: IrClass?, val kind: BridgeDirectionKind) {
+    companion object {
+        val NONE = BridgeDirection(null, BridgeDirectionKind.NONE)
+        const val RETURN_INDEX = 0
+        const val DISPATCH_RECEIVER_INDEX = 1
+        const val EXTENSION_RECEIVER_INDEX = 2
+        fun mapParameterIndex(index: Int) = index + 3
+        fun unmapParameterIndex(index: Int) = index - 3
```
SvyatoslavScherbina, line 117: Please consider using `inline class` for indices.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
homuroll: Why not? Are you implying that there is no need in a bridge for this case?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
homuroll: I don't see any handling of `Nothing` there.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
homuroll: Indeed it is not needed, I didn't expect `unbox` to correctly work on `null` :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
SvyatoslavScherbina: Well, now that you say it, I realize that it might be required because of special handling for `Nothing` return type. This indeed complicates everything. I need some time to rethink it then :)

SvyatoslavScherbina: I guess we should keep checking `returnType.isNothing || returnType.isUnit` conditions in the same place, e.g. in `DataLayout.kt`. The remaining logic is probably equal to boxing.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -75,55 +74,120 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
         val result = StringBuilder()
         array.forEach {
-            result.append(when (it) {
-                BridgeDirection.FROM_VALUE_TYPE -> 'U' // unbox
-                BridgeDirection.TO_VALUE_TYPE   -> 'B' // box
-                BridgeDirection.NOT_NEEDED      -> 'N' // none
+            result.append(when (it.kind) {
+                BridgeDirectionKind.BOX -> 'B'
+                BridgeDirectionKind.UNBOX -> 'U'
+                BridgeDirectionKind.NONE -> 'N'
```
SvyatoslavScherbina, line 190: Previous implementation kinda guaranteed that we don't get clashing names for different bridges in the same class. I'm not sure about current one.

homuroll: There is no difference, I just renamed enum values.

homuroll: The only potential clash that I see is for bridges returning `Nothing`, but for those the return type will be different for each bridge, so still no real clash.

SvyatoslavScherbina: > I just renamed enum values.

Before: if `bridgeDirection1.toString() == bridgeDirection2.toString()`, then `bridgeDirection1 == bridgeDirection2`.
After: not true anymore.

SvyatoslavScherbina: On the other hand, I can't find potential clashes in this version, because binary function signatures include a lot of information.

homuroll: Ah, I got it. There is a difference indeed. Still, the difference is in parameter types, which is included in the signature.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -75,55 +75,127 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
+    }
+}
+
+private fun IrType.isNullableNothing() =
+        isNullable() && classifierOrNull?.isClassWithFqName(KotlinBuiltIns.FQ_NAMES.nothing) == true
+
+private fun IrFunction.bridgeDirectionToAt(overriddenFunction: IrFunction, index: ParameterIndex): BridgeDirection {
+    val irClass = overriddenFunction.typeKindAt(index).irType?.erasure()
+    return when {
+        index == ParameterIndex.RETURN_INDEX && returnType.isNothing() && !overriddenFunction.returnType.isNothing() ->
```
SvyatoslavScherbina, line 152: Still `isNothing` and other cases checked here. Is this required?

homuroll: These are special cases this entire PR was originally created for. I don't see how to move them to, say, `TypeKind`, because they depend on both types (`this` and `overriddenFunction`). The only alternative would be to add `TypeKind.Nothing`, `TypeKind.NullableNothing` and `TypeKind.Pointer`. What do you think, will it be better? I myself think either one will do.

SvyatoslavScherbina: I still don't find handling `Nothing` here necessary.
For example, if you have `Nothing` vs `Any` types, then `TypeKind`s would be different, and the generic machinery would decide that a boxing/unboxing is required. Do you see any issues with this?

homuroll: Ok, this one is true, but what about `Nothing` vs `Int`? They both will become `TypeKind.ValueType`, but a bridge is needed.

homuroll: Here is the table for return slot:
```
return:
    Nothing, Nothing -> none
    Nothing, [-Nothing] -> unbox
    Nothing?, Pointer -> unbox
    Unit, Unit -> none
    Unit, * -> box
    [-Nothing, -Unit], ValueType -> unbox
    ValueType, ValueType -> none
    ValueType, Ref -> box
```
Probably it will look ugly however we implement it.

homuroll: Hmm, now that I am looking at it, looks like the row `Nothing?, Pointer -> unbox` can be removed..

homuroll: Yeah, I think, I've got it.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt
```diff
@@ -75,55 +74,113 @@ internal val IrClass.isArrayWithFixedSizeItems: Boolean
...
+    }
+}
+
+private fun IrFunction.bridgeDirectionToAt(overriddenFunction: IrFunction, index: ParameterIndex): BridgeDirection {
+    val kind = typeWithKindAt(index).kind
+    val (irClass, otherKind) = overriddenFunction.typeWithKindAt(index)
+    if ((kind == TypeKind.VOID || kind == TypeKind.REFERENCE) && otherKind != kind)
+        return BridgeDirection(irClass?.erasure(), BridgeDirectionKind.UNBOX)
+    if (kind == TypeKind.VALUE_TYPE && otherKind == TypeKind.REFERENCE)
+        return BridgeDirection(null /* Erase to [Any?] */, BridgeDirectionKind.BOX)
```
SvyatoslavScherbina, line 151: I'm not sure that `kind == TypeKind.VALUE_TYPE && otherKind == TypeKind.VOID` case is impossible.

Consider the following example:
```kotlin
open class Base1<T : Int> {
    fun foo(): T { 
        TODO()
    }
}

interface Base2 {
    fun foo(): Nothing
}

class Derived : Base1<Nothing>(), Base2 {
}
```

It requires a bridge and probably meets the condition above.

homuroll: Indeed.

PR #4259, opened 2020-06-26
IEEE754: hypot with NaN & Inf and unary minus with floating point operand

PavelPunegov:
1. According to the https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.math/hypot.html it should return `+Inf` if any of the operand is `Inf`, and `NaN` if any is `NaN`.
Tests show that on some platforms either 1-st or 2-nd case doesn't show the desired result as we use C `hypot()` function.
Such cases are required only if C compiler supports IEC 559, but clang doesn't. Failures are found on linux-mips32 and iOS arm64 targets.

2. Floating point unary operator should use LLVM's `fneg` instruction, because `0 - arg` doens't work if `arg = 0.0`, returning `0.0` instead of `-0.0`. See https://llvm.org/docs/LangRef.html#fneg-instruction

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
@@ -699,10 +699,10 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
     private fun FunctionGenerationContext.emitUnaryMinus(args: List<LLVMValueRef>): LLVMValueRef {
         val first = args[0]
         val destTy = first.type
-        val const0 = makeConstOfType(destTy, 0)
         return if (destTy.isFloatingPoint()) {
-            fsub(const0, first)
+            fneg(first)
         } else {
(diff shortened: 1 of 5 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IntrinsicGenerator.kt
```diff
@@ -699,10 +699,10 @@ internal class IntrinsicGenerator(private val environment: IntrinsicGeneratorEnv
     private fun FunctionGenerationContext.emitUnaryMinus(args: List<LLVMValueRef>): LLVMValueRef {
         val first = args[0]
         val destTy = first.type
-        val const0 = makeConstOfType(destTy, 0)
         return if (destTy.isFloatingPoint()) {
-            fsub(const0, first)
+            fneg(first)
```
sbogolepov, line 703: > Simple negation from 0 doesn't work in case -0.0 argument.
Do we have a test for that?

PavelPunegov: Yes, I've added back "runtime_math_harmony" test that found this issue.

PR #4263, opened 2020-06-30
[IR] Refactored GET_FIELD, SET_FIELD for inline classes + test

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt
```diff
@@ -243,34 +240,31 @@ private class InlineClassTransformer(private val context: Context) : IrBuildingT
...
-            // TODO: it is better to get rid of functions setting such fields.
-            val startOffset = expression.startOffset
-            val endOffset = expression.endOffset
-            IrBlockImpl(startOffset, endOffset, irBuiltIns.unitType).apply {
-                statements.addIfNotNull(expression.receiver)
-                statements += expression.value
-                statements += IrCallImpl(startOffset, endOffset, irBuiltIns.nothingType, symbols.ThrowNullPointerException)
-                statements += IrGetObjectValueImpl(startOffset, endOffset, irBuiltIns.unitType, irBuiltIns.unitClass)
-            }
+        return if (expression.symbol.owner.parentClassOrNull?.isInlined() == true) {
```
SvyatoslavScherbina, line 263: What was wrong with previous implementation? It was intended to preserve all valid IR nodes and their possible side effects.

homuroll: Frankly, I was stuck while looking on that code, a comment would've been nice. Anyhow, is it really needed? AFAIK, it is not allowed to have mutable field in an inline class, so SET_FIELD can only be in the primary constructor, which is not code generated.

SvyatoslavScherbina: > Frankly, I was stuck while looking on that code, a comment would be nice.

Fair enough.

> Anyhow, is it really needed?

Well, this was just the most safe way to handle this IR node. Throwing an exception still makes sense here, to ensure this code never gets reachable.

PR #4264, opened 2020-07-01
Optimize reference counting in relaxed memory model.

slowcoders:
Eliminated redundant UpdateStackRef and UpdateReturnRef calls when using Relaxed memory model.
As a result, the execution speed of hash_map0 test is improved 20%.  
The whole test results are exactly same as previous relaxed model. (10 test fails)


SvyatoslavScherbina:
"Relaxed memory model" mode enabled by `-memory-model=relaxed` compiler flag is fundamentally broken at the moment. I don't believe it is usable in non-trivial number of real projects (for example, cyclic garbage is not collected in this mode); if you have counterexamples to this statements, please share with us. In fact this particular implementation is abandoned.

Moreover, we plan to rework this code completely. See more details: https://blog.jetbrains.com/kotlin/2020/07/kotlin-native-memory-management-roadmap/

That's why I'm closing this PR.

slowcoders:
Hmm. I see.
Would you let me know when you are going to release new memory model?


2020년 7월 21일 (화) 오후 7:06, SvyatoslavScherbina <notifications@github.com>님이
작성:

SvyatoslavScherbina:
We don't have specific ETA yet. Please stay tuned, you won't miss it.

PR #4268, opened 2020-07-03
Fix framework tests in iOS

PavelPunegov:
* Fix name in the path
* Don't copy static framework to the application bundle

### DIFF
--- a/backend.native/tests/build.gradle
+++ b/backend.native/tests/build.gradle
@@ -4108,7 +4108,7 @@ Task frameworkTest(String name, Closure<FrameworkTest> configurator) {
                         }
 
                         extraOpts fr.bitcode ? "-Xembed-bitcode" : "-Xembed-bitcode-marker"
-                        extraOpts fr.isStatic ? "-Xstatic-framework" : ""
+                        if (fr.isStatic) extraOpts "-Xstatic-framework"
                         extraOpts fr.opts
                         extraOpts project.globalTestArgs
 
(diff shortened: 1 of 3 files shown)


backend.native/tests/build.gradle
```diff
@@ -4108,6 +4108,7 @@ Task frameworkTest(String name, Closure<FrameworkTest> configurator) {
                         }
 
                         extraOpts fr.bitcode ? "-Xembed-bitcode" : "-Xembed-bitcode-marker"
+                        extraOpts fr.isStatic ? "-Xstatic-framework" : ""
```
SvyatoslavScherbina, line 4111: Is adding empty flag has the same effect as adding no flags?

SvyatoslavScherbina: Well, adding empty argument is definitely different from not adding arguments at all.
The former might have unpredictable impact. For example:
```bash
$ konanc "" foo.kt
```
->
```text
error: source file or directory not found:
```

PavelPunegov: Yes, I found out that Konan plugin runs konanc main with arguments in the `.lst` argfile

PavelPunegov: Parser skips empty lines here checking for `isWhitespace`: https://github.com/JetBrains/kotlin/blob/master/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/preprocessCommandLineArguments.kt#L64
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-whitespace.html

SvyatoslavScherbina: My intention was to demonstrate that adding empty argument is error-prone and should be avoided in general, no matter what particular implementation does.

Please avoid adding `""`.

PR #4270, opened 2020-07-06
Improve cache linkage

homuroll:
It is bad to link more libraries than it is really needed. For the whole world compilation there is LllvmImports, which tracks all dependencies on bitcode level, but there has not been such a mechanism for compilations against cached libraries.


ilmat192:
My approve above is about the platform libs generation part

PR #4272, opened 2020-07-06
update: docs for CocoaPods dependencies

katia-energizer:
* Described scenarios on how to add dependencies on Pod libraries stored in the CocoaPods repository or locally as well as use a Kotlin/Native Gradle project as a CocoaPods dependency (Kotlin Pod).
* Fixed code highlighting in Concurrency


COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+You can manage Pod dependencies right in Intellij IDEA enjoying its benefits of working with code, such as code
+ highlighting and completion. You can build the whole Kotlin project with Gradle as a whole not switching to Xcode. 
+ Go to Xcode only when you need to write Swift/Objective-C code or run your application on a simulator or device.
  
- - When building from Xcode, you can use CocoaPods libraries without writing
- .def files manually and setting cinterop tool parameters. In this case, all required parameters can be
- obtained from the Xcode project configured by CocoaPods.
+Depending on your purpose, you can add dependencies between:
+* [A Kotlin project and a Pod library from the CocoaPods repository](#add-a-dependency-on-a-pod-library-from-the-cocoapods-repository)
+* [A Kotlin project and an Objective-C Pod library stored locally](#add-a-dependency-on-an-objective-c-pod-library-stored-locally)
```
SvyatoslavScherbina, line 13: How "Objective-C Pod library" here is different from "Pod library" above?

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
-  - `homepage`
+To use this dependency from Kotlin code, import a package `cocoapods.<library-name>`.
+In the example above, it's `cocoapods.AFNetworking`.
+
+### Add a dependency on an Objective-C Pod library stored locally
+
+1. Add a dependency on a Pod library stored locally with `pod()` to `build.gradle.kts` (`build.gradle`) of your
+ project.  
+As the third argument, specify the path to `Podspec` of the local Pod using `projectDir.resolve("..")`.  
+    > You can also add dependencies as subspecs.
```
SvyatoslavScherbina, line 134: Do I understand correctly that subspecs work only for locally-stored pods?

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
 
-2. Run the `podspec` task. The podspec file described above will be generated.
+### Add a dependency between a Kotlin Pod and Xcode project with one target
 
-    > In order to avoid compatibility issues during an Xcode build, the plugin requires using
-    a [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html). 
-    To generate the wrapper automatically during execution of the `podspec` task,
-    run it with the parameter `-Pkotlin.native.cocoapods.generate.wrapper=true`.
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
+2. Add the path to your `Podfile` with `podfile = projectDir.resolve("..")` to `build.gradle.kts` (`build.gradle`)
```
SvyatoslavScherbina, line 187: This step might be confusing, it makes sense to describe its purpose.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
  
- - When building from Xcode, you can use CocoaPods libraries without writing
- .def files manually and setting cinterop tool parameters. In this case, all required parameters can be
- obtained from the Xcode project configured by CocoaPods.
+Depending on your purpose, you can add dependencies between:
+* [A Kotlin project and a Pod library from the CocoaPods repository](#add-a-dependency-on-a-pod-library-from-the-cocoapods-repository)
+* [A Kotlin project and an Objective-C Pod library stored locally](#add-a-dependency-on-an-objective-c-pod-library-stored-locally)
+* [A Kotlin Pod and an Xcode project with one target](#add-a-dependency-between-a-kotlin-pod-and-xcode-project-with-one-target) 
+or [several targets](#add-a-dependency-between-a-kotlin-pod-with-an-xcode-project-with-several-targets)
+* [Several Kotlin Pods and an Xcode project](#add-dependencies-between-multiple-kotlin-pods-and-an-xcode-project)
```
SvyatoslavScherbina, line 16: There is no sections for multiple Xcode projects. We should at least mention whether it is supported.

COCOAPODS.md
zoldater: I think that the difference in storage place: Cocoapods repo vs user's filesystem. In both cases only Objective-C libraries could be used.

COCOAPODS.md
zoldater: Current state of Cocoapods Integration doesn't support multiple Xcode projects.

COCOAPODS.md
zoldater: No, this syntax is available also for remote pods.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
-this Xcode project. If necessary, this framework is automatically rebuilt during Xcode build process.
+2. Re-import the project.
 
-## Workflow
+You can find a sample project [here](https://github.com/zoldater/KotlinWithLocalObjCPods). 
+ 
+## Use a Kotlin Gradle project as a CocoaPods dependency
 
-To import a Kotlin/Native module in an existing Xcode project:
+You can use a Kotlin/Native Gradle project as a CocoaPods dependency (Kotlin Pod). Such a representation provides the
```
zoldater, line 163: More precisely, Kotlin Pod is Multiplatform Project with Native targets. It also requires multiplatform plugin to be applied.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+ following advantages:
 
-0. Make sure that you have CocoaPods [installed](https://guides.cocoapods.org/using/getting-started.html#installation).
-We recommend using CocoaPods 1.6.1 or later.
+ * Such a dependency can be included in a `Podfile` of an Xcode project and automatically built (and rebuilt)
+ along with this project. It simplifies importing to Xcode by removing a need to write corresponding Gradle tasks and
+  Xcode build steps manually.
```
zoldater, line 168: May be useful to say that Kotlin pod dependency is included in Podfile by it's name and path to project directory (it contains generated .podspec file)

COCOAPODS.md
zoldater: Agree. Description may be simple, like "to maintain a consistent state of Xcode project and Kotlin Pod dependencies"

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+            pod("AFNetworking", "~> 4.0.0")
+            podfile = projectDir.resolve("../severalTargetsXcodeProject/Podfile") // specify the path to Podfile
+        }
+    }
+    ```
+    
     </div>
 
-4. Run `pod install` for the Xcode project.
+2. Add the name and path of the Kotlin Pod you want to include in the Xcode project to `Podfile`.
```
zoldater, line 254: Numeration 2 -> 5, also next point

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
-of the multiplatform plugin documentation). The `cocoapods { ... }` code block allows you to
-add dependencies on CocoaPods libraries.
+### Add dependencies between multiple Kotlin Pods and an Xcode project
 
-<div class="sample" markdown="1" theme="idea" mode="kotlin">
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
+2. Create a hierarchical structure of your Kotlin projects - a root and child projects.
+3. Add the path to your `Podfile` with `podfile = projectDir.resolve("..")` to `build.gradle.kts` (`build.gradle`) of
+ the root project.
+4. Specify that you don’t need a `Podspec` for this root project - `noPodspec()`.
```
zoldater, line 287: This is only necessary if root project is not represent Kotlin Pod

COCOAPODS.md
ilmat192: We cannot automatically call `pod install` in this case. But AFAIK a user can depend on the same Kotlin pod from multiple Xcode projects given that he calls `pod install` manually for each of these projects. Is't this the case?

zoldater: Of course yes, this case may be supported by manual `pod install` calls.

COCOAPODS.md
SvyatoslavScherbina: >  In both cases only Objective-C libraries could be used.

Then we should stick to the same wording here. Otherwise this can be understood as "published Swift pods are supported, local ones aren't", which is not true.

COCOAPODS.md
SvyatoslavScherbina: Then the document is kinda misleading here: subspecs are mentioned under "stored locally" section, and the example shows only subspecs of locally-stored pods.

zoldater: Agree

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
-To import a Kotlin/Native module in an existing Xcode project:
+You can use a Kotlin/Native Gradle project as a CocoaPods dependency (Kotlin Pod). Such a representation provides the
+ following advantages:
 
-0. Make sure that you have CocoaPods [installed](https://guides.cocoapods.org/using/getting-started.html#installation).
-We recommend using CocoaPods 1.6.1 or later.
+ * Such a dependency can be included in a `Podfile` of an Xcode project and automatically built (and rebuilt)
+ along with this project. It simplifies importing to Xcode by removing a need to write corresponding Gradle tasks and
+  Xcode build steps manually.
+ * When building from Xcode, you can use CocoaPods libraries without writing
```
ilmat192, line 169: This is rather an advantage of the whole CocoaPods plugin. I mean representing a Kotlin module as a pod and adding it to Xcode project dependencies *was* required to consume pod dependencies in the Kotlin module without manual def-files/cinterop configuration. But with Yaroslav's changes this is not necessary anymore and a user can easily consume pods without an Xcode project.

May be fully remove this paragraph and don't scare users with things like def-files or cinterop?

COCOAPODS.md
ilmat192, line 187: (minor) It's better to replace `podfile = projectDir.resolve("..")` with `podfile = project.file(..)`. The second construction is more common in Gradle.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+    platform :ios, '9.0'
+    
+    target 'ios-app' do
+            pod 'kotlin_library', :path => '../kotlin-library'
     end
     ```
+    
+    </div>
 
+4. Re-import the project.
```
ilmat192, line 223: May be add that we call `pod install` for the user's Podfile under the hood at this step? Probably a user doesn't expect this so it's better to warn him.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+
+### Add a dependency between a Kotlin Pod with an Xcode project with several targets
+
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
+2. Add the path to your `Podfile` with `podfile = projectDir.resolve("..")` to `build.gradle.kts` (`build.gradle`) of
+ your Kotlin project.
+3. Add dependencies to Pod libraries that you want to use in your project with `pod()`.
+4. For each target, specify the minimum target version for the Pod library.
```
ilmat192, line 231: @zoldater, can't this be needed for an Xcode project with a single target?

COCOAPODS.md
ilmat192: Then it worth mentioning because currently users may think that this case is fully unsupported.

COCOAPODS.md
katia-energizer: Agree :). Will remove Objective-C and use Pod library everywhere

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+3. Re-import the project.
 
-## Interoperability
+You can find a sample project [here](https://github.com/zoldater/severalXcodeTargetsDemo).
 
-The CocoaPods plugin also allows using CocoaPods libraries without manual configuring cinterop
-parameters (see the [corresponding section](https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#cinterop-support)
-of the multiplatform plugin documentation). The `cocoapods { ... }` code block allows you to
-add dependencies on CocoaPods libraries.
+### Add dependencies between multiple Kotlin Pods and an Xcode project
```
SvyatoslavScherbina, line 281: We probably should mention that the resulting Kotlin pods are independent (i.e. declarations included to more than one pods get duplicated).

COCOAPODS.md
katia-energizer: Let's add a subspec example for remote pods as well then. I didn't realize that. Thanks! Will ask @zoldater for such an example.

COCOAPODS.md
katia-energizer: OK. Will add an explanation.

COCOAPODS.md
katia-energizer: I can add a note about this without providing a procedure. Does it work?

COCOAPODS.md
katia-energizer: Thx. Will add.

COCOAPODS.md
katia-energizer: Let's add.

COCOAPODS.md
katia-energizer: @zoldater Should we update sample projects and docs?

COCOAPODS.md
SvyatoslavScherbina: I believe a note is enough here.

COCOAPODS.md
zoldater: Sometimes generating of synthetic project fails for the reason of it's absence.

COCOAPODS.md
zoldater: In point 2 we decided to add some explanation about `podfile` setting

COCOAPODS.md
zoldater: Remote Subspec dependency added in [KotlinWithLocalObjCPods exapmle](https://github.com/zoldater/KotlinWithLocalObjCPods/blob/master/kotlin-library/build.gradle.kts)

COCOAPODS.md
zoldater: I prefer this syntax because it shows more implicitly that entered path  is related to `projectDir`. As these 2 constructions are semantically equivalent, we can replace it.

COCOAPODS.md
ilmat192: So this setting makes sense for both use cases. May be mention it separately? Like "You may get an error that a dependency pod requires a higher minimum deployment target. In this case you can specify the correct minimum deployment target in the cocoapods block". @zoldater, is this warning ok or it's worth adding something?

zoldater: Ok, implementation is able to log this warning too

COCOAPODS.md
SvyatoslavScherbina: Also this configuration has known issues, e.g. it requires a user to disable compiler caches.

COCOAPODS.md
katia-energizer: Done

COCOAPODS.md
katia-energizer: Added a note and example to a procedure on adding a remote Pod. Updated the example with local pods - included a remote Pod added via a subspec as well

COCOAPODS.md
katia-energizer: Added a note after the list of cases. @SvyatoslavScherbina and @ilmat192 , please take a look if it's ok.

COCOAPODS.md
katia-energizer: Updated a description of a Kotlin Pod. There is information about the multiplatform plugin in the section about _Installing plugins._

COCOAPODS.md
katia-energizer: Updated the syntax - replaced `podfile = projectDir.resolve("..")` with `podfile = project.file(..)`

katia-energizer: @zoldater will update sample projects.

COCOAPODS.md
katia-energizer: Added this info.

COCOAPODS.md
katia-energizer: Removed this paragraph as we discussed. :)

COCOAPODS.md
katia-energizer: Added this info.

COCOAPODS.md
katia-energizer: Fixed. Thx.

COCOAPODS.md
```diff
@@ -1,150 +1,343 @@
...
+    
+    ```ruby
+    $ sudo gem install cocoapods
+    ```
+    
+    </div>
 
-For an example of CocoaPods integration, refer to the
-[`cocoapods`](https://github.com/JetBrains/kotlin-native/tree/master/samples/cocoapods) sample.
+2. Install the `cocoapods-generate` plugin.
```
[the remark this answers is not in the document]
katia-energizer: added

COCOAPODS.md
katia-energizer: Added a step to specify the min target version in case of an Xcode project with one target; updated the example.

COCOAPODS.md
katia-energizer: Added the explanation.

COCOAPODS.md
katia-energizer: As discussed with @zoldater, I've added a note with a workaround in case of an error when using multiple Kotlin pods with an Xcode project.

katia-energizer: @SvyatoslavScherbina, could you clarify your suggestion? 

> We probably should mention that the resulting Kotlin pods are independent (i.e. declarations included to more than one pods get duplicated).

Why is it important for a user? When should they use this info?

COCOAPODS.md
```diff
@@ -1,150 +1,379 @@
...
+or [`use_frameworks!`](https://guides.cocoapods.org/syntax/podfile.html#use_frameworks_bang)
+directive.
+{:.note}
+
+### Add a dependency between a Kotlin Pod and Xcode project with one target
+
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
+2. Add the path to your `Podfile` with `podfile = project.file(..)` to `build.gradle.kts` (`build.gradle`) 
+of your Kotlin project.  
+    This step helps synchronize an Xcode project and Kotlin Pod dependencies.
```
ilmat192, line 214: May be explicitly say that this synchronization is performed by calling `pod install` for the user's Xcode project? 

I mean a user may add some third-party dependencies to Podfile, re-import the Gradle project to IDEA and unexpectedly get Xcode dependencies updated too. It may be a WTF for the user so it worth warning him. I think that just saying that `pod install` is called on IDEA re-import is enough.

COCOAPODS.md
ilmat192: It looks good to me.

ilmat192: It looks good to me.

COCOAPODS.md
```diff
@@ -103,7 +107,9 @@ steps are required.
 ### Add a dependency on a Pod library from the CocoaPods repository
 
 1. Add dependencies on a Pod library that you want to use from the CocoaPods repository with `pod()`  to `build.gradle.kts` 
-(`build.gradle`) of your project.
+(`build.gradle`) of your project.  
+    > You can also add dependencies as subspecs.
```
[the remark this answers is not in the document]
katia-energizer: updated

COCOAPODS.md
```diff
@@ -1,150 +1,379 @@
...
+    </div>
+    
+5. Re-import the project.
 
 When applied, the CocoaPods plugin does the following:
 
-1. Adds both `debug` and `release` frameworks as output binaries for all iOS and macOS targets.
-2. Creates a `podspec` task which generates a [podspec](https://guides.cocoapods.org/syntax/podspec.html)
-file for the given project.
+* Adds both `debug` and `release` frameworks as output binaries for all iOS and macOS targets.
```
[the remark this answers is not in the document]
katia-energizer: added

COCOAPODS.md
katia-energizer: Updated the explanation as discussed with @zoldater.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
-to Kotlin/Native. This feature allows you to represent a Kotlin/Native Gradle-project as a
-CocoaPods dependency. Such a representation provides the following advantages:
+Kotlin/Native provides integration with the [CocoaPods dependency manager](https://cocoapods.org/). You can add
+dependencies on Pod libraries stored in the CocoaPods repository or locally as well as use a multiplatform project with 
+native targets as a CocoaPods dependency (Kotlin Pod).
 
- - Such a dependency can be included in a Podfile of an Xcode project and automatically built (and rebuilt)
- along with this project. As a result, importing to Xcode is simplified since there is no need to
- write corresponding Gradle tasks and Xcode build steps manually.
+You can manage Pod dependencies right in Intellij IDEA enjoying its benefits of working with code, such as code
```
B1ggDave, line 7: You can manage Pod dependencies directly in IntelliJ IDEA and enjoy all the additional features such as code highlighting and completion. You can build the whole Kotlin project with Gradle and not ever have to switch to Xcode. 
Use Xcode only when you need to write Swift/Objective-C code or run your application on a simulator or device.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+ highlighting and completion. You can build the whole Kotlin project with Gradle as a whole not switching to Xcode. 
+ Go to Xcode only when you need to write Swift/Objective-C code or run your application on a simulator or device.
  
- - When building from Xcode, you can use CocoaPods libraries without writing
- .def files manually and setting cinterop tool parameters. In this case, all required parameters can be
- obtained from the Xcode project configured by CocoaPods.
+Depending on your purpose, you can add dependencies between:
```
B1ggDave, line 11: Depending on your project and purposes, you can add dependencies between:
* [A Kotlin project and a Pod library from the CocoaPods repository](#add-a-dependency-on-a-pod-library-from-the-cocoapods-repository)
* [A Kotlin project and a Pod library stored locally](#add-a-dependency-on-a-pod-library-stored-locally)
* [A Kotlin Pod and an Xcode project with one target](#add-a-dependency-between-a-kotlin-pod-and-xcode-project-with-one-target) 
or [several targets](#add-a-dependency-between-a-kotlin-pod-with-an-xcode-project-with-several-targets)
* [Several Kotlin Pods and an Xcode project](#add-dependencies-between-multiple-kotlin-pods-and-an-xcode-project)

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+
+    <div class="sample" markdown="1" theme="idea" mode="ruby" data-highlight-only>
+    
+    ```ruby
+    $ sudo gem install cocoapods-generate
+    ```
+    
+    </div>
+    
+3. In `build.gradle.kts` (or `build.gradle`) of your IDEA project, apply the CocoaPods plugin as well as the Kotlin
```
B1ggDave, line 44: 3. In `build.gradle.kts` (or `build.gradle`) of your IntelliJ IDEA project, apply the CocoaPods plugin as well as the Kotlin Multiplatform plugin it is based on.

" apply the CocoaPods plugin as well as the Kotlin Multiplatform plugin it is based on."
this could be clearer with what it means? -- apply how, what is the Multiplatform plugin it is based on?

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
-2. Creates a `podspec` task which generates a [podspec](https://guides.cocoapods.org/syntax/podspec.html)
-file for the given project.
+* Adds both `debug` and `release` frameworks as output binaries for all macOS, iOS, tvOS, and watchOS targets.
+* Creates a `podspec` task which generates a [Podspec](https://guides.cocoapods.org/syntax/podspec.html)
+file for the project.
 
-The podspec generated includes a path to an output framework and script phases which automate building
-this framework during a build process of an Xcode project. Some fields of the podspec file can be
-configured using the `kotlin.cocoapods { ... }` code block.
+The `Podspec` file includes a path to an output framework and script phases which automate building
```
B1ggDave, line 95: The `Podspec` file includes a path to an output framework and script phases that automate building this framework during the build process of an Xcode project.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
-        summary = "Some description for a Kotlin/Native module"
-        homepage = "Link to a Kotlin/Native module homepage"
-
-        // The name of the produced framework can be changed.
-        // The name of the Gradle project is used here by default.
-        frameworkName = "my_framework"
+You can add dependencies between a Kotlin project and Pod libraries [stored in the CocoaPods repository](#add-a-dependency-on-a-pod-library-from-the-cocoapods-repository) 
+and [stored locally](#add-a-dependency-on-a-pod-library-stored-locally).
+
+[Complete initial configuration](#install-the-cocoapods-dependency-manager-and-plugin), and when you add a new dependency,
```
B1ggDave, line 103: [Complete the initial configuration](#install-the-cocoapods-dependency-manager-and-plugin), and when you add a new dependency and re-import the project in IntelliJ IDEA, the new dependency will be added automatically. There are no additional steps required.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+        }
     }
-}
+    ```
+    
+    </div>
+
+2. Re-import the project.
+
+To use these dependencies from Kotlin code, import packages `cocoapods.<library-name>`.
```
B1ggDave, line 134: To use these dependencies from the Kotlin code, import the packages `cocoapods.<library-name>`.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
-A version of the Gradle project is used as a value for the `version` field.
-Fields`summary` and `homepage` can be configured using the `cocoapods` code block.
+### Add a dependency on a Pod library stored locally
 
-This podspec file can be referenced from a [Podfile](https://guides.cocoapods.org/using/the-podfile.html)
-of an Xcode project. After that the framework built from the Kotlin/Native module can be used from
-this Xcode project. If necessary, this framework is automatically rebuilt during Xcode build process.
+1. Add a dependency on a Pod library stored locally with `pod()` to `build.gradle.kts` (`build.gradle`) of your
+ project.  
+As the third parameter, specify the path to `Podspec` of the local Pod using `project.file(..)`.
```
B1ggDave, line 151: As the third parameter

where does third come from?

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
-    run it with the parameter `-Pkotlin.native.cocoapods.generate.wrapper=true`.
+</div>
 
-3. Add a reference to the generated podspec in a Podfile of the Xcode project.
+You can find a sample project [here](https://github.com/Kotlin/kotlin-with-cocoapods-sample).
+ 
+## Use a Kotlin Gradle project as a CocoaPods dependency
 
-    <div class="sample" markdown="1" theme="idea" mode="ruby">
+You can use a multiplatform project with native targets as a CocoaPods dependency (Kotlin Pod). You can include such a dependency
```
B1ggDave, line 194: You can use a Kotlin Multiplatform project with native targets as a CocoaPods dependency (Kotlin Pod). You can include such a dependency

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+  in a `Podfile` of an Xcode project  by its name and path to the project directory containing the generated `Podspec`.
```
B1ggDave, line 195: in the `Podfile` of an Xcode project by its name and path to the project directory containing the generated `Podspec`. 
This dependency will be automatically built (and rebuilt) along with this project. 
Such an approach simplifies importing to Xcode by removing a need to write corresponding Gradle tasks and Xcode build steps manually.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+
+> To correctly import the dependencies into the Kotlin/Native module, the
+`Podfile` must contain either [`use_modular_headers!`](https://guides.cocoapods.org/syntax/podfile.html#use_modular_headers_bang)
+or [`use_frameworks!`](https://guides.cocoapods.org/syntax/podfile.html#use_frameworks_bang)
+directive.
+{:.note}
+
+### Add a dependency between a Kotlin Pod and Xcode project with one target
+
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
```
B1ggDave, line 212: 1. Create an Xcode project with a `Podfile` if you haven’t done so yet.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
     end
     ```
+    
+    </div>
+
+5. Re-import the project.
+
+### Add a dependency between a Kotlin Pod with an Xcode project with several targets
+
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
```
B1ggDave, line 258: 1. Create an Xcode project with a `Podfile` if you haven’t done so yet.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+2. Add the path to your Xcode project `Podfile` with `podfile = project.file(..)` to `build.gradle.kts` (`build.gradle`) of
+ your Kotlin project.  
+    This step helps synchronize your Xcode project with Kotlin Pod dependencies by calling `pod install` for your `Podfile`.
+3. Add dependencies to Pod libraries that you want to use in your project with `pod()`.
```
B1ggDave, line 262: 3. Add dependencies to the Pod libraries that you want to use in your project with `pod()`.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+            pod("AFNetworking", "~> 4.0.0")
+            podfile = project.file("../severalTargetsXcodeProject/Podfile") // specify the path to Podfile
+        }
+    }
+    ```
+    
     </div>
 
-4. Run `pod install` for the Xcode project.
+5. Add the name and path of the Kotlin Pod you want to include in the Xcode project to `Podfile`.
```
B1ggDave, line 286: 5. Add the name and path of the Kotlin Pod you want to include in the Xcode project to the `Podfile`.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+You can find a sample project [here](https://github.com/Kotlin/multitarget-xcode-with-kotlin-cocoapods-sample).
 
-The CocoaPods plugin also allows using CocoaPods libraries without manual configuring cinterop
-parameters (see the [corresponding section](https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#cinterop-support)
-of the multiplatform plugin documentation). The `cocoapods { ... }` code block allows you to
-add dependencies on CocoaPods libraries.
+### Add dependencies between multiple Kotlin Pods and an Xcode project
 
-<div class="sample" markdown="1" theme="idea" mode="kotlin">
+1. Create an Xcode project with `Podfile` if you haven’t done this yet.
```
B1ggDave, line 315: 1. Create an Xcode project with a `Podfile` if you haven’t done so yet.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+        }
     }
-}
-```
+    ```
+    
+    </div>
 
-</div>
+5. Add dependencies to Pod libraries that you want to use to the build script of your child projects using `pod()`.
```
B1ggDave, line 337: 5. Add dependencies to the Pod libraries that you want to use to the build script of your child projects using `pod()`.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
     }
-}
-```
+    ```
+    
+    </div>
 
-</div>
+5. Add dependencies to Pod libraries that you want to use to the build script of your child projects using `pod()`.  
+    > Do NOT add a path to `Podfile` in your child projects.
```
B1ggDave, line 338: > Do NOT add a path to the `Podfile` in your child projects.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+            pod("AFNetworking", "~> 4.0.0")
+        }
+    }
+    ```
+    
+    </div>
 
-To use these dependencies from a Kotlin code, import a package `cocoapods.<library-name>`.
-In the example above, it's `cocoapods.AFNetworking`.
+6. Add names and paths of the Kotlin Pods you want to include in the Xcode project to `Podfile`.
```
B1ggDave, line 356: 6. Add the names and paths of the Kotlin Pods you want to include in the Xcode project to the `Podfile`.

COCOAPODS.md
```diff
@@ -1,150 +1,381 @@
...
+    
+    </div>
 
-> __Important:__ To correctly import the dependencies into the Kotlin/Native module, the
-Podfile must contain either [`use_modular_headers!`](https://guides.cocoapods.org/syntax/podfile.html#use_modular_headers_bang)
-or [`use_frameworks!`](https://guides.cocoapods.org/syntax/podfile.html#use_frameworks_bang)
-directive.
+7. Re-import the project.
+
+>If you get an error about a duplicate symbol when building your Xcode project, add `kotlin.native.cacheKind=none` to
```
B1ggDave, line 375: >If you get a duplicate symbol error when building your Xcode project, add `kotlin.native.cacheKind=none` to 
>the `gradle.properties` and re-import the project.

CONCURRENCY.md
```diff
@@ -207,4 +219,4 @@ that are potentially parts of shared cyclic data should be zeroed out once no lo
  If atomic reference value is attempted to be set to non-frozen value runtime exception is thrown.
```
B1ggDave, line 219: If you attempt to set an atomic reference value to non-frozen, a value runtime exception will be thrown.

CONCURRENCY.md
```diff
@@ -207,4 +219,4 @@ that are potentially parts of shared cyclic data should be zeroed out once no lo
...
 
  Freezable atomic reference is similar to the regular atomic reference, but until frozen behaves like regular box
```
B1ggDave, line 221: Freezable atomic reference is similar to the regular atomic reference, but until frozen behaves like a regular box

what is a regular box?

COCOAPODS.md
SvyatoslavScherbina, line 375: The whole advice is not exactly correct.
1. The same issue can have different consequences, not only duplicate symbol errors.
2. AFAIK, using multiple pods almost always gets affected by the issue, so the "if you get an error" part doesn't make much sense, it should be ok to recommend adding the line unconditionally.

COCOAPODS.md
arpie42: to use to the build script?  --> to use the build script

COCOAPODS.md
SvyatoslavScherbina: 👍

COCOAPODS.md
katia-energizer: updated links to sample projects.

COCOAPODS.md
katia-energizer: CocoaPods plugin is based on Multiplatform plugin. That's why we need to apply both. I can remove "it is based on" if it seems confusing. The idea is to explain somehow why you need to apply multiplatform plugin but it's not critical.

COCOAPODS.md
katia-energizer: from the example. project.file() is the third parameter after the name and version.
`pod("pod_dependency", "1.0", project.file("../pod_dependency/pod_dependency.podspec"))`

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
katia-energizer: Discussed with @SvyatoslavScherbina a couple of hours ago. Decided to remove this scenario from our blog post and docs.

COCOAPODS.md
katia-energizer: Removed this scenario since it has issues and users may face an unexpected behaviour.

COCOAPODS.md
katia-energizer: Removed the scenario

COCOAPODS.md
katia-energizer: Removed the scenario.

CONCURRENCY.md
katia-energizer: I haven't updated content in this article. Just fixed formatting. No proofread is required for that article for now. Thx!

CONCURRENCY.md
katia-energizer: I haven't updated content in this article. Will request proofread separately. 
I just fixed formatting there.

COCOAPODS.md
katia-energizer: As we agreed, I've removed the whole scenario.

COCOAPODS.md
katia-energizer: ✅

COCOAPODS.md
B1ggDave: Thanks! I think, as it is not described that it is based on it right now, it is superfluous information and "it is based on" is a good thing to cut as they don't need to know this and they may be like me... easily confused :)

COCOAPODS.md
B1ggDave: Ahh perfect, thanks!

COCOAPODS.md
katia-energizer: ✅

PR #4275, opened 2020-07-09
External tests runner fixes

PavelPunegov:
1. Make link options (`-l lib`) be unique.
2. Add test directive parsing for -Xexpect-actual-linker

### DIFF
--- a/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
+++ b/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
@@ -220,6 +220,10 @@ class RunExternalTestGroup extends JavaExec {
                 line.split(" ").toList().forEach { flags.add("-Xopt-in=$it") }
             }
         }
+        def expectActualLinker = findLinesWithPrefixesRemoved(text, "// EXPECT_ACTUAL_LINKER")
+        if (expectActualLinker.size() != 0) {
+            flags.add("-Xexpect-actual-linker")
+        }
     }
 
     static String markMutableObjects(String text) {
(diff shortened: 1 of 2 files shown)


build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -505,23 +509,31 @@ fun runTest() {
                     List<TestModule> orderedModules = DFS.INSTANCE.topologicalOrder(modules.values()) { module ->
                         module.dependencies.collect { modules[it] }.findAll { it != null }
                     }
-                    def libsFlags = []
+                    List<String> libs = new ArrayList<String>()
```
ilmat192, line 512: Can't we just use a `Set` here to get rid of `libs.unique()`?

build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -505,23 +509,31 @@ fun runTest() {
...
                         if (!module.isDefaultModule()) {
                             def klibModulePath = "${executablePath()}.${module.name}.klib"
-                            libsFlags += module.linkDependencies(executablePath())
+                            libs.addAll(module.dependencies)
                             runCompiler(compileList.findAll { it.module == module }.collect { it.path },
-                                    klibModulePath, flags + ["-p", "library"] + libsFlags)
+                                    klibModulePath, flags + ["-p", "library"] +
+                                    libs.unique()
+                                            .collect { "${executablePath()}.${it}.klib" }
+                                            .collectMany { ["-l", it] }
```
ilmat192, line 521: (minor) IMHO folding these two collects into one:
```
collectMany { ["-l", "${executablePath()}.${it}.klib"] }
```
will make this code more readable. But it's just an IMHO.

PR #4276, opened 2020-07-09
Special case for floating-point pow() function

PavelPunegov:
Add a special case for floating-point pow() function: in case of ±0 the function should return 1.0


PavelPunegov:
See also #4259

PR #4284, opened 2020-07-13
Fix imports after compiler update

ilmat192:


ilmat192:
Fixed by https://github.com/JetBrains/kotlin-native/pull/4283

PR #4286, opened 2020-07-15
Restore cycle detector

projedi:


runtime/src/main/cpp/Memory.h
```diff
@@ -329,6 +329,12 @@ struct MetaObjHeader {
     // Strong reference to the counter object.
     ObjHeader* counter_;
   } WeakReference;
+  struct {
+    // Leak detector's previous list element.
+    ObjHeader* previous_;
+    // Leak detector's next list element.
+    ObjHeader* next_;
+  } LeakDetector;
```
SvyatoslavScherbina, line 337: This is kinda space-inefficient. Can we avoid this?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -956,6 +1054,7 @@ ALWAYS_INLINE void runDeallocationHooks(ContainerHeader* container) {
       cyclicRemoveAtomicRoot(obj);
     }
 #endif  // USE_CYCLIC_GC
+    CycleDetector::instance().removeCandidateIfNeeded(obj);
```
projedi, line 1057: That's a mistake. The instance should not be created at all if leak checker is disabled. Will fix.

runtime/src/main/cpp/Memory.h
projedi: Replaced it with a global hashmap and a list, which are only allocated if someone asks for this leak checker (this isn't currently the case, will fix).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2022,6 +2121,7 @@ OBJ_GETTER(allocInstance, const TypeInfo* type_info) {
 #endif  // USE_GC
   auto container = ObjectContainer(state, type_info);
   ObjHeader* obj = container.GetPlace();
+  CycleDetector::insertCandidateIfNeeded(obj);
```
SvyatoslavScherbina, line 2124: The meaning is quite similar to what `cyclicAddAtomicRoot` does, so it might be better to place `insertCandidateIfNeeded` near `cyclicAddAtomicRoot`.
(For example, `cyclicAddAtomicRoot` must be after `rememberNewContainer` to prevent data race, the same might be required for `insertCandidateIfNeeded` too).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2708,6 +2808,118 @@ void shareAny(ObjHeader* obj) {
...
+  }
+}
+
+// static
+CycleDetectorRootset CycleDetector::collectRootset() {
+  auto& detector = instance();
+  CycleDetectorRootset rootset;
+  LockGuard<SimpleMutex> guard(detector.lock_);
+  for (auto* candidate: detector.candidateList_) {
+    rootset.roots.push_back(candidate);
```
SvyatoslavScherbina, line 2829: Maybe skip non-frozen candidates here?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2708,6 +2808,118 @@ void shareAny(ObjHeader* obj) {
...
+    rootset.heldRefs.emplace_back(candidate);
+    traverseReferredObjects(candidate, [&rootset, candidate](KRef field) {
+      rootset.rootToFields[candidate].push_back(field);
+      rootset.heldRefs.emplace_back(field);
```
SvyatoslavScherbina, line 2833: I suspect a data race here: this thread reads field, other thread nullifies the field and performs GC, which removes the original field value.
We have a lock in atomic reference to protect from such a race. Using it here would probably be an overkill, but we should at least add a comment about this.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2708,6 +2808,118 @@ void shareAny(ObjHeader* obj) {
...
+    traverseFields(obj, [&toVisit, &currentPath](KRef field) {
+      auto path = currentPath;
+      path.push_back(field);
+      toVisit.emplace_back(std::move(path));
+    });
+  };
+
+  appendFieldsToVisit(root, KRefList(1, root));
+
+  KStdUnorderedSet<KRef> seen;
```
SvyatoslavScherbina, line 2867: Should we add `root` to `seen` initially?

SvyatoslavScherbina: I see: probably doesn't make sense, because if `node == root`, then we return and don't check if it is already seen.

backend.native/tests/runtime/memory/cycle_detector.kt
```diff
@@ -0,0 +1,167 @@
...
+        assertEquals(3, cycles.size)
+        assertArrayEquals(arrayOf(atomic3, atomic3.value!!, atomic3), GC.findCycle(cycles[0])!!)
+        assertArrayEquals(arrayOf(atomic2, atomic2.value!!, atomic2), GC.findCycle(cycles[1])!!)
+        assertArrayEquals(arrayOf(atomic1, atomic1), GC.findCycle(cycles[2])!!)
+    } finally {
+        atomic1.value = null
+        atomic2.value = null
+        atomic3.value = null
+    }
+}
```
SvyatoslavScherbina, line 167: Tests for `FreezableAtomicReference` would be useful.

SvyatoslavScherbina: I mean the test that would check that `FreezableAtomicReference` aren't reported as leaking cycles.

runtime/src/main/cpp/Memory.cpp
projedi: Yes, there's no need to add `root` to `seen`. Also, there's no problem in doing so. So I'll add it, since it makes a bit more sense this way.

runtime/src/main/cpp/Memory.cpp
projedi: Okay, so I left a comment for now, but I'm very unhappy about this overlooked data race.

SvyatoslavScherbina: I guess it is ok for this cycle detector tool.

PR #4288, opened 2020-07-16
Make WorkerBoundReference initially unfrozen

projedi:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2600,6 +2600,10 @@ void freezeSubgraph(ObjHeader* root) {
 
   MEMORY_LOG("Freeze subgraph of %p\n", root)
 
+  if (root->type_info() == theWorkerBoundReferenceTypeInfo) {
+    FreezeWorkerBoundReference(root);
```
SvyatoslavScherbina, line 2604: Does this work for non-root `WorkerBoundReference`, found in subgraph to be frozen?

projedi: Missed it, thanks!

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2458,6 +2464,17 @@ bool clearSubgraphReferences(ObjHeader* root, bool checked) {
   return true;
 }
 
+inline void freezeContainer(ContainerHeader* container) {
+  // Run class-specific pre-freezing code.
+  traverseContainerObjects(container, [](ObjHeader* obj) {
+    if (obj->type_info() == theWorkerBoundReferenceTypeInfo) {
+      FreezeWorkerBoundReference(obj);
```
SvyatoslavScherbina, line 2471: I'm afraid doing this here might be incorrect.
For `freezeAcyclic` it might result into some objects not being unmarked.
For `freezeCyclic` -- incorrectly freezing WBR's value etc.

Generally the entire freezing machinery doesn't expect object graph to change.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -48,10 +48,16 @@ public class WorkerBoundReference<out T : Any>(value: T) {
      * The referenced value or null if referred object is not frozen and current worker is different from the one created [this].
      */
     val valueOrNull: T?
-        get() = @Suppress("UNCHECKED_CAST") (derefWorkerBoundReference(ptr) as T?)
+        get() = if (valueBeforeFreezing != null) valueBeforeFreezing else @Suppress("UNCHECKED_CAST") (derefWorkerBoundReference(ptr) as T?)
```
SvyatoslavScherbina, line 51:
```suggestion
        get() = valueBeforeFreezing ?: @Suppress("UNCHECKED_CAST") (derefWorkerBoundReference(ptr) as T?)
```
?

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -48,10 +48,16 @@ public class WorkerBoundReference<out T : Any>(value: T) {
...
 
     /**
      * Worker that [value] is bound to.
      */
     val worker: Worker = Worker.current
+
+    @ExportForCppRuntime("Kotlin_WorkerBoundReference_doFreeze")
+    private fun doFreeze() {
```
SvyatoslavScherbina, line 59: This and other related names are kinda misleading: this function doesn't freeze anything, it just prepares the object to freezing.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -48,10 +48,16 @@ public class WorkerBoundReference<out T : Any>(value: T) {
...
+        ptr = createWorkerBoundReference(valueBeforeFreezing!!)
+        valueBeforeFreezing = null
```
SvyatoslavScherbina, line 61: The implementation should probably be idempotent, since freezing can fail after the hooks are already applied. Then next attempt to freeze will run the hooks again.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2598,6 +2631,10 @@ void freezeSubgraph(ObjHeader* root) {
   ContainerHeader* rootContainer = root->container();
   if (isPermanentOrFrozen(rootContainer)) return;
 
+  MEMORY_LOG("Run freeze hooks on subgraph of %p\n", root);
+
+  runFreezeHooksRecursive(root);
```
SvyatoslavScherbina, line 2636:
```kotlin

class A(val value: Any?)

fun main() {
    val wbr = WorkerBoundReference(42)
    val a = A(wbr)
    wbr.freeze()
    a.freeze()
}
```

runtime/src/main/cpp/Memory.cpp
```diff
@@ -2598,6 +2631,10 @@ void freezeSubgraph(ObjHeader* root) {
   ContainerHeader* rootContainer = root->container();
   if (isPermanentOrFrozen(rootContainer)) return;
 
+  MEMORY_LOG("Run freeze hooks on subgraph of %p\n", root);
+
```
SvyatoslavScherbina, line 2635: Note that running freeze hooks here is not entirely correct too:
freezing can fail (see below). In this case WBRs would be transitioned into "pre-frozen" state and make cycles with non-frozen WBRs leak.
But I suggest leaving the implementation as is for now, because failing freeze is typically used during development (to debug unexpected freezing).
Let's just document this for now.

SvyatoslavScherbina: By "document" I mean "leave a comment to the source code" of course, not the documentation for WBR :)

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: Yep, thanks for catching!

runtime/src/main/cpp/Memory.cpp
projedi: Thanks!

PR #4302, opened 2020-07-24
Use ElasticSearch as data storage for performance monitoring system

LepilkinaElena:


tools/benchmarks/shared/src/main/kotlin/report/BenchmarksReport.kt
```diff
@@ -30,50 +26,62 @@ interface JsonSerializable {
...
+    val benchmarks = structBenchmarks(benchmarksList)
 
-    override fun toJson(): String {
+    var buildNumber: String? = null
+
+    override fun serializeFields(): String {
+        val buildNumberField = buildNumber?.let {
+            """,
+            "buildNumber": "$buildNumber"
+        """
```
PavelPunegov, line 79: A broken indent here. I see that methods like `trimIndent()` are not used while writing multiline strings, so the resulting JSON would be hard to read due to inconsistant indents.

tools/benchmarksAnalyzer/src/main/kotlin/main.kt
```diff
@@ -76,11 +65,28 @@ object TeamCityConnector: Connector() {
     }
 }
 
+object DBServerConnector : Connector() {
+    override val connectorPrefix = ""
+    val serverUrl = "https://kotlin-native-perf-summary.labs.jb.gg"
```
PavelPunegov, line 70: Why are all urls hardcoded? I think it could be a property defined by TC instance. And I think for the security reasons it's better to not show the internal URL.

tools/performance-server/src/main/kotlin/database/BenchmarksIndexesDispatcher.kt
```diff
@@ -0,0 +1,323 @@
...
+            {
+                "_source": false,
+                ${featureFilter?.let {
+            """
+                "query": {
+                    "bool": {
+                        "must": [ ${it(featureValue)} ]
+                    }
+                }, """
+        } ?: ""}
```
PavelPunegov, line 257: I think making your own function like:
```
inline fun <T> T.str(block: (T) -> String): String =
   if (this != null) block(this)
   else ""
```  
may improve readability and remove cluttering with several `let` and `?: ""`

tools/performance-server/src/main/kotlin/database/BenchmarksIndexesDispatcher.kt
```diff
@@ -0,0 +1,323 @@
...
+        """
+
+        return getIndex(featureValue).search(queryDescription, listOf("aggregations")).then { responseString ->
+            val dbResponse = JsonTreeParser.parse(responseString).jsonObject
+            val aggregations = dbResponse.getObjectOrNull("aggregations") ?: error("Wrong response:\n$responseString")
+            buildNumbers?.let {
+                val buckets = aggregations.getObjectOrNull("builds")?.getObjectOrNull("buckets")
+                        ?: error("Wrong response:\n$responseString")
+                buildNumbers.map {
+                    it to listOf(buckets.getObject(it).getObject("metric_build").getObject("metric_samples").getObject("buckets").getObject("samples").getObject("geom_mean").getPrimitive("value").double)
```
PavelPunegov, line 317: I think it's better to make such chains multiline. See https://kotlinlang.org/docs/reference/coding-conventions.html#chained-call-wrapping

tools/benchmarksAnalyzer/src/main/kotlin/main.kt
LepilkinaElena: >  I think it could be a property defined by TC instance. 

It can't be, this code is also compiled and used in different way. As JS it produces on our server detailed information (frontend JS can't read some local files or variables), it also is used as command line tool on developers machine and creating extra environmental variables or files is overkilling, I think.

PR #4304, opened 2020-07-24
[Interop] Mark declaration as Deprecated if failed to import

sbogolepov:


Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -152,46 +152,44 @@ internal class ModuleMetadataEmitter(
...
-                            }
-                            km.getterAnnotations += when (kind) {
-                                is PropertyStub.Kind.Val -> kind.getter.annotations.map { it.map() }
-                                is PropertyStub.Kind.Var -> kind.getter.annotations.map { it.map() }
-                                is PropertyStub.Kind.Constant -> emptyList()
-                            }
-                            if (kind is PropertyStub.Kind.Constant) {
-                                km.compileTimeValue = kind.constant.mapToAnnotationArgument()
-                            }
+                    val (kind, annotations) = element.bridgeSupportedKind?.let { it to element.annotations }
```
SvyatoslavScherbina, line 172: We might have an additional case with properties, when getter is available but setter is not. Could you check this?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -152,46 +152,44 @@ internal class ModuleMetadataEmitter(
...
-                            km.returnType = element.returnType.map()
-                            km.uniqId = data.uniqIds.uniqIdForFunction(element)
-                        }
+                    val annotations = if (bridgeBuilderResult.nativeBridges.isSupported(element)) {
+                        element.annotations
                     } else {
-                        null
+                        listOf(AnnotationStub.Deprecated.unableToImport)
+                    }
+                    KmFunction(element.flags, element.name).also { km ->
```
SvyatoslavScherbina, line 160: Not clear for me: text emitter rendered "unsupported" declarations as non-external ones.
Metadata emitter doesn't seem to do so. Is this true? If yes, does our infrastructure detect such a difference?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
```diff
@@ -271,23 +276,21 @@ private class MappingExtensions(
...
-                        is PropertyStub.Kind.Var -> Flag.Property.IS_VAR
-                        is PropertyStub.Kind.Constant -> Flag.Property.IS_CONST
-                    }
-            ) or modality.flags
-        }
+    fun PropertyStub.getFlags(kind: PropertyStub.Kind): Flags {
+        return flagsOfNotNull(
+                Flag.IS_PUBLIC,
+                Flag.Property.IS_DECLARATION,
+                Flag.HAS_ANNOTATIONS.takeIf { annotations.isNotEmpty() },
```
SvyatoslavScherbina, line 283: What if original property doesn't have annotations, while our "unsupported" one has `@Deprecated`?

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: Nice catch: test infrastructure doesn't check flags of `KmFunction`.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/StubIrMetadataEmitter.kt
sbogolepov: d761925f77cd8c22391aabd36074a3158b7c07fd

PR #4307, opened 2020-07-28
Use ForeignException wrapper to handle native exception [KT-35056]

knebekaizer:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -713,6 +713,24 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
     }
 
+    private fun createForeignException(landingpadResult: LLVMValueRef, exceptionHandler: ExceptionHandler): LLVMValueRef {
+        val exceptionRecord = extractValue(landingpadResult, 0, "er")
+
+        // __cxa_begin_catch returns pointer to C++ exception object.
+        val exceptionRawPtr = call(context.llvm.cxaBeginCatchFunction, listOf(exceptionRecord))
+
+        // ObjC expects NSException** here, so do dereference
+        val id = LLVMBuildLoad(builder, bitcast(kInt8PtrPtr, exceptionRawPtr, ""), "")!!
```
SvyatoslavScherbina, line 723: `load` is a shortcut for `LLVMBuildLoad`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -713,6 +713,24 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
         return exceptionPtr
     }
 
+    private fun createForeignException(landingpadResult: LLVMValueRef, exceptionHandler: ExceptionHandler): LLVMValueRef {
+        val exceptionRecord = extractValue(landingpadResult, 0, "er")
+
+        // __cxa_begin_catch returns pointer to C++ exception object.
+        val exceptionRawPtr = call(context.llvm.cxaBeginCatchFunction, listOf(exceptionRecord))
+
+        // ObjC expects NSException** here, so do dereference
```
SvyatoslavScherbina, line 722: What happens with other C++ exceptions here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -713,6 +713,24 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+        val id = LLVMBuildLoad(builder, bitcast(kInt8PtrPtr, exceptionRawPtr, ""), "")!!
+
+        // This will take care of ARC - need to be done in the catching scope, i.e. before __cxa_end_catch
+        val exception = call(context.ir.symbols.createForeignException.owner.llvmFunction,
+                listOf(bitcast(kInt8Ptr, id, "")),  // cast to NativePtr aka kInt8Ptr
```
SvyatoslavScherbina, line 727: Why do we need a bitcast here? Isn't `id` already a `kInt8Ptr`?

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+import kotlin.RuntimeException
+import kotlin.native.internal.NativePtr
```
SvyatoslavScherbina, line 8: Unused import.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -713,6 +713,24 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                Lifetime.LOCAL, exceptionHandler)
```
SvyatoslavScherbina, line 728: This lifetime doesn't look correct here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -627,25 +627,25 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
             }
 
-            appendingTo(fatalForeignExceptionBlock) {
-                val exceptionRecord = extractValue(landingpad, 0)
-                call(context.llvm.cxaBeginCatchFunction, listOf(exceptionRecord))
-                terminate()
+            appendingTo(forwardNativeExceptionBlock) {
+                val exception = createForeignException(landingpad, codeContext.exceptionHandler)
+                codeContext.genThrow(exception)
+                unreachable()
```
SvyatoslavScherbina, line 648: `unreachable` shouldn't be required here, `genThrow` is expected to take care of that.

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
```diff
@@ -0,0 +1,10 @@
...
+
+public class ForeignException internal constructor(val objCException: Any?)  : RuntimeException()
```
SvyatoslavScherbina, line 10: Can `objCException` actually be `null`?

knebekaizer: There is no guarantee: think of `throw (NSException*)nullptr;` in user code.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
knebekaizer: C++ exception now safely passed and can be caught in other C++ frame.  Control flow reaches `resume`, as we don't have universal `clause null` anymore ( aka `catch (...)` ) .

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
SvyatoslavScherbina: I see. Makes sense.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
SvyatoslavScherbina: I'm not sure that it would work this way. Do Kotlin `catch` blocks safely pass C++ exceptions?
Also passing C++ exceptions through Kotlin code like this is not desirable at the moment, as we discussed earlier.

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+import kotlin.RuntimeException
+
+public class ForeignException internal constructor(val nativeException: Any?)  : RuntimeException()
```
SvyatoslavScherbina, line 9: 1. Why `RuntimeException`?
2. It makes sense to add some additional info to `ForeignException` from `nativeException`, e.g. the message.

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -44,6 +45,10 @@ fun ThrowInvalidReceiverTypeException(klass: KClass<*>): Nothing {
     throw RuntimeException("Unexpected receiver type: " + (klass.qualifiedName ?: "noname"))
 }
 
+@ExportForCppRuntime
+fun CreateForeignException(payload: NativePtr): Throwable
+        = ForeignException(interpretObjCPointerOrNull<Any?>(payload))
```
SvyatoslavScherbina, line 50: Typically we place all interop-specific code to `Interop/Runtime`.
Also `ForeignException` should probably be moved there too.

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+package kotlin.native
+
+import kotlin.RuntimeException
+
+public class ForeignException internal constructor(val objCException: Any?)  : RuntimeException()
```
SvyatoslavScherbina, line 9: If this class becomes Obj-C-specific, it should be named accordingly.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -625,28 +625,46 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
...
+                )
+                condBr(isObjCException, forwardNativeExceptionBlock, fatalForeignExceptionBlock)
+            }
 
             appendingTo(forwardKotlinExceptionBlock) {
                 // Rethrow Kotlin exception to real handler.
                 codeContext.genThrow(extractKotlinException(landingpad))
             }
 
+            appendingTo(forwardNativeExceptionBlock) {
```
SvyatoslavScherbina, line 657: This part doesn't make sense (and actually doesn't work) for non-Apple platforms yet.

runtime/src/main/kotlin/kotlin/native/ForeignException.kt
knebekaizer: The class itself if platform-agnostic. Compiler implementation currently supports NSException only but this could be extended in future. I revert field name to nativeException as it may hold generic payload.

knebekaizer: Class is supposed to be platform-agnostic. Currently implementation of compiler part creates ForeignException wrapper for NSException only and terminates on other exceptions but this may be extended in future. Renamed field to nativeExveption to make naming consistent.

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -7,6 +7,8 @@ package kotlin.native.internal
 
 import kotlin.internal.getProgressionLastElement
 import kotlin.reflect.KClass
+import kotlinx.cinterop.interpretObjCPointerOrNull
+import kotlinx.cinterop.ForeignException
```
SvyatoslavScherbina, line 11: Unused imports

runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt
```diff
@@ -43,6 +45,10 @@ fun ThrowKotlinNothingValueException(): Nothing {
 fun ThrowInvalidReceiverTypeException(klass: KClass<*>): Nothing {
     throw RuntimeException("Unexpected receiver type: " + (klass.qualifiedName ?: "noname"))
 }
+//
+//@ExportForCppRuntime
+//fun CreateForeignException(payload: NativePtr): Throwable
+//        = ForeignException(interpretObjCPointerOrNull<Any?>(payload))
```
SvyatoslavScherbina, line 51: Dead code

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/ForeignException.kt
```diff
@@ -0,0 +1,21 @@
...
+        kotlin_ObjCExport_ExceptionDetails(nativeException)
+    }?: ""
+
+    // Current implementation expects NSException type only, which is ensured by CodeGenerator.
+    @SymbolName("Kotlin_ObjCExport_ExceptionDetails")
+    private external fun kotlin_ObjCExport_ExceptionDetails(nativeException: Any): String?
+}
+
+@ExportForCppRuntime
+fun CreateForeignException(payload: NativePtr): Throwable
```
SvyatoslavScherbina, line 20: Please make it non-public.

SvyatoslavScherbina:
Tests are failing.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -306,7 +310,11 @@ internal fun KotlinStubs.generateObjCCall(
         receiver: ObjCCallReceiver,
         arguments: List<IrExpression?>
 ) = builder.irBlock {
-    val callBuilder = KotlinToCCallBuilder(builder, this@generateObjCCall, isObjCMethod = true)
+    val exceptionMode = ForeignExceptionMode.byValue(
```
SvyatoslavScherbina, line 313: Does this handle the case when manifest doesn't contain the property? This is required to make the compiler able to consume older "cinterop klibs".

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2261,11 +2262,11 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
     private fun call(function: IrFunction, llvmFunction: LLVMValueRef, args: List<LLVMValueRef>,
                      resultLifetime: Lifetime): LLVMValueRef {
-        val exceptionHandler = if (function.hasAnnotation(RuntimeNames.filterExceptions)) {
-            functionGenerationContext.filteringExceptionHandler(currentCodeContext)
-        } else {
-            currentCodeContext.exceptionHandler
-        }
+
+        val exceptionHandler = function.descriptor.annotations.findAnnotation(RuntimeNames.filterExceptions)?.let {
+            val foreignExceptionMode = ForeignExceptionMode.byValue(it.properValue("mode"))
```
SvyatoslavScherbina, line 2267: What happens if annotation is missing a mode value here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
SvyatoslavScherbina: (Should be ok, I probably misread the `byValue` implementation).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
SvyatoslavScherbina: (same as https://github.com/JetBrains/kotlin-native/pull/4307/commits/73ff5c4a5ccba17ed27a7f28a2e560d1a9815630#r484859093)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
knebekaizer: Yes it does. `ForeignExceptionMode.byValue` returns default value in that case (i.e. "terminate") . BTW this case occasionally tested a lot because all standard frameworks compiled without that option so they handled as before.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2261,11 +2262,11 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
 
     private fun call(function: IrFunction, llvmFunction: LLVMValueRef, args: List<LLVMValueRef>,
                      resultLifetime: Lifetime): LLVMValueRef {
-        val exceptionHandler = if (function.hasAnnotation(RuntimeNames.filterExceptions)) {
-            functionGenerationContext.filteringExceptionHandler(currentCodeContext)
-        } else {
-            currentCodeContext.exceptionHandler
-        }
+
+        val exceptionHandler = function.descriptor.annotations.findAnnotation(RuntimeNames.filterExceptions)?.let {
```
SvyatoslavScherbina, line 2266: Using descriptors should be avoided. Also `properValue` implementation is strange, I would avoid using it too. Please try to use `function.findAnnotation(RuntimeNames.filterExceptions)?.getAnnotationValueOrNull<String>("mode")`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -306,7 +310,11 @@ internal fun KotlinStubs.generateObjCCall(
...
+            method.module.konanLibrary?.manifestProperties?.getProperty(ForeignExceptionMode.manifestKey)
```
SvyatoslavScherbina, line 314: A compiler's IR "feature" might affect this code in a harmful way.

There are "fake overrides".  When you call a method that is defined in a superclass, compiler IR encodes it as a call to a "fake override" in a subclass. E.g.
```kotlin
class NSObjectImpl : NSObject()

fun main() {
    NSObjectImpl().description()
}
```
In this case the compiler synthesizes "fake override" method `description` in `NSObjectImpl`, and IR contains `IrCall` to this `NSObjectImpl.description`. So the code here may accidentally inspect this method and query the manifest for the library containing this method, which is not even an interop library.

This is usually addressed with `method.resolveFakeOverride(allowAbstract = true)`.

backend.native/tests/interop/objc/allocException.kt
```diff
@@ -18,7 +18,13 @@ fun exc_handler(x: Any?) : Unit {
 }
 
 fun main() {
+    // This does not work anymore, as NSException propagated as ForeignException
```
SvyatoslavScherbina, line 21: I guess the new test version works with old behaviour too, but can be reverted now.
Please don't forget to revert if you find it necessary.

SvyatoslavScherbina (approved):

backend.native/tests/interop/objc/allocException.kt
knebekaizer: I'd rather keep it as is. This test is intended to validate that snippets runs "as expected", as opposite to segmentation fault (as it was before), or another OS level fault. Both variants are expected now with different cinterop options, so both cases are accepted in this test. Let's have another test to validate different behavior with different options.

SvyatoslavScherbina (approved):

PR #4308, opened 2020-07-28
[codegenerator] wrap local function generation in original lambda's file scope

vvlevchenko:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -706,27 +706,33 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-                                            "unsupported call of reified inlined function `${declaration.fqNameForIrSerialization}`").llvm),
-                                    Lifetime.IRRELEVANT)
-                            return@usingVariableScope
-                        }
-                        when (body) {
-                            is IrBlockBody -> body.statements.forEach { generateStatement(it) }
-                            is IrExpressionBody -> generateStatement(body.expression)
-                            is IrSyntheticBody -> throw AssertionError("Synthetic body ${body.kind} has not been lowered")
-                            else -> TODO(ir2string(body))
+        val isLocalFunction = declaration.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA
```
homuroll, line 709: So, only inline lambdas can reference other file?

vvlevchenko: If it referenced from this file we just do additional switch to same file. perhaps it better to avoid of course, but don't break anything.

homuroll: So, may be the name for the variable should be `isInlineLambda`?

vvlevchenko: i could be misleading because it's lmbda but not inlined one. So maybe `isNotInlinedLambda`?

PR #4310, opened 2020-07-28
Compile cpp code with -Wall

projedi:


runtime/src/main/cpp/KString.cpp
```diff
@@ -44,7 +44,7 @@ OBJ_GETTER(utf8ToUtf16Impl, const char* rawString, const char* end, uint32_t cha
   if (rawString == nullptr) RETURN_OBJ(nullptr);
   ArrayHeader* result = AllocArrayInstance(theStringTypeInfo, charCount, OBJ_RESULT)->array();
   KChar* rawResult = CharArrayAddressOfElementAt(result, 0);
-  auto convertResult = conversion(rawString, end, rawResult);
+  conversion(rawString, end, rawResult);
```
SvyatoslavScherbina, line 47: Does the compiler complain about ignoring return value?

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1511,32 +1509,6 @@ inline void releaseHeapRef(const ObjHeader* header) {
     releaseHeapRef<Strict>(const_cast<ContainerHeader*>(container));
 }
 
-// We use first slot as place to store frame-local arena container.
```
SvyatoslavScherbina, line 1514: Maybe put these under `#if 0`?

runtime/src/main/cpp/KString.cpp
projedi: Unused variable `convertResult`.

SvyatoslavScherbina: I get this. I mean, if we remove this variable, is the compiler totally ok with ignoring return value?

projedi: Well, the function is not marked [`nodiscard`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) and has a trivial destructor, so it's ok to ignore the return value.

SvyatoslavScherbina: I see. Thanks for explanation!

runtime/src/main/cpp/Memory.cpp
projedi: I disabled `-Wunused-function` for now.

common/src/hash/cpp/Base64.cpp
```diff
@@ -136,7 +136,10 @@ int DecodeBase64(
   size_t buf = 0, len = 0;
 
   while (in < end) {
-    unsigned char c = kDecode[*in++];
+    // In base64 characters are guaranteed to be ASCII, and so are positive
+    // even if char is a signed type.
```
knebekaizer, line 140: Consider runtime check, and terminate if the condition fails. It is a public function, so `in` may be an arbitrary byte array out of our control, not Base64 sequence. In fact, there is no guarantee in caller code, see HashUtils.kt.

knebekaizer, line 140: There is no guarantee it is actually Base64. It is safe anyway because kDecode has 256 elements, but the comment is a bit mispointing. Consider rewording.

projedi: Good catch, thanks! Reworded.

PR #4311, opened 2020-07-29
[KT-40508] Respect libraryPaths in metadata interop

sbogolepov:


backend.native/tests/interop/embedStaticLibraries/embedStaticLibraries.def
```diff
@@ -1,3 +1,2 @@
-staticLibraries = \
-    backend.native/tests/build/embedStaticLibraries/3.a \
-    backend.native/tests/build/embedStaticLibraries/4.a
+libraryPaths = backend.native/tests/build/embedStaticLibraries
```
SvyatoslavScherbina, line 1: It would be great to test getting library path from both def file and arguments (new tests aren't necessary, patching this test should be enough). But it's ok do this later in a separate PR.

PR #4314, opened 2020-07-30
Added extra GC call before analyzing subgraph during freeze operation…

LepilkinaElena:
…s to decrease analyzed objects number

### DIFF
--- a/runtime/src/main/cpp/Memory.cpp
+++ b/runtime/src/main/cpp/Memory.cpp
@@ -1679,7 +1679,7 @@ void garbageCollect(MemoryState* state, bool force) {
       GC_LOG("||| GC: collectCyclesDuration = %lld\n", cyclicGcEndTime - cyclicGcStartTime);
     #endif
     auto cyclicGcDuration = cyclicGcEndTime - cyclicGcStartTime;
-    if (state->gcErgonomics && cyclicGcDuration > kGcCollectCyclesMinimumDuration &&
+    if (!force && state->gcErgonomics && cyclicGcDuration > kGcCollectCyclesMinimumDuration &&
         double(cyclicGcDuration) / (cyclicGcStartTime - state->lastCyclicGcTimestamp + 1) > kGcCollectCyclesLoadRatio) {
       increaseGcCollectCyclesThreshold(state);
       GC_LOG("Adjusting GC collecting cycles threshold to %lld\n", state->gcCollectCyclesThreshold);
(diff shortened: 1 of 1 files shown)


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2394,6 +2404,9 @@ bool clearSubgraphReferences(ObjHeader* root, bool checked) {
   auto state = memoryState;
   auto* container = root->container();
 
+  // Free cyclic garbage to decrease number of analyzed objects.
+  checkIfForceCyclicGcNeeded(state);
```
SvyatoslavScherbina, line 2408: It makes sense to place this after  `if (isShareable(container))`.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1997,6 +1997,16 @@ inline void checkIfGcNeeded(MemoryState* state) {
   }
 }
 
+inline void checkIfForceCyclicGcNeeded(MemoryState* state) {
+  if (state != nullptr && state->toFree->size() > kMaxToFreeSizeThreshold) {
+    // To avoid GC trashing check that at least 10ms passed since last GC.
+    if (konan::getTimeMicros() - state->lastGcTimestamp > 10 * 1000) {
+      GC_LOG("Calling GC from checkIfForceCyclicGcNeeded: %d\n", state->toFree->size())
+      garbageCollect(state, true);
```
SvyatoslavScherbina, line 2005: I guess forced cycle collection shouldn't trigger increasing collect cycles threshold in `garbageCollect`.

LepilkinaElena: Only collect cycles threshold ?
I now think about same logic for gcThreshold.

SvyatoslavScherbina: Well, makes sense.

LepilkinaElena: I made such logic only for increasing based on time.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1997,6 +1997,16 @@ inline void checkIfGcNeeded(MemoryState* state) {
   }
 }
 
+inline void checkIfForceCyclicGcNeeded(MemoryState* state) {
+  if (state != nullptr && state->toFree->size() > kMaxToFreeSizeThreshold) {
```
SvyatoslavScherbina, line 2001: It might make sense to check `toFree != nullptr` too.

PR #4315, opened 2020-07-31
Objc header generation public api

aydarin:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjcExportHeaderGeneratorMobile.kt
```diff
@@ -0,0 +1,46 @@
...
+        private val warningCollector: ObjCExportWarningCollector,
+        objcGenerics: Boolean
+) : ObjCExportHeaderGenerator(moduleDescriptors, mapper, namer, objcGenerics) {
+
+    companion object {
+        fun createInstance(
+                configuration: ObjCExportLazy.Configuration,
+                warningCollector: ObjCExportWarningCollector,
+                builtIns: KotlinBuiltIns,
+                objcGenerics: Boolean,
```
SvyatoslavScherbina, line 21: `objcGenerics` is also available in `configuration`.

PR #4317, opened 2020-07-31
[test][debugger] non debug compilation with debug static framework should contains resolvable symbols

vvlevchenko:


backend.native/tests/debugger/src/test/kotlin/org/jetbrains/kotlin/native/test/debugger/Driver.kt
```diff
@@ -58,6 +66,11 @@ class ToolDriver(
         val out = dwarfProcess.takeIf { it.process.exitValue() == 0 }?.stdout ?: error(dwarfProcess.stderr)
         DwarfUtilParser().parse(StringReader(out)).tags.toList().processor()
     }
+
+    fun swiftc(output: Path, swiftSrc: Path, vararg args: String) {
```
PavelPunegov, line 70: I think `swiftc` here should specify target in the same way as in https://github.com/JetBrains/kotlin-native/blob/ddc9428d4f59004dce55a30cb5860bff26acc6e5/build-tools/src/main/kotlin/org/jetbrains/kotlin/Utils.kt#L231
Also what about bitcode support?

vvlevchenko: That would be good, not to pass project everywhere but at least have it for logging and so on, but unfortunately test infrastructure designed in the way that we totaly unaware about gradle at this level :/

PR #4319, opened 2020-08-04
Compile C++ code with -Wextra

projedi:


runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -71,7 +71,7 @@ CSSourceInfoRef (*CSSymbolOwnerGetSourceInfoWithAddress)(
 
 const char* (*CSSourceInfoGetPath)(CSSourceInfoRef info);
 
-uint32_t (*CSSourceInfoGetLineNumber)(CSSourceInfoRef info);
+int32_t (*CSSourceInfoGetLineNumber)(CSSourceInfoRef info);
```
projedi, line 74: [Here](https://opensource.apple.com/source/IOKitTools/IOKitTools-97/ioclasscount.tproj/ioclasscount.c) it's used with `printf`'s `%d`, so it's probably signed.

SvyatoslavScherbina: On the other hand: https://github.com/mozilla/positron/blob/4b0595d06faa1617bd989a8c47231576f30ea6b2/widget/cocoa/nsCocoaDebugUtils.h#L77

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -417,7 +414,7 @@ void Kotlin_ByteArray_setFloatAt(KRef thiz, KInt index, KFloat value) {
 
 void Kotlin_ByteArray_setDoubleAt(KRef thiz, KInt index, KDouble value) {
   ArrayHeader* array = thiz->array();
-  if (static_cast<uint32_t>(index + 7) >= array->count_) {
+  if (index < 0 || static_cast<uint32_t>(index) + 7 >= array->count_) {
```
SvyatoslavScherbina, line 417: Please ensure that we have tests for the code changed in this file.

runtime/src/main/cpp/KString.cpp
```diff
@@ -752,8 +754,9 @@ OBJ_GETTER(Kotlin_String_plusImpl, KString thiz, KString other) {
   RuntimeAssert(other != nullptr, "other cannot be null");
   RuntimeAssert(thiz->type_info() == theStringTypeInfo, "Must be a string");
   RuntimeAssert(other->type_info() == theStringTypeInfo, "Must be a string");
-  KInt result_length = thiz->count_ + other->count_;
-  if (result_length < thiz->count_ || result_length < other->count_) {
+  uint32_t result_length = thiz->count_ + other->count_;
+  if (result_length < thiz->count_ || result_length < other->count_ ||
+      result_length > static_cast<uint32_t>(std::numeric_limits<int32_t>::max())) {
```
SvyatoslavScherbina, line 759: Can we reduce the number of checks here?

runtime/src/main/cpp/Natives.cpp
```diff
@@ -49,7 +49,7 @@ OBJ_GETTER0(Kotlin_native_internal_undefined) {
 }
 
 void* Kotlin_interop_malloc(KLong size, KInt align) {
-  if (size > SIZE_MAX) {
+  if (size < 0 || static_cast<size_t>(size) > std::numeric_limits<size_t>::max()) {
```
SvyatoslavScherbina, line 52: Is this correct?
`KLong` might be longer than `size_t`, so `static_cast<size_t>(size)` would drop significant bits in this case, wouldn't it?

runtime/src/debug/cpp/SourceInfo.cpp
projedi: A couple of experiments showed that for files larger than 3 billion lines of code, some debug info (notably, symbol location) isn't being read (or not even generated, never checked). For files with 2 billion lines of code, everything is fine.

Nevertheless, restored `uint32_t` here and added loud checks if it ever overflows `int32_t` boundaries.

runtime/src/main/cpp/Natives.cpp
projedi: That was me being silly. Added a cast to an appropriate unsigned type.

runtime/src/main/cpp/KString.cpp
projedi: Good catch, thanks!

runtime/src/debug/cpp/SourceInfo.cpp
```diff
@@ -156,6 +156,7 @@ extern "C" struct SourceInfo Kotlin_getSourceInfo(void* addr) {
         uint32_t lineNumber = CSSourceInfoGetLineNumber(ref);
         if (lineNumber == 0)
           return 0;
+        RuntimeAssert(lineNumber <= static_cast<uint32_t>(std::numeric_limits<int32_t>::max()), "Unexpectedly huge line number");
```
SvyatoslavScherbina, line 159: I'm not sure we should have such assertions in this code.
Motivation: on the one hand, the entire approach used in this code is fragile, so we actually can get somewhat incorrect data. On the other hand, this code is used in error handling. If we allow incorrect data, it will be displayed in stack traces, giving us more information than if we just crash here.

projedi: Makes sense, removed the asserts.

runtime/src/main/cpp/KString.cpp
```diff
@@ -734,12 +736,12 @@ OBJ_GETTER(Kotlin_String_replace, KString thiz, KChar oldChar, KChar newChar, KB
   KChar* resultRaw = CharArrayAddressOfElementAt(result, 0);
   if (ignoreCase) {
     KChar oldCharLower = towlower_Konan(oldChar);
-    for (KInt index = 0; index < count; ++index) {
+    for (uint32_t index = 0; index < count; ++index) {
```
knebekaizer, line 739: [Minor] Again, it's a bit susceptible regarding (nvisible) type of count_.
`decltype(thiz->count_)` is exact but noisy. Other variants may be simple `while (count--)` or possibly less fragile `for (auto end = thizRaw + thiz->count_; thizRaw < end; )`

runtime/src/main/cpp/Arrays.cpp
projedi: Good thing you noticed: we didn't have enough coverage indeed. I've added tests on out-of-bounds indexing and on out-of-bounds `addressOf` (on pinned objects). The former found problems with the `ByteArray::(get|set)(Type)At` implementation missing `index < 0`, and the latter also found a lack of `boundsCheck` [here](https://github.com/JetBrains/kotlin-native/pull/4319/files#diff-2fdb85cd4e883cd81e8bb6abdec07409R664)

common/src/hash/cpp/Names.cpp
```diff
@@ -29,7 +29,7 @@ constexpr uint32_t PrintableHexSize(uint32_t input_length) {
 
 void PrintableHex(const uint8_t* data, uint32_t data_length, char* hex) {
   static const char* hex_digits = "0123456789ABCDEF";
-  for(int i = 0; i < data_length; ++i) {
+  for (uint32_t i = 0; i < data_length; ++i) {
```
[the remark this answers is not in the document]
projedi: These for loops need to be improved indeed. Maybe a range-for would even be better here? Something like `for (auto i: iota(0, length)) { … }` where `iota` is something like [this](https://en.cppreference.com/w/cpp/ranges/iota_view) from C++20. Unfortunately the standard `iota` wouldn't do, because `iota` allows bounds to have different types and so the type of `i` would be `int`, but we can create our own simple iterator that would enforce bounds type equality.

At any rate, I suggest to postpone this refactoring a bit. At least with the current warning setup, we would get a compile time error if `data_length` becomes signed.

runtime/src/main/cpp/KString.cpp
projedi: Let's postpone for now, answered [here](https://github.com/JetBrains/kotlin-native/pull/4319#discussion_r468648150)

backend.native/tests/runtime/collections/array5.kt
```diff
@@ -0,0 +1,814 @@
...
+    }
+    assertFailsWith<ArrayIndexOutOfBoundsException> {
+        arr.setUByteAt(Int.MIN_VALUE, 1U)
+    }
+}
+
+@Test fun byteArrayGetChar() {
+    val arr = ByteArray(10) { 0 }
+    assertFailsWith<ArrayIndexOutOfBoundsException> {
+        arr.getCharAt(10)
```
SvyatoslavScherbina, line 461: Note that `10` is not a minimal positive index causing `ArrayIndexOutOfBoundsException`. `9` should too.
Same below.

projedi: Thanks, added some positive cases as well.

PR #4320, opened 2020-08-04
[caches] Restored the topological order of libraries' initializers

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/ContextUtils.kt
```diff
@@ -438,6 +438,9 @@ internal class Llvm(val context: Context, val llvmModule: LLVMModuleRef) {
             context.config.cachedLibraries.getLibraryCache(it) == null
         }
         val set = (allNonCachedDependencies + allCachedBitcodeDependencies).toSet()
+        // This list is used in particular to build the libraries' initializers chain.
+        // The initializers must be called in the topological order, so make sure that the
+        // libraries list being returned is also toposorted.
```
SvyatoslavScherbina, line 443: A comment is required for `appendStaticInitializers` too.

PR #4322, opened 2020-08-09
Optimize vtable building [KT-40652]

LepilkinaElena:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/ClassLayoutBuilder.kt
```diff
@@ -286,6 +286,7 @@ internal class ClassLayoutBuilder(val irClass: IrClass, val context: Context, va
 
         val methods = irClass.sortedOverridableOrOverridingMethods
         val newVtableSlots = mutableListOf<OverriddenFunctionInfo>()
+        val overridedVtableSlots = mutableMapOf<IrSimpleFunction, OverriddenFunctionInfo>()
```
homuroll, line 289: overrided -> overridden (it is an irregular verb).

PR #4325, opened 2020-08-11
Disable CycleDetector on WASM

projedi:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -164,6 +164,10 @@ class ScopedRefHolder {
   KRef obj_ = nullptr;
 };
 
+// CycleDetector internally uses static local with runtime initialization,
+// which requires atomics. Atomics are not available on WASM.
+#ifndef KONAN_WASM
```
SvyatoslavScherbina, line 169: Maybe define a dedicated macro, like `USE_CYCLE_DETECTOR`?

PR #4335, opened 2020-08-19
Untie Kotlin/Native from predefined LLVM distribution

sbogolepov:


vvlevchenko (approved):

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/TargetProperties.kt
```diff
@@ -19,19 +19,67 @@ package org.jetbrains.kotlin.konan.properties
 import org.jetbrains.kotlin.konan.target.*
 
 fun Properties.hostString(name: String, host: KonanTarget): String?
-    = this.propertyString(name, host.name)
+    = this.resolvablePropertyString(name, host.name)
```
ilmat192, line 22: We also access `llvmHome` in [Gradle plugin](https://github.com/JetBrains/kotlin/blob/4794297640db25923b53d4f471c98418f79888af/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/nativeToolRunners.kt#L107) to add LLVM's DLLs to `PATH`. We will need to expand $-values there too.

So may be move the `resolvableProperty...` API to the main Kotlin repo near to `propertyList` and `propertyString` ?

We also provide access to these properties via [`Distribution.properties`](https://github.com/JetBrains/kotlin/blob/4794297640db25923b53d4f471c98418f79888af/native/utils/src/org/jetbrains/kotlin/konan/target/Distribution.kt#L46) so this API may be useful for other users e.g. for IDE.

PR #4336, opened 2020-08-19
Handle some special characters in deprecation message in ObjCExport

SvyatoslavScherbina:
Fix [KT-39206](https://youtrack.jetbrains.com/issue/KT-39206).

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
@@ -1237,7 +1237,23 @@ private fun Deprecation.toDeprecationAttribute(): String {
     // TODO: consider avoiding code generation for unavailable.
 
     val message = this.message.orEmpty()
-            .replace("\n", "\\n")
 
-    return "$attribute(\"$message\")"
+    return "$attribute(${quoteAsCStringLiteral(message)})"
+}
+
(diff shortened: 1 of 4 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1237,6 +1237,7 @@ private fun Deprecation.toDeprecationAttribute(): String {
     // TODO: consider avoiding code generation for unavailable.
 
     val message = this.message.orEmpty()
+            .replace("\n", "\\n")
```
knebekaizer: OK as a temporary solution. Any other escaped symbol still fails.
TODO: Consider another approach: getting message "as is"  is dangerous and brute force consistent escaping is complicate.

knebekaizer (approved):
OK as a temporary solution. Any other escaped symbol still fails.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
SvyatoslavScherbina: > OK as a temporary solution.

Exactly! That's why I've added a second commit that handles a couple of other symbols. Have you reviewed it?

> TODO: Consider another approach: getting message "as is" is dangerous and brute force consistent escaping is complicate.

Could you clarify?

knebekaizer (approved):

PR #4340, opened 2020-08-24
Turn on common RangeContainsLowering

LepilkinaElena:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
```diff
@@ -225,6 +225,12 @@ internal val innerClassPhase = makeKonanFileLoweringPhase(
         prerequisite = setOf(defaultParameterExtentPhase)
 )
 
+internal val rangeContainsLoweringPhase = makeKonanFileLoweringPhase(
```
homuroll, line 228: Shouldn't we mark `rangeContainsLoweringPhase` a prerequisite for `forLoopsLoweringPhase`?

LepilkinaElena: I don't see such dependency in JVM. Why should they be connected? Current optimization isn't connected with ranges in loops. And `ForLoopsLowering` replace them with induction variables.

homuroll (approved):

PR #4342, opened 2020-08-25
Fix macOS minimum version in README

LouisCAD:
10.11 is El Capitan while 11.0 is still in beta Big Sur.


SvyatoslavScherbina (approved):
Thanks!

PR #4346, opened 2020-09-01
Equals sam function references

homuroll:
Fixes https://youtrack.jetbrains.com/issue/KT-39800 and https://youtrack.jetbrains.com/issue/KT-39798


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
```diff
@@ -159,7 +158,7 @@ internal class BridgesBuilding(val context: Context) : ClassLoweringPass {
                 val irBuilder = context.createIrBuilder(declaration.symbol, declaration.startOffset, declaration.endOffset)
                 declaration.body = irBuilder.irBlockBody(declaration) {
                     buildTypeSafeBarrier(declaration, declaration, typeSafeBarrierDescription)
-                    body.statements.forEach { +it }
+                    (body as IrBlockBody).statements.forEach { +it }
```
SvyatoslavScherbina, line 161: This lowering is before returns insertion, so the former shouldn't be affected by the latter. What is this patch for then?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReturnsInsertionLowering.kt
```diff
@@ -31,11 +32,17 @@ internal class ReturnsInsertionLowering(val context: Context) : FileLoweringPass
...
-                val body = declaration.body ?: return
-                if (declaration is IrConstructor || declaration.returnType == context.irBuiltIns.unitType) {
-                    val irBuilder = context.createIrBuilder(declaration.symbol, declaration.endOffset, declaration.endOffset)
-                    irBuilder.run {
-                         (body as IrBlockBody).statements += irReturn(irGetObject(symbols.unit))
+                context.createIrBuilder(declaration.symbol, declaration.endOffset, declaration.endOffset).run {
+                    when (val body = declaration.body) {
+                        is IrExpressionBody -> {
+                            declaration.body = IrBlockBodyImpl(body.startOffset, body.endOffset) {
+                                statements += irReturn(body.expression)
```
SvyatoslavScherbina, line 39: What exactly is this required for?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -365,17 +365,17 @@ internal val allLoweringsPhase = NamedCompilerPhase(
                             localFunctionsPhase,
                             foldConstantLoweringPhase,
                             tailrecPhase,
+                            singleAbstractMethodPhase,
```
SvyatoslavScherbina: What is this reording required for?

runtime/src/main/kotlin/kotlin/native/internal/KFunctionImpl.kt
```diff
@@ -7,19 +7,28 @@ package kotlin.native.internal
 
 import kotlin.reflect.KFunction
 import kotlin.reflect.KType
+import kotlin.reflect.KClass
```
SvyatoslavScherbina, line 10: Unused import

runtime/src/main/kotlin/kotlin/native/internal/KFunctionImpl.kt
```diff
@@ -7,19 +7,28 @@ package kotlin.native.internal
...
     override fun equals(other: Any?): Boolean {
         if (other !is KFunctionImpl<*>) return false
-        return fqName == other.fqName && bound == other.bound && receiver == other.receiver
+        return fqName == other.fqName && receiver == other.receiver
+                && arity == other.arity && flags == other.flags
     }
 
-    override fun hashCode(): Int {
-        return (fqName.hashCode() * 31 + if (bound) 1 else 0) * 31 + receiver.hashCode()
+    private fun horner(x: Int, vararg coeffs: Int): Int {
```
SvyatoslavScherbina, line 23: The name is kinda confusing.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -94,6 +94,32 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
...
+            // This is needed not only for optimization purposes (producing one less class)
+            // but also for implementing correct equals/hashCode methods.
+            override fun visitTypeOperator(expression: IrTypeOperatorCall): IrExpression {
+                if (expression.operator == IrTypeOperator.SAM_CONVERSION) {
+                    val invokable = expression.argument
+                    val reference = if (invokable is IrFunctionReference) {
+                        invokable
+                    } else if (invokable is IrBlock && (invokable.origin.isLambda)
+                            && invokable.statements.last() is IrFunctionReference) {
+                        invokable.statements.dropLast(1).forEach { it.transform(this, null) }
```
SvyatoslavScherbina, line 112: This code assumes that `transform` returns the original object, which is not always true.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -240,12 +301,32 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
...
+                buildInvokeMethod(suspendFunctionClass.getInvokeFunction()).also {
+                    if (isKSuspendFunction)
+                        it.overriddenSymbols += functionClass.getInvokeFunction().symbol
+                }
+            }
+            samSuperType?.let { superTypes += it }
+            val sam = samSuperClass?.functions?.single { it.owner.modality == Modality.ABSTRACT }
+            if (sam != null) {
+                if (sam.owner.extensionReceiverParameter != null)
+                    buildInvokeMethod(sam.owner)
```
SvyatoslavScherbina, line 320: Why can't we reuse the existing invoke in this case as below?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -94,6 +94,32 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
...
             }
 
+            // Handle SAM conversions which wrap a function reference:
+            //     class sam$n(private val receiver: R) : Interface { override fun method(...) = receiver.target(...) }
+            //
+            // This is actually very common, as `Interface { something }` is a local function + a SAM-conversion
+            // of a reference to it into an implementation.
+            //
+            // This is needed not only for optimization purposes (producing one less class)
+            // but also for implementing correct equals/hashCode methods.
```
SvyatoslavScherbina, line 104: This part is not exactly clear.
Does equals/hashCode work correctly for SAM wrappers over anything except function references, i.e. when this optimization is not applied? It should. If it does, then why the optimization here is required for equals/hashCode?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
homuroll: Because earlier we assumed that `IrExpressionBody` is not present in IR anywhere. This is not true anymore. Here we're lowering that condition to only the functions for which `typeSafeBarrierDescription` is not null.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReturnsInsertionLowering.kt
homuroll: Some of the common lowerings produce `IrExpressionBody`, JVM/JS correctly handle it, so why wouldn't we do the same?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: For handling SAM conversions to function references (which is done in `FunctionReferenceLowering`).

runtime/src/main/kotlin/kotlin/native/internal/KFunctionImpl.kt
homuroll: Sure, I'll rename to a more general `evaluatePolynom`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
homuroll: We can, but some work needs to be done in bridges building algorithm - it doesn't expect extension receiver to hop into value parameters.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
homuroll: Yes, equals/hashCode work correctly for SAM wrappers, but the thing is that if we perform SAM conversion after function references lowering, then for function references to the same function two different implementations will be generated and, hence, references to their invoke functions (which will go into SAM wrappers) will be different as well.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
SvyatoslavScherbina: Do you mean that if `typeSafeBarrierDescription` is not null, then the body is not `IrExpressionBody`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReturnsInsertionLowering.kt
SvyatoslavScherbina: I'm trying to ask: which lowerings? It is completely not clear for me how is this related to the main purpose of the PR. Does it enable these lowerings producing `IrExpressionBody`? Or does this patch fix unrelated bugs?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
SvyatoslavScherbina: This explains placing `singleAbstractMethodPhase` after `functionReferencePhase`.
I'm asking about moving bunch of lowerings in e3386e26b15a23f55194683cde4ac6366a4237af.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
SvyatoslavScherbina: What do you mean by "two different implementations will be generated"? What is an implementation here? Could you provide an example?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -94,6 +94,34 @@ internal class CallableReferenceLowering(val context: Context): FileLoweringPass
...
+            override fun visitTypeOperator(expression: IrTypeOperatorCall): IrExpression {
+                if (expression.operator == IrTypeOperator.SAM_CONVERSION) {
+                    val invokable = expression.argument
+                    val reference = if (invokable is IrFunctionReference) {
+                        invokable
+                    } else if (invokable is IrBlock && (invokable.origin.isLambda)
+                            && invokable.statements.last() is IrFunctionReference) {
+                        val statements = invokable.statements
+                        for (i in 0 until statements.size - 1)
```
SvyatoslavScherbina, line 113: Btw, the code below returns `transformFunctionReference(reference, expression.typeOperand)`. Does it drop all statements but last in this case then?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReturnsInsertionLowering.kt
homuroll: SAM lowering, in particular (equals/hashCode from it, which weren't generated before this PR).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/BridgesBuilding.kt
homuroll: Yes.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
homuroll: I tried to remove it and all tests still pass, looks like it is not needed (that's weird, because the reason why I added it was falling tests, may be some other things fixed them). Still, as an optimization to not build two wrappers (one over function reference and the other - for SAM conversion) it is still valid.

homuroll: Ah, I figured out why everything works: SAM wrapper in `equals` calls other equals which is `KFunctionImpl.equals', which works correctly in this case.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
homuroll: Huh, indeed, this means that there is a dependency on LocalFunctionsLowering here. Because actually there are exactly two statements here: IrComposite (that was left of IrFunction) and IrFunctionReference.

SvyatoslavScherbina: Please either don't rely on this here or check this explicitly (with proper comment).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
SvyatoslavScherbina: > Still, as an optimization to not build two wrappers (one over function reference and the other - for SAM conversion) it is still valid.

I agree. But this requires different comment.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
SvyatoslavScherbina: I see. Ok, let's keep it this way for now.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
homuroll: Looks like other movements not needed (all the relevant tests seem to be passing), probably this was an artefact of some experiments.

homuroll: Reordered. The final order reflects this chain of dependencies: BuiltInOperatorLowering -> SAMLowering -> FunRefLowering -> LocalDeclLowering.

SvyatoslavScherbina: Should this affect `testProcessorPhase`?

SvyatoslavScherbina: Ah, I see.

SvyatoslavScherbina (approved):

PR #4348, opened 2020-09-02
[IR] Adopted FlattenStringConcatenation lowering

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanLoweringPhases.kt
```diff
@@ -359,5 +365,6 @@ internal val ifNullExpressionsFusionPhase = makeKonanFileLoweringPhase(
 internal val foldConstantLoweringPhase = makeKonanFileOpPhase(
         { context, irFile -> FoldConstantLowering(context).lower(irFile) },
         name = "FoldConstantLowering",
-        description = "Constant Folding"
+        description = "Constant Folding",
+        prerequisite = setOf(flattenStringConcatenationPhase)
```
SvyatoslavScherbina, line 369: Why so?

homuroll: Hmm, it's actually not that necessary, just `FoldConstantLowering` folds more string concatenations after `FlattenStringConcatenation`.

SvyatoslavScherbina (approved):

PR #4349, opened 2020-09-03
[Linker] Pre-link static caches

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -170,8 +175,36 @@ internal class Linker(val context: Context) {
...
+        val config = context.config
+        return when {
+            config.produce == CompilerOutputKind.STATIC_CACHE -> {
+                // Do not link static cache dependencies.
+                LinkerInput(objectFiles, CachesToLink(emptyList(), caches.dynamic), emptyList(), cachingInvolved)
+            }
+            config.cacheSupport.disableCachesPreLink || caches.static.isEmpty() -> {
+                LinkerInput(objectFiles, caches, emptyList(), cachingInvolved)
+            }
+            else -> {
```
SvyatoslavScherbina, line 192: We don't actually need prelinking caches when producing anything but static libraries/frameworks.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -186,11 +195,18 @@ open class MacOSBasedLinker(targetProperties: AppleConfigurables)
 
     override fun filterStaticLibraries(binaries: List<String>) = binaries.filter { it.isUnixStaticLib }
 
-    override fun linkCommands(objectFiles: List<ObjectFile>, executable: ExecutableFile,
-                              libraries: List<String>, linkerArgs: List<String>,
-                              optimize: Boolean, debug: Boolean, kind: LinkerOutputKind,
-                              outputDsymBundle: String,
-                              needsProfileLibrary: Boolean): List<Command> {
+    override fun preLinkCommands(objectFiles: List<ObjectFile>, output: ObjectFile): List<Command> =
+        Command(linker).apply {
```
SvyatoslavScherbina, line 200: I've encountered some issues when not specifying some of the common flags like -arch for ld -r. We don't have caches for non-x86_64-arches though.

Have you checked ld flags when Xcode is prelinking?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -300,6 +300,19 @@ private fun selectFrameworkType(
...
 
+private fun parsePreLinkCachesValue(
+        configuration: CompilerConfiguration,
+        value: String?
+): Boolean? = when (value) {
+        "enable" -> true
+        "disable" -> false
+        null -> null
+        else -> {
+            configuration.report(ERROR, "Unsupported `-pre-link-caches` value: $value. Possible values are 'enable'/'disable'")
```
SvyatoslavScherbina, line 311: Should be an `-X*` flag too.

SvyatoslavScherbina (approved):

PR #4351, opened 2020-09-03
Use LazyIR for cached interop classes

sbogolepov:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/EnumSpecialDescriptorsFactory.kt
```diff
@@ -8,38 +8,146 @@ package org.jetbrains.kotlin.backend.konan
...
+                    .associate { it.value.name to it.index }
+                    .toMap()
+
+    private fun findItemGetterSymbol(): IrSimpleFunctionSymbol =
+            symbols.array.functions.single { it.descriptor.name == Name.identifier("get") }
+
+    private fun valuesArrayType(enumClass: IrClass): IrType =
+            symbols.array.typeWith(enumClass.defaultType)
+
+    // We can't move property getter to the top-level scope.
```
homuroll, line 109: Why is that, BTW?

sbogolepov: Accessors should have the same parent as the property. IR validator catches it. 
Also it would a really surprising behaviour for IR-processing code :)

homuroll: I meant to turn it from being a getter to being a usual function and probably removing the property also. But looks like it will be an overkill anyway.

homuroll (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/EnumSpecialDescriptorsFactory.kt
sbogolepov: Ah, ok. Yes, it would make the PR needlessly big.

PR #4352, opened 2020-09-03
[debug][bindings][python] don't build synthetic for null values (KT-41291)

vvlevchenko:


kishmakov:
This fix fixes one scenario. Still has different scenario, where it is not enough.

PR #4362, opened 2020-09-08
Generalize finalization with Cleaner

projedi:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -1193,6 +1213,21 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
                         putValueArgument(3, jobPointer)
                     }
                 }
+                IntrinsicType.CREATE_CLEANER -> {
+                    val irCallableReference = expression.getValueArgument(1)
+                    if (irCallableReference == null || !irCallableReference.isNonCapturingLambda()) {
+                        context.reportCompilationError(
```
SvyatoslavScherbina, line 1219: @homuroll FYI: one more compilation error reported from backend, please make sure it doesn't get lost when merging this PR with your one.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+ * @param block must not capture anything
+ */
+// TODO: Consider just annotating the lambda argument rather than using intrinsic.
+@TypedIntrinsic(IntrinsicType.CREATE_CLEANER)
+external fun <T> createCleaner(argument: T, block: (T) -> Unit): Cleaner
+
+/**
+ * Schedule GC on a worker that executes Cleaner blocks.
+ */
+fun scheduleGCOnCleanerWorker(): Future<Unit> =
```
SvyatoslavScherbina, line 139: This function is error-prone: it returns `Future` which should be disposed (`.result`) explicitly.

backend.native/tests/runtime/basic/cleaner_basic.kt
```diff
@@ -0,0 +1,188 @@
...
+        cleaner.freeze()
+        cleanerWeak = WeakReference(cleaner)
+        assertFalse(called.value)
+    }()
+
+    GC.collect()
+
+    assertNull(cleanerWeak!!.value)
+    tryWithTimeout(3) {
+        GC.collect()  // Collect local stack (from previous iteration)
```
SvyatoslavScherbina, line 118: Enclosing lambda gets inlined, so the code in the block can actually retain some strong references to the objects. Is this intended?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+ * the cleaner is deallocated, which can't happen until `LeakingResourceWrapper`
+ * is deallocated. So the requirements on object deallocations are contradictory
+ * in this case, which can't be handled gracefully. The cleaner's block
+ * is not executed then, and cleaner and its argument might leak
+ * (depending on the implementation).
+ *
+ * [block] should not use `@ThreadLocal` globals, because it may
+ * be executed on a different thread.
+ *
+ * If [block] throws an exception, it gets silently ignored.
```
SvyatoslavScherbina, line 118: I'm not sure it should work this way, but we haven't designed this case properly.
Let's just mention that it is not specified, and throwing an exception might do anything, from nothing to program termination?

runtime/src/main/cpp/Cleaner.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace {
+
+bool cleanersDisabled = false;
+
+void disposeCleaner(KRef thiz) {
+    if (atomicGet(&cleanersDisabled)) {
+        if (Kotlin_cleanersLeakCheckerEnabled()) {
+            konan::consoleErrorf(
+                    "Cleaner %p was disposed after main()\n"
```
SvyatoslavScherbina, line 23: It's not always `main()`. Maybe use more general wording like `during program exit`?

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -55,12 +57,17 @@ public class WorkerBoundReference<out T : Any>(value: T) {
...
 
+    private lateinit var cleaner: Cleaner
+
     @ExportForCppRuntime("Kotlin_WorkerBoundReference_freezeHook")
     private fun freezeHook() {
         // If this hook was already run, do nothing.
         if (valueBeforeFreezing == null)
             return
         ptr = createWorkerBoundReference(valueBeforeFreezing!!)
+        cleaner = createCleaner(ptr) {
```
SvyatoslavScherbina, line 68: It might make some sense to reuse the same lambda object for all WBRs as an optimization.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+
+package kotlin.native.internal
+
+import kotlin.native.concurrent.*
+import kotlinx.cinterop.StableRef
+import kotlinx.cinterop.asStableRef
+
+public interface Cleaner
+
+object CleanerWorker {
```
SvyatoslavScherbina, line 14: Shouldn't be public.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private val objHolder = StableRef.create(obj as Any)
+
+    @ExportForCppRuntime("Kotlin_CleanerImpl_clean")
+    private fun clean() {
+        // It's externally guaranteed that this is called only if cleanerWorker was
+        // not yet shut down.
+        val cleanPackage = Pair(cleanObj, objHolder).freeze()
+        CleanerWorker.worker.execute(TransferMode.SAFE, { cleanPackage }) { (cleanObj, objHolder) ->
```
SvyatoslavScherbina, line 54: Doesn't the future leak here?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+ * they will be run. Moreover, it depends on `Platform.isCleanersLeakCheckerActive`.
+ * With the checker enabled, cleaners will be run (and therefore not reported as
+ * unexecuted cleaners); with the checker disabled - they might not get run.
+ *
+ * @param argument must be shareable
+ * @param block must not capture anything
+ */
+// TODO: Consider just annotating the lambda argument rather than using intrinsic.
+@TypedIntrinsic(IntrinsicType.CREATE_CLEANER)
+external fun <T> createCleaner(argument: T, block: (T) -> Unit): Cleaner
```
SvyatoslavScherbina, line 134: Please add `@ExperimentalStdlibApi`.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,142 @@
...
+
+@ExportForCompiler
+private fun <T> createCleanerImpl(argument: T, block: (T) -> Unit): Cleaner {
+    if (!argument.isShareable())
+        throw IllegalArgumentException("$argument must be shareable")
+
+    return CleanerImpl(argument, block.freeze())
+}
+
+/**
```
SvyatoslavScherbina, line 77: The API should probably be at the beginning of the file to simplify reading.

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1069,8 +1070,8 @@ ALWAYS_INLINE void runDeallocationHooks(ContainerHeader* container) {
   ObjHeader* obj = reinterpret_cast<ObjHeader*>(container + 1);
   for (uint32_t index = 0; index < container->objectCount(); index++) {
     auto* type_info = obj->type_info();
-    if (type_info == theWorkerBoundReferenceTypeInfo) {
-      DisposeWorkerBoundReference(obj);
+    if (type_info == theCleanerImplTypeInfo) {
+        DisposeCleaner(obj);
```
SvyatoslavScherbina, line 1074: Running Kotlin code (`Kotlin_CleanerImpl_clean`) during GC is not great (see e.g. recent recursive GC issue).

backend.native/tests/runtime/basic/cleaner_basic.kt
projedi: It gets inlined, yes, and the resulting code should look like
```kotlin
while (true) {
    try {
        GC.collect()
        // retain on stack and possibly throw an exception
        return
    } catch(…) { … }
}
```
So, if the exception happens, the newly retained objects will be out of scope by the next `while` iteration and `GC.collect()` should collect them.

SvyatoslavScherbina: I see. Ok.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: Forgot about leaking `Future`s. Thanks!

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
projedi: Lambda isn't even needed here. Replaced with function reference.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: Right, thanks!

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
homuroll: Ok, thanks for notifying.

runtime/src/main/kotlin/kotlin/native/concurrent/WorkerBoundReference.kt
```diff
@@ -55,12 +57,16 @@ public class WorkerBoundReference<out T : Any>(value: T) {
...
+    private lateinit var cleaner: Cleaner
+
     @ExportForCppRuntime("Kotlin_WorkerBoundReference_freezeHook")
     private fun freezeHook() {
         // If this hook was already run, do nothing.
         if (valueBeforeFreezing == null)
             return
         ptr = createWorkerBoundReference(valueBeforeFreezing!!)
+        @OptIn(ExperimentalStdlibApi::class)
+        cleaner = createCleaner(ptr, ::cleanWorkerBoundReference)
```
SvyatoslavScherbina, line 69: Function reference is an object too (see https://github.com/JetBrains/kotlin-native/pull/4362#discussion_r498040064)

projedi: Removed `Cleaner` from `WorkerBoundReference` altogether.

runtime/src/main/cpp/Memory.cpp
projedi: Yep, moved scheduling to C++

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
```
SvyatoslavScherbina, line 2: Wrong year :)

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+// TODO: Consider just annotating the lambda argument rather than using intrinsic.
+@ExperimentalStdlibApi
+@TypedIntrinsic(IntrinsicType.CREATE_CLEANER)
+external fun <T> createCleaner(argument: T, block: (T) -> Unit): Cleaner
+
+/**
+ * Perform GC on a worker that executes Cleaner blocks.
+ */
+@ExperimentalStdlibApi
+fun performGCOnCleanerWorker() =
```
SvyatoslavScherbina, line 79: Do we really need this as public API? There are a couple of options for exposing it only for tests.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+
+private object CleanerWorker {
+    // Worker is an int. Putting it inside an object to make sure it's initialized
+    // lazily.
+    val worker = Worker.start(errorReporting = false, name = "Cleaner worker")
+}
+
+@ExportForCppRuntime("Kotlin_CleanerImpl_shutdownCleanerWorker")
+private fun shutdownCleanerWorker(executeScheduledCleaners: Boolean) {
+    // TODO: This starts the worker even if no cleaners were ever created.
```
SvyatoslavScherbina, line 92: Should be fixed before merging.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+@ExperimentalStdlibApi
+fun performGCOnCleanerWorker() =
+    CleanerWorker.worker.execute(TransferMode.SAFE, {}) {
+        GC.collect()
+    }.result
+
+private object CleanerWorker {
+    // Worker is an int. Putting it inside an object to make sure it's initialized
+    // lazily.
+    val worker = Worker.start(errorReporting = false, name = "Cleaner worker")
```
SvyatoslavScherbina, line 87: Btw, does it cause a bogus leaking workers report when compiling to anything but program?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+
+@SymbolName("Kotlin_Any_isShareable")
+external private fun Any?.isShareable(): Boolean
+
+@ExportForCompiler
+private fun <T> createCleanerImpl(argument: T, block: (T) -> Unit): Cleaner {
+    if (!argument.isShareable())
+        throw IllegalArgumentException("$argument must be shareable")
+
+    val ref = StableRef.create(argument as Any)
```
SvyatoslavScherbina, line 112: Why is it required? `argument` is shared.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    val clean = {
```
SvyatoslavScherbina, line 114: Redundant object here. Why not just store both `argument` and `block` into `CleanerImpl`?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
SvyatoslavScherbina: I think I get this. To ensure it doesn't get deallocated at the same time, right?

runtime/src/main/cpp/Cleaner.cpp
```diff
@@ -0,0 +1,59 @@
...
+            konan::consoleErrorf(
+                    "Cleaner %p was disposed during program exit\n"
+                    "Use `Platform.isCleanersLeakCheckerActive = false` to avoid this check.\n",
+                    thiz);
+            RuntimeCheck(false, "Terminating now");
+        }
+        return;
+    }
+
+    WorkerSchedule(thiz->worker, thiz->clean);
```
SvyatoslavScherbina, line 37: `WorkerSchedule` seems to use `CreateStablePointer` (~`addHeapRef`) internally.
We better make absolutely sure this is safe during (that phase of) GC.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
SvyatoslavScherbina: I guess this simplifies implementation a bit. Well, we can probably keep it that way for now.

runtime/src/main/cpp/Cleaner.cpp
SvyatoslavScherbina: Alternatively we could make a buffer that gets processed at the end of GC, where it should be more safe (at least it is similar to `incrementStack` at the beginning of GC).

runtime/src/main/cpp/Memory.cpp
```diff
@@ -1063,14 +1064,26 @@ void freeAggregatingFrozenContainer(ContainerHeader* container) {
   MEMORY_LOG("Freeing subcontainers done\n");
 }
 
+// Not inlining this call as it affects deallocation performance for
+// all types.
+NO_INLINE RUNTIME_NOTHROW void runFinalizers(ObjHeader* obj) {
+    auto* type_info = obj->type_info();
+    if (type_info == theCleanerImplTypeInfo) {
+        DisposeCleaner(obj);
```
SvyatoslavScherbina, line 1072: What exactly prevents `obj->clean` from being dead at this point?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: We only report leaking workers when compiling to a program.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: Replaced with plain `CreateStablePointer`

runtime/src/main/cpp/Memory.cpp
projedi: Fixed by putting the entire `clean` lambda inside stable pointer.

runtime/src/main/cpp/Cleaner.cpp
projedi: Fixed by making `WorkerSchedule` accept a stable pointer to a job. Stable pointer is made at the `createCleaner` and just passed to `WorkerSchedule` during GC.

runtime/src/main/cpp/Cleaner.cpp
```diff
@@ -0,0 +1,127 @@
...
+constexpr KInt kCleanerWorkerInitializing = -1;
+constexpr KInt kCleanerWorkerShutdown = -2;
+
+KInt globalCleanerWorker = kCleanerWorkerUninitialized;
+
+void disposeCleaner(CleanerImpl* thiz) {
+    auto worker = atomicGet(&globalCleanerWorker);
+    RuntimeAssert(worker != kCleanerWorkerUninitialized, "Cleaner worker must've been initialized by now");
+    if (worker == kCleanerWorkerInitializing) {
+        while (atomicGet(&globalCleanerWorker) == kCleanerWorkerInitializing) {
```
SvyatoslavScherbina, line 33: Shouldn't we assign this to `worker`?

runtime/src/main/cpp/Cleaner.cpp
```diff
@@ -0,0 +1,127 @@
...
+constexpr KInt kCleanerWorkerUninitialized = 0;
+constexpr KInt kCleanerWorkerInitializing = -1;
+constexpr KInt kCleanerWorkerShutdown = -2;
+
+KInt globalCleanerWorker = kCleanerWorkerUninitialized;
+
+void disposeCleaner(CleanerImpl* thiz) {
+    auto worker = atomicGet(&globalCleanerWorker);
+    RuntimeAssert(worker != kCleanerWorkerUninitialized, "Cleaner worker must've been initialized by now");
+    if (worker == kCleanerWorkerInitializing) {
```
SvyatoslavScherbina, line 32: Can this really be `true` here? When creating a cleaner, runtime waits until the worker gets initialized.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+        block(argument)
+    }.freeze()
+
+    // Make sure there's an extra reference to clean, so it's definitely alive when CleanerImpl is destroyed.
+    val cleanPtr = createStablePointer(clean)
+
+    // Make sure cleaner worker is initialized.
+    getCleanerWorker()
+
+    return CleanerImpl(cleanPtr)
```
SvyatoslavScherbina, line 124: Does it make sense to freeze it here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
```diff
@@ -1194,6 +1221,21 @@ private class InteropTransformer(val context: Context, override val irFile: IrFi
...
+                IntrinsicType.CREATE_CLEANER -> {
+                    val irCallableReference = expression.getValueArgument(1)
+                    if (irCallableReference == null || !irCallableReference.isNonCapturingFunction()) {
+                        context.reportCompilationError(
+                                "${function.fqNameForIrSerialization} must take an unbound, non-capturing function or lambda",
+                                irFile, expression
+                        )
+                    }
+
+                    builder.irCall(symbols.createCleanerImpl).apply {
```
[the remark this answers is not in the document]
projedi: Because intrinsics are expected to be replaced. It's handled here:
https://github.com/JetBrains/kotlin-native/blob/c2d05acf710d6c88a768f747760c400bd3f417e2/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt#L2153
I can workaround it by introducing predicate whether intrinsic must be replaced or should be evaluated with its original expression, but do we need it?

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: Might as well

runtime/src/main/cpp/Cleaner.cpp
projedi: True. Fixed.

runtime/src/main/cpp/Cleaner.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+#endif
+}
+
+void ShutdownCleaners(bool executeScheduledCleaners) {
+    KInt worker = 0;
+    do {
+        worker = atomicGet(&globalCleanerWorker);
+        RuntimeAssert(
+                worker != kCleanerWorkerShutdown, "Cleaner worker must not be shutdown twice") if (worker == kCleanerWorkerUninitialized) {
```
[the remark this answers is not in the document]
projedi: Yep, fixed.

runtime/src/main/kotlin/kotlin/native/internal/Cleaner.kt
projedi: Hid under a different Experimental flag.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InteropLowering.kt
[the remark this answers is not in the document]
SvyatoslavScherbina: I guess the more straightforward solution would be just to make `createCleaner` a regular function and add a compiler check (which is going to be moved to a special pass soon, and then to the frontend).

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Worker.cpp
```diff
@@ -475,6 +492,25 @@ class State {
     }
   }
 
+  void waitNativeWorkerTerminationUnlocked(KInt id) {
```
SvyatoslavScherbina, line 495: Code duplication here is suspicious. Can we avoid it? For example, always take an `id` parameter, with `id == -1` meaning "all workers"? Or, probably more C++-styled solution, take a lambda filtering worker ids.

runtime/src/main/cpp/Worker.cpp
```diff
@@ -475,6 +492,25 @@ class State {
...
+      pthread_t threadToWait = 0;
+      {
+          Locker locker(&lock_);
+
+          for (auto& kvp : terminating_native_workers_) {
+              RuntimeAssert(!pthread_equal(kvp.second, pthread_self()), "Native worker is joining with itself");
+              if (kvp.first == id) {
+                  threadToWait = kvp.second;
+                  break;
```
SvyatoslavScherbina, line 504: `waitNativeWorkersTerminationUnlocked` above also clears `terminating_native_workers_`. Shouldn't we remove the worker from that registry here too?

runtime/src/main/cpp/Worker.cpp
projedi: Good idea, thanks!

runtime/src/main/cpp/Worker.cpp
projedi: We should, fixed.

PR #4367, opened 2020-09-10
[KT-41655] Fix handling of globals of forward-declared type

sbogolepov:


backend.native/tests/interop/incomplete_types/main.kt
```diff
@@ -0,0 +1,8 @@
+import library.*
+import kotlinx.cinterop.*
+
+fun main() {
+    println(s.ptr)
+    println(u.ptr)
+    println(array)
+}
```
SvyatoslavScherbina, line 8: Testing actual contents would be great.

backend.native/tests/build.gradle
```diff
@@ -3574,6 +3574,12 @@ createInterop("auxiliaryCppSources") {
     it.extraOpts "-Xsource-compiler-option", "-std=c++17"
 }
 
+createInterop("incomplete_types") {
+    it.defFile 'interop/incomplete_types/library.def'
+    it.headers "$projectDir/interop/incomplete_types/library.h"
+    it.extraOpts "-Xcompile-source", "$projectDir/interop/incomplete_types/library.cpp"
```
SvyatoslavScherbina, line 3580: Wow, I didn't expect this! I was expecting one more bloated code fragment compiling a dynamic library, this approach is much better :)

sbogolepov: Eating own food :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/cgen/CBridgeGen.kt
```diff
@@ -834,9 +834,14 @@ private fun KotlinStubs.mapType(
 
     isObjCReferenceType(type) -> ObjCReferenceValuePassing(symbols, type, retained = retained)
 
+    type.isForwardDeclaration() -> TrivialValuePassing(symbols.nativePtrType, CTypes.voidPtr, isForwardDeclaration = true)
```
SvyatoslavScherbina, line 837: Is it really required to handle this ad hoc? Adjusting the condition covering non-forward structs should be enough. Moreover, that condition should actually handle forward declarations on a general basis, so I guess something is wrong with the IR here.

sbogolepov: You are right, the problem was in type translation. I pushed the (probably) correct fix and will invite a member of middle-end team to check.

sbogolepov: Checked ✅

backend.native/tests/interop/incomplete_types/main.kt
SvyatoslavScherbina: By contents I mean contents of the global, e.g. `s.name` (can be obtained with a dedicated getter function) :)
The contents should also be non-trivial (i.e. not null).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/serialization/KonanIrlinker.kt
```diff
@@ -186,30 +185,12 @@ internal class KonanIrLinker(
...
 
-        private fun getIrFile(packageFragment: PackageFragmentDescriptor): IrFile {
-            val fqn = packageFragment.fqName
-            return packageToFileMap.getOrPut(packageFragment.fqName) {
-                val fileSymbol = IrFileSymbolImpl(packageFragment)
-                IrFileImpl(NaiveSourceBasedFileEntryImpl("forward declarations for $fqn"), fileSymbol, packageFragment.fqName).also {
-                    moduleFragment.files.add(it)
-                }
+        private fun buildForwardDeclarationStub(descriptor: ClassDescriptor): IrClass {
+            return stubGenerator.generateClassStub(descriptor).also {
```
SvyatoslavScherbina, line 189: I generally have no idea what is the difference here, so my approve of this patch doesn't mean that I've actually verified it.

sbogolepov: The difference is following. Previously we constructed IR class for descriptor "by hand" and forgot to translate its supertypes (which caused the bug). Now we use Lazy IR mechanism here which automatically translates supertypes. 
This change was checked by @romanart.
As a little positive side-effect, forward declarations are removed from `context.irModule` which saves us a couple of milliseconds during compile-time :)

SvyatoslavScherbina: I understand this part. What I don't understand is why didn't we use Lazy IR mechanism before?

sbogolepov: Probably because we didn't use it before IR Linker refactoring (when `buildForwardDeclarationStub` logic was moved from common part).

SvyatoslavScherbina (approved):
Please don't forget to [improve the test](https://github.com/JetBrains/kotlin-native/pull/4367#discussion_r487101793).

backend.native/tests/interop/incomplete_types/main.kt
SvyatoslavScherbina: The test still doesn't actually check that `s`, `u` and `array` represent the original C globals, because it doesn't examine the contents of the globals in Kotlin before it overwrites them.
For example, getters could return pointers to wrong parts of the memory, and the test wouldn't detect this.

PR #4369, opened 2020-09-10
Adapt to IR change to allow setting of parameters for default stubs.

madsager:


madsager:
This is not ready for review. At this point I'm just making sure that the changes in

https://github.com/JetBrains/kotlin/pull/3691

are going to work for native as well.

I'll mark this a real non-draft pull request when ready for review.

madsager:
Updated after iterating on the kotlin side change. We should get this reviewed as well, so that we can land both when we agree on the changes.

@udalov @romanart

PR #4372, opened 2020-09-14
[typo] Fix NSNumber markdown table typo

masters3d:


SvyatoslavScherbina:
This is not a typo. Kotlin primitives types are represented as `NSNumber` or primitive types in both Swift and Obj-C.

masters3d:
The table doesn’t look formatted correctly.

SvyatoslavScherbina:
Feel free to fix it if you think so.

masters3d:
This is the fix. \ vs |

SvyatoslavScherbina:
No it is not. See https://github.com/JetBrains/kotlin-native/pull/4372#issuecomment-701927178.
"Swift" column should contain **both** "Primitive type" and "NSNumber".

masters3d:
it was confusing to me that objective column was empty. looks like primitive types are not supported in objc then.

PR #4373, opened 2020-09-14
[IR] Rewrote debug output using Context.log

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -429,3 +429,12 @@ private fun MemberScope.getContributedClassifier(name: String) =
 
 private fun MemberScope.getContributedFunctions(name: String) =
         this.getContributedFunctions(Name.identifier(name), NoLookupLocation.FROM_BUILTINS)
+
+internal class LogBuilder(val context: Context) {
```
SvyatoslavScherbina, line 433: The name is somewhat confusing here: this class doesn't build anything, i.e. it doesn't create a new entity. Maybe `Logger` or `Log`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -429,3 +429,12 @@ private fun MemberScope.getContributedClassifier(name: String) =
...
+    operator fun String.unaryPlus() = context.log { this }
+}
+
+internal fun Context.buildLog(messageBuilder: LogBuilder.() -> Unit) {
```
SvyatoslavScherbina, line 437: `logMultiple`, `logBatch`?

sbogolepov: +1, `buildX` is used for functions that produce value of type X.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
sbogolepov: Or just use `StringBuilder`. It doesn't have fancy + operator overloading, but have other formatting functions that might be useful in the future. Another option is to wrap `StringBuilder` inside `LogBuilder`. Either way, builder will allow to avoid calling `println` for each string.

homuroll: I think it could be an implementation detail, I like using '+' operator.

sbogolepov (approved):

SvyatoslavScherbina (approved):

PR #4377, opened 2020-09-14
[KT-38779] Add gles31.def to Android platform libs

natario1:
Adds GLES 3.1 APIs to Android. https://developer.android.com/ndk/guides/stable_apis#opengl_es_10_

As discussed in [KT-38779](https://youtrack.jetbrains.com/issue/KT-38779) this doesn't add GLES 3.2 because K/N is linked to Android API level 21.

### DIFF
--- a/platformLibs/src/platform/android/gles31.def
+++ b/platformLibs/src/platform/android/gles31.def
@@ -1,0 +1,5 @@
+depends = glesCommon posix
+package = platform.gles31
+headers = GLES3/gl31.h GLES3/gl3ext.h GLES3/gl3platform.h
+headerFilter = GLES3/**
+linkerOpts = -lGLESv3
--- a/platformLibs/src/platform/android/gles31.def
+++ b/platformLibs/src/platform/android/gles31.def
@@ -1,5 +1,5 @@
-depends = glesCommon posix
+depends = gles3 glesCommon posix
 package = platform.gles31
-headers = GLES3/gl31.h GLES3/gl3ext.h GLES3/gl3platform.h
+headers = GLES3/gl31.h
 headerFilter = GLES3/**
 linkerOpts = -lGLESv3


platformLibs/src/platform/android/gles31.def
```diff
@@ -0,0 +1,5 @@
+depends = glesCommon posix
+package = platform.gles31
+headers = GLES3/gl31.h GLES3/gl3ext.h GLES3/gl3platform.h
```
ilmat192, line 3: The `gles3.def` already includes `GLES3/gl3ext.h` and `GLES3/gl3platform.h`. Having the same headers in different platform libraries may be a problem because in this case each C symbol from these headers will be represented by several Kotlin symbols.

In the same time `gl31.h` includes all declarations from `gl3.h`. So I think that there is no need to create a separate platform library for OpenGL ES 3.1. Instead we can just update the already existing `gles3` platform library to version 3.1. We can do this by replacing `gl3.h` with `gl31.h` in `gles3.def`.

What do you think about this?

natario1: @ilmat192 as I see it the fact that both gl3.h and gl31.h link against the same binary (`libGLESv3`) is just an implementation detail. GLES developers would typically choose a version (be it 2.0, 3.0, 3.1, ...) and stick with it, e.g. with an `import platform.glXX.*` on top. Which is why gl31.h has all gl3.h APIs as you said.

So IMO finding GLES 3.1 APIs under `platform.gles3` would be weird. See for example how in Android GLES31 extends GLES30 but is still a separate class https://developer.android.com/reference/android/opengl/GLES31 

That said, we can do they way you suggested if you prefer. Let me know

ilmat192: @natario1 Thank you for the clarification. Well, I have no strong objections against adding gles31 as a separate platform library. If this approach is more familiar for GLES developers, let's add it. 

As for `GLES3/gl3ext.h` and `GLES3/gl3platform.h`, I think that we can remove them from this def-file and add a dependency on `gles3` instead, like it is done for `posix` and `glesCommon`. Could you please add this change?

natario1: @ilmat192 done! These two files are actually empty as far as I can see, so it should be fine either way.

ilmat192 (approved):

PR #4390, opened 2020-09-16
Added missing note about Gradle wrapper required

wild-lynx:
...for the CocoaPods plugin.

### DIFF
--- a/COCOAPODS.md
+++ b/COCOAPODS.md
@@ -94,6 +94,11 @@ file for the project.
 The `Podspec` file includes a path to an output framework and script phases that automate building this framework during 
 the build process of an Xcode project.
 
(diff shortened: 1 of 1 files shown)


COCOAPODS.md
```diff
@@ -94,6 +94,11 @@ file for the project.
 The `Podspec` file includes a path to an output framework and script phases that automate building this framework during 
 the build process of an Xcode project.
 
+6. Generate Gradle wrapper. In order to avoid compatibility issues during an Xcode build, the plugin requires using
```
katia-energizer, line 97: I would rephrase like this:

> 6. Generate the [Gradle Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html), which helps avoid compatibility issues during an Xcode build.

koshachy:
Fixed in #4507

PR #4393, opened 2020-09-17
Add failing test for multiple static frameworks cache isolation

SvyatoslavScherbina:


PavelPunegov (approved):

backend.native/tests/build.gradle
```diff
@@ -4390,6 +4390,30 @@ if (isAppleTarget(project)) {
...
+            bitcode = true
+            isStatic = true
+            opts = ['-Xstatic-framework', "-Xpre-link-caches=enable"]
+        }
+        framework('SecondStatic') {
+            artifact = 'Second'
+            sources = ['framework/multiple/framework2', 'framework/multiple/shared']
+            bitcode = true
+            isStatic = true
+            opts = ['-Xstatic-framework', "-Xpre-link-caches=enable"]
```
sbogolepov, line 4411: Why do you pass `isStatic = true` together with `-Xstatic-framework`?

sbogolepov (approved):

backend.native/tests/build.gradle
SvyatoslavScherbina: Because `isStatic` doesn't actually enable passing this flag. See also `testObjCExportStatic`.

PR #4398, opened 2020-09-21
[IR] Supported non-reified typeOf

homuroll:


homuroll:
Kotlin part (includes turning on the tests): https://github.com/JetBrains/kotlin/compare/rr/igoryan/native_non_reified_typeof.

SvyatoslavScherbina:
> Kotlin part (includes turning on the tests): https://github.com/JetBrains/kotlin/compare/rr/igoryan/native_non_reified_typeof.

How does this work? These tests have `// WITH_REFLECT`, and our test machinery seems to ignore such tests: https://github.com/JetBrains/kotlin-native/blob/f76cf52c81109503cea1bd4591cd874a24507ee3/build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy#L421

runtime/src/main/kotlin/kotlin/native/internal/KTypeParameterImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+ * that can be found in the LICENSE file.
+ */
+
+package kotlin.native.internal
+
+import kotlin.reflect.*
+
+internal data class KTypeParameterImpl(
+        override val name: String,
+        val containerFqName: String,
```
SvyatoslavScherbina, line 12: Is it used?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
```diff
@@ -25,11 +30,17 @@ internal class PreInlineLowering(val context: Context) : BodyLoweringPass {
...
+
     private val asserts = symbols.asserts
     private val enableAssertions = context.config.configuration.getBoolean(KonanConfigKeys.ENABLE_ASSERTIONS)
 
     override fun lower(irBody: IrBody, container: IrDeclaration) {
-        irBody.transformChildrenVoid(object : IrBuildingTransformer(context) {
+        var symbolOwner = container
+        while (symbolOwner !is IrSymbolOwner)
+            symbolOwner = symbolOwner.parent as IrDeclaration
+        val builder = context.createIrBuilder(symbolOwner.symbol, irBody.startOffset, irBody.endOffset)
```
SvyatoslavScherbina, line 42: Might be incorrect for local functions

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
```diff
@@ -38,6 +49,10 @@ internal class PreInlineLowering(val context: Context) : BodyLoweringPass {
                 if (!enableAssertions && expression.symbol in asserts) {
                     assert(expression.type.isUnit())
                     return IrCompositeImpl(expression.startOffset, expression.endOffset, expression.type)
+                } else if (expression.symbol.owner.isTypeOfIntrinsic()) {
+                    val type = expression.getTypeArgument(0)
+                            ?: org.jetbrains.kotlin.backend.konan.error(container.file, expression, "missing type argument")
+                    return with (kTypeGenerator) { builder.at(expression).irKType(type, needExactTypeParameters = true) }
```
SvyatoslavScherbina, line 55: But how does typeOf substitute reified type parameters now, if it gets lowered before inline?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,63 +10,140 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
+        irKTypeImpl(
+                kClassifier = kClassifier,
+                irTypeArguments = type.arguments,
+                isMarkedNullable = type.hasQuestionMark,
+                needExactTypeParameters = needExactTypeParameters,
+                seenTypeParameters = seenTypeParameters
+        )
+    } catch (t: RecursiveBoundsException) {
+        if (needExactTypeParameters)
+            this@KTypeGenerator.context.reportCompilationError(t.message!!)
```
SvyatoslavScherbina, line 73: It should report this error with source location.
Maybe introduce a class encapsulating all generation parameters and state?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
SvyatoslavScherbina: I mean:
```kotlin
inline fun <reified T> t() = kotlin.reflect.typeOf<T>()

fun main() {
    println(t<String>())
}
```
prints String on JVM. Does this work with Native on this branch?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,63 +10,140 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
+    }
+
+    private fun IrBuilderWithScope.irKVariance(variance: Variance) =
+            IrGetEnumValueImpl(
+                    startOffset, endOffset,
+                    symbols.kVariance.defaultType,
+                    symbols.kVariance.owner.declarations.filterIsInstance<IrEnumEntry>()[variance.ordinal].symbol
+            )
+
+    private fun IrBuilderWithScope.irKTypeList(
```
SvyatoslavScherbina, line 121: Looks quite similar to the function below. Can we share more code here?

runtime/src/main/kotlin/kotlin/native/internal/KTypeParameterImpl.kt
homuroll: It is used in the `equals` and `hashCode`.

SvyatoslavScherbina: I see, you are right.
But then I have another question: should `equals` and `hashCode` use `upperBounds`? This seems quite inefficient. Also not sure about `variance` or `isReified` either.
What does JVM implementation do?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
homuroll: My bad. The reason why I moved it before the inliner was non-passing test, where `reified` parameters aren't expected to be substituted at some places, I'll rethink, what is going on there.

homuroll:
Regarding filtering out tests with `WITH_REFLECT` in them: why not turn them on? JS for example doesn't skip them, also, some of the tests are already marked with either `TARGET_BACKEND: JVM` or `IGNORE_BACKEND: NATIVE`, so considering that currently quite a big number of external tests fail anyway, I think it couldn't be much worse.

runtime/src/main/kotlin/kotlin/native/internal/KTypeParameterImpl.kt
homuroll: Good point. I copied this from JS implementation, but JVM only uses `name` and `container` for `equals`/`hashCode`, which is logical.

backend.native/tests/codegen/ktype/nonReified.kt
```diff
@@ -0,0 +1,78 @@
...
+    val l = foo<Int>()
+    assertEquals(List::class, l.classifier)
+    val t = l.arguments.single().type!!.classifier
+    assertTrue(t is KTypeParameter)
+    assertFalse((t as KTypeParameter).isReified)
+    assertEquals("T", (t as KTypeParameter).name)
+}
+
+class C<T> {
+    @OptIn(kotlin.ExperimentalStdlibApi::class)
```
SvyatoslavScherbina, line 25: It should be enough to place this annotation on the entire file.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,95 +10,165 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
 
-internal class KTypeGenerator(
-        private val context: KonanBackendContext,
-        private val eraseTypeParameters: Boolean = false
-) {
+internal class KTypeGenerator(val context: KonanBackendContext, val irFile: IrFile, val irElement: IrElement) {
     private val symbols = context.ir.symbols
 
-    fun IrBuilderWithScope.irKType(type: IrType): IrExpression = if (type !is IrSimpleType) {
+    fun IrBuilderWithScope.irKType(type: IrType, needExactTypeParameters: Boolean = false): IrExpression {
```
SvyatoslavScherbina, line 40: Is there a reason for not putting `needExactTypeParameters` into class property?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,95 +10,165 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
+    }
 
-            is IrStarProjection -> irCall(symbols.kTypeProjectionStar.owner.getter!!).apply {
-                dispatchReceiver = irGetObject(symbols.kTypeProjectionCompanion)
-            }
+    private fun IrBuilderWithScope.irKVariance(variance: Variance) =
+            IrGetEnumValueImpl(
+                    startOffset, endOffset,
+                    symbols.kVariance.defaultType,
+                    symbols.kVariance.owner.declarations.filterIsInstance<IrEnumEntry>()[variance.ordinal].symbol
```
SvyatoslavScherbina, line 120: This approach may break for a number of reasons. Finding by name would be more reliable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
homuroll: No, indeed, will do.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
homuroll: Rewrote.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
homuroll: Done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
```diff
@@ -29,19 +37,36 @@ internal class PreInlineLowering(val context: Context) : BodyLoweringPass {
     private val enableAssertions = context.config.configuration.getBoolean(KonanConfigKeys.ENABLE_ASSERTIONS)
 
     override fun lower(irBody: IrBody, container: IrDeclaration) {
-        irBody.transformChildrenVoid(object : IrBuildingTransformer(context) {
+        var symbolOwner = container
+        while (symbolOwner !is IrSymbolOwner)
```
SvyatoslavScherbina, line 41: Looks somewhat cryptic. Could you provide an example for this case, in a comment?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,95 +10,160 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
+        try {
+            val kClassifier = when (val classifier = type.classifier) {
+                is IrClassSymbol -> irKClass(classifier)
+                is IrTypeParameterSymbol -> {
+                    if (classifier.owner.isReified && context.leaveReifiedForLater) {
+                        // Leave as is for reification.
+                        return irCall(symbols.typeOf).apply { putTypeArgument(0, type) }
+                    }
+
+                    context.leaveReifiedForLater = false
```
SvyatoslavScherbina, line 69: What is this required for?

homuroll: To leave upper bounds of non-reified type parameters as is, even if they are reified themselves (as JVM does).

backend.native/tests/codegen/ktype/nonReified.kt
```diff
@@ -0,0 +1,78 @@
...
+@Test
+fun test_reifiedUpperBound() {
+    val l = reifiedUpperBound<Any, Any>()
+    assertEquals(List::class, l.classifier)
+    val r = l.arguments.single().type!!.classifier
+    assertTrue(r is KTypeParameter)
+    assertFalse((r as KTypeParameter).isReified)
+    assertEquals("R", (r as KTypeParameter).name)
+    val t = (r as KTypeParameter).upperBounds.single().classifier
+    assertTrue(t is KTypeParameter)
```
SvyatoslavScherbina, line 75: So reified upper bound is not substituted here. Should really work this way? What is the motivation?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
SvyatoslavScherbina: I see.
Shouldn't we revert this after `irKTypeParameter` below then?

backend.native/tests/codegen/ktype/nonReified.kt
SvyatoslavScherbina: (I've posted this question before I saw your answer [above](https://github.com/JetBrains/kotlin-native/pull/4398#discussion_r508338726). I guess the answer explains it here too.)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
homuroll: Indeed. Previously it was just a parameter. I guess I'll revert this since now `BuildingContext` contains only two values.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PreInlineLowering.kt
homuroll: Well, technically, `IrDeclaration` doesn't inherit `IsSymbolOwner` but in reality look like it does. I'll remove the loop then.

SvyatoslavScherbina: One more thing to consider here: `container` is not an arbitrary `IrDeclaration`, but it is an `IrDeclaration` with a body.
So it should be enough to make sure that all declarations with bodies are symbol owners.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/ReflectionSupport.kt
```diff
@@ -10,94 +10,183 @@ import org.jetbrains.kotlin.backend.konan.KonanBackendContext
...
+        try {
+            val kClassifier = when (val classifier = type.classifier) {
+                is IrClassSymbol -> irKClass(classifier)
+                is IrTypeParameterSymbol -> {
+                    if (classifier.owner.isReified && leaveReifiedForLater) {
+                        // Leave as is for reification.
+                        return irCall(symbols.typeOf).apply { putTypeArgument(0, type) }
+                    }
+
+                    irKTypeParameter(classifier.owner, leaveReifiedForLater = false, seenTypeParameters = seenTypeParameters)
```
SvyatoslavScherbina, line 72: > `leaveReifiedForLater = false`

A comment explaining this would be extremely useful.

SvyatoslavScherbina (approved):

SvyatoslavScherbina (approved):

PR #4404, opened 2020-09-25
Rework invalid LLVM module reporting

SvyatoslavScherbina:
* Dump module to file instead of stderr
* Report errors in an exception message instead of stdout (useful when running in Gradle daemon)

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/VerifyModule.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/VerifyModule.kt
@@ -1,0 +1,95 @@
+package org.jetbrains.kotlin.backend.konan.llvm
+
+import kotlinx.cinterop.*
+import llvm.*
+import java.io.File
+
+internal fun verifyModule(llvmModule: LLVMModuleRef, current: String = "") = memScoped {
+    val errorRef = allocPointerTo<ByteVar>()
+    errorRef.value = null
+
+    val verificationResult = LLVMVerifyModule(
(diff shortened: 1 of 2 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/VerifyModule.kt
```diff
@@ -0,0 +1,95 @@
...
+    appendLine("Verification errors:")
+
+    val maxLines = 12
+
+    lines.take(maxLines).forEach {
+        appendLine("    $it")
+    }
+
+    if (lines.size > maxLines) {
+        appendLine("    ...")
```
sbogolepov, line 72: May be add something like "full error is available at ${moduleDumpFile.absolutePath}"?

sbogolepov: It is not obvious that error is appended to module file.

SvyatoslavScherbina: This exact error message fragment would require this code to depend on the `moduleDumpFile`.
I guess things would be more simple and reliable if the message was more like "full error is available at the LLVM module dump file". What do you think?

sbogolepov: Yes, it will work.

sbogolepov (approved):

PR #4408, opened 2020-09-28
Support for Xcode 12

sbogolepov:


konan/konan.properties
```diff
@@ -26,7 +26,7 @@ airplaneMode = false
 # if true, ignores Xcode version check.
 ignoreXcodeVersionCheck = false
 
-minimalXcodeVersion=11.0
+minimalXcodeVersion=12.0
```
SvyatoslavScherbina, line 29: Is it? Does compiler work with Xcode 11?

konan/konan.properties
```diff
@@ -111,22 +111,22 @@ clangDebugFlags.ios_arm32 = -O0
 clangDynamicFlags.ios_arm32 =
 linkerNoDebugFlags.ios_arm32 = -S
 linkerDynamicFlags.ios_arm32 = -dylib
-linkerKonanFlags.ios_arm32 = -lSystem -lc++ -lobjc -framework Foundation -sdk_version 13.5
+linkerKonanFlags.ios_arm32 = -lSystem -lc++ -lobjc -framework Foundation -sdk_version 14.0
```
SvyatoslavScherbina, line 114: Shouldn't we bump macOS SDK version too?

tools/scripts/update_apple_frameworks.sh
```diff
@@ -0,0 +1,191 @@
+#! /bin/bash
```
SvyatoslavScherbina, line 1: Please add `set -e` to ensure the script exits after first failure

tools/scripts/update_apple_frameworks.sh
```diff
@@ -0,0 +1,191 @@
...
+
+DEFS_FILE=$(mktemp)
+FRAMEWORKS_DEV_FILE=$(mktemp)
+FRAMEWORKS_SIM_FILE=$(mktemp)
+FRAMEWORKS_COMMON_FILE=$(mktemp)
+ls $DEFS | grep .def | cut -d '.' -f 1 > $DEFS_FILE
+ls $FRAMEWORKS_DEV | grep .framework | cut -d '.' -f 1 > $FRAMEWORKS_DEV_FILE
+ls $FRAMEWORKS_SIM | grep .framework | cut -d '.' -f 1 > $FRAMEWORKS_SIM_FILE
+
+# Some frameworks might be available for device but not for framework, and vice versa.
```
SvyatoslavScherbina, line 54:
```suggestion
# Some frameworks might be available for device but not for simulator, and vice versa.
```

tools/scripts/update_apple_frameworks.sh
```diff
@@ -0,0 +1,191 @@
...
+# Platform library should contain the common part.
+comm -12 $FRAMEWORKS_DEV_FILE $FRAMEWORKS_SIM_FILE > $FRAMEWORKS_COMMON_FILE
+ABSENT=$(comm -13 $DEFS_FILE $FRAMEWORKS_COMMON_FILE)
```
SvyatoslavScherbina, line 57: Is this correct? If a framework is available in Swift on device, we should make it available in Kotlin on device too.

tools/scripts/update_apple_frameworks.sh
```diff
@@ -0,0 +1,191 @@
...
+        UNAVAILABLE+=($FRAMEWORK_NAME)
+        return
+    fi
+    # DriverKit is C++. Drop it.
+    if [[ $(cat $JSON | jq '.metadata.platforms[] | .name' | grep DriverKit) ]]
+    then
+        DRIVER_KIT+=($FRAMEWORK_NAME)
+        return
+    fi
+    if [[ ! $(cat $JSON | jq '.metadata.platforms[] | .name' | grep $OS_NAME) ]]
```
SvyatoslavScherbina, line 89: Could you provide an example for this case?

platformLibs/src/platform/ios/AudioUnit.def
```diff
@@ -0,0 +1,6 @@
+depends = AudioToolbox CFNetwork CoreAudioTypes CoreFoundation Foundation Security darwin posix
+language = Objective-C
+package = platform.AudioUnit
+modules = AudioUnit
+compilerOpts = -framework AudioUnit
+linkerOpts = -framework AudioUnit
```
SvyatoslavScherbina: There is something wrong with AudioUnit framework. IIRC, it is effectively empty, because it just imports headers from AudioToolbox. Please check.

SvyatoslavScherbina: Now I see that PR no longer contains this file. Might be a glitch in IDEA GitHub integration.

SvyatoslavScherbina: No, IDEA is ok, it just gets moved to tvOS later.

platformLibs/src/platform/osx/AGL.def.disabled
```diff
@@ -0,0 +1,6 @@
+language = Objective-C
+package = platform.AGL
+modules = AGL
+compilerOpts = -framework AGL
+linkerOpts = -framework AGL
+#Disabled: Framework without module
```
SvyatoslavScherbina, line 6: I'm not sure it is the right reason. But IIRC, the framework is deprecated. Maybe mention this instead?

platformLibs/src/platform/osx/ColorSync.def
```diff
@@ -0,0 +1,6 @@
+depends = CoreFoundation darwin posix
+language = Objective-C
+package = platform.ColorSync
+modules = ColorSync
+compilerOpts = -framework ColorSync
+linkerOpts = -framework ColorSync
```
SvyatoslavScherbina, line 6: Isn't it a part of `ApplicationServices` platform library at the moment?

platformLibs/src/platform/osx/InputMethodKit.def.disabled
```diff
@@ -0,0 +1,6 @@
+language = Objective-C
+package = platform.InputMethodKit
+modules = InputMethodKit
+compilerOpts = -framework InputMethodKit
+linkerOpts = -framework InputMethodKit
+#Disabled: Framework without module
```
SvyatoslavScherbina, line 6: Probably deprecated too.

platformLibs/src/platform/osx/Intents.def
```diff
@@ -0,0 +1,7 @@
+depends = ApplicationServices CFNetwork ColorSync CoreFoundation CoreGraphics CoreLocation CoreServices CoreText DiskArbitration Foundation IOKit ImageIO Security darwin libkern osx posix
+language = Objective-C
+package = platform.Intents
+modules = Intents
+compilerOpts = -framework Intents
+linkerOpts = -framework Intents
+#Disabled: Unavailable
```
SvyatoslavScherbina, line 7: Doesn't seem correct.

platformLibs/src/platform/osx/System.def.disabled
```diff
@@ -0,0 +1,6 @@
+language = Objective-C
+package = platform.System
+modules = System
+compilerOpts = -framework System
+linkerOpts = -framework System
+#Disabled: Swift-only framework
```
SvyatoslavScherbina, line 6: Not sure about this: the framework doesn't contain any headers or Swift modules.

konan/konan.properties
sbogolepov: macOS 11 is not released yet.

tools/scripts/update_apple_frameworks.sh
sbogolepov: ExposureNotification on tvOS, OSLog on iOS and tvOS.

platformLibs/src/platform/ios/AudioUnit.def
sbogolepov: Thanks, looks like commit squashing went wrong.

platformLibs/src/platform/osx/ColorSync.def
sbogolepov: 🤦‍♂️

sbogolepov: Looks like we should remove it from `ApplicationServices` to make things simpler, but it will be a breaking change.

platformLibs/src/platform/osx/Intents.def
sbogolepov: True. `Intents` and `IntentsUI` frameworks are united under SiriKit. I fixed it manually for other OS, but missed it here.

platformLibs/src/platform/osx/System.def.disabled
sbogolepov: At the same time, it _will_ be a Swift-only framework starting from macOS 11 https://developer.apple.com/documentation/system?language=objc.

platformLibs/src/platform/osx/AGL.def.disabled
sbogolepov: Yeah, sure. Sadly, there is no obvious way to distinguish deprecated frameworks automatically.

konan/konan.properties
sbogolepov: It will, but platform libraries will be inconsistent with actually available SDKs.

SvyatoslavScherbina: Will this be a problem if the code doesn't use new APIs?

konan/konan.properties
SvyatoslavScherbina: Of course. But Xcode 12 includes SDK for macOS 10.15.6.

platformLibs/src/platform/osx/ColorSync.def
SvyatoslavScherbina: Well, things are complicated here.
`ApplicationServices` imports the entire `ColorSync` (and even has the symlink to it under `Frameworks/`).
Anyway, it is a breaking change of course, so we shouldn't do this at the moment anyway.

konan/konan.properties
sbogolepov: No, but it will be a problem from UX point of view (declaration is available in IDE when it actually isn't). OTOH, I agree with you that forcing users to switch to Xcode 12 isn't a good move.

platformLibs/src/platform/osx/ColorSync.def
sbogolepov: Ok, I'll disable it.

SvyatoslavScherbina (approved):

PR #4412, opened 2020-09-30
[KT-42275] Eliminate recursive GC calls

LepilkinaElena:


runtime/src/main/cpp/Memory.cpp
```diff
@@ -2117,15 +2117,16 @@ void updateHeapRefIfNull(ObjHeader** location, const ObjHeader* object) {
 inline void checkIfGcNeeded(MemoryState* state) {
   if (state != nullptr && state->allocSinceLastGc > state->allocSinceLastGcThreshold) {
     // To avoid GC trashing check that at least 10ms passed since last GC.
-    if (konan::getTimeMicros() - state->lastGcTimestamp > 10 * 1000) {
+    if (konan::getTimeMicros() - state->lastGcTimestamp > 10 * 1000 && state->gcSuspendCount == 0) {
```
projedi, line 2120: Nit: Maybe move this check to the `if` above just like in `checkIfForceCyclicGcNeeded`?

projedi (approved):

SvyatoslavScherbina (approved):

PR #4417, opened 2020-10-05
Don't allow GC in BackRefFromAssociatedObject::tryAddRef

SvyatoslavScherbina:
This function is invoked while a non-recursive os_lock is taken,
and GC can provoke Obj-C object deallocation taking the same lock
recursively.

### DIFF
--- a/backend.native/tests/interop/objc/tests/tryRetainGC.kt
+++ b/backend.native/tests/interop/objc/tests/tryRetainGC.kt
@@ -1,0 +1,25 @@
+import kotlinx.cinterop.*
+import kotlin.test.*
+import objcTests.*
+
+@Test
+fun testTryRetainGC() {
+    kotlin.native.internal.GC.collect()
+    val weakRefHolder = WeakRefHolder()
+    createGarbageNSObjects(weakRefHolder)
+    weakRefHolder.obj = object : NSObject() {}
+    // Loading weak ref takes a lock. If K/N runtime runs GC while the lock is taken,
+    // then it releases garbage objects and thus Obj-C runtime might take a recursive lock
+    // and abort in _os_unfair_lock_recursive_abort.
+    weakRefHolder.loadManyTimes()
+}
+
+private fun createGarbageNSObjects(weakRefHolder: WeakRefHolder) {
+    autoreleasepool {
+        repeat(100) {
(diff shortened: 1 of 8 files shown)


backend.native/tests/interop/objc/tests/tryRetainGC.kt
```diff
@@ -0,0 +1,25 @@
+import kotlinx.cinterop.*
+import kotlin.test.*
+import objcTests.*
+
+@Test
```
projedi, line 5: I don't see this test added to the `backend.native/tests/build.gradle`. Is it intentional?

SvyatoslavScherbina: It is not required for `backend.native/tests/interop/objc/tests/*`

backend.native/tests/interop/objc/tests/tryRetainGC.m
```diff
@@ -0,0 +1,11 @@
+#import "tryRetainGC.h"
+#import <Foundation/NSArray.h>
+
+@implementation WeakRefHolder
+-(void)loadManyTimes {
+    NSMutableArray* array = [NSMutableArray new];
+    for (int i = 0; i < 10000; ++i) {
+        [array addObject:self.obj];
```
projedi, line 8: So, during this `self.obj`, Obj-C runtime takes a lock (1); then calls `_tryRetain`; we intercept that and call `BackRefFromAssociatedObject::tryAddRef`; there we first increment refcount of Kotlin associated object; and then enqueue decrementing it back. And the problem is that this enqueuing could previously induce GC (because of a lot of queued decrements); and in `createGarbageNSObject` we've created a bunch of objects that happen to take the same lock (1) during their deallocation?

backend.native/tests/interop/objc/tests/tryRetainGC.kt
projedi: Ah, yes. I see.

backend.native/tests/interop/objc/tests/tryRetainGC.m
SvyatoslavScherbina: Yes.

projedi (approved):

knebekaizer (approved):

PR #4420, opened 2020-10-05
Run typeOf tests and exclude tests

PavelPunegov:


build-tools/src/main/groovy/org/jetbrains/kotlin/KonanTest.groovy
```diff
@@ -335,11 +336,10 @@ class RunExternalTestGroup extends JavaExec {
             }
             // now replace all package usages in full qualified names
             def res = ""                      // filesToCompile
-            def vars = new HashSet<String>()  // variables that has the same name as a package
             text.eachLine { line ->
                 packages.each { pkg ->
                     // line contains val or var declaration or function parameter declaration
-                    if ((line =~ ~/va(l|r) *$pkg *\=/) || (line =~ ~/fun .*\(\n?\s*$pkg:.*/)) {
+                    if ((line =~ ~/va(l|r) *$pkg *(get\(\))? *\=/) || (line =~ ~/fun .*\(\n?\s*$pkg:.*/)) {
```
ilmat192, line 342: Can't the line contain a type of the property? Like
```
val $pkg: Foo get() = ...
```

ilmat192 (approved):

SvyatoslavScherbina (approved):

PR #4424, opened 2020-10-07
Prevent Obj-C dealloc accessing Kotlin counterpart 

SvyatoslavScherbina:
#KT-41811 Fixed.

Also add GC assertion for add ref for reclaimed object.

### DIFF
--- a/runtime/src/main/cpp/MemorySharedRefs.hpp
+++ b/runtime/src/main/cpp/MemorySharedRefs.hpp
@@ -54,12 +54,14 @@ class BackRefFromAssociatedObject {
 
   void releaseRef();
 
+  void detach();
+
   // Error if called from the wrong worker with non-frozen obj_.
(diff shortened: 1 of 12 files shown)


runtime/src/main/cpp/MemorySharedRefs.hpp
```diff
@@ -54,14 +54,12 @@ class BackRefFromAssociatedObject {
...
   // Error if called from the wrong worker with non-frozen obj_.
   template <ErrorPolicy errorPolicy>
   ObjHeader* ref() const;
 
-  inline bool permanent() const {
-    return obj_->permanent(); // Safe to query from any thread.
-  }
-
  private:
   ObjHeader* obj_;
```
projedi, line 64: Maybe a comment that `obj_` might be null after `detach`?

projedi (approved):

knebekaizer (approved):

PR #4430, opened 2020-10-08
Use non-local ObjC-export mapping and naming in ObjcExportHeaderGeneratorMobile

fkistner:
We're getting duplicate stubs in local mode, but the problem does not seem to occur in non-local mode.


SvyatoslavScherbina (approved):
Please be careful, non-local mode has a huge overhead.

SvyatoslavScherbina (approved):

PR #4431, opened 2020-10-08
Use analogous prefix mechanism in `ObjCExportLazy` to the default of `ObjCExportNamer`

fkistner:
There doesn't seem to be any point to reimplement the same behavior in `ObjCExportLazy.Configuration` implementation.
Am I missing anything? Under what circumstances is a different behavior necessary?

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportLazy.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportLazy.kt
@@ -7,12 +7,12 @@ package org.jetbrains.kotlin.backend.konan.objcexport
 
 import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.analyzer.ModuleInfo
-import org.jetbrains.kotlin.descriptors.konan.isNativeStdlib
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.descriptors.annotations.Annotations
 import org.jetbrains.kotlin.descriptors.impl.PropertyDescriptorImpl
 import org.jetbrains.kotlin.descriptors.impl.SimpleFunctionDescriptorImpl
+import org.jetbrains.kotlin.descriptors.konan.isNativeStdlib
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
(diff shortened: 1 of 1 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportLazy.kt
```diff
@@ -430,13 +431,15 @@ internal fun createNamerConfiguration(configuration: ObjCExportLazy.Configuratio
...
+        /** @see ObjCExportNamerImpl */
         override fun getAdditionalPrefix(module: ModuleDescriptor): String? {
             if (module.isStdlib()) return "Kotlin"
 
             // Note: incorrect for compiler since it doesn't store ModuleInfo to ModuleDescriptor.
             val moduleInfo = module.getCapability(ModuleInfo.Capability) ?: return null
             if (configuration.isIncluded(moduleInfo)) return null
-            return abbreviate(configuration.getCompilerModuleName(moduleInfo))
+
+            return module.namePrefix
```
SvyatoslavScherbina, line 442: IIRC, I had some issues with this approach when used with IDEA.
The issues were caused by different approaches to `ModuleDescriptor.name` for source modules: compiler combines all source sets of a single framework and uses the framework's name for the module, while IDEA creates a `ModuleDescriptor` for each IDEA module (~ Gradle source set) and uses IDEA-project-structure-specific names for these modules.

fkistner: Ah, alright, I see.
But this would only apply to `module.getCapability(ModuleInfo.Capability) is ModuleSourceInfo`, correct?

Can you clarify for me, whether the following invariant holds here?
`module.getCapability(ModuleInfo.Capability)?.let { it.name == module.name } != false`
I.e. does the `ModuleInfo` always have an equal `name` as the `ModuleDescriptor` it was obtained from?

SvyatoslavScherbina: > But this would only apply to module.getCapability(ModuleInfo.Capability) is ModuleSourceInfo, correct?

I have no idea.

> Can you clarify for me, whether the following invariant holds here?
module.getCapability(ModuleInfo.Capability)?.let { it.name == module.name } != false
I.e. does the ModuleInfo always have an equal name as the ModuleDescriptor it was obtained from?

Same here. In any case, the compiler and the IDE have different ideas for `ModuleInfo.name`, so not sure how it would help here.

fkistner: Well the problem is that the interface only provides the `ModuleInfo`, but not the `ModuleDescriptor`. `ObjCExport` uses the `ModuleDescriptor` name, if the names are not equivalent, I can't use `ModuleInfo` to calculate the same name…

PR #4435, opened 2020-10-09
Runtime unit tests

ilmat192:


runtime/src/test_support/cpp/CompilerGenerated.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the LICENSE file.
+ */
+
+#ifndef RUNTIME_COMPILER_GENERATED_H
+#define RUNTIME_COMPILER_GENERATED_H
+
+#define THROW_NOT_IMPLEMENTED throw std::runtime_error("Not implemented for tests");
```
projedi, line 9: Maybe we can do without this macros? We can define a plain function `void ThrowNotImplementedForTests() { … }` here. However, I don't see anything wrong with repeating `throw std::runtime_error(...)` at each site either.

.gitignore
```diff
@@ -62,6 +62,7 @@ llvmCoverageMappingC/CMakeLists.txt
 tools/performance-server/node_modules
 tools/performance-server/server
 tools/performance-server/ui/js
+runtime/cmake-build-debug/
```
projedi, line 65: Why is this needed?

runtime/src/test_support/cpp/CompilerGenerated.h
ilmat192: Fixed by inlining the macro.

.gitignore
ilmat192: CMake generates this directory during CLion import

projedi (approved):

PR #4437, opened 2020-10-12
Add crashing test on KT-41811

projedi:
Follow-up to #4424

### DIFF
--- a/backend.native/tests/interop/objc/tests/kt41811.kt
+++ b/backend.native/tests/interop/objc/tests/kt41811.kt
@@ -81,3 +81,29 @@ private fun createGarbageDeallocLoadWeak(weakRef: WeakReferenceProtocol) {
         assertSame(obj, weakDeallocLoadWeak!!.referent)
     }
 }
+
+@Test
+fun testKT41811WithGlobal() {
+    // Attempt to make the state predictable:
+    kotlin.native.internal.GC.collect()
+
+    deallocRetainReleaseDeallocated = false
+    assertFalse(deallocRetainReleaseDeallocated)
+
+    autoreleasepool {
+        {
+            globalDeallocRetainRelease = object: DeallocRetainRelease() {}
+        }()
+        // Clean up local DeallocRetainRelease on Kotlin side
+        kotlin.native.internal.GC.collect()
+        // And drop the last reference to DeallocRetainRelease from ObjC global scope.
+        globalDeallocRetainRelease = null
+    }
+
+    assertFalse(deallocRetainReleaseDeallocated)
+
+    // This will dispose DeallocRetainRelease on Kotlin side, which will cause dealloc
+    // on ObjC side, which will immediately schedule another disposal on the Kotlin side.
+    kotlin.native.internal.GC.collect()
+    assertTrue(deallocRetainReleaseDeallocated)
+}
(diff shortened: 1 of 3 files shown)


backend.native/tests/interop/objc/tests/kt41811.kt
```diff
@@ -81,3 +81,29 @@ private fun createGarbageDeallocLoadWeak(weakRef: WeakReferenceProtocol) {
...
+        kotlin.native.internal.GC.collect()
+        // And drop the last reference to DeallocRetainRelease from ObjC global scope.
+        globalDeallocRetainRelease = null
+    }
+
+    assertFalse(deallocRetainReleaseDeallocated)
+
+    // This will dispose DeallocRetainRelease on Kotlin side, which will cause dealloc
+    // on ObjC side, which will immediately schedule another disposal on the Kotlin side.
+    kotlin.native.internal.GC.collect()
```
projedi, line 107: Crashing occurs because `freeContainer` ran `scheduleDestroyContainer` which runs `setNextLink`: https://github.com/JetBrains/kotlin-native/blob/0a5e64f6175f63db833e32e88df2e5c715ee8fc2/runtime/src/main/cpp/Memory.h#L282-L284
which pretty much overwrites `ObjHeader`; so the next time someone accesses the object (e.g. checks its type info), it contains garbage.

Other tests don't crash, because during their run of `freeContainer` `scheduleDestroyContainer` isn't called: https://github.com/JetBrains/kotlin-native/blob/0a5e64f6175f63db833e32e88df2e5c715ee8fc2/runtime/src/main/cpp/Memory.cpp#L1108-L1109 Here buffered flag is reset because of `GC.collect()` at line 98.

backend.native/tests/interop/objc/tests/kt41811.kt
```diff
@@ -81,3 +81,29 @@ private fun createGarbageDeallocLoadWeak(weakRef: WeakReferenceProtocol) {
...
+        // Clean up local DeallocRetainRelease on Kotlin side
+        kotlin.native.internal.GC.collect()
+        // And drop the last reference to DeallocRetainRelease from ObjC global scope.
+        globalDeallocRetainRelease = null
+    }
+
+    assertFalse(deallocRetainReleaseDeallocated)
+
+    // This will dispose DeallocRetainRelease on Kotlin side, which will cause dealloc
+    // on ObjC side, which will immediately schedule another disposal on the Kotlin side.
```
SvyatoslavScherbina, line 106: The comment should be more conditional: mention that this happens if the fix isn't applied.
See e.g. "if the dealloc applied addRef/releaseRef to reclaimed Kotlin object" above.

backend.native/tests/interop/objc/tests/kt41811.kt
```diff
@@ -81,3 +81,29 @@ private fun createGarbageDeallocLoadWeak(weakRef: WeakReferenceProtocol) {
...
+    // Attempt to make the state predictable:
+    kotlin.native.internal.GC.collect()
+
+    deallocRetainReleaseDeallocated = false
+    assertFalse(deallocRetainReleaseDeallocated)
+
+    autoreleasepool {
+        {
+            globalDeallocRetainRelease = object: DeallocRetainRelease() {}
+        }()
```
SvyatoslavScherbina, line 96: The code below should be moved out of the `autoreleasepool`. Otherwise, if the code above actually uses autoreleasepool, the latter can emit a release-ref at the end of the scope, and the object won't be reclaimed by the first GC call.

SvyatoslavScherbina (approved):

SvyatoslavScherbina (approved):

PR #4441, opened 2020-10-13
Export kdocs to framework header

knebekaizer:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -18,6 +24,8 @@ object StubRenderer {
                 }
                 +"*/"
             }
+            this.descriptor?.findKDoc()?.let { +it.render() }
```
SvyatoslavScherbina, line 27: Can it be merged with `this.comment` above?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -18,6 +24,8 @@ object StubRenderer {
                 }
                 +"*/"
             }
+            this.descriptor?.extractKDocString()?.let { +it }
```
SvyatoslavScherbina, line 27: So, AFAIU, this just copies Kotlin comment to Obj-C headers. I guess this is ok in most cases, but
1. Are there simple cases when this doesn't work (i.e. because of different comment syntax in Kotlin and Obj-C)?
2. How hard would it be to extract "the contents" of Kotlin comment and "wrap" it properly as Obj-C comment?

The latter should also help merging this comment with one emitted above, which somewhat makes sense.

SvyatoslavScherbina, line 27: This part should probably go before `this.comment?.contentLines`, because the latter is a `@note` or a `@warning`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -202,3 +210,19 @@ internal fun formatGenerics(buffer: Appendable, generics:List<String>) {
...
+        val psi = source.getPsi()
+        if (psi is KtDeclaration) {
+            if (psi is KtPrimaryConstructor)
+                return null  // to be rendered with class itself
+            val kdoc = psi.docComment
+            if (kdoc != null) {
+                return kdoc.getDefaultSection().parent.text
+            }
+        }
+    }
```
SvyatoslavScherbina, line 224: The part above is quite similar to [`fun DeclarationDescriptor.findKDoc(): KDocTag?` in kotlin repo](https://github.com/JetBrains/kotlin/pull/3978/files#diff-27ccd26943ccb07ece6106b91696eee8cf9e627c6c531f2904fe6cf4ca0f2323R90).
Also the behaviour should match. So it makes some sense to keep only one of these functions.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
knebekaizer: 1. KDoc is a legal comment in ObjC, I believe. So it  shouldn't cause any compilation issues. However KDoc-specific syntax is not the exact equivalent of HeaderDoc syntax used by Apple. OTOH Doxygen seems to be compatibe with our format.  So, the comment will be legal, readable and partially supported by the common documentation tools.

2. Full syntax translation between KDoc and other tools is hardly possible and not a goal probably. As for simple formatting like new lines, spaces, etc it may be not so trivial too. Particularly leading spaces and ` *. ` at each line not always easy to interpret correctly and/or replace or remove for correct representation. Consider a simple block 
```
/********************************
 * A legal and fancy KDoc here  *
 ********************************/     
```
that may easily become misformatted.

I consider any non-trivial re-formatting as dangerous. At a minimal risk and efforts we can replace KDoc-specific leading `/**` with HeaderDoc compatible `/*!` which is also Doxygen-compatible. However, in that way we'll pretend to provide HeaderDoc syntax inside of it, which may be too brave assumption. Does it make sense? Note that Doxygen supports both formats anyway.

knebekaizer: Done.

SvyatoslavScherbina: Thank you for the detailed examination!
I guess placing the kdoc as is should be ok for now, especially considering the complexity brought by any potential modifications.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
SvyatoslavScherbina: (I guess we would avoid this for now, see https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r558416537)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -6,18 +6,32 @@
...
+                    + it.removeSurrounding("/**", "*/")
+                    +"" // Probably makes the following comment more readable.
+                }
                 comment.contentLines.forEach {
                     +" $it"
                 }
                 +"*/"
             }
+                    // Let's try to keep non-trivial kdoc formatting intact, when there is no other comments here
+                    ?: kDocString?.let { +it }
```
SvyatoslavScherbina, line 33: Generally the resulting list is a list of lines, while this code might add a "multiline" string to it. This is probably not critical, but following the established convention is still desirable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -6,18 +6,32 @@
...
+                kDocString?.let {
+                    + it.removeSurrounding("/**", "*/")
+                    +"" // Probably makes the following comment more readable.
+                }
                 comment.contentLines.forEach {
                     +" $it"
                 }
                 +"*/"
             }
+                    // Let's try to keep non-trivial kdoc formatting intact, when there is no other comments here
```
SvyatoslavScherbina, line 32: Actually, your explanation in https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r558416537 is quite convincing, and if the related tooling (e.g. documentation tooltips in Xcode) can properly handle two comments (could you please check this?), then we probably should prefer keeping the kdoc formatting always intact by emitting two comments.

knebekaizer: As I can see, XCode 12 (at least) does handle two HeaderDoc comments correctly. Particularly in my sample the exported kdoc rendered as "Summary" in Quick Help (without any tag it goes as `@brief` by default) and `@note` section rendered as Note, as expected. 
So I'd prefer the most conservative approach here, i.e. avoiding any reformatting of KDoc whenever possible.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
knebekaizer: Fixed.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
SvyatoslavScherbina: Totally agree.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
SvyatoslavScherbina: (This is again the case)

SvyatoslavScherbina:
Everything except https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r559664462, https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r561839353 and some tests is lgtm.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -18,6 +22,10 @@ object StubRenderer {
                 }
                 +"*/"
             }
+
+            // Let's try to keep non-trivial kdoc formatting intact
+            this.descriptor?.extractKDocString()?.let { +it }
```
SvyatoslavScherbina, line 27: Sorry, one more concern:
when the flag is not enabled, this would work differently for "two-stage" and "single-stage" compilation. The latter is generally not used in Gradle projects, but the inconsistency is still not entirely ok, as well as enabling the "experimental" feature by default in single-stage mode.

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -181,6 +181,9 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var libraryToAddToCache: String? = null
 
+    @Argument(value = "-Xexport-kdoc", description = "Export KDoc in framework header")
```
SvyatoslavScherbina, line 184: This description needs updating too.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -6,9 +6,13 @@
...
 
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptorWithSource
+import org.jetbrains.kotlin.backend.common.serialization.extractSerializedKdocString
+import org.jetbrains.kotlin.backend.common.serialization.metadata.findKDocString
 
 object StubRenderer {
-    fun render(stub: Stub<*>): List<String> = collect {
+    fun render(stub: Stub<*>, shouldExportKDoc: Boolean): List<String> = collect {
```
SvyatoslavScherbina, line 15: 1. This is in fact public API used externally, its signature should not be modified like this.
2. `Stub` rendering should not essentially depend on any parameters: `Stub` class is already the structured internal representation for header fragments, including comments. So it is `ObjCExportTranslator`'s responsibility to decide if kdocs should be added to framework comments.

Fixing 2 might be complicated at the moment, because there is no single code fragment which builds comments for `Stub`s.
So, as a trade-off (involving technical debt), we could keep `Stub`s rendering _privately_ depend on `shouldExportKDoc`, while keeping _public_ API intact.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1012,8 +1012,9 @@ abstract class ObjCExportHeaderGenerator internal constructor(
         }
         add("")
 
+        val shouldExportKDoc = (this@ObjCExportHeaderGenerator as? ObjCExportHeaderGeneratorImpl)?.context?.shouldExportKDoc() ?: false
```
SvyatoslavScherbina, line 1015: Somewhat dirty. Adding an `abstract` or `open val` for this flag (and overriding it in that subclass) should be ok.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -18,22 +22,28 @@ object StubRenderer {
                 }
                 +"*/"
             }
+
+            // Let's try to keep non-trivial kdoc formatting intact
+            if (shouldExportKDoc) {
+                this.descriptor?.extractKDocString()?.let { +it }
```
SvyatoslavScherbina, line 28: Comments order is still wrong: https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r543413440

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -6,18 +6,32 @@
...
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptorWithSource
+import org.jetbrains.kotlin.backend.common.serialization.extractSerializedKdocString
+import org.jetbrains.kotlin.backend.common.serialization.metadata.findKDocString
 
 object StubRenderer {
-    fun render(stub: Stub<*>): List<String> = collect {
+    fun render(stub: Stub<*>): List<String> = render_(stub, stub.shouldExportKDoc)
+    private fun render_(stub: Stub<*>, shouldExportKDoc_: Boolean): List<String> = collect {
```
SvyatoslavScherbina, line 16: Underscore is neither required in the name nor matches the Kotlin naming convention.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubRenderer.kt
```diff
@@ -6,18 +6,32 @@
...
+            val kDoc = if (shouldExportKDoc_) {
+                this.descriptor?.extractKDocString()
+            } else null
+            kDoc?.let {
                 +"" // Probably makes the output more readable.
+                +it // Let's try to keep non-trivial kdoc formatting intact
+            }
+
+            this.comment?.let { comment ->
+                kDoc?: let { +"" } // Probably makes the output more readable.
```
SvyatoslavScherbina, line 27: Formatting: space is required before `?:`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/stubs.kt
```diff
@@ -20,6 +20,7 @@ abstract class Stub<out D : DeclarationDescriptor>(val name: String, val comment
         get() = ((descriptor as? DeclarationDescriptorWithSource)?.source as? PsiSourceElement)?.psi
     open val isValid: Boolean
         get() = descriptor?.module?.isValid ?: true
+    internal open val shouldExportKDoc = false
```
SvyatoslavScherbina, line 23: Adding this property doesn't make much sense if you take the flag as a parameter to `render`.
Also associated refactorings shouldn't be required either.

SvyatoslavScherbina: I mean that it is enough to pass `shouldExportKDoc` to `render()` from `ObjCExportHeaderGenerator` instead.

SvyatoslavScherbina (approved):
Everything except https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r583432080 and https://github.com/JetBrains/kotlin-native/pull/4441#discussion_r574419361 lgtm.

Adding tests would also be great, but it's ok to make a separate PR for this later.

PR #4448, opened 2020-10-16
Started to eliminate expensive fqNames usages in IR

LepilkinaElena:


LepilkinaElena:
Connected with PR https://github.com/JetBrains/kotlin/pull/3815

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/NewIrUtils.kt
```diff
@@ -17,24 +17,29 @@ import org.jetbrains.kotlin.ir.expressions.IrConstructorCall
...
 import org.jetbrains.kotlin.ir.types.classifierOrFail
 import org.jetbrains.kotlin.ir.types.isMarkedNullable
 import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
-val IrField.fqNameForIrSerialization: FqName get() = this.parent.fqNameForIrSerialization.child(this.name)
+private fun IrClass.isClassTypeWithSignature(signature: IdSignature.PublicSignature): Boolean {
+    if (!symbol.isPublicApi || !symbol.signature.isPublic) return false
```
SvyatoslavScherbina, line 29: Why is this required?

LepilkinaElena: Symbols which isn't public don't have signature, there is an error inside `signature` implementation for private classes.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/StaticObjects.kt
```diff
@@ -86,22 +84,22 @@ internal fun StaticData.createInitializer(type: IrClass, vararg fields: ConstVal
...
+        "backing" to NullPointer(kObjHeader))
 
     // Now sort these values according to the order of fields returned by getFields()
     // to match the sorting order of the real ArrayList().
-    val sorted = linkedMapOf<String, ConstValue>()
+    val sorted = mutableListOf<ConstValue>()
     context.getLayoutBuilder(arrayListClass).fields.forEach {
-        val fqName = it.fqNameForIrSerialization.asString()
-        sorted.put(fqName, arrayListFields[fqName]!!)
+        if (it.parent == arrayListClass) {
```
SvyatoslavScherbina, line 97: Maybe make it a `require`? Now it seems to just skip fields from superclasses (if any), so this code would silently break if a superclass actually gets a field, leading to later vague LLVM failure.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/InlineClasses.kt
```diff
@@ -267,17 +292,24 @@ private object IrTypeInlineClassesSupport : InlineClassesSupport<IrClass, IrType
...
             clazz.constructors.firstOrNull { it.isPrimary }?.valueParameters?.single()?.type
                     ?: clazz.declarations.filterIsInstance<IrProperty>().single { it.backingField != null }.backingField!!.type
+
+    override fun getPackageFqName(clazz: IrClass) =
+            clazz.packageFqName
+
+    override fun getName(clazz: IrClass): Name? =
+            clazz.name
+
+    override fun isNestedClass(clazz: IrClass): Boolean = clazz.parent !is IrPackageFragment
```
SvyatoslavScherbina, line 314: This is not exactly correct. For example, it would return `true` for local classes too, and it is not clear if local classes are actually nested.
The more straightforward way would be to make `isTopLevel` function instead.

SvyatoslavScherbina: (The semantics is also different from descriptor version, the proposal above should fix this too).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ObjCInterop.kt
```diff
@@ -34,8 +34,11 @@ import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
 
 internal val interopPackageName = InteropFqNames.packageName
 internal val objCObjectFqName = interopPackageName.child(Name.identifier("ObjCObject"))
+internal val objCObjectIdSignature = getPublicSignature(interopPackageName, "ObjCObject")
```
SvyatoslavScherbina, line 37: Can we avoid duplication here by implementing a function like `getTopLevelPublicSignature(fqName: FqName)`

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/NewIrUtils.kt
SvyatoslavScherbina: I see, makes sense.
But this explains only `!symbol.isPublicApi` part. What is `!symbol.signature.isPublic` required for?

SvyatoslavScherbina: Also, the entire pattern `if (symbol.isPublic) symbol.signature` looks like an unsafe leaking abstraction to me.
More safe and encapsulated solution would probably be to define an extension property like `symbol.publicSignatureOrNull` or `symbol.signatureOrNull` near the `IrSymbol` definition.
Need an opinion from @udalov on this though.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/InlineClasses.kt
```diff
@@ -233,14 +253,21 @@ internal object KotlinTypeInlineClassesSupport : InlineClassesSupport<ClassDescr
...
     override fun getInlinedClassUnderlyingType(clazz: ClassDescriptor): KotlinType =
             clazz.unsubstitutedPrimaryConstructor!!.valueParameters.single().type
+
+    override fun getPackageFqName(clazz: ClassDescriptor) =
+            clazz.findPackage().fqName
+
+    override fun getName(clazz: ClassDescriptor) =
+            clazz.name
+
+    override fun isTopLevelClass(clazz: ClassDescriptor): Boolean = clazz.containingDeclaration !is ClassDescriptor
```
SvyatoslavScherbina, line 270: Now this one is wrong in corner cases :)
`clazz.containingDeclaration is PackageFragmentDescriptor` should work ok.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
```diff
@@ -186,8 +188,11 @@ internal class RTTIGenerator(override val context: Context) : ContextUtils {
             vector128Type to 10
     )
 
-    private fun getInstanceSize(classType: LLVMTypeRef?, className: FqName) : Int {
-        val elementType = arrayClasses.get(className.asString())
+    private fun getElementType(irClass: IrClass): LLVMTypeRef? =
+            if (irClass.symbol.isPublicApi) arrayClasses[irClass.symbol.signature as IdSignature.PublicSignature] else null
```
SvyatoslavScherbina, line 192: Why is the cast required here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ir/NewIrUtils.kt
LepilkinaElena: > !symbol.signature.isPublic

It was made for checking that info is consistent inside class, I can remove it without any problems I think.

> symbol.publicSignatureOrNull or symbol.signatureOrNull near the IrSymbol definition.

Good point! Thank you.

udalov: My opinion is that we should remove `isPublic` and make `signature` nullable, it would simplify everything. But not in these changes of course.

LepilkinaElena: > make signature nullable

If so I wouldn't make now it as separate function and just fix this when signature will be nullable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/RTTIGenerator.kt
LepilkinaElena: Because field is `val signature: IdSignature` and error without cast is  `Type inference failed. The value of the type parameter K should be mentioned in input types (argument types, receiver type or expected type). Try to specify it explicitly.`

SvyatoslavScherbina (approved):

homuroll (approved):

PR #4449, opened 2020-10-16
[KT-40670] Override konan.properties from CLI

sbogolepov:


backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
```diff
@@ -278,6 +278,13 @@ class K2NativeCompilerArguments : CommonCompilerArguments() {
     )
     var preLinkCaches: String? = null
 
+    @Argument(
+            value = "-Xoverride-konan-properties",
+            valueDescription = "key1=value1;key2=value2;...",
+            description = "Override konan.properties.values"
+    )
+    var overrideKonanProperties: String? = null
```
LepilkinaElena, line 286: May be use `Array<String>?`. Argument annotation has a delimiter field, it can be set to ";". Then parsing will be easier.

PavelPunegov (approved):

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2Native.kt
```diff
@@ -251,6 +251,7 @@ class K2Native : CLICompiler<K2NativeCompilerArguments>() {
                 }
                 put(DISABLE_FAKE_OVERRIDE_VALIDATOR, arguments.disableFakeOverrideValidator)
                 putIfNotNull(PRE_LINK_CACHES, parsePreLinkCachesValue(configuration, arguments.preLinkCaches))
+                put(OVERRIDE_KONAN_PROPERTIES, parseOverrideKonanProperties(arguments, configuration))
```
LepilkinaElena, line 254: May be save same api as called before `parsePreLinkCachesValue`?

backend.native/cli.bc/src/org/jetbrains/kotlin/cli/bc/K2NativeCompilerArguments.kt
sbogolepov: Didn't know that delimiter is adjustable. Thanks!

LepilkinaElena (approved):

PR #4451, opened 2020-10-19
Remove unused Runtime suspend/resume

projedi:
Reverts #1677

### DIFF
--- a/runtime/src/main/cpp/MemorySharedRefs.cpp
+++ b/runtime/src/main/cpp/MemorySharedRefs.cpp
@@ -17,14 +17,12 @@ OBJ_GETTER(DescribeObjectForDebugging, KConstNativePtr typeInfo, KConstNativePtr
 namespace {
 
 inline bool isForeignRefAccessible(ObjHeader* object, ForeignRefContext context) {
-  if (!Kotlin_hasRuntime()) {
     // So the object is either unowned or shared.
     // In the former case initialized runtime is required to throw exceptions
     // in the latter case -- to provide proper execution context for caller.
     Kotlin_initRuntimeIfNeeded();
-  }
 
-  return IsForeignRefAccessible(object, context);
+    return IsForeignRefAccessible(object, context);
 }
 
 RUNTIME_NORETURN inline void throwIllegalSharingException(ObjHeader* object) {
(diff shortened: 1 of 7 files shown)


runtime/src/main/cpp/MemorySharedRefs.cpp
```diff
@@ -17,14 +17,12 @@ OBJ_GETTER(DescribeObjectForDebugging, KConstNativePtr typeInfo, KConstNativePtr
 namespace {
 
 inline bool isForeignRefAccessible(ObjHeader* object, ForeignRefContext context) {
-  if (!Kotlin_hasRuntime()) {
     // So the object is either unowned or shared.
```
SvyatoslavScherbina, line 20:
```suggestion
    // If runtime has not been initialized on this thread, then the object is either unowned or shared.
```

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -108,11 +96,16 @@ RuntimeState* initRuntime() {
...
+  result->status = RuntimeStatus::kRunning;
   return result;
 }
 
 void deinitRuntime(RuntimeState* state) {
-  ResumeMemory(state->memoryState);
+  RuntimeAssert(state->status == RuntimeStatus::kRunning, "Runtime must be in the running state");
+  state->status = RuntimeStatus::kDestroying;
+  // This may be called after TLS is zeroed out, so ::memoryState in Memory cannot be trusted.
+  RestoreMemory(state->memoryState);
```
SvyatoslavScherbina, line 108: Thing to consider when reworking this: there is a chance that this and further TLS accesses cause memory leak, at least on some platforms.
If TLS is zeroed out, then it is also deallocated. So an access to TLS allocates it again, and I'm not sure thread deinitialization machinery would take care of it.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Runtime.cpp
projedi: I agree. pthread's TLS behaviour (in zeroing out TLS before running `pthread_key_t` destructors) turns out to be rather unhelpful for us.

PR #4459, opened 2020-10-22
Update mimalloc up to version 1.6.7

LepilkinaElena:


konan/konan.properties
```diff
@@ -652,7 +652,7 @@ linkerDynamicFlags.mingw_x64 = -shared
 linkerKonanFlags.mingw_x64 =-static-libgcc -static-libstdc++ \
   -Xclang -flto-visibility-public-std -Wl,--dynamicbase \
   -Wl,-Bstatic,--whole-archive -lwinpthread -Wl,--no-whole-archive,-Bdynamic \
-  -Wl,--defsym,__cxa_demangle=Konan_cxa_demangle
+  -Wl,--defsym,__cxa_demangle=Konan_cxa_demangle -lbcrypt
```
sbogolepov, line 655: I suggest to avoid adding this dependency unconditionally. You can add a separate property (e.g. `mimallocLinkerDependencies.mingw_x64`) and add it in `Linker.kt` if we're compiling with mimalloc enabled.

build-tools/src/main/kotlin/org/jetbrains/kotlin/bitcode/CompileToBitcode.kt
```diff
@@ -70,7 +70,8 @@ open class CompileToBitcode @Inject constructor(
                 Language.C ->
                     // Used flags provided by original build of allocator C code.
                     listOf("-std=gnu11", "-O3", "-Wall", "-Wextra", "-Wno-unknown-pragmas",
```
projedi, line 72: What about moving these flags to `runtime/build.gradle.kts`? There's already a `compilerArgs` override for `mimalloc`.
Here we can just leave
```kotlin
Language.C ->
    listOf("-std=gnu11", "-Werror", "-O3",
            "-Wall", "-Wextra")
```
to mimic C++ flags below (except for optimization level and language standard).

runtime/src/mimalloc/c/include/mimalloc-internal.h
```diff
@@ -221,27 +234,120 @@ static inline size_t _mi_wsize_from_size(size_t size) {
...
 
+// Does malloc satisfy the alignment constraints already?
+static inline bool mi_malloc_satisfies_alignment(size_t alignment, size_t size) {
+  return (alignment == sizeof(void*) || (alignment == MI_MAX_ALIGN_SIZE && size > (MI_MAX_ALIGN_SIZE/2)));
+}
 
-/* -----------------------------------------------------------
-  The thread local default heap
------------------------------------------------------------ */
+// Overflow detecting multiply
```
projedi, line 242: We used to have some custom handling for `mi_mul_overflow` https://github.com/JetBrains/kotlin-native/blob/19317fa7269f2f17624e043368469f0fa6e1c568/runtime/src/mimalloc/c/include/mimalloc-internal.h#L169 is it not needed anymore?

runtime/src/mimalloc/c/static.c
```diff
@@ -4,10 +4,14 @@ This is free software; you can redistribute it and/or modify it under the
 terms of the MIT license. A copy of the license can be found in the file
 "licenses/third_party/mimalloc_LICENSE.txt" at the root of this distribution.
 -----------------------------------------------------------------------------*/
-// Copyright 2019-2020 JetBrains s.r.o.
-
-#if !KONAN_MI_MALLOC
```
projedi, line 9: Is it not needed anymore?

projedi (commented):
It'd also be great to add some sort of README to mimalloc specifying where to get this code (e.g repo address, version and/or commit hash). Also, some steps for how to update mimalloc, and how to introduce changes would be superb. For the latter maybe something like:

To ease updating mimalloc, follow these rules when introducing changes:
* To add code, do
```c
#if defined(KONAN_MI_MALLOC)
<new code>
#endif // KONAN_MI_MALLOC
```
* To delete code, do
```c
#if !defined(KONAN_MI_MALLOC)
<code to delete>
#endif // KONAN_MI_MALLOC
```
* To modify code, do
```c
#if !defined(KONAN_MI_MALLOC)
<current code>
#else // KONAN_MI_MALLOC
<modified code>
#endif // KONAN_MI_MALLOC
```
or
```c
#if defined(KONAN_MI_MALLOC)
<modified code>
#else // KONAN_MI_MALLOC
<current code>
#endif // KONAN_MI_MALLOC
```

ilmat192 (approved):
LGTM modulo comments from other reviewers.

runtime/src/mimalloc/c/include/mimalloc-internal.h
LepilkinaElena: It's needed, I now made another hack, better to use this one, you're right. Fixed.

runtime/src/mimalloc/c/static.c
LepilkinaElena: Yes, this file is just excluded now.

projedi (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -137,6 +137,7 @@ internal class Linker(val context: Context) {
         }
 
         val needsProfileLibrary = context.coverage.enabled
+        val mimallocEnabled = config.get(KonanConfigKeys.ALLOCATION_MODE) == "mimalloc"
```
projedi, line 140: There's an edge case: if you enabled `mimalloc` via command line on a target that doesn't support it, we would complain and link standard allocator instead: https://github.com/JetBrains/kotlin-native/blob/a710a478a522b1ba03e66d3dd85f2d5a8213e83d/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt#L120

sbogolepov (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
LepilkinaElena: Good point.

projedi (approved):

PR #4461, opened 2020-10-22
Fix tests target usage

PavelPunegov:


backend.native/tests/build.gradle
```diff
@@ -4270,8 +4270,11 @@ dynamicTest("produce_dynamic") {
                 "Error handler: kotlin.Error: Expected error\n"
 }
 
+println("HOSTNAME: ${project.hostName}")
```
projedi, line 4273: Should we leave them?

backend.native/tests/build.gradle
```diff
@@ -4309,7 +4311,7 @@ dynamicTest("interop_cleaners_second_thread") {
 }
 
 dynamicTest("interop_cleaners_leak") {
-    disabled = (project.testTarget != null && project.testTarget != project.hostName)
+    disabled = (project.target.name != project.hostName)
```
projedi, line 4314: I merged in another two tests that use `std::thread` here: https://github.com/JetBrains/kotlin-native/pull/4494/files#diff-d3c79e0be6518d77481d6b73a22db4f79ab351158fda879bc88bb6b237fd6512R4330 I checked that the tests pass fine with `project.target.name != project.hostName` on everything but mingw.

backend.native/tests/build.gradle
```diff
@@ -931,7 +939,6 @@ standaloneTest("cleaner_in_main_without_checker") {
 standaloneTest("cleaner_leak_release") {
     enabled = !project.globalTestArgs.contains('-g') && (project.testTarget != 'wasm32') // Cleaners need workers
     source = "runtime/basic/cleaner_leak.kt"
-    goldValue = ""
```
SvyatoslavScherbina, line 934: Is removing `goldValue` still required?

SvyatoslavScherbina (approved):

backend.native/tests/build.gradle
PavelPunegov: Actually not, I'm going to rebase and squash some commits. Also this setting goldValue is not needed as this is an optional nullable parameter.

SvyatoslavScherbina: > Also this setting goldValue is not needed as this is an optional nullable parameter.

`goldValue = ""` isn't the same as `goldValue = null`.
The former checks that output is empty (and this might be the original intention), the latter doesn't.

projedi: > and this might be the original intention

It is.

PR #4467, opened 2020-10-27
Improve compatibility with newer GCC+Clang, resolve some warnings

serebit:
This PR includes three modifications:
- Resolve `kotlin.native.home` and `org.jetbrains.kotlin.native.home` in the `backend.native:tests` package
- Update C and C++ files to improve their compatibility with newer versions of GCC/Glibc and Clang, while retaining compatibility with the current versions
- Resolve several compiler warnings in Kotlin source files

### DIFF
--- a/build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
+++ b/build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
@@ -10,14 +10,14 @@ import kotlin.math.min
  * Examples of numeric version strings: "12.4.1.2", "9", "0.5".
  */
 private fun compareStringsAsVersions(version1: String, version2: String): Int {
-    val version1 = version1.split('.').map { it.toInt() }
-    val version2 = version2.split('.').map { it.toInt() }
-    val minimalLength = min(version1.size, version2.size)
+    val splitVersion1 = version1.split('.').map { it.toInt() }
+    val splitVersion2 = version2.split('.').map { it.toInt() }
+    val minimalLength = min(splitVersion1.size, splitVersion2.size)
     for (index in 0 until minimalLength) {
-        if (version1[index] < version2[index]) return -1
-        if (version1[index] > version2[index]) return 1
+        if (splitVersion1[index] < splitVersion2[index]) return -1
+        if (splitVersion1[index] > splitVersion2[index]) return 1
     }
-    return version1.size.compareTo(version2.size)
+    return splitVersion1.size.compareTo(splitVersion2.size)
 }
 
 /**
--- a/runtime/src/main/kotlin/kotlin/text/regex/Lexer.kt
+++ b/runtime/src/main/kotlin/kotlin/text/regex/Lexer.kt
@@ -457,7 +457,7 @@ internal class Lexer(val patternString: String, flags: Int) {
             // Word/whitespace/digit.
             'w', 's', 'd', 'W', 'S', 'D' -> {
                 lookAheadSpecialToken = AbstractCharClass.getPredefinedClass(
-                        String(pattern, prevNonWhitespaceIndex, 1),
+                        pattern.concatToString(prevNonWhitespaceIndex, prevNonWhitespaceIndex + 1),
                         false
                 )
                 lookAhead = 0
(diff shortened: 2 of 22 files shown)


runtime/src/main/cpp/Memory.cpp
```diff
@@ -3133,17 +3133,17 @@ bool TryAddHeapRef(const ObjHeader* object) {
   return tryAddHeapRef(object);
 }
 
-void ReleaseHeapRefStrict(const ObjHeader* object) {
+__attribute__((nothrow)) void ReleaseHeapRefStrict(const ObjHeader* object) {
```
SvyatoslavScherbina, line 3136: What's wrong with `RUNTIME_NOTHROW` macro?

SvyatoslavScherbina:
Please rebase this PR to master branch first.

serebit:
> Please rebase this PR to master branch first.

Done.

build-tools/src/main/kotlin/org/jetbrains/kotlin/XcRunRuntimeUtils.kt
```diff
@@ -9,9 +9,9 @@ import kotlin.math.min
  * Compares two strings assuming that both are representing numeric version strings.
  * Examples of numeric version strings: "12.4.1.2", "9", "0.5".
  */
-private fun compareStringsAsVersions(version1: String, version2: String): Int {
-    val version1 = version1.split('.').map { it.toInt() }
-    val version2 = version2.split('.').map { it.toInt() }
+private fun compareStringsAsVersions(version1String: String, version2String: String): Int {
```
sbogolepov, line 12: It's obvious to caller that `version1String` has String type, so "String" suffix is meaningless. If you want to avoid name shadowing, it is better to rename local variable.

runtime/src/main/cpp/Memory.h
```diff
@@ -107,7 +107,7 @@ extern "C" {
 
 #define OBJ_RESULT __result__
 #define OBJ_GETTER0(name) ObjHeader* name(ObjHeader** OBJ_RESULT)
-#define OBJ_GETTER(name, ...) ObjHeader* name(__VA_ARGS__, ObjHeader** OBJ_RESULT)
+#define OBJ_GETTER(name, ...) RUNTIME_NOTHROW ObjHeader* name(__VA_ARGS__, ObjHeader** OBJ_RESULT)
```
projedi, line 110: I'm not sure that's correct. For example, see `Kotlin_Array_get`
https://github.com/JetBrains/kotlin-native/blob/1aef9da517f5b56a31e213929e6416d0751e96ab/runtime/src/main/cpp/Arrays.cpp#L99-L103
It calls `boundCheck`
https://github.com/JetBrains/kotlin-native/blob/1aef9da517f5b56a31e213929e6416d0751e96ab/runtime/src/main/cpp/Arrays.cpp#L37-L44
which does throw an exception. And since Kotlin/Native exceptions are actually C++ exceptions:
https://github.com/JetBrains/kotlin-native/blob/1aef9da517f5b56a31e213929e6416d0751e96ab/runtime/src/main/cpp/Exceptions.cpp#L209-L218
this may result in `Kotlin_Array_get` terminating the program instead of throwing an exception.

projedi (approved):

sbogolepov (approved):

SvyatoslavScherbina:
@serebit the tests are failing. Please run `./gradlew :backend.native:tests:sanity` and fix the failures.

konan/konan.properties
```diff
@@ -727,4 +727,4 @@ runtimeDefinitions.wasm32 = KONAN_WASM=1 KONAN_NO_FFI=1 KONAN_NO_THREADS=1 \
   KONAN_INTERNAL_NOW=1 KONAN_NO_MEMMEM KONAN_NO_CTORS_SECTION=1
 
 # The version of Kotlin/Native compiler
-compilerVersion=@compilerVersion@
\ No newline at end of file
+compilerVersion=@compilerVersion@
```
PavelPunegov, line 730: It's beter to avoid changing files by adding new lines without actual changes to content of the file.

PavelPunegov:
According to the test runs `:backend.native:tests:harmonyRegexTest` started to fail

runtime/src/main/kotlin/kotlin/text/regex/Lexer.kt
```diff
@@ -457,7 +457,7 @@ internal class Lexer(val patternString: String, flags: Int) {
             // Word/whitespace/digit.
             'w', 's', 'd', 'W', 'S', 'D' -> {
                 lookAheadSpecialToken = AbstractCharClass.getPredefinedClass(
-                        String(pattern, prevNonWhitespaceIndex, 1),
+                        pattern.concatToString(prevNonWhitespaceIndex, 1),
```
PavelPunegov, line 460: It looks like it should be `pattern.concatToString(prevNonWhitespaceIndex, prevNonWhitespaceIndex+1)`

PavelPunegov (changes requested):

konan/konan.properties
serebit: Yeah, I agree. This happened because I have IntelliJ set to automatically add a newline to EOF when saving files, and I did test modifications to konan.properties, so it saved with a linefeed at EOF. I can revert that.

serebit:
> @serebit the tests are failing. Please run `./gradlew :backend.native:tests:sanity` and fix the failures.

Will do, thanks for the heads up.

serebit:
Just ran all tests again locally after modifications, no failures. Should be good to go.

SvyatoslavScherbina:
@serebit please avoid force-pushing the branch. Adding review fixes as separate commits makes reviewing them easier.

PavelPunegov (approved):

PR #4472, opened 2020-10-27
Wrap worker jobs to autoreleasepool

SvyatoslavScherbina:
#KT-42822 Fixed.

### DIFF
--- a/backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
+++ b/backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
@@ -43,6 +43,29 @@ fun testExecuteAfter0InWorkerProcessQueue() =
 fun testExecuteAfter0InMainToWorkerNoYield() =
         test(ExecuteAfter0, InMainToWorker, NoYield)
 
+@Test
+fun testExecuteAfter10InMainPark() =
+        testExecuteAfter10(InMain, Park)
+
+@Test
+fun testExecuteAfter10InMainProcessQueue() =
+        testExecuteAfter10(InMain, ProcessQueue)
+
+@Test
+fun testExecuteAfter10InWorkerPark() =
+        testExecuteAfter10(InWorker, Park)
+
+@Test
+fun testExecuteAfter10InWorkerProcessQueue() =
+        testExecuteAfter10(InWorker, ProcessQueue)
+
+@Test
+fun testExecuteAfter10InMainToWorkerNoYield() =
+        testExecuteAfter10(InMainToWorker, NoYield)
+
+private fun testExecuteAfter10(context: Context, yieldMethod: Yield) =
+        test(ExecuteAfter10(yieldMethod), context, yieldMethod)
+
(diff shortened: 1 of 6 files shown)


runtime/src/main/cpp/ObjCMMAPI.h
```diff
@@ -12,6 +12,22 @@
 
 extern "C" ALWAYS_INLINE void Kotlin_ObjCExport_releaseAssociatedObject(void* associatedObject);
 
+namespace konan {
```
projedi, line 15: I think we'll need to specify how we use namespaces.

backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
```diff
@@ -0,0 +1,153 @@
...
+
+        // autorelease pool is processed after the job is finished, so the object should be deallocated;
+        // Checking in a job to make sure previous job is completely processed:
+        execute {
+            assertTrue(deallocated.value)
+        }
+    }
+}
+
+interface ExecuteMethod<Job> {
```
projedi, line 70: According to usage, maybe `Result` is a better name than `Job`

backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
```diff
@@ -0,0 +1,153 @@
+import kotlin.native.concurrent.*
+import kotlinx.cinterop.*
+import kotlin.test.*
+import objcTests.*
+
+@Test
+fun testExecute1() =
```
projedi, line 7: I think names like `testExecuteInMainPark` will look better in test failures report.

projedi (approved):

backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
SvyatoslavScherbina: `Result` is kinda ambiguous here. Maybe `Future`? Since it might clash with `Future` class, I'd go with `F`. What do you think?

projedi: `F` is fine. `JobFuture` is another alternative.

SvyatoslavScherbina: Done.

backend.native/tests/interop/objc/tests/workerAutoreleasePool.kt
SvyatoslavScherbina: Done.

runtime/src/main/cpp/ObjCMMAPI.h
SvyatoslavScherbina: Yes.

SvyatoslavScherbina:
I've also added tests for `executeAfter` with actual timeout, please take a look.

projedi (approved):

PR #4477, opened 2020-10-29
Introduce CheckAbiCompatibility task.

sbogolepov:
Since Kotlin/Native is moving towards stable ABI it is important to check between releases that there are no unexpected breaking changes.
${target}CheckAbiCompatibility task allows to compare `klib signatures` dumps of stdlib and platforms libs from current distribution and the given one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ import org.jetbrains.kotlin.KotlinBuildPusher
 import org.jetbrains.kotlin.CollisionDetector
 import org.jetbrains.kotlin.CollisionTransformer
 import org.jetbrains.kotlin.CompilationDatabaseKt
+import org.jetbrains.kotlin.CompareDistributionSignatures
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -778,3 +779,14 @@ task compdb(type: Copy) {
...
     into "$projectDir"
 }
+
+if (project.hasProperty("anotherDistro")) {
+    targetList.each { targetName ->
+        task "${targetName}CheckAbiCompatibility"(type: CompareDistributionSignatures) {
+            dependsOn "${targetName}PlatformLibs"
+
+            target = targetName
+            oldDistribution = project.findProperty("anotherDistro")
```
PavelPunegov, line 789: Should we add `onMismatchMode = FAIL` here? As soon as these tasks are not run regulary they are not going to fail on annotations making noise. If they are intended to be used as tasks to be run during the release, I think it's better if they fail to make release manager pay attention to them.

PavelPunegov (approved):

build.gradle
sbogolepov: Makes sense, thanks!

PR #4479, opened 2020-10-29
Remove unneeded ArraysTest.cpp

projedi:


.idea/vcs.xml
```diff
@@ -16,5 +16,6 @@
   </component>
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/runtime/googletest" vcs="Git" />
```
SvyatoslavScherbina, line 19: I have no idea about the consequences of this patch, so generally I can't verify it.

SvyatoslavScherbina (approved):

.idea/vcs.xml
projedi: IDEA just keeps automatically adding this change after you run runtime tests. If you never have run those (so, don't have `runtime/googletest` checked out), then IDEA will complain that there's an invalid VCS mapping, but that's it.

SvyatoslavScherbina: Probably ok for now.

PR #4480, opened 2020-10-29
Remove unneeded zeroing TLS

projedi:


backend.native/tests/runtime/workers/worker_threadlocal_no_leak.kt
```diff
@@ -0,0 +1,19 @@
...
+
+@ThreadLocal
+var x = Any()
+
+fun main() {
+    val worker = Worker.start()
+
+    worker.execute(TransferMode.SAFE, {}) {
+        println(x)  // Make sure x is initialized
+    }
```
SvyatoslavScherbina, line 16: Shouldn't we get the result to dispose the future here?
How does this work otherwise?

SvyatoslavScherbina: (I guess `Unit` just doesn't "leak")

projedi: I step in it every time… Fixed, thanks!

SvyatoslavScherbina (approved):

PR #4481, opened 2020-10-29
Put FrameOverlay into MemoryState

projedi:


runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -2441,8 +2442,8 @@ void enterFrame(ObjHeader** start, int parameters, int count) {
   MEMORY_LOG("EnterFrame %p: %d parameters %d locals\n", start, parameters, count)
   FrameOverlay* frame = reinterpret_cast<FrameOverlay*>(start);
   if (Strict) {
-    frame->previous = currentFrame;
-    currentFrame = frame;
+    frame->previous = ::memoryState->currentFrame;
```
SvyatoslavScherbina, line 2445: Does this affect performance? Same question for `leaveFrame` of course.

projedi: Honestly, wouldn't think that simply adding a constant offset to the pointer before reading/writing would have a measurable impact. But it does have a negative one. Probably because it rejuggled registers in the resulting code.

Another time, then.

SvyatoslavScherbina: > simply adding a constant offset to the pointer before reading/writing

Is it?
`MemoryState` is stored by pointer in a thread local, so doesn't moving `currentFrame` to `memoryState` actually add a memory access?

projedi: Currently: you read `currentFrame` address into a register, and then use it.
Proposed: you read `memoryState` address into a register, then add a constant offset to it, and then use it.

That's what was going on in the assembly.

projedi: Nope, my bad: misread the assembly. It does dereference `memoryState` and then adds an offset to it, so we indeed have one more indirection, which explains the slowdown.

PR #4482, opened 2020-10-29
Add a switch to destroy runtime only on shutdown

projedi:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2387,6 +2389,38 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
         llvmDebugSelector.setLinkage(LLVMLinkage.LLVMExternalLinkage)
     }
 
+    private fun appendGlobal(name: String, value: ConstValue) {
```
SvyatoslavScherbina, line 2392: The name is somewhat vague. `overrideRuntimeGlobal` should be more clear.

backend.native/tests/interop/migrating_main_thread/main.cpp
```diff
@@ -8,18 +8,26 @@
 #include <cassert>
 #include <thread>
 
+constexpr int kInitialValue = 0;
+constexpr int kNewValue = 0;
```
SvyatoslavScherbina, line 12: Why are values the same?

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -31,6 +31,12 @@ struct InitNode {
   InitNode* next;
 };
 
+RUNTIME_WEAK DestroyRuntimeMode Kotlin_destroyRuntimeMode = DESTROY_RUNTIME_ON_SHUTDOWN;
```
SvyatoslavScherbina, line 34: Needs a comment about overriding this global by the compiler.

runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -2005,22 +2005,49 @@ MemoryState* initMemory() {
...
+      // Nothing to do.
+      break;
+  }
+  if (firstRuntime) {
 #if USE_CYCLIC_GC
     cyclicInit();
 #endif  // USE_CYCLIC_GC
     memoryState->isMainThread = true;
   }
+  if (!firstRuntime && firstMemoryState) {
```
SvyatoslavScherbina, line 2022: This part looks totally cryptic to me. Which problem does it solve?

projedi: "Migrating" main thread.

Suppose we had a main thread, then this thread dies, and then later another thread appears and initializes Kotlin runtime on it.
In legacy mode this new thread becomes the main thread, and all the globals are freshly initialized on it.
I mimic this behaviour in the new mode: this new thread becomes the main thread, globals are not freshly initialized, but are now accessible on it. For the latter to work correctly I use `synchronize` in `deinitMemory` to publish all refcounts of globals and their object subgraphs, and `synchronize` here to acquire them.

If the main thread died while there were other threads, then in legacy mode globals become inaccessible (until all the threads die and a new one is born), and so it is in the new mode.

For experimental MM this behaviour is unneeded, because globals will be accessible from any thread regardless.

SvyatoslavScherbina: Doesn't sound like worth fixing to me.

> Suppose we had a main thread, then this thread dies

Things are already pretty broken here. When K/N runtime considers some thread "main" while it is actually not a platform main thread, it is weird and in fact a bug. And platform main thread usually doesn't die.
Changing "main" thread during single execution is awkward too.

On the other hand, even if the solution is correct, it is still quite hard to reason about, considering all the memory model restrictions and GC implementation details. This just adds one more way for migrating thread-local objects to other threads.

backend.native/tests/interop/migrating_main_thread/main.cpp
projedi: Oh, wow. Fixed.

runtime/src/legacymm/cpp/Memory.cpp
projedi: Ok, removed this logic.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -2378,6 +2379,7 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
         LLVMSetSection(llvmUsedGlobal.llvmGlobal, "llvm.metadata")
     }
 
+    // TODO: Consider migrating `KonanNeedDebugInfo` to the `appendGlobal` mechanism from below.
```
SvyatoslavScherbina, line 2382: Obsolete name `appendGlobal` in the comment.

SvyatoslavScherbina (approved):

ilmat192 (approved):

PR #4493, opened 2020-11-02
Forbid subclasses of NativePointed to have backing fields 

SvyatoslavScherbina:
(this crashes the compiler otherwise)

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/SpecialBackendChecksTraversal.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/SpecialBackendChecksTraversal.kt
@@ -317,7 +317,7 @@ private class BackendChecker(val context: Context, val irFile: IrFile) : IrEleme
             )
         }
 
-        if (callee.returnType.getInlinedClassNative()?.descriptor == interop.nativePointed)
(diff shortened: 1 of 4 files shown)


homuroll (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/SpecialBackendChecksTraversal.kt
```diff
@@ -317,10 +317,19 @@ private class BackendChecker(val context: Context, val irFile: IrFile) : IrEleme
             )
         }
 
-        if (callee.returnType.getInlinedClassNative()?.descriptor == interop.nativePointed)
+        if (callee.returnType.isNativePointed(symbols))
             reportError(expression, "Native interop types constructors must not be called directly")
     }
 
+    override fun visitField(declaration: IrField) {
```
homuroll, line 324: BTW,  what's the status of fake override fields (are they still present in IR)? Can we encounter one here?

SvyatoslavScherbina: Nice catch!
I guess we can't: https://github.com/JetBrains/kotlin/commit/c015463926315c6bba1d68ed1f51363ec1792af9
But added a check to avoid possible problems in future.

PR #4494, opened 2020-11-02
Tests on migrating main thread and memory leaks with interop

projedi:


SvyatoslavScherbina (approved):

backend.native/tests/build.gradle
```diff
@@ -4327,6 +4327,25 @@ dynamicTest("interop_cleaners_leak") {
     flags = ['-Xopt-in=kotlin.native.internal.InternalForKotlinNative']
 }
 
+dynamicTest("interop_migrating_main_thread") {
+    // TODO: Fix it.
```
projedi, line 4331: These tests suffer from the same problem as all `dynamicTest`s above. They are getting collectively fixed in https://github.com/JetBrains/kotlin-native/pull/4461

SvyatoslavScherbina (approved):

PR #4495, opened 2020-11-03
[psi2ir] Clear FE bindingContext after initial analysis

homuroll:
It should not be used anywhere in the BE


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/PsiToIr.kt
```diff
@@ -165,6 +166,10 @@ internal fun Context.psiToIr(
             expectDescriptorToSymbol = if (expectActualLinker) expectDescriptorToSymbol else null
     )
 
+    val originalBindingContext = bindingContext as? CleanableBindingContext
+            ?: error("BindingContext should be cleanable in K/N IR to avoid leaking memory: $bindingContext")
+    originalBindingContext.clear()
```
SvyatoslavScherbina, line 171: Some of the code below also has access to `bindingContext`, however probably doesn't actually use it.
But still it would be more safe to clear it at the end of this phase.

JVM backend seems to clear binding context later than here too.

Also it makes some sense to invalidate `bindingContext` in `Context` too, i.e. to have `bindingContext = null` here.

SvyatoslavScherbina (approved):

PR #4497, opened 2020-11-03
Extract Kotlin_shutdownRuntime

projedi:


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -76,7 +76,44 @@ inline bool isValidRuntime() {
...
+    ScopedInitializingRuntime() { atomicAdd(&initializingRuntimesCount, 1); }
+
+    ~ScopedInitializingRuntime() { atomicAdd(&initializingRuntimesCount, -1); }
+
+    static bool IsInitializing() { return atomicGet(&initializingRuntimesCount) > 0; }
+
+    ScopedInitializingRuntime(const ScopedInitializingRuntime&) = delete;
+    ScopedInitializingRuntime(ScopedInitializingRuntime&&) = delete;
+    ScopedInitializingRuntime& operator=(const ScopedInitializingRuntime&) = delete;
+    ScopedInitializingRuntime& operator=(ScopedInitializingRuntime&&) = delete;
```
SvyatoslavScherbina, line 99: Is there a C++ magic spell that would make this automatically? :)
Maybe extracting a `Scope*` superclass or something.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -149,6 +186,63 @@ void Kotlin_deinitRuntimeIfNeeded() {
...
+    // Cleaners are now done, disallow new runtimes.
+    lastStatus = compareAndSwap(&globalRuntimeStatus, kGlobalRuntimeShuttingDown, kGlobalRuntimeShutdown);
+    RuntimeAssert(lastStatus == kGlobalRuntimeShuttingDown, "Must be in ShuttingDown state");
+
+    // Spin until all runtimes have fully initialized.
+    while (ScopedInitializingRuntime::IsInitializing()) {
+    }
+
+    // TODO: If we add early return at the top, this if would be unneeded.
+    if (Kotlin_memoryLeakCheckerEnabled() || Kotlin_memoryLeakCheckerEnabled()) {
```
SvyatoslavScherbina, line 228: Why check the same twice?

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -149,6 +186,63 @@ void Kotlin_deinitRuntimeIfNeeded() {
...
+        // First make sure workers are gone.
+        WaitNativeWorkersTermination();
+
+        // Now check for existence of any other runtimes.
+        // `aliveRuntimesCount` can only go down, because we forbade new runtimes initialization.
+        auto otherRuntimesCount = atomicGet(&aliveRuntimesCount) - 1;
+        RuntimeAssert(otherRuntimesCount >= 0, "Cannot be negative");
+        if (otherRuntimesCount > 0) {
+            konan::consoleErrorf("Cannot run checkers when there are %d alive runtimes at the shutdown", otherRuntimesCount);
+            konan::abort();
```
SvyatoslavScherbina, line 238: Can probably break existing working code (e.g. if the thread that was not started as a worker is still running?)

runtime/src/main/cpp/Runtime.cpp
projedi: Probably. It's also enough to delete just copy-ctor + copy-assignment, since that automatically deletes move-ctor + move-assignment. However, I don't see anything wrong with being explicit here.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -149,6 +186,63 @@ void Kotlin_deinitRuntimeIfNeeded() {
...
+
+    // Stop cleaner worker. Only execute the cleaners if checker is enabled.
+    ShutdownCleaners(Kotlin_cleanersLeakCheckerEnabled());
+
+    // Cleaners are now done, disallow new runtimes.
+    lastStatus = compareAndSwap(&globalRuntimeStatus, kGlobalRuntimeShuttingDown, kGlobalRuntimeShutdown);
+    RuntimeAssert(lastStatus == kGlobalRuntimeShuttingDown, "Must be in ShuttingDown state");
+
+    // Spin until all runtimes have fully initialized.
+    while (ScopedInitializingRuntime::IsInitializing()) {
```
SvyatoslavScherbina, line 224: One more potential dead lock at the termination.
Also more complications below.
Do we absolutely need all this stuff?

runtime/src/main/cpp/Runtime.cpp
SvyatoslavScherbina: > However, I don't see anything wrong with being explicit here.

Too explicit. This is basically a boilerplate.

runtime/src/main/cpp/Runtime.cpp
projedi: Simplified a bit.

runtime/src/main/cpp/Runtime.cpp
projedi: You're right, removed `konan::abort` for now.

runtime/src/main/cpp/Runtime.cpp
projedi: My mistake, meant to check for cleaners checker.

runtime/src/main/cpp/Runtime.cpp
projedi: Removed the class completely.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -76,15 +76,30 @@ inline bool isValidRuntime() {
...
-  SetKonanTerminateHandler();
   RuntimeState* result = konanConstructInstance<RuntimeState>();
   if (!result) return kInvalidRuntime;
   RuntimeCheck(!isValidRuntime(), "No active runtimes allowed");
   ::runtimeState = result;
+
+  bool firstRuntime = atomicAdd(&aliveRuntimesCount, 1) == 1;
+  auto lastStatus = compareAndSwap(&globalRuntimeStatus, kGlobalRuntimeUninitialized, kGlobalRuntimeRunning);
+  if (lastStatus == kGlobalRuntimeShutdown) {
+      konan::consoleErrorf("Kotlin runtime was shut down. Cannot create new runtimes\n");
```
SvyatoslavScherbina, line 96: Error message is not actionable: "runtime" term is used only internally, so this phrase generally means nothing for users.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -76,15 +76,30 @@ inline bool isValidRuntime() {
...
+      konan::abort();
```
SvyatoslavScherbina, line 97: I'm still not sure we should crash in this case.
Current implementation probably works.

runtime/src/main/cpp/Runtime.cpp
```diff
@@ -149,6 +164,42 @@ void Kotlin_deinitRuntimeIfNeeded() {
...
+    // TODO: If we add early return at the top, this if would be unneeded.
+    if (Kotlin_memoryLeakCheckerEnabled() || Kotlin_cleanersLeakCheckerEnabled()) {
+        // First make sure workers are gone.
+        WaitNativeWorkersTermination();
+
+        // Now check for existence of any other runtimes.
+        auto otherRuntimesCount = atomicGet(&aliveRuntimesCount) - 1;
+        RuntimeAssert(otherRuntimesCount >= 0, "Cannot be negative");
+        if (otherRuntimesCount > 0) {
+            konan::consoleErrorf("Cannot run checkers when there are %d alive runtimes at the shutdown", otherRuntimesCount);
```
SvyatoslavScherbina, line 194: Checkers are enabled by default. This means that this error message would bother all users, including those who aren't interested in it.

SvyatoslavScherbina (approved):

PR #4498, opened 2020-11-05
Add facility to restrict extra classes in Mobile ObjC Generator

fkistner:
- Implement way to restrict extra classes from other modules in Mobile ObjC generator
- Split generation into smaller steps
- Add API to retrieve stubs and foundation imports

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
@@ -6,8 +6,10 @@
 package org.jetbrains.kotlin.backend.konan.objcexport
 
 import org.jetbrains.kotlin.backend.konan.*
+import org.jetbrains.kotlin.backend.konan.Context
 import org.jetbrains.kotlin.backend.konan.descriptors.getPackageFragments
 import org.jetbrains.kotlin.backend.konan.descriptors.isInterface
+import org.jetbrains.kotlin.backend.konan.getExportedDependencies
 import org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator
 import org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportBlockCodeGenerator
 import org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportCodeGenerator
@@ -18,10 +20,7 @@ import org.jetbrains.kotlin.ir.util.SymbolTable
 import org.jetbrains.kotlin.konan.exec.Command
 import org.jetbrains.kotlin.konan.file.File
 import org.jetbrains.kotlin.konan.file.createTempFile
-import org.jetbrains.kotlin.konan.target.AppleConfigurables
-import org.jetbrains.kotlin.konan.target.CompilerOutputKind
-import org.jetbrains.kotlin.konan.target.Family
-import org.jetbrains.kotlin.konan.target.KonanTarget
+import org.jetbrains.kotlin.konan.target.*
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.name.isSubpackageOf
@@ -33,12 +32,6 @@ internal class ObjCExportedInterface(
         val headerLines: List<String>,
         val namer: ObjCExportNamer,
         val mapper: ObjCExportMapper
-)
-
-data class ObjCExportedStubs(
-    val classForwardDeclarations: Set<String>,
-    val protocolForwardDeclarations: Set<String>,
-    val stubs: List<Stub<*>>
 )
 
 internal class ObjCExport(val context: Context, symbolTable: SymbolTable) {
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
@@ -6,10 +6,8 @@
 package org.jetbrains.kotlin.backend.konan.objcexport
 
 import org.jetbrains.kotlin.backend.konan.*
-import org.jetbrains.kotlin.backend.konan.Context
 import org.jetbrains.kotlin.backend.konan.descriptors.getPackageFragments
 import org.jetbrains.kotlin.backend.konan.descriptors.isInterface
-import org.jetbrains.kotlin.backend.konan.getExportedDependencies
 import org.jetbrains.kotlin.backend.konan.llvm.CodeGenerator
 import org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportBlockCodeGenerator
 import org.jetbrains.kotlin.backend.konan.llvm.objcexport.ObjCExportCodeGenerator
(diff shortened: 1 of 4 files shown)


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
```diff
@@ -34,6 +35,12 @@ internal class ObjCExportedInterface(
         val mapper: ObjCExportMapper
 )
 
+data class ObjCExportedStubs(
```
SvyatoslavScherbina, line 38: This file is a wrong place for this class.
The file generally covers high-level workflow for generating Obj-C export in compiler. And the latter doesn't even use this class.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1054,19 +1046,36 @@ abstract class ObjCExportHeaderGenerator internal constructor(
...
         // to stabilize name mangling.
+        translateBaseDeclarations()
+        translatePackageFragments()
+        translateExtraClasses()
+    }
 
-        stubs += translator.generateBaseDeclarations()
+    fun translateExtraClasses() {
+        while (extraClassesToTranslate.isNotEmpty()) {
+            val descriptor = extraClassesToTranslate.iterator().run { next().also { remove() } }
```
SvyatoslavScherbina, line 1068: Is using iterator here an optimization?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1054,19 +1046,36 @@ abstract class ObjCExportHeaderGenerator internal constructor(
...
     protected abstract fun reportWarning(method: FunctionDescriptor, text: String)
 
     protected open fun getAdditionalImports(): List<String> = emptyList()
 
-
-    fun translateModule(): List<Stub<*>> {
+    fun translateModule() {
         // TODO: make the translation order stable
         // to stabilize name mangling.
+        translateBaseDeclarations()
```
SvyatoslavScherbina, line 1061: If your intention is to translate modules separately, then you would end up with same "base declaration" in all translation results. Can this be a problem?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -1054,19 +1046,36 @@ abstract class ObjCExportHeaderGenerator internal constructor(
...
+        translatePackageFragments()
+        translateExtraClasses()
```
SvyatoslavScherbina, line 1063: Does this actually add any stubs for `ObjcExportHeaderGeneratorMobile(restrictToLocalModules = true)`?
If no and this is intended, then it makes some of sense to use only `translatePackageFragments` in this case (considering https://github.com/JetBrains/kotlin-native/pull/4498#discussion_r518068321), don't patch `translateExtraClasses` and change the name and the meaning of  `shouldTranslateExtraClass` to something like `shouldTranslateExtraClassEarly`. The implementation and other usages also suggest the latter.

fkistner: Verified: `translateExtraClasses` is still adding stubs for `ObjcExportHeaderGeneratorMobile(restrictToLocalModules = true)`. Seems to happen very rarely, but I am seeing an inner classes being translated in this way.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
fkistner: No, we don't call this method (except when we do want to write a full header to disk). We currently call the more fine-granular methods directly.
But thinking about it, I guess we don't need this much granularity, so I propose we reduce visibility for `translatePackageFragments` and `translateExtraClasses` and introduce a method `translateModuleDeclarations`, which calls both of them (but doesn't unnecessarily couple us to these fine-granular methods).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
fkistner: Yes, just something small I noticed. The previous version required the calculation of the descriptors hashcode, whereas this doesn't.
If you prefer the previous version, please advise.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExport.kt
fkistner: 👌

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
SvyatoslavScherbina: > If you prefer the previous version, please advise.

Absolutely, I do prefer!
AFAIK, descriptor's `hashCode` is identity hash code, so it shouldn't cost much. On the other hand, this new variant is barely readable to me.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
SvyatoslavScherbina: Ok.
But in any case, we didn't supposed it to be this way.

SvyatoslavScherbina (approved):

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
fkistner: 👌

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
fkistner: Could you please clarify? Would you like me to file a bug about the inner classes not being generated by `translatePackageFragments`?

SvyatoslavScherbina: No, I don't think this is necessary for now.
Adding a comment makes more sense.

fkistner: Added doc comment.

SvyatoslavScherbina (approved):

PR #4500, opened 2020-11-05
Interop generators use symbol owner too early.

alexander-gorshenev:
Interop generators use symbol owner too early.
We should take parameter list size from the descriptor


alexander-gorshenev:
Already fixed in master

PR #4504, opened 2020-11-06
Implement CharSequence?.contentEquals(CharSequence?)

qurbonzoda:
Fixes https://youtrack.jetbrains.com/issue/KT-42840


qurbonzoda:
development has moved to kotlin repo

PR #4517, opened 2020-11-11
Filter out __cxa_demangle definition linker options in tests

PavelPunegov:


build-tools/src/main/kotlin/org/jetbrains/kotlin/KotlinNativeTest.kt
```diff
@@ -476,7 +477,10 @@ open class KonanDynamicTest : KonanStandaloneTest() {
                 needsProfileLibrary = false,
                 mimallocEnabled = false
         )
-        commands.forEach {
+        commands.map { cmd ->
+            // Filter out linker option that defines __cxa_demangle because Konan_cxa_demangle is not defined in tests.
+            Command(cmd.argsWithExecutable.filterNot { it.contains("--defsym") || it.contains("Konan_cxa_demangle") })
```
projedi, line 482: Ideally, I think, we'd need to split linker flags in `konan.properties` into toolchain flags and runtime flags. The latter being additional flags for the former. And the former can then be used for dynamic tests here and for other cpp targets (which are currently being built with gradle c & cpp plugins).

projedi (approved):

PR #4518, opened 2020-11-12
Thread registry

projedi:


runtime/src/mm/cpp/Memory.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+struct MemoryState {
+    kotlin::mm::ThreadData data;
+
+    ALWAYS_INLINE static MemoryState* from(kotlin::mm::ThreadData* data) { return wrapper_cast(MemoryState, data, data); }
+};
+
+MemoryState* InitMemory() {
+    auto* data = kotlin::mm::ThreadRegistry::instance().RegisterCurrentThread();
+    return MemoryState::from(data);
```
SvyatoslavScherbina, line 22: Somewhat fragile. It makes some sense to add static assertions checking there are no more fields in MemoryState.

runtime/src/main/cpp/SingleLockList.hpp
```diff
@@ -0,0 +1,120 @@
...
+        LockGuard<Mutex> guard(mutex_);
+        if (root_) {
+            root_->previous = node.get();
+        }
+        node->next = std::move(root_);
+        root_ = std::move(node);
+        return result;
+    }
+
+    // You can only `Erase` `Value`s that were returned by `Emplace`. Trying
```
SvyatoslavScherbina, line 70: Would this be more type safe if we returned pointer to `Node` from `Emplace` instead?

runtime/src/mm/cpp/Memory.cpp
projedi: Left a single member inside `MemoryState` and left a comment explaining that it's a wrapper and asking not to add anything else to it. Should be sufficient.

runtime/src/main/cpp/SingleLockList.hpp
projedi: Done, less magic is involved now.

SvyatoslavScherbina (approved):

ilmat192 (approved):

PR #4521, opened 2020-11-12
[interop] Optimized unmanaged memory alloc/free

homuroll:


Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmNativeMem.kt
```diff
@@ -95,17 +95,79 @@ internal object nativeMemUtils {
...
+
+    // Chunk layout: [chunk size,...padding...,diff to start,aligned data start,.....data.....]
     fun alloc(size: Long, align: Int): NativePointed {
-        val address = unsafe.allocateMemory(
-                if (size == 0L) 1L else size // It is a hack: `sun.misc.Unsafe` can't allocate zero bytes
-        )
+        val chunkHeaderSize = 8 // chunk size + alignment hop size.
+        val totalChunkSize = chunkHeaderSize + size + align
+        val ptr = chunkHeaderSize + when {
+            totalChunkSize <= 2048 -> allocFromFreeList(totalChunkSize.toInt(), 8, smallChunks)
```
LepilkinaElena, line 113: May be save save all these numbers as named constant fields in case if they are changed in future? They all are repeated both in allocation and free functions.
```
private const val chunkHeaderSize = 8 (may be even 2 * Int.SIZE_BYTES)
private const val maxSmallChunkSize = 2048
...
```

May be also replacing 4 with `Int.SIZE_BYTES` made this code a bit easier for me at least.

LepilkinaElena (approved):

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmNativeMem.kt
LepilkinaElena: Also may be better to connect padding with long array sizes `paddingSize = maxSmallChunkSize / arraySize`

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmNativeMem.kt
```diff
@@ -95,17 +95,88 @@ internal object nativeMemUtils {
...
+        require(paddedSize > 0 && paddedSize % align == 0)
+        val index = paddedSize / align - 1
+        unsafe.putLong(chunk, freeList[index])
+        freeList[index] = chunk
+    }
 
-        if (address % align != 0L) TODO(align.toString())
-        return interpretOpaquePointed(address)
+    private const val RawChunkSize: Long = 4 * 1024 * 1024
+    private val rawChunks = mutableListOf<NativePtr>()
```
SvyatoslavScherbina, line 158: IIUC, allocated raw chunks aren't freed.
If this is true, then memory in fact leaks here, which might be especially undesirable when running the compiler in the Gradle daemon process.

homuroll: Damn, you're right, native memory is not freed automatically after finishing compilation.

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt
```diff
@@ -34,6 +34,8 @@ public object nativeHeap : NativeFreeablePlacement {
     override fun alloc(size: Long, align: Int) = nativeMemUtils.alloc(size, align)
 
     override fun free(mem: NativePtr) = nativeMemUtils.free(mem)
+
+    fun freeAll() = nativeMemUtils.freeAll()
```
SvyatoslavScherbina, line 38: This undesirably affects cinterop API for Native. And despite its name, the function actually does nothing.
On the other hand, it doesn't seem necessary here, in common part.

homuroll: Ok, extracted to common utils visible for both `JvmNativeMem` and `Context`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Context.kt
```diff
@@ -347,6 +348,14 @@ internal class Context(config: KonanConfig) : KonanBackendContext(config) {
         llvmDisposed = true
     }
 
+    private var nativeMemFreed = false
+
+    fun freeNativeMem() {
+        if (nativeMemFreed) return
+        NativeMemoryAllocator.freeAll()
```
SvyatoslavScherbina, line 355: What happens if there are other thread using the compiler at the moment?

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmNativeMem.kt
```diff
@@ -95,19 +96,20 @@ internal object nativeMemUtils {
...
     }
 
-    fun free(mem: NativePtr) {
+    fun freeRaw(mem: NativePtr) {
         unsafe.freeMemory(mem)
     }
 
+    fun alloc(size: Long, align: Int) = interpretOpaquePointed(NativeMemoryAllocator.alloc(size, align))
+
+    fun free(mem: NativePtr) = NativeMemoryAllocator.free(mem)
```
SvyatoslavScherbina, line 111: So the allocator is not thread-safe anymore. Might be dangerous when using the compiler in parallel in Gradle daemon.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanDriver.kt
```diff
@@ -29,6 +29,7 @@ fun runTopLevelPhases(konanConfig: KonanConfig, environment: KotlinCoreEnvironme
         toplevelPhase.cast<CompilerPhase<Context, Unit, Unit>>().invokeToplevel(context.phaseConfig, context, Unit)
     } finally {
         context.disposeLlvm()
+        context.freeNativeMem()
```
SvyatoslavScherbina, line 32: Please consider making two `finally` blocks here, to make sure that native memory will be freed even if `disposeLlvm` is failed.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeMem.kt
```diff
@@ -118,14 +118,22 @@ internal object nativeMemUtils {
...
 
     fun alloc(size: Long, align: Int): NativePointed {
+        return interpretOpaquePointed(allocRaw(size, align))
+    }
+
+    fun free(mem: NativePtr) {
+        freeRaw(mem)
+    }
+
+    fun allocRaw(size: Long, align: Int): NativePtr {
```
SvyatoslavScherbina, line 128: Should be `internal` (probably same for `JvmNativeMem.kt`).

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeMem.kt
```diff
@@ -118,14 +118,22 @@ internal object nativeMemUtils {
...
         val ptr = malloc(size, align)
         if (ptr == nativeNullPtr) {
             throw OutOfMemoryError("unable to allocate native memory")
         }
-        return interpretOpaquePointed(ptr)
+        return ptr
     }
 
-    fun free(mem: NativePtr) {
+    fun freeRaw(mem: NativePtr) {
```
SvyatoslavScherbina, line 136: `private`?

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt
```diff
@@ -395,7 +395,7 @@ private object EmptyCString: CValues<ByteVar>() {
     override val align get() = 1
 
     private val placement =
-            interpretCPointer<ByteVar>(nativeMemUtils.alloc(1, 1).rawPtr)!!.also {
+            interpretCPointer<ByteVar>(nativeMemUtils.allocRaw(1, 1))!!.also {
```
SvyatoslavScherbina, line 398: (Not related to this PR but) this one probably leaks too. Although this happens once per classloader, so might be ok.

SvyatoslavScherbina (approved):

Interop/Runtime/src/main/kotlin/kotlinx/cinterop/Utils.kt
homuroll: Yes, it is, but it's only 8 byte (after padding), so probably doesn't matter.

Interop/Runtime/src/native/kotlin/kotlinx/cinterop/NativeMem.kt
homuroll: I'll make both `allocRaw` and `freeRaw` internal for uniformity.

SvyatoslavScherbina (approved):

PR #4524, opened 2020-11-13
Add kotlin.build.useIR property to compile project with JVM IR

udalov:
The change in Devirtualization.kt is needed to workaround JVM IR issue
reported at KT-43286.


udalov:
We're planning to enable JVM IR for building Kotlin soon, with Kotlin/Native to follow, and this change will help in trying it out beforehand. After we switch the backend to JVM IR, we can keep this option as a way to fallback, if it's needed.

PavelPunegov (approved):

PavelPunegov (approved):

PR #4525, opened 2020-11-13
Add CppSupport with backported utilities

projedi:


runtime/src/main/cpp/UtilsTest.cpp
```diff
@@ -46,21 +28,21 @@ class PinnedImpl : private kotlin::Pinned {
 } // namespace
 
 TEST(UtilsTest, MoveOnlyImpl) {
-    static_assert(is_nothrow_default_constructible_v<MoveOnlyImpl>, "Must be nothrow default constructible");
-    static_assert(is_nothrow_destructible_v<MoveOnlyImpl>, "Must be nothrow destructible");
-    static_assert(!is_copy_constructible_v<MoveOnlyImpl>, "Must not be copy constructible");
-    static_assert(!is_copy_assignable_v<MoveOnlyImpl>, "Must not be copy assignable");
-    static_assert(is_nothrow_move_constructible_v<MoveOnlyImpl>, "Must be nothrow move constructible");
-    static_assert(is_nothrow_move_assignable_v<MoveOnlyImpl>, "Must be nothrow move assignable");
+    static_assert(kotlin::is_nothrow_default_constructible_v<MoveOnlyImpl>, "Must be nothrow default constructible");
```
knebekaizer, line 31: I'd rather suggest `using namespace kotlin` in the scope of cpp file, for lower noise. When upgraded to C++17 we may be able to keep the main source intact by replacing it with `using namespace std`.
See also the comment at CppSupport.h

runtime/src/main/cpp/CppSupport.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef RUNTIME_CPP_SUPPORT_H
+#define RUNTIME_CPP_SUPPORT_H
+
+#include <type_traits>
+#include <memory>
+
+// A collection of backported utilities from future C++ versions.
+
+namespace kotlin {
```
knebekaizer, line 14: Should it be `kotlin::std` namespace?

runtime/src/main/cpp/CppSupport.hpp
```diff
@@ -0,0 +1,49 @@
...
+constexpr bool is_move_constructible_v = std::is_move_constructible<T>::value;
+template <typename T>
+constexpr bool is_move_assignable_v = std::is_move_assignable<T>::value;
+template <typename T>
+constexpr bool is_nothrow_move_constructible_v = std::is_nothrow_move_constructible<T>::value;
+template <typename T>
+constexpr bool is_nothrow_move_assignable_v = std::is_nothrow_move_assignable<T>::value;
+
+} // namespace kotlin
+
```
knebekaizer, line 48: With some `#if __cplusplus < 201703L` and `using ...` we may switch between `std::` and `kotlin::std::` in the header depending on c++ version available for the current target. The point is: let the single code base to use C++17 `std` where available, and custom backport otherwise.

runtime/src/main/cpp/UtilsTest.cpp
projedi: To be honest, I prefer explicitness (one of the reasons I rather liked going fully qualified names in cpp instead of namespace blocks). And I don't see anything bad in having to replace `kotlin::*` with `std::*` in the future, because it's clear from the code, if we're using our utilities or the standard ones.

However, `using namespace kotlin` is fine IMO, since the entire runtime is supposed to be inside this namespace.
 OTOH `using namespace kotlin::[some_module]` is less fine, because we have many modules and this'd make it harder to understand at a glance where does the symbol come from.
Finally, `using namespace std` is even less fine, at least because of existence of our utilities with the same name as the standard ones, so name collisions are unavoidable.

runtime/src/main/cpp/CppSupport.hpp
projedi: Together with https://github.com/JetBrains/kotlin-native/pull/4525#discussion_r525952842 I think not. However I'd be fine putting it into `namespace kotlin::std_support` or something like that.

runtime/src/main/cpp/CppSupport.hpp
projedi: Unfortunately that won't work: `__cplusplus` is defined by a compiler, not a standard library, and we're currently in a situation where our compiler is doing C++14, but not all stdlibs have full C++14.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/CppSupport.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <memory>
+
+// A collection of backported utilities from future C++ versions.
+
+namespace kotlin {
+
+////////////////////////// C++14 //////////////////////////
+
+template <typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args... args) {
```
knebekaizer, line 20: I guess this does not work for move-only parameter type. Should it be `make_unique(Args&&... args)` and then `forward` ?

projedi: Right you are, fixed.

knebekaizer (approved):

SvyatoslavScherbina (approved):

PR #4542, opened 2020-11-19
Tracking globals

projedi:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -420,15 +420,22 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-                                        )
                                         if (irField.storageKind == FieldStorageKind.SHARED_FROZEN)
                                             freeze(initialization, currentCodeContext.exceptionHandler)
-                                        storeAny(initialization, address, false)
+                                        initialization
+                                    } else {
+                                        null
+                                    }
+                                    if (irField.type.binaryTypeIsReference()) {
+                                        call(context.llvm.initGlobalFunction, listOf(address, initialValue ?: kNullObjHeaderPtr))
```
SvyatoslavScherbina, line 436: Might affect (startup) performance and code size for old GC. Debug binaries matter too.
I guess it should be easy to avoid generating redundant code by checking "memory model" above.

runtime/src/main/cpp/Memory.h
```diff
@@ -142,11 +142,14 @@ OBJ_GETTER(AllocInstance, const TypeInfo* type_info) RUNTIME_NOTHROW;
...
 
-OBJ_GETTER(InitInstance,
-    ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*));
+OBJ_GETTER(InitThreadLocalSingleton, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*));
 
-OBJ_GETTER(InitSharedInstance,
-    ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*));
+OBJ_GETTER(InitSingleton, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*));
+
+// `initialValue` may be `nullptr`, which signifies that the appropriate initial value was already
```
SvyatoslavScherbina, line 149: Note: for statically-initialized read-only globals it might make sense to avoid registering at all, because the value is not going to be garbage collected anyway.
This optimization is not required at the moment, but we should at least have this machinery flexible enough.

runtime/src/mm/cpp/ThreadData.hpp
```diff
@@ -17,14 +18,25 @@ namespace mm {
 // Pin it in memory to prevent accidental copying.
 class ThreadData final : private Pinned {
 public:
+    enum class State {
+        kRunning,
+        kWaitGC,
+    };
```
SvyatoslavScherbina, line 24: The resulting thread state machinery is going to be different anyway. Maybe don't expose and even implement it at the moment, and make a predicate like `isWaitingForGC` instead?

runtime/src/main/cpp/Memory.h
projedi: It was relatively easy to add, so done.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
projedi: Right, fixed.

runtime/src/mm/cpp/ThreadData.hpp
projedi: Sure, replaced with a predicate.

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -432,10 +429,15 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
...
-                                        call(context.llvm.initGlobalFunction, listOf(address, initialValue ?: kNullObjHeaderPtr))
-                                    } else if(initialValue != null) {
-                                        store(initialValue, address)
+                                    val needRegistration =
+                                            context.memoryModel == MemoryModel.EXPERIMENTAL && // only for the new MM
+                                                    irField.type.binaryTypeIsReference() && // only for references
+                                                    (initialValue != null || // which are initialized from heap object
+                                                            !irField.isFinal) // or are not final
+                                    if (needRegistration) {
+                                        call(context.llvm.initAndRegisterGlobalFunction, listOf(address))
```
ilmat192, line 438: Shouldn't the argument be `listOf(address, initialValue)`?

projedi: Right, thanks!

ilmat192 (approved):

PR #4544, opened 2020-11-20
Keep MemoryState empty.

projedi:


runtime/src/main/cpp/Utils.hpp
```diff
@@ -54,25 +54,6 @@ class Pinned {
...
-// allows to cast from `SomeValue*` to `SomeWrapper*` as no-op. It only works
-// if `SomeWrapper` is standard layout and `value` is the first non-static data member.
-// See https://en.cppreference.com/w/cpp/language/data_members#Standard_layout
-//
-// Useful for exporting SomeType under a different name (e.g. exporting inner C++ class
-// as public C struct).
-#define wrapper_cast(Wrapper, inner, field) \
-    /* With -O2 the lambda is replaced with a cast in the bitcode. */ \
-    [inner]() { \
-        static_assert(std_support::is_standard_layout_v<Wrapper>, #Wrapper " must be standard layout"); \
```
projedi, line 71: Turns out, `std::is_standard_layout` can produce different results with different stdlibs, so it cannot be trusted.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Utils.hpp
projedi: Thankfully, this was because of different representations of another standard library type, namely `std::list`. But still, this makes the former approach brittle.

ilmat192 (approved):

PR #4546, opened 2020-11-23
Fix samples

PavelPunegov:
* Use kotlin version from the kotlin-native project
* Make CopySample work for gradle.kts scripts
* Use lazy task API

### DIFF
--- a/build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
+++ b/build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
@@ -8,8 +8,9 @@ import java.io.File
 
 /**
  * A task that copies samples and replaces direct repository URLs with ones provided by the cache-redirector service.
+ * This task also adds kotlin compiler repository from the project's gradle.properties file.
  */
-open class CopySamples: Copy() {
+open class CopySamples : Copy() {
     @InputDirectory
     var samplesDir: File = project.file("samples")
 
@@ -17,6 +18,7 @@ open class CopySamples: Copy() {
         from(samplesDir) {
             it.exclude("**/*.gradle.kts")
             it.exclude("**/*.gradle")
+            it.exclude("**/gradle.properties")
         }
         from(samplesDir) {
             it.include("**/*.gradle")
@@ -32,6 +34,23 @@ open class CopySamples: Copy() {
             it.filter { line ->
                 val repo = line.trim()
                 replacements[repo]?.let { r -> line.replace(repo, r) } ?: line
+            }
+        }
+        from(samplesDir) {
+            it.include("**/gradle.properties")
+
+            val kotlinVersion = project.property("kotlinVersion") as? String
+                ?: throw IllegalArgumentException("Property kotlinVersion should be specified in the root project")
+            val kotlinCompilerRepo = project.property("kotlinCompilerRepo") as? String
+                ?: throw IllegalArgumentException("Property kotlinCompilerRepo should be specified in the root project")
+
+            it.filter { line ->
+                when {
+                    line.startsWith("kotlin_version") -> "kotlin_version=$kotlinVersion"
+                    line.startsWith("#kotlinCompilerRepo") || line.startsWith("kotlinCompilerRepo") ->
+                        "kotlinCompilerRepo=$kotlinCompilerRepo"
+                    else -> line
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


ddolovov (approved):

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
```diff
@@ -42,14 +60,22 @@ open class CopySamples: Copy() {
         return this
     }
 
-    companion object {
-        val replacements = listOf(
-            "mavenCentral()" to "maven { url 'https://cache-redirector.jetbrains.com/maven-central' }",
```
ilmat192, line 47: Minor: There is a syntax working in both `*.gradle` and `*.kts`:

```
 maven { setUrl("...") }
```

It seems possible to use it in the `replacement` list and get rid of `replacementsWithWrapper` and duplicated code in the copy task.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
```diff
@@ -3,35 +3,53 @@ package org.jetbrains.kotlin
...
             it.filter { line ->
-                replacements.forEach { (repo, replacement) ->
-                    if (line.contains(repo)) {
-                        return@filter line.replace(repo, replacement)
-                    }
+                val repo = line.trim()
+                replacements[repo]?.let { r -> line.replace(repo, r) } ?: line
+            }
+        }
+        from(samplesDir) {
```
ilmat192, line 39: This logic works on CI only. But what if I want to run samples locally against a compiler built from sources? May be additionally bump Kotlin version in the samples to e.g. 1.4.20-RC?

This particular change seems good to me though. It is useful because it allows us to check samples against a fresh version of the Gradle plugin on CI.

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
PavelPunegov: Thanks, `setUrl` works nice here. But it's not possible to get rid of the duplicated code in the copy task beacuse in `.gradle` and `.gradle.kts` maven setters are different. You still have to wrap original URL, or change them to `setUrl` in all samples, or rewrite them all to `kts`

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
PavelPunegov: > May be additionally bump Kotlin version in the samples to e.g. 1.4.20-RC?
It bumps during the release, and it still won't work in case something have changed in kotlin-native repo, while latest RC/eap/M* doesn't have appropriate changes.

> But what if I want to run samples locally against a compiler built from sources?
If you build with a composite build you'll be able to run it locally, but it won't work for gradle plugin unless you have manually published it locally, then I guess this version could be changed to something like `1.4.30-SNAPSHOT-255`

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
ilmat192: > You still have to wrap original URL, or change them to setUrl in all samples, or rewrite them all to kts

But we don't need to find and replace the whole `maven { url '<some url>' }` expression. It seems enough to replace only the URL used in such expressions just like it is done in current master. The only special cases are `mavenCentral()` and `jcenter()`. In these cases we can use the `setUrl` method to avoid creating different replacements for `*.gradle` and `*.kts`.

PavelPunegov: Yes, it is enough to replace URL only. My fault. I'll fix this

build-tools/src/main/kotlin/org/jetbrains/kotlin/CopySamples.kt
```diff
@@ -34,6 +32,22 @@ open class CopySamples: Copy() {
...
+            it.include("**/gradle.properties")
+
+            val kotlinVersion = project.property("kotlinVersion") as? String
+                ?: throw IllegalArgumentException("Property kotlinVersion should be specified in the root project")
+            val kotlinCompilerRepo = project.property("kotlinCompilerRepo") as? String
+                ?: throw IllegalArgumentException("Property kotlinCompilerRepo should be specified in the root project")
+
+            it.filter { line ->
+                when {
+                    line.contains("kotlin_version") ->  "kotlin_version=$kotlinVersion"
```
ilmat192, line 45: (minor) The "kotlin_version" substring may be a part of a comment, so may be use `startsWith` instead of `contains`? The same for "kotlinCompilerRepo" below.

ilmat192 (approved):

ilmat192 (approved):

PR #4551, opened 2020-11-24
Tweak TLS init/deinit

projedi:


runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -584,15 +583,78 @@ class ForeignRefManager {
...
+        RuntimeAssert(offset + index < size_, "Out of bound in TLS access");
+        lastKey_ = key;
+        lastOffset_ = offset;
+        return storage_ + offset + index;
+    }
+
+private:
+    using Map = KStdUnorderedMap<Key, std::pair<int, int>>;
+
+    Map* map_ = nullptr;
```
SvyatoslavScherbina, line 642: Redudant indirection here?

SvyatoslavScherbina: On the other hand, this is not in fact different from the previous version, so should be ok.

SvyatoslavScherbina (approved):

ilmat192 (approved):

runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -584,15 +583,78 @@ class ForeignRefManager {
...
+        RuntimeAssert(it != map_->end(), "Must be there");
+        int offset = it->second.first;
+        RuntimeAssert(offset + index < size_, "Out of bound in TLS access");
+        lastKey_ = key;
+        lastOffset_ = offset;
+        return storage_ + offset + index;
+    }
+
+private:
+    using Map = KStdUnorderedMap<Key, std::pair<int, int>>;
```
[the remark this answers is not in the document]
projedi: Good idea, fixed in https://github.com/JetBrains/kotlin-native/pull/4554

PR #4552, opened 2020-11-24
[cli] Supported running cinterop in daemon

homuroll:


common/src/env/cpp/Env.cpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+#include "Env.h"
+
+#include <cstdlib>
+
+#ifdef __APPLE__
+
+void setEnv(const char* name, const char* value) {
+  setenv(name, value, 1);
+}
```
SvyatoslavScherbina, line 13: Is this different from default implementation (after `#else`)?

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
```diff
@@ -60,9 +59,20 @@ private fun getVariableCType(type: KType): CType<*>? {
...
 
-private val structTypeCache = ConcurrentHashMap<Class<*>, CType<*>>()
+private class Caches {
+    val structTypeCache = mutableMapOf<Class<*>, CType<*>>()
+    val createdStaticFunctions = mutableMapOf<Class<*>, CPointer<CFunction<*>>>()
+}
+
+private val cachesHolder = ThreadLocal<Caches>()
+private val caches: Caches
+    get() = cachesHolder.get() ?: Caches().also { cachesHolder.set(it) }
```
SvyatoslavScherbina, line 69: This thread local should be initialized explicitly, not on demand.
Motivation: this would help to detect bugs like accidental (and thus e.g. leaking) initialization after deinitialization.

utilities/cli-runner/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/main.kt
```diff
@@ -40,5 +42,17 @@ private fun mainImpl(args: Array<String>, konancMain: (Array<String>) -> Unit) {
 
 fun main(args: Array<String>) = mainImpl(args, ::konancMain)
 
-fun daemonMain(args: Array<String>) = mainImpl(args, ::konancMainForGradle)
+private fun setupClangEnv() {
+    setEnv("LIBCLANG_NOTHREADS", "1")
```
SvyatoslavScherbina, line 46: Is `LIBCLANG_NOTHREADS` actually required?

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
SvyatoslavScherbina: (same for native allocator thread local)

utilities/cli-runner/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/main.kt
```diff
@@ -40,5 +42,17 @@ private fun mainImpl(args: Array<String>, konancMain: (Array<String>) -> Unit) {
...
+    setEnv("LIBCLANG_DISABLE_CRASH_RECOVERY", "1")
+}
+
+fun daemonMain(args: Array<String>) {
+    setupClangEnv()
+    try {
+        mainImpl(args, ::konancMainForGradle)
```
SvyatoslavScherbina, line 53: One major problem with running cinterop in daemon might be:
it prints messages to stdout/stderr, and these won't be displayed to user. (That's why we use different compiler entry points for CLI and daemon). Please make sure that cinterop and libclang don't do that.

common/src/env/cpp/Env.cpp
homuroll: Copy/paste artefact - I'll merge macos/linux impls.

utilities/cli-runner/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/main.kt
homuroll: Seems that it's not, will remove.

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
```diff
@@ -60,9 +59,20 @@ private fun getVariableCType(type: KType): CType<*>? {
...
 
-private fun getStructCType(structClass: KClass<*>): CType<*> = structTypeCache.computeIfAbsent(structClass.java) {
+fun clearJvmCallbackCaches() {
+    cachesHolder.remove()
```
SvyatoslavScherbina, line 72: I guess we have a memory/resource leak here: cache entries might require manual dispose.

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
SvyatoslavScherbina: This is still relevant.

homuroll: Done.

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -7,15 +7,23 @@ package org.jetbrains.kotlin.konan.util
...
+internal var allocatorHolder: NativeMemoryAllocator? = null
+
+inline fun <R> usingNativeMemoryAllocator(block: () -> R): R {
+    val allocator = NativeMemoryAllocator()
+    allocatorHolder = allocator
+    return try {
+        block()
+    } finally {
+        allocator.freeAll()
+        allocatorHolder = null
```
SvyatoslavScherbina, line 20: How would this work for parallel compilation?

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -7,15 +7,30 @@ package org.jetbrains.kotlin.konan.util
...
-    allocatorHolder.get()?.freeAll()
-    allocatorHolder.remove()
+inline fun <R> usingNativeMemoryAllocator(block: () -> R): R {
+    if (counter.getAndIncrement() == 0) {
+        allocatorHolder = NativeMemoryAllocator()
+    }
+    return try {
+        block()
+    } finally {
+        if (counter.decrementAndGet() == 0) {
```
SvyatoslavScherbina, line 24: What if other thread enters `usingNativeMemoryAllocator` and reaches `block()` while this thread is after the line above (and before the line below)?

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -6,15 +6,37 @@
...
+@PublishedApi
+internal var counter = AtomicInteger(0)
+
+inline fun <R> usingNativeMemoryAllocator(block: () -> R): R {
+    counter.getAndIncrement()
+    do {
+        val reference = allocatorHolder.reference
+        val stamp = allocatorHolder.stamp
+        val allocator = reference ?: NativeMemoryAllocator()
+    } while (!allocatorHolder.compareAndSet(reference, allocator, stamp, stamp + 1))
```
SvyatoslavScherbina, line 24: Using non-atomic counter and holder under single lock should be much simpler and more reliable.
Also similar data race probably needs to be fixed in JvmCallbacks.

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
```diff
@@ -435,10 +474,13 @@ private fun ffiCreateCif(returnType: ffi_type, paramTypes: List<ffi_type>): ffi_
         -3L -> throw Error("libffi error occurred")
     }
 
+    caches.addCif(res)
```
SvyatoslavScherbina, line 477: Memory probably leaks when `ffiCreateCif0` allocates memory but `ffi_prep_cif` fails.

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
```diff
@@ -453,6 +495,8 @@ private fun ffiCreateClosure(ffiCif: ffi_cif, impl: FfiClosureImpl): NativePtr {
         -1L -> throw Error("libffi error occurred")
     }
 
+    caches.addClosure(res)
```
SvyatoslavScherbina, line 498: `userDataGlobalRef` in `ffiCreateClosure0` seems to leak.
Closure itself seems to leak if `ffi_prep_closure_loc` fails.

SvyatoslavScherbina:
It's not clear how to handle `getCurrentEnv` in callbacks.c properly: it attaches current thread to the JVM "as daemon", and the thread is never detached, so some JVM internals might leak.

SvyatoslavScherbina:
One more thing in `callbacks.c` that might require some updates is `checkException`.
It aborts the entire process, which might be harmful in case of parallel compilation. First three usages of `checkException` are probably ok to ruin the entire process (since they might happen only in some quite exceptional situations), but the last one checks for exceptions thrown from user callbacks, which might of course happen.
It is completely unclear how to handle this in a general way though.

Interop/Runtime/src/jvm/kotlin/kotlinx/cinterop/JvmCallbacks.kt
```diff
@@ -60,9 +61,44 @@ private fun getVariableCType(type: KType): CType<*>? {
...
+        createdTypeStructs.forEach { ffiFreeTypeStruct0(it) }
+        createdCifs.forEach { ffiFreeCif0(it) }
+        createdClosures.forEach { ffiFreeClosure0(it) }
+    }
+}
+
+@PublishedApi
+internal val cachesDisposeHelper = ThreadSafeDisposableHelper({ Caches() }, { it.disposeFfi() })
+
+inline fun <R> usingJvmCallbacks(block: () -> R) = cachesDisposeHelper.usingDisposable(block)
```
SvyatoslavScherbina, line 96: Please make its name more specific, like `usingJvmCInteropCallbacks`.

Interop/StubGenerator/src/main/kotlin/org/jetbrains/kotlin/native/interop/gen/jvm/main.kt
```diff
@@ -60,7 +62,11 @@ fun main(args: Array<String>) {
     val arguments = FullCInteropArguments()
     arguments.argParser.parse(args)
     val flavorName = arguments.flavor
-    processCLib(flavorName, arguments, InternalInteropOptions(arguments.generated, arguments.natives))
+    usingNativeMemoryAllocator {
+        usingJvmCallbacks {
+            processCLib(flavorName, arguments, InternalInteropOptions(arguments.generated, arguments.natives))
+        }
+    }
```
SvyatoslavScherbina, line 69: `usingNativeMemoryAllocator { usingJvmCallbacks { processCLib(...) } }` is used twice so should be extracted to a function.

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -7,15 +7,41 @@ package org.jetbrains.kotlin.konan.util
...
 
-private val allocatorHolder = ThreadLocal<NativeMemoryAllocator>()
-val nativeMemoryAllocator: NativeMemoryAllocator
-    get() = allocatorHolder.get() ?: NativeMemoryAllocator().also { allocatorHolder.set(it) }
+class ThreadSafeDisposableHelper<T>(val create: () -> T, val dispose: (T) -> Unit) {
+    @Volatile
+    var counter = 0
+    @Volatile
+    var holder: T? = null
+    val lock = Any()
```
SvyatoslavScherbina, line 15: Please make `counter` and `lock` properties and `holder` setter internal.

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -7,15 +7,41 @@ package org.jetbrains.kotlin.konan.util
...
+                    dispose(holder!!)
+                    holder = null
+                }
+            }
+        }
+    }
 }
 
+@PublishedApi
+internal val allocatorDisposeHelper = ThreadSafeDisposableHelper({ NativeMemoryAllocator() }, { it.freeAll() })
```
SvyatoslavScherbina, line 38: Please make `freeAll` internal.

utilities/basic-utils/src/main/kotlin/NativeMemoryAllocator.kt
```diff
@@ -38,9 +64,12 @@ class NativeMemoryAllocator {
     private val smallChunks = LongArray(ChunkBucketSize)
     private val mediumChunks = LongArray(ChunkBucketSize)
     private val bigChunks = LongArray(ChunkBucketSize)
+    private val lock = Any()
```
SvyatoslavScherbina, line 67: Note: using lock when compilation is actually parallel might have much bigger overhead, e.g. due to biased locking in JVM.

utilities/cli-runner/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/main.kt
```diff
@@ -40,5 +42,11 @@ private fun mainImpl(args: Array<String>, konancMain: (Array<String>) -> Unit) {
 
 fun main(args: Array<String>) = mainImpl(args, ::konancMain)
 
-fun daemonMain(args: Array<String>) = mainImpl(args, ::konancMainForGradle)
+private fun setupClangEnv() {
+    setEnv("LIBCLANG_DISABLE_CRASH_RECOVERY", "1")
+}
 
+fun daemonMain(args: Array<String>) {
+    usingNativeMemoryAllocator { setupClangEnv() }
```
SvyatoslavScherbina, line 50: It would be great to avoid additional `usingNativeMemoryAllocator`: it allocates a huge block of memory here.

Can `setupClangEnv` be moved to `prepareTool`, just before `System.load`? This would also probably allow us to drop this environment from CLI scripts and other places eventually.
On the other hand, in this case we would disable any possibility to avoid running `setupClangEnv`, so if it has a problem and fails for someone we won't be able to provide a workaround. But I guess making it more like `if (has no env) setEnv()` should be safe enough.

Alternatively, to eliminate redundant `usingNativeMemoryAllocator` here we should consider wrapping the entire `main` and `daemonMain` to `usingNativeMemoryAllocator` (and remove it from other places).

PR #4553, opened 2020-11-25
Update kotlinx.cli

LepilkinaElena:


sbogolepov (approved):

utilities/cli-runner/src/main/kotlin/org/jetbrains/kotlin/cli/utilities/GeneratePlatformLibraries.kt
```diff
@@ -146,7 +146,7 @@ fun generatePlatformLibraries(args: Array<String>) {
 
     val logger = Logger(if (verbose) Logger.Level.VERBOSE else Logger.Level.NORMAL)
 
-    val cacheInfo = cacheDirectory?.let { CacheInfo(it, cacheKind, cacheArgs) }
+    val cacheInfo = cacheDirectory?.let { CacheInfo(it, cacheKind.visibleName, cacheArgs) }
```
ilmat192, line 149: Here we rely on the fact that `CacheKind` entries have the same names as `CompilerOutputKind` entries. `CompilerOutputKind` entries are very unlikely to be renamed, but may be implement `CacheKind` as it is done in the gradle plugin [here](https://github.com/JetBrains/kotlin-native/blob/83b8240dbb974370bf88955e6da47ccdcf7c68fe/tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanCacheTask.kt#L13)? And then write something like `cacheKind.outputKind.visibleName` here?

ilmat192 (approved):

PR #4563, opened 2020-12-01
Cleanup ClangArgs and update linux_x64 toolchain

sbogolepov:


PavelPunegov (approved):

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -37,146 +37,174 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
...
+    }.flatten()
+
+    private val osVersionMin: String
+            get() {
+                require(configurables is AppleConfigurables)
+                return configurables.osVersionMin
             }
 
-    val commonClangArgs = listOf("-B$binDir", "-fno-stack-protector") + extraHostClangArgs
+    val specificClangArgs: List<String> = when (target) {
```
SvyatoslavScherbina, line 72: Probably should be private (e.g. to ensure nobody relies on this, since the contents are changed in this PR).

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -37,146 +37,174 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
...
+
+        KonanTarget.MACOS_X64 -> listOf(
+                "--sysroot=$absoluteTargetSysRoot",
+                "-mmacosx-version-min=$osVersionMin"
+        )
+
+        KonanTarget.IOS_ARM32 -> listOf(
+                "-stdlib=libc++",
+                "-arch", "armv7",
+                "--sysroot=$absoluteTargetSysRoot",
```
SvyatoslavScherbina: Are you absolutely sure that `--sysroot` is equivalent to `-isysroot`?

konan/konan.properties
```diff
@@ -697,7 +703,7 @@ clangNooptFlags.mingw_x64 = -O1
 linkerNoDebugFlags.mingw_x64 = -Wl,-S
 linkerDynamicFlags.mingw_x64 = -shared
 linkerKonanFlags.mingw_x64 =-static-libgcc -static-libstdc++ \
-  -Xclang -flto-visibility-public-std -Wl,--dynamicbase \
+  -Wl,--dynamicbase \
```
SvyatoslavScherbina, line 706: Why do you find it obsolete?

sbogolepov: I left a link in a commit message: https://github.com/msys2/MINGW-packages/issues/5321

SvyatoslavScherbina: Ok, makes sense.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
sbogolepov: AFAIU, yes. But the devil is in the details, I'l recheck Clang source.

LepilkinaElena: If I remember right on Darwin platforms `-isysroot` is exactly the same as `--sysroot`, in other there are some changes.

LepilkinaElena (approved):

SvyatoslavScherbina (approved):

PR #4567, opened 2020-12-01
StableRef registry

projedi:


ilmat192 (approved):

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -75,3 +93,57 @@ extern "C" RUNTIME_NOTHROW void ClearTLS(MemoryState* memory) {
...
+    auto* node = static_cast<mm::StableRefRegistry::Node*>(pointer);
+    ObjHeader* object = **node;
+    RETURN_OBJ(object);
+}
+
+extern "C" RUNTIME_NOTHROW OBJ_GETTER(AdoptStablePointer, void* pointer) {
+    auto* threadData = mm::ThreadRegistry::Instance().CurrentThreadData();
+    auto* node = static_cast<mm::StableRefRegistry::Node*>(pointer);
+    ObjHeader* object = **node;
+    mm::StableRefRegistry::Instance().UnregisterStableRef(threadData, node);
```
SvyatoslavScherbina, line 118: `object` is probably not a local root here. So it would be more safe to assign it to "return slot" before unregistering the stable ref.

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -75,3 +93,57 @@ extern "C" RUNTIME_NOTHROW void ClearTLS(MemoryState* memory) {
...
+    RETURN_OBJ(object);
+}
+
+extern "C" ForeignRefContext InitLocalForeignRef(ObjHeader* object) {
```
SvyatoslavScherbina, line 122: It should probably return `nullptr` instead.
IIRC, this function is generally a hack to make a "foreign ref" that has owner but don't actually retain anything, i.e. not a strong reference.
Since we won't need to restrict ownership, returning `nullptr` should be ok.
If unsure, please keep it as stub instead for now.

SvyatoslavScherbina (approved):
Everything else LGTM.

runtime/src/mm/cpp/Memory.cpp
projedi: Good point, thanks!

runtime/src/mm/cpp/Memory.cpp
projedi: I'm indeed unsure. Will keep it as it is then.

SvyatoslavScherbina: Keeping it as is might be harmful, it might lead to memory leaks later.
Let's make it a `RuntimeCheck`-stub again.

projedi: Ok, done.

SvyatoslavScherbina (approved):

PR #4570, opened 2020-12-02
Inline properties accessors for release builds

LepilkinaElena:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
```diff
@@ -83,8 +83,7 @@ internal class InitializersLowering(val context: CommonBackendContext) : ClassLo
 
                     // We shall keep initializer for constants for compile-time instantiation.
                     declaration.initializer =
-                            if (initExpression is IrConst<*> &&
-                                    (initExpression.type.isPrimitiveType() || initExpression.type.isString())) {
+                            if (initExpression is IrConst<*>) {
```
SvyatoslavScherbina, line 86: What exactly is this required for?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InlineClassPropertyAccessorsLowering.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import org.jetbrains.kotlin.backend.common.*
+import org.jetbrains.kotlin.backend.konan.*
+import org.jetbrains.kotlin.ir.declarations.*
+import org.jetbrains.kotlin.ir.expressions.*
+import org.jetbrains.kotlin.ir.visitors.*
+
+/**
+ * Inline accessors of inline class properties.
+ * TODO: this lowering is needed because common lowering PropertyAccessorInlineLowering
```
SvyatoslavScherbina, line 16: Also PropertyAccessorInlineLowering produces access to backing field, but accessing backing fields of inline classes should be lowered by this moment.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
LepilkinaElena: https://github.com/JetBrains/kotlin/blob/b81ec2c8f05df39cfb7918d7c2032f0a1c8c6df9/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/optimizations/PropertyAccessorInlineLowering.kt#L51

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InlineClassPropertyAccessorsLowering.kt
LepilkinaElena: Fix comment?

SvyatoslavScherbina: Yes, if we don't manage to meld this lowering into InlineClassTransformer.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt
```diff
@@ -309,6 +309,24 @@ private class InlineClassTransformer(private val context: Context) : IrBuildingT
         return declaration
     }
 
+    override fun visitCall(expression: IrCall): IrExpression {
+        expression.transformChildrenVoid(this)
+        if (context.shouldOptimize()) {
+            val property = expression.symbol.owner.correspondingPropertySymbol?.owner ?: return expression
+
+            property.parent.also {
```
sbogolepov, line 317: Minor: it might be more appropriate to use `let` here. See https://kotlinlang.org/docs/reference/scope-functions.html#function-selection.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -468,6 +469,8 @@ internal fun PhaseConfig.konanPhasesConfig(config: KonanConfig) {
         disableUnless(buildDFGPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
         disableUnless(devirtualizationPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
         disableUnless(escapeAnalysisPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
+        disableUnless(propertyAccessorInlinePhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
+        disableUnless(inlineClassPropertyAccessorInlinePhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
```
sbogolepov: Will it compile? :)

LepilkinaElena: Already fixed :)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt
```diff
@@ -309,6 +309,24 @@ private class InlineClassTransformer(private val context: Context) : IrBuildingT
...
     }
 
+    override fun visitCall(expression: IrCall): IrExpression {
+        expression.transformChildrenVoid(this)
+        if (context.shouldOptimize()) {
+            val property = expression.symbol.owner.correspondingPropertySymbol?.owner ?: return expression
+
+            property.parent.let {
+                if (it is IrClass && it.isInline && property.backingField != null) {
+                    expression.dispatchReceiver?.let {
```
SvyatoslavScherbina, line 319: Doesn't this shadow `it`?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt
```diff
@@ -309,6 +309,24 @@ private class InlineClassTransformer(private val context: Context) : IrBuildingT
...
+                        it.type = expression.type
```
SvyatoslavScherbina, line 320: This might be harmful, because resulting expression would have type that doesn't match its "inferred" type.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
SvyatoslavScherbina: So this code expects that initializer of `const` property is `IrConst`. This is probably true, but the expectation still makes this code fragile. Additionally checking for `property.isConst` should make this code more reliable.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -468,6 +469,7 @@ internal fun PhaseConfig.konanPhasesConfig(config: KonanConfig) {
         disableUnless(buildDFGPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
         disableUnless(devirtualizationPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
         disableUnless(escapeAnalysisPhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
+        disableUnless(propertyAccessorInlinePhase, getBoolean(KonanConfigKeys.OPTIMIZATION))
```
SvyatoslavScherbina, line 472: Please add comment describing the motivation (separate compilation and debug information concerns). Same for inline classes.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt
LepilkinaElena: But what you suggest, make a copy?

SvyatoslavScherbina: I suggest returning it as is.

LepilkinaElena: Then we need to return to the first version and make separate lowering to run it after devirtualization. Should I revert last commits?

SvyatoslavScherbina: > Then we need to return to the first version and make separate lowering to run it after devirtualization

Why do you find this necessary? What exactly goes wrong if we don't patch the type here?

LepilkinaElena: https://github.com/JetBrains/kotlin-native/blob/b4d5f092a138b9aed2f459e44774c5c0ef18cca7/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/Devirtualization.kt#L1463

In case 
```
inline class A(val data: String)

...

if (A("a").data == "b") ..
```

we can't find devirtualized function equals, because if don't change type we are trying to find `equals(A, String)` and get wrong receiver type. DFG for this call has wrong receiver type. If stay type we can make this change only after building DFG phase.

SvyatoslavScherbina: I see.
Well, in this case using `reinterpret` intrinsic should help. See the example here: https://github.com/JetBrains/kotlin-native/blob/e4d75ccf41cba6116f85269f478cd951c9c4acde/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt#L259

Also I guess we might have more problems like this, e.g. due to https://github.com/JetBrains/kotlin-native/blob/e4d75ccf41cba6116f85269f478cd951c9c4acde/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/Autoboxing.kt#L440

LepilkinaElena: Ok, I was trying not to use extra calls.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/ToplevelPhases.kt
```diff
@@ -391,6 +391,7 @@ internal val bitcodePhase = NamedCompilerPhase(
                 ghaPhase then
                 RTTIPhase then
                 generateDebugInfoHeaderPhase then
+                propertyAccessorInlinePhase then
```
sbogolepov, line 394: It is also useful to state why we put this phase in pipeline here.

sbogolepov (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
```diff
@@ -84,7 +84,7 @@ internal class InitializersLowering(val context: CommonBackendContext) : ClassLo
                     // We shall keep initializer for constants for compile-time instantiation.
                     declaration.initializer =
                             if (initExpression is IrConst<*> &&
-                                    (initExpression.type.isPrimitiveType() || initExpression.type.isString())) {
+                                    declaration.correspondingPropertySymbol?.owner?.isConst == true) {
```
SvyatoslavScherbina, line 87: This fix [still](https://github.com/JetBrains/kotlin-native/pull/4570#discussion_r536036700) relies on the fact that if the property is `const`, then its initializer is `IrConst`.
Imagine what would happen if this wasn't true. If we had a `const` property with non-`IrConst` initializer, this code wouldn't keep the `declaration.initializer` in its place. Then https://github.com/JetBrains/kotlin/blob/b81ec2c8f05df39cfb7918d7c2032f0a1c8c6df9/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/optimizations/PropertyAccessorInlineLowering.kt#L51 would break.

But now I see that this way it is easier to duplicate the initializer. So let's keep it so, but please add the comment to the code.

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/InitializersLowering.kt
LepilkinaElena: > If we had a const property with non-IrConst initializer

Is this possible? 

> But now I see that this way it is easier to duplicate the initializer. So let's keep it so, but please add the comment to the code.

So we should try to delete `initExpression is IrConst<*>`  and take into account only the second condition?

SvyatoslavScherbina (approved):

PR #4580, opened 2020-12-07
Update toolchains for the rest of Linux targets

sbogolepov:


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -340,7 +340,11 @@ class GccBasedLinker(targetProperties: GccConfigurables)
 
         val isMips = target == KonanTarget.LINUX_MIPS32 || target == KonanTarget.LINUX_MIPSEL32
         val dynamic = kind == LinkerOutputKind.DYNAMIC_LIBRARY
-        val crtPrefix = if (configurables.target == KonanTarget.LINUX_ARM64) "usr/lib" else "usr/lib64"
+        val crtPrefix = when (val bitness = configurables.target.architecture.bitness) {
+            32 -> "usr/lib"
+            64 -> "usr/lib64"
+            else -> error("Unexpected target bitness: $bitness")
+        }
```
SvyatoslavScherbina, line 352: So this patch effectively changes crt prefix from `usr/lib` to `usr/lib64` for linux_arm64? Could you provide more details?

LepilkinaElena (approved):

SvyatoslavScherbina (approved):

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
sbogolepov: Actually, it can be reduced to unconditional `/usr/lib`, because `/usr/lib64` is an alias to `/usr/lib` in our sysroot. 
Now I'm thinking that it is better to be moved to `konan.properties`, because `crt*` may have a different location (e.g. `/usr/lib/x86_64-linux-gnu/crt1.o` on Ubuntu 20.04)

SvyatoslavScherbina (approved):

PR #4581, opened 2020-12-07
Thread states in runtime

ilmat192:


runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+std::string illegalStateSwitchMessage(mm::ThreadState oldState, mm::ThreadState newState) {
+    return std::string("Illegal thread state switch. Old state: ") + stateToString(oldState)
+        + ". New state: " + stateToString(newState);
+}
+
+} // namespace
+
+// Switches the state of the current thread to `newState` and returns the previous state.
+ALWAYS_INLINE RUNTIME_NOTHROW mm::ThreadState mm::SwitchThreadState(mm::ThreadState newState) {
+    auto threadData = ThreadRegistry::Instance().CurrentThreadData();
```
projedi, line 47: Let's pass it by argument? Here and in `ThreadStateGuard` constructor.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+std::string illegalStateSwitchMessage(mm::ThreadState oldState, mm::ThreadState newState) {
+    return std::string("Illegal thread state switch. Old state: ") + stateToString(oldState)
+        + ". New state: " + stateToString(newState);
+}
+
+} // namespace
+
+// Switches the state of the current thread to `newState` and returns the previous state.
+ALWAYS_INLINE RUNTIME_NOTHROW mm::ThreadState mm::SwitchThreadState(mm::ThreadState newState) {
```
projedi, line 46: Let's use `noexcept` here leaving `RUNTIME_NOTHROW` for C functions?

runtime/src/mm/cpp/ThreadState.hpp
```diff
@@ -0,0 +1,42 @@
...
+private:
+    ThreadState oldState_;
+};
+
+} // namespace mm
+} // namespace kotlin
+
+extern "C" {
+
+ALWAYS_INLINE RUNTIME_NOTHROW void Kotlin_mm_switchThreadStateNative();
```
projedi, line 37: I think this declaration will have to go into `main/cpp/Memory.h`, then it'll have to be used in `main/cpp/CompilerExport.cpp` (to make sure the compiler sees this declaration) and then be implemented in both `legacymm/cpp/Memory.cpp` (as noop) and in `mm/cpp/Memory.cpp`.

runtime/src/mm/cpp/ThreadState.hpp
```diff
@@ -0,0 +1,42 @@
...
+#ifndef RUNTIME_MM_THREAD_STATE_H
+#define RUNTIME_MM_THREAD_STATE_H
+
+#include <Common.h>
+#include <Utils.hpp>
+
+namespace kotlin {
+namespace mm {
+
+enum ThreadState {
```
projedi, line 15: Let's use `enum class`?

runtime/src/mm/cpp/ThreadStateTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+        auto initialState = threadData->state();
+        EXPECT_EQ(mm::ThreadState::kRunnable, initialState);
+        {
+            mm::ThreadStateGuard guard(mm::ThreadState::kNative);
+            EXPECT_EQ(mm::ThreadState::kNative, threadData->state());
+        }
+        EXPECT_EQ(initialState, threadData->state());
+    });
+    t.join();
+}
```
projedi, line 51: We can also try to check if assertions, well, assert: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#death-tests I haven't used them myself yet, though.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+    auto threadData = ThreadRegistry::Instance().CurrentThreadData();
+    auto oldState = threadData->state();
+    // TODO(perf): Mesaure the impact of this assert in debug and opt modes.
+    RuntimeAssert(isStateSwitchAllowed(oldState, newState),
+                  illegalStateSwitchMessage(oldState, newState).c_str());
```
projedi, line 51: Aside: I think we need `RuntimeAssert` with `printf`-style formatting (for consistency with `consolePrintf` at least)…

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+    threadData->setState(newState);
+    return oldState;
+}
+
+ALWAYS_INLINE void mm::AssertThreadState(mm::ThreadState expected) {
```
projedi, line 56: Let's also pass `threadData` by argument?

runtime/CMakeLists.txt
```diff
@@ -17,76 +17,9 @@ include_directories(${CMAKE_SOURCE_DIR}/../common/src/hash/headers)
...
-	src/main/cpp/dtoa/fltparse.cpp
-	src/main/cpp/math/fmod.cpp
-	src/main/cpp/math/fmodf.cpp
-	src/main/cpp/math/scalbn.cpp
-	src/main/cpp/snprintf/snprintf.cpp
-	src/profile_runtime/cpp/ProfileRuntime.cpp
-	src/relaxed/cpp/MemoryImpl.cpp
-	src/release/cpp/SourceInfo.cpp
-	src/strict/cpp/MemoryImpl.cpp
+file(GLOB_RECURSE SOURCES *.cpp *.c)
```
projedi, line 20: Cannot judge this change.

runtime/src/mm/cpp/ThreadData.hpp
```diff
@@ -29,10 +33,15 @@ class ThreadData final : private Pinned {
 
     ThreadLocalStorage& tls() noexcept { return tls_; }
 
+    ThreadState state() noexcept { return state_; }
```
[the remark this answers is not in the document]
ilmat192: Fixed

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+bool isStateSwitchAllowed(mm::ThreadState oldState, mm::ThreadState newState) {
+    switch (oldState) {
+        case mm::kRunnable:
+            return newState == mm::kNative;
+        case mm::kNative:
+            return newState == mm::kRunnable;
+    }
+}
+
+// TODO: Is there a more general way to get an enum name in runtime?
```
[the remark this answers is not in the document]
ilmat192: Ok, let's keep the `switch`.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+std::string stateToString(mm::ThreadState state) {
```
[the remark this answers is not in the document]
ilmat192: Done.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "ThreadData.hpp"
+#include "ThreadState.hpp"
+
+using namespace kotlin;
+
+namespace {
+
+// TODO: Can we make it constexpr? Is it worthwhile at all?
+bool isStateSwitchAllowed(mm::ThreadState oldState, mm::ThreadState newState) {
```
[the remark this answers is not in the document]
ilmat192: Done.

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: Done

runtime/src/mm/cpp/ThreadState.hpp
ilmat192: Done

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: Done

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: Done

runtime/CMakeLists.txt
SvyatoslavScherbina: +1

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+    switch (oldState) {
```
SvyatoslavScherbina, line 15: Seems awkward. What's wrong with `return oldState != newState`?

runtime/src/mm/cpp/ThreadState.cpp
SvyatoslavScherbina: +1

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: I suppose that we will have more than 2 states and this is just a template for a more complex check. But currently it's indeed just `oldState != newState`. Fixed.

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: Done.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#include "ThreadData.hpp"
+#include "ThreadState.hpp"
+
+using namespace kotlin;
+
+namespace {
+
+// TODO: Can we make it constexpr? Is it worthwhile at all?
```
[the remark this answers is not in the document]
ilmat192: > at the call site the arguments are not constexpr, so nothing will change even if you make it constexpr

Probably `constexpr` was not a correct term. I meant making the `newState` known statically, e.g. by moving it to a template parameter:

```cpp
ThreadState SwitchThreadState<ThreadState newState>(ThreadData* threadData) noexcept {
    ...
    isStateSwitchAllowed<newState>(oldState);
}
```

The motivation is to make this check faster. But probably we don't need it and LLVM's inlining + constant propagation will do all the work instead of us. Let me double check that.

runtime/src/mm/cpp/ThreadStateTest.cpp
```diff
@@ -13,39 +13,65 @@
...
+            mm::ThreadStateGuard guard(threadData, mm::ThreadState::kNative);
             EXPECT_EQ(mm::ThreadState::kNative, threadData->state());
         }
         EXPECT_EQ(initialState, threadData->state());
     });
     t.join();
 }
+
+TEST(ThreadStateDeathTest, StateAsserts) {
+    testing::FLAGS_gtest_death_test_style="threadsafe";
```
projedi, line 54: Why is this needed?

runtime/src/mm/cpp/ThreadState.cpp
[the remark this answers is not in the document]
projedi: Right, also since `oldState` is definitely known only at runtime, and you need (1) load, (2) atomic read, I doubt that getting `newState` from a register would have a noticeable impact.

SvyatoslavScherbina (approved):

runtime/src/mm/cpp/ThreadStateTest.cpp
ilmat192: To execute a death test, GTest forks the test process. Since forking in a multithreaded environment may cause problems, GTest [supports](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-it-works) two death test styles: `fast` and `threadsafe`:
 - In the fast mode, the tested statement is executed right after the fork. If there are more than one thread at the moment of the fork (like in our case), GTest shows a corresponding warning.
 - In the thread safe mode, GTest restarts the test binary with additional flags to run only this particular death test. 

These particular tests seem to be safe for the `fast` mode but I still enabled the `threadsafe` mode to avoid GTest's warnings. The thread safe mode is 3 times slower for these tests but it's still a matter of tens of milliseconds on my machine (13 ms in the fast mode against 40 ms in the threadsafe mode).

There are also alternatives:

1. Move registering the thread to the `EXPECT_DEATH` and get rid of threads at all:

```C++
EXPECT_DEATH({
        // EXPECT_DEATH forks the test process so it's ok to register the main thread.
        auto* threadData = mm::ThreadRegistry::Instance().RegisterCurrentThread()->Get();
        mm::AssertThreadState(threadData, mm::ThreadState::kNative);
    }, "runtime assert: Unexpected thread state. Expected: NATIVE. Actual: RUNNABLE");
```

2. Create a `ThreadData` manually instead of calling `RegisterCurrentThread`. Not sure if it's a good idea though.

What is your opinion? Which alternative is better?

runtime/src/main/cpp/KAssert.cpp
```diff
@@ -1,29 +1,31 @@
...
-  // TODO: produce stacktrace and such.
-  char buf[1024];
-  if (location != nullptr)
-      konan::snprintf(buf, sizeof(buf), "%s: runtime assert: %s\n", location, message);
-  else
-      konan::snprintf(buf, sizeof(buf), "runtime assert: %s\n", message);
-  konan::consoleErrorUtf8(buf, konan::strnlen(buf, sizeof(buf)));
-  konan::abort();
+RUNTIME_NORETURN void RuntimeAssertFailed(const char* location, const char* format, ...) {
+    char buf[1024];
```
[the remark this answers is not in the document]
ilmat192: Doesn't it mean that this thread local buffer will always be allocated at the app startup (at least for a debug app)?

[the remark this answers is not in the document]
projedi: True, it'll sacrifice 1KiB for each thread. Also, I wasn't very clear about "no-memory". What I meant is "when we're nearing stack overflow".

runtime/src/mm/cpp/ThreadStateTest.cpp
projedi: I see. Let's make `threadsafe` a default by putting it into `TestLauncher.cpp`?

projedi (approved):

PR #4582, opened 2020-12-07
 Enable and fix macOS Hypervisor platform lib for Xcode 12.2

SvyatoslavScherbina:


platformLibs/src/platform/osx/Hypervisor.def
```diff
@@ -0,0 +1,10 @@
+depends = darwin posix
+language = Objective-C
+package = platform.Hypervisor
+
+# Using headers instead of modules to workaround incorrect module definition in Xcode 12.2:
```
sbogolepov, line 5: Is there any kind of public issue on Apple bug tracker for it? If so, it's better to be referenced here.

sbogolepov (approved):

platformLibs/src/platform/osx/Hypervisor.def
SvyatoslavScherbina: To the best of my knowledge, there is no public bug tracker for this.

PR #4583, opened 2020-12-07
Sketch mm::ObjectFactory

projedi:


runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,280 @@
...
+public:
+    // This class does not know its size at compile-time.
+    class Node : private Pinned {
+    public:
+        ~Node() = default;
+
+        static void operator delete(void* ptr) noexcept { konanFreeMemory(ptr); }
+
+        // Note: This can only be trivially destructible data, as nobody can invoke its destructor.
+        void* Data() noexcept { return this + 1; }
```
SvyatoslavScherbina, line 36: This function doesn't align the result. What is `dataAlignment` then?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+    private:
+        friend class ObjectFactoryStorage;
+
+        Node() noexcept = default;
+
+        static void* operator new(size_t size, size_t dataSize, size_t dataAlignment) noexcept {
+            size_t allocSize = AlignUp(size + dataSize, dataAlignment);
```
SvyatoslavScherbina, line 44: What exactly is this required for?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,280 @@
...
+// uses `konanAllocMemory` and `konanFreeMemory`
+// TODO: Consider merging with `MultiSourceQueue` somehow.
+class ObjectFactoryStorage : private Pinned {
+public:
+    // This class does not know its size at compile-time.
+    class Node : private Pinned {
+    public:
+        ~Node() = default;
+
+        static void operator delete(void* ptr) noexcept { konanFreeMemory(ptr); }
```
SvyatoslavScherbina, line 33: Can it be private too?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,280 @@
...
+        if (!previousNode->next_) {
+            last_ = previousNode;
+        }
+
+        AssertCorrectUnsafe();
+        return previousNode->next_.get();
+    }
+
+    // Expects `mutex_` to be held by the current thread.
+    void AssertCorrectUnsafe() const noexcept {
```
SvyatoslavScherbina, line 196: It makes some sense to mark it as `ALWAYS_INLINE`.
Motivation: when compiling release binaries, it is possible that LLVM optimizes this function only after (trying) inlining it. So if it makes a decision about inlining before optimizing this to no-op, it might decide to avoid inlining.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: It's invoked by `std::default_delete` (in `std::unique_ptr`), so it either have to be public or be friends with it (which essentially makes it public). The alternative - create a custom deleter, which, I think, is an overkill.

SvyatoslavScherbina: Ok, thank you for the explanation.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Right, reworked aligning. Now it should be more clear and correct.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Good point, done.

runtime/src/main/cpp/Alignment.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <cstdint>
+
+namespace kotlin {
+
+constexpr inline size_t AlignUp(size_t size, size_t alignment) {
+    return (size + alignment - 1) & ~(alignment - 1);
+}
+
+inline void* AlignUp(void* ptr, size_t alignment) {
+    return reinterpret_cast<void*>(reinterpret_cast<size_t>(ptr) + (alignment - 1) & -alignment);
```
SvyatoslavScherbina, line 19: Maybe some parens?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,295 @@
...
+            size_t alignment = std::max(alignof(Node), dataAlignment);
+            size_t size = sizeof(Node) + dataSize;
+            size_t allocSize = AlignUp(size, alignment);
+            void* ptr = konanAllocMemory(allocSize);
+            if (!ptr) {
+                // TODO: Try doing GC first.
+                konan::consoleErrorf("Out of memory trying to allocate %zu. Aborting.\n", allocSize);
+                konan::abort();
+            }
+            void* alignedPtr = AlignUp(ptr, alignment);
```
SvyatoslavScherbina, line 57: What is this variable required for?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,295 @@
...
+            RuntimeAssert(
+                    reinterpret_cast<uintptr_t>(alignedPtr) + size <= reinterpret_cast<uintptr_t>(ptr) + allocSize,
+                    "Aligning %p (with size %zu) to %p overflowed allocated size %zu", ptr, size, alignedPtr, allocSize);
+            size_t dataOffset = AlignUp(sizeof(Node), dataAlignment);
+            void* data = static_cast<uint8_t*>(ptr) + dataOffset;
+            RuntimeAssert(IsAligned(data, dataAlignment), "data=%p is not aligned to %zu", data, dataAlignment);
+            auto* nodePtr = new (ptr) Node(data);
+            return std::unique_ptr<Node>(nodePtr);
```
SvyatoslavScherbina, line 65: The entire implementation above looks kinda cryptic, but doesn't look correct to me.
For example:
1. It seems to rely on the allocator providing pointers aligned by `alignment`, which might be not true.
2. Consider the case dataSize = 112, dataAlignment = 128. IIUC, the implementation would allocate 128 bytes with `data` pointing to the end of the region, and no assertion would fire (if the allocator returns memory aligned by 128).

It makes some sense to make the implementation simpler by specializing it to dataAlignment = 8 for now. For example, in this case node size and alignment would be aligned by dataAlignment.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: I just realized I made a mistake (in `delete`). This code is used to convert a possibly unaligned `ptr` from allocator into a definitely aligned pointer `alignedPtr`. The mistake is that I free `alignedPtr` - not `ptr`.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Oh, that's another mistake. Of course it should be `alignedPtr` instead of `ptr` in the call to `new`.

projedi: > Consider the case dataSize = 112, dataAlignment = 128

This'll fail at the second assertion: `dataSize` must itself be aligned by `dataAlignment`.

projedi: Reworked again and made `dataAlignment` a compile-time parameter.

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -74,6 +74,20 @@ extern "C" void RestoreMemory(MemoryState*) {
...
 }
 
+extern "C" RUNTIME_NOTHROW OBJ_GETTER(AllocInstance, const TypeInfo* typeInfo) {
+    auto* threadData = mm::ThreadRegistry::Instance().CurrentThreadData();
+    auto* object = threadData->objectFactoryThreadQueue().CreateObject(typeInfo);
+    RETURN_OBJ(object);
+}
+
+extern "C" OBJ_GETTER(AllocArrayInstance, const TypeInfo* typeInfo, int32_t elements) {
+    RuntimeAssert(elements >= 0, "elements cannot be negative");
```
SvyatoslavScherbina, line 84: It should throw an exception if `elements < 0`.

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,312 @@
...
+            size_t dataSizeAligned = AlignUp(dataSize, DataAlignment);
+            size_t totalAlignment = std::max(alignof(Node), DataAlignment);
+            size_t totalSize = AlignUp(sizeof(Node) + dataSizeAligned, totalAlignment);
+            void* ptr = konanAllocAlignedMemory(totalSize, totalAlignment);
+            if (!ptr) {
+                // TODO: Try doing GC first.
+                konan::consoleErrorf("Out of memory trying to allocate %zu. Aborting.\n", totalSize);
+                konan::abort();
+            }
+            RuntimeAssert(IsAligned(ptr, totalAlignment), "Allocator returned unaligned to %zu pointer %p", totalAlignment, ptr);
```
SvyatoslavScherbina, line 70: One more thing to check: `ptr + dataOffset + dataSize <= totalSize`.

runtime/src/mm/cpp/ObjectFactoryTest.cpp
```diff
@@ -0,0 +1,554 @@
...
+            } else {
+                ++it;
+            }
+        }
+    }
+
+    auto actual = Collect<int>(storage);
+
+    EXPECT_THAT(actual, testing::ElementsAre(1, 2));
+}
```
SvyatoslavScherbina, line 255: Maybe also a test that inserts only one element and deletes it?

SvyatoslavScherbina (approved):

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Checked `dataOffset + dataSize <= totalSize`

SvyatoslavScherbina (approved):

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+    private:
+        friend class ObjectFactoryStorage;
+
+        Node() noexcept = default;
+
+        static void* operator new(size_t size, size_t dataSize) noexcept {
+            size_t dataSizeAligned = AlignUp(dataSize, DataAlignment);
+            size_t totalAlignment = std::max(alignof(Node), DataAlignment);
+            size_t totalSize = AlignUp(sizeof(Node) + dataSizeAligned, totalAlignment);
```
ilmat192, line 65: Why do we need to `AlignUp` when computing the total size? Isn't it enough to call

```
konanAllocAlignedMemory(sizeof(Node) + dataSizeAligned, totalAlignment);
```
instead?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+private:
+    // Expects `mutex_` to be held by the current thread.
+    Node* EraseUnsafe(Node* previousNode) noexcept {
+        RuntimeAssert(root_ != nullptr, "Must not be empty");
+        AssertCorrectUnsafe();
+
+        if (previousNode == nullptr) {
+            // Deleting the root.
+            auto newRoot = std::move(root_->next_);
```
ilmat192, line 211: Why is it impossible to move it directy: `root_ = std::move(root_->next_)`?

runtime/src/mm/cpp/ObjectFactory.cpp
```diff
@@ -0,0 +1,61 @@
...
+    auto* object = static_cast<ObjHeader*>(node.Data());
+    object->typeInfoOrMeta_ = const_cast<TypeInfo*>(typeInfo);
+    return object;
+}
+
+ArrayHeader* mm::ObjectFactory::ThreadQueue::CreateArray(const TypeInfo* typeInfo, uint32_t count) noexcept {
+    RuntimeAssert(typeInfo->IsArray(), "Must be an array");
+    uint32_t arraySize = static_cast<uint32_t>(-typeInfo->instanceSize_) * count;
+    // Note: array body is aligned, but for size computation it is enough to align the sum.
+    size_t allocSize = AlignUp(sizeof(ArrayHeader) + arraySize, kObjectAlignment);
```
ilmat192, line 28: Is this correct that `instanceSize` already includes a padding needed to properly align each element of the array (like `sizeof` does for C++ classes) so we don't need to think about this padding when computing the `allocSize`?

projedi: It should, yes. At least this matches what's happening in `legacymm`.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Mimicking c++ invariant that `sizeof(T)` is always aligned to `alignof(T)`.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Fixed, thanks!

ilmat192 (approved):

PR #4587, opened 2020-12-09
Restore runtimeState value in deinitRuntime

projedi:
#KT-42172 Fixed

### DIFF
--- a/runtime/src/main/cpp/Runtime.cpp
+++ b/runtime/src/main/cpp/Runtime.cpp
@@ -140,7 +140,9 @@ RuntimeState* initRuntime() {
 void deinitRuntime(RuntimeState* state, bool destroyRuntime) {
   RuntimeAssert(state->status == RuntimeStatus::kRunning, "Runtime must be in the running state");
   state->status = RuntimeStatus::kDestroying;
-  // This may be called after TLS is zeroed out, so ::memoryState in Memory cannot be trusted.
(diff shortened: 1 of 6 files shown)


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -140,7 +140,9 @@ RuntimeState* initRuntime() {
 void deinitRuntime(RuntimeState* state, bool destroyRuntime) {
   RuntimeAssert(state->status == RuntimeStatus::kRunning, "Runtime must be in the running state");
   state->status = RuntimeStatus::kDestroying;
-  // This may be called after TLS is zeroed out, so ::memoryState in Memory cannot be trusted.
+  // This may be called after TLS is zeroed out, so ::runtimeState and ::memoryState in Memory cannot be trusted.
+  // TODO: This may in fact reallocate TLS without guarantees that it'll be deallocated again.
+  ::runtimeState = state;
```
SvyatoslavScherbina, line 145: Zeroing it after `deinitRuntime` in `Kotlin_deinitRuntimeCallback` might also make sense.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Runtime.cpp
projedi: Good idea, thanks!

SvyatoslavScherbina (approved):

PR #4588, opened 2020-12-09
[objc][codegen] Fixed bug with calling fake override interface methods

homuroll:


runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -943,8 +945,21 @@ static void throwIfCantBeOverridden(Class clazz, const KotlinToObjCMethodAdapter
       insertOrReplace(methodTable, adapter->nameSignature, const_cast<void*>(adapter->kotlinImpl));
       RuntimeAssert(adapter->vtableIndex == -1, "");
 
-      if (adapter->itableIndex != -1 && superITable != nullptr)
+      if (adapter->itableIndex != -1 && superITable != nullptr) {
+        // Because of fake overrides [adapter->interfaceId] might not be equal to [typeInfo->classId_].
```
SvyatoslavScherbina, line 949: This is not entirely correct. Yes, the particular bug we've encountered features a fake override. But
* It is probably triggered by a bug in ObjCExportCodeGenerator: the latter doesn't "resolve fake overrides" when computing `functionName`, so the machinery doesn't find the adapter among the inherited ones and generates an adapter for super interface's method.
* Generally that machinery doesn't guarantee that it would only emit adapters matching the type. Avoiding emitting adapters that are inherited from super types is only an optimization. We are free to remove or rework it. And even if we don't, there might be cases when it doesn't guarantee that adapters match type.

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -943,8 +945,21 @@ static void throwIfCantBeOverridden(Class clazz, const KotlinToObjCMethodAdapter
...
+        auto interfaceId = adapter->interfaceId;
+        int interfaceVTableSize = adapter->itableSize;
+        RuntimeAssert(interfaceVTableSize >= 0, "");
+        auto interfaceVTablesIt = interfaceVTables.find(interfaceId);
+        if (interfaceVTablesIt == interfaceVTables.end()) {
+          interfaceVTables.emplace(interfaceId, KStdVector<VTableElement>(interfaceVTableSize));
+        } else {
+          auto const& interfaceVTable = interfaceVTablesIt->second;
+          RuntimeAssert(interfaceVTable.size() == static_cast<size_t>(interfaceVTableSize), "");
+        }
```
SvyatoslavScherbina, line 959: Looks pretty much like a copy-paste from the code above. It should be possible to extract a (lambda) function for this code.

runtime/src/main/cpp/ObjCExport.mm
```diff
@@ -943,8 +948,12 @@ static void throwIfCantBeOverridden(Class clazz, const KotlinToObjCMethodAdapter
       insertOrReplace(methodTable, adapter->nameSignature, const_cast<void*>(adapter->kotlinImpl));
       RuntimeAssert(adapter->vtableIndex == -1, "");
 
-      if (adapter->itableIndex != -1 && superITable != nullptr)
+      if (adapter->itableIndex != -1 && superITable != nullptr) {
+        // In general, [adapter->interfaceId] might not be equal to [typeInfo->classId_].
+        auto interfaceId = adapter->interfaceId;
```
SvyatoslavScherbina, line 953: Can also be used in `addToITable` below.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/ObjCExport.mm
SvyatoslavScherbina: > a bug in ObjCExportCodeGenerator: the latter doesn't "resolve fake overrides" when computing functionName, so the machinery doesn't find the adapter among the inherited ones and generates an adapter for super interface's method.

JFTR: I've rechecked the ObjCExportCodeGenerator and am not sure anymore that it has such a bug. 

Apparently our virtual call machinery calls interface fake override methods using `.functionName` of a fake override, not of its "real" super method. This means that ObjCExportCodeGenerator generally has to generate additional "reverse adapter" for the fake override, using its own `functionName`. While `itablePlace` is reused from super interface.
It can use `itablePlace = null` though, but this would be just an optimization (see also above).

PR #4593, opened 2020-12-10
Cleanup Stubs.cpp

projedi:


runtime/src/mm/cpp/Memory.cpp
```diff
@@ -63,6 +69,10 @@ extern "C" void DeinitMemory(MemoryState* state, bool destroyRuntime) {
     mm::ThreadRegistry::Instance().Unregister(FromMemoryState(state));
 }
 
+extern "C" void RestoreMemory(MemoryState*) {
```
SvyatoslavScherbina, line 73: How does new MM solve the problem which is workarounded by `RestoreMemory`?
IIUC, the actual solution would require a lot of additional efforts, so it might be safer to implement thus function properly for now.

projedi: By not having to access the thread local during deinitialization. Which by itself is made possible by deinitialization not doing much, and, crucially, not executing arbitrary (indeed, any) Kotlin code.

runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -72,6 +72,10 @@
 
 namespace {
 
+bool IsStrictMemoryModel() noexcept {
```
SvyatoslavScherbina, line 75: Maybe add `ALWAYS_INLINE`?

runtime/src/relaxed/cpp/MemoryImpl.cpp
```diff
@@ -9,7 +9,7 @@
 
 extern "C" {
 
-const bool IsStrictMemoryModel = false;
+const MemoryModel CurrentMemoryModel = MemoryModel::kStrict;
```
SvyatoslavScherbina, line 12: Is it?

runtime/src/mm/cpp/Memory.cpp
SvyatoslavScherbina: > not executing arbitrary (indeed, any) Kotlin code.

Exactly! We don't have code for this yet, so it is not safe to add code assuming we do.
Would implementing `RestoreMemory` be hard?

projedi: > Would implementing RestoreMemory be hard?

Okay, how about this: I'll implement `RestoreMemory`, but I'll add a flag into `ThreadState` to signal when it's in the termination phase. And then I'll add a `RuntimeAssert` to `ThreadRegistry::CurrentThreadData` that checks that the returned thread is not in the termination phase. If we ever get Kotlin code in deinitialization, we will be able to quickly allow it by disabling the assert, until we figure out if we want to keep it this way or move Kotlin code execution to somewhere else.

runtime/src/main/cpp/KAssertTest.cpp
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+#include "KAssert.h"
+
+#include "gtest/gtest.h"
+
+TEST(TODODeathTest, EmptyTODO) {
+    EXPECT_DEATH({
+        TODO();
+    }, "Unimplemented");
```
SvyatoslavScherbina, line 13: Is it possible to test that it prints a location too?

runtime/src/mm/cpp/Memory.cpp
SvyatoslavScherbina: >  If we ever get Kotlin code in deinitialization, we will be able to quickly allow it by disabling the assert

Not if this happens in user's code.

SvyatoslavScherbina: On the other hand, it is hard to imagine cases when this could happen with new MM.
So let's keep it implemented this way, but add a note to our plan.

runtime/src/relaxed/cpp/MemoryImpl.cpp
projedi: Oh. Thanks!

SvyatoslavScherbina (approved):

ilmat192 (approved):

PR #4596, opened 2020-12-11
Enable caches for linux_x64

sbogolepov:


homuroll (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt
```diff
@@ -112,14 +112,19 @@ internal val IrClass.kotlinObjCClassInfoSymbolName: String
...
+fun IrFunction.computeSymbolName() = with(KonanBinaryInterface) { symbolName }.replaceSpecialSymbols()
 
-val IrClass.typeInfoSymbolName get() = with(KonanBinaryInterface) { typeInfoSymbolName }
+fun IrField.computeSymbolName() = with(KonanBinaryInterface) { symbolName }.replaceSpecialSymbols()
+
+fun IrClass.computeTypeInfoSymbolName() = with(KonanBinaryInterface) { typeInfoSymbolName }.replaceSpecialSymbols()
+
+private fun String.replaceSpecialSymbols() =
+        // '@' is used for symbol versioning in GCC: https://gcc.gnu.org/wiki/SymbolVersioning.
+        this.replace("@", "__")
```
SvyatoslavScherbina, line 127: So here we generally hope that this won't make symbol names clash?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/optimizations/DataFlowIR.kt
```diff
@@ -639,7 +638,7 @@ internal object DataFlowIR {
 
             assert(irField.parent !is IrClass) { "All local properties initializers should've been lowered" }
             val attributes = FunctionAttributes.IS_GLOBAL_INITIALIZER or FunctionAttributes.RETURNS_UNIT
-            val symbol = FunctionSymbol.Private(privateFunIndex++, module, -1, attributes, null, null, takeName { "${irField.symbolName}_init" })
+            val symbol = FunctionSymbol.Private(privateFunIndex++, module, -1, attributes, null, null, takeName { "${irField.computeSymbolName()}_init" })
```
SvyatoslavScherbina, line 641: (This change probably belongs to other commit).

konan/konan.properties
```diff
@@ -400,14 +401,16 @@ targetSysRoot.linux_x64 = $gccToolchain.linux_x64/x86_64-unknown-linux-gnu/sysro
...
 targetCpu.linux_x64 = x86-64
-clangFlags.linux_x64 = -cc1 -target-cpu $targetCpu.linux_x64 -emit-obj -disable-llvm-optzns -x ir
+clangFlags.linux_x64 = -cc1 -target-cpu $targetCpu.linux_x64 -emit-obj -disable-llvm-optzns -x ir \
+   -ffunction-sections -fdata-sections
 clangNooptFlags.linux_x64 = -O1
-clangOptFlags.linux_x64 = -O3 -ffunction-sections
+clangOptFlags.linux_x64 = -O3
 clangDebugFlags.linux_x64 = -O0
 clangDynamicFlags.linux_x64 = -mrelocation-model pic
+clangStaticFlags.linux_x64 = -mrelocation-model pic
```
SvyatoslavScherbina, line 410: Might be related: `OptimizationPipeline.kt` also has relocation mode. It is set to "default", and AFAIK it is not PIC on Linux.

konan/konan.properties
```diff
@@ -68,7 +68,8 @@ cacheableTargets.macos_x64 = \
   ios_x64 \
   ios_arm64
 
-cacheableTargets.linux_x64 =
+cacheableTargets.linux_x64 = \
+  linux_x64
```
SvyatoslavScherbina, line 72: Btw, did hidden visibility (`makeVisibilityHiddenLikeLlvmInternalizePass`) for static caches work as expected? I mean, does it help to make cached symbols non-exported in final executable or .so?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt
sbogolepov: Yes. The tricky part here are functions that contain `@` in their name. Maybe make it more robust and readable, e.g. `__at__`?

konan/konan.properties
sbogolepov: Yes, good catch. Looks like I'll have to add a separate properties for relocation mode.

konan/konan.properties
sbogolepov: Yes. More than that, it helps linker to cope with absence of some runtime declarations like `Kotlin_ObjCInterop_getTypeInfoForClass`.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/BinaryInterface.kt
SvyatoslavScherbina: Well, maybe. On the other hand, functions that contain `@` in their name probably could clash with extension functions before this patch. So let's avoid complicated mangling for now. `__at__` or similar things should be just ok.

konan/konan.properties
sbogolepov: Done.

SvyatoslavScherbina (approved):
Everything [else](https://github.com/JetBrains/kotlin-native/pull/4596#discussion_r544097648) is ok.

PR #4599, opened 2020-12-14
Post-toolchain update fixes

sbogolepov:


backend.native/tests/build.gradle
```diff
@@ -3872,7 +3872,9 @@ standaloneTest("interop_objc_allocException") {
 interopTest("interop0") {
     disabled = (project.testTarget == 'wasm32') ||         // No interop for wasm yet.
                (project.testTarget == 'linux_mips32') ||   //  st_uid of '/' is not equal to 0 when using qemu
-               (project.testTarget == 'linux_mipsel32')
+               (project.testTarget == 'linux_mipsel32') ||
+               (project.testTarget == 'linux_arm32_hfp') ||
+               (project.testTarget == 'linux_arm64')
```
PavelPunegov, line 3877: Why should this test be excluded on arm32/64?

sbogolepov: The same reason as for MIPS. Yes, it disables this test for real Raspberry Pi, but I think it's a fine trade-off, since number of interop is quite big already.

PavelPunegov (approved):

PR #4602, opened 2020-12-15
Add ExtraObjectData

projedi:


runtime/src/mm/cpp/ExtraObjectData.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+using namespace kotlin;
+
+// static
+mm::ExtraObjectData& mm::ExtraObjectData::Install(ObjHeader* object) noexcept {
+    if (object->has_meta_object()) {
+        return mm::ExtraObjectData::FromMetaObjHeader(object->meta_object());
+    }
+
+    TypeInfo* typeInfo = object->typeInfoOrMeta_;
```
SvyatoslavScherbina, line 23: What if this pointer is already an ExtraObjectData?

runtime/src/mm/cpp/ExtraObjectData.cpp
```diff
@@ -0,0 +1,58 @@
...
+    RuntimeCheck(!hasPointerBits(typeInfo, OBJECT_TAG_MASK), "Object must not be tagged");
+
+    auto* data = new ExtraObjectData(typeInfo);
```
SvyatoslavScherbina, line 26: This doesn't use `konanAllocMemory`, does it?

runtime/src/mm/cpp/ExtraObjectDataTest.cpp
```diff
@@ -0,0 +1,74 @@
...
+    TypeInfo typeInfo;
+    typeInfo.typeInfo_ = &typeInfo;
+    ObjHeader object;
+    object.typeInfoOrMeta_ = &typeInfo;
+
+    ASSERT_FALSE(object.has_meta_object());
+
+    auto& extraData = mm::ExtraObjectData::Install(&object);
+
+    EXPECT_TRUE(object.has_meta_object());
```
SvyatoslavScherbina, line 28: Just curious: why `ASSERT_FALSE` but `EXPECT_TRUE`?

projedi: In `googletest` there are 2 kinds of test assertions: `EXPECT_*` and `ASSERT_*`. The former record the error and move on, the latter record the error and stop the test (technically, just execute the `return` statement). So, I typically use `ASSERT_*` as a precondition: when there's no point to run the test if it does not hold.

runtime/src/mm/cpp/ExtraObjectData.cpp
projedi: Nope, I treat it as "system data", so use the default allocator for it.

runtime/src/mm/cpp/ExtraObjectData.cpp
projedi: Right, a race, fixed.

SvyatoslavScherbina: Please try to add a test that would detect this race.

projedi: `ExtraObjectDataTest.ConcurrentInstall` should trigger it. However, according to the assembly (on x64), `TypeInfo* typeInfo = object->typeInfoOrMeta_;` didn't generate another load from `object`, it reused the load from `object->has_meta_object()`, so in practice this race cannot occur (on x64).

runtime/src/mm/cpp/ExtraObjectData.cpp
projedi: I'll separately convert all the new code (and some old, using `std::make_unique` and `std::unique_ptr`) to the `konan*` stuff.

runtime/src/mm/cpp/ExtraObjectData.cpp
SvyatoslavScherbina: > it reused the load from `object->has_meta_object()`

Did the latter get inlined?

projedi: Yes.

runtime/src/main/cpp/Memory.h
```diff
@@ -38,19 +38,29 @@ struct MetaObjHeader;
...
+  }
+
   const TypeInfo* type_info() const {
     return clearPointerBits(typeInfoOrMeta_, OBJECT_TAG_MASK)->typeInfo_;
   }
 
   bool has_meta_object() const {
-    auto* typeInfoOrMeta = clearPointerBits(typeInfoOrMeta_, OBJECT_TAG_MASK);
-    return (typeInfoOrMeta != typeInfoOrMeta->typeInfo_);
+      return AsMetaObject(typeInfoOrMeta_) != nullptr;
```
SvyatoslavScherbina, line 56: This might be suboptimal: does the compiler optimize this code by eliminating the redundant conditional jump here?
I mean, if `typeInfoOrMeta != typeInfoOrMeta->typeInfo_` does the compiler assume that `typeInfoOrMeta != nullptr` to avoid checking this explicitly here?

SvyatoslavScherbina (approved):

runtime/src/main/cpp/Memory.h
projedi: Looking at the code:
```cpp
    if (!has_meta_object()) {
        return nullptr;
    }
    return (void*)1;
```
With the old `has_meta_object`:
```llvm
  %2 = bitcast %struct.ObjHeader* %0 to i64*
  %3 = load i64, i64* %2, align 8, !tbaa !4
  %4 = and i64 %3, -4
  %5 = inttoptr i64 %4 to %struct.TypeInfo*
  %6 = getelementptr inbounds %struct.TypeInfo, %struct.TypeInfo* %5, i64 0, i32 0
  %7 = load %struct.TypeInfo*, %struct.TypeInfo** %6, align 8, !tbaa !9
  %8 = icmp eq %struct.TypeInfo* %7, %5
  %9 = select i1 %8, i8* null, i8* inttoptr (i64 1 to i8*)
  ret i8* %9
```
With the new:
```llvm
  %2 = bitcast %struct.ObjHeader* %0 to i64*
  %3 = load i64, i64* %2, align 8, !tbaa !4
  %4 = and i64 %3, -4
  %5 = inttoptr i64 %4 to %struct.TypeInfo*
  %6 = getelementptr inbounds %struct.TypeInfo, %struct.TypeInfo* %5, i64 0, i32 0
  %7 = load %struct.TypeInfo*, %struct.TypeInfo** %6, align 8, !tbaa !9
  %8 = icmp ne %struct.TypeInfo* %7, %5
  %9 = icmp ne i64 %4, 0
  %10 = and i1 %9, %8
  %11 = select i1 %10, i8* inttoptr (i64 1 to i8*), i8* null
  ret i8* %11
}
```
So, it doesn't generate an extra jump, but does generate an extra `and`. Same in the x64 assembly.

SvyatoslavScherbina: Also an extra comparison. But this might be ok.

runtime/src/mm/cpp/ExtraObjectData.cpp
projedi: https://github.com/JetBrains/kotlin-native/pull/4611

ilmat192 (approved):

PR #4606, opened 2020-12-17
Stacktraces in runtime asserts

ilmat192:


runtime/src/main/cpp/KAssert.h
```diff
@@ -25,17 +25,17 @@
...
 
 inline RUNTIME_NORETURN void TODOImpl(const char* location) {
     RuntimeAssertFailed(location, "Unimplemented");
 }
 
-// TODO: Support format string when `RuntimeAssertFailed` supports it.
-inline RUNTIME_NORETURN void TODOImpl(const char* location, const char* message) {
-    RuntimeAssertFailed(location, message);
+template <typename ... Args>
+inline RUNTIME_NORETURN void TODOImpl(const char* location, const char* message, Args ... args) {
```
projedi, line 37: The usual pattern for forwarding is:
```cpp
template <typename... Args>
void f(Args&&... args) {
   g(std::forward<Args>(args)...);
}
```
Note the use of `&&` in the argument list, and `std::forward` where forwarding happens. This pattern allows copying the arguments that should be copied and moving the arguments that should be moved.

runtime/src/main/cpp/KAssert.cpp
```diff
@@ -4,9 +4,107 @@
...
+#include "ExecFormat.h"
+#else
+// Glibc backtrace() function.
+#include <execinfo.h>
+#endif // USE_GCC_UNWIND
+#endif // !KONAN_NO_EXCEPTIONS
+
+
+// The stacktrace printing below partially duplicates the logic that forms stacktraces for Kotlin exceptions (see Exceptions.cpp).
+// TODO: Unify this code with the exceptions machinery.
```
projedi, line 24: Took a stab at it: https://github.com/JetBrains/kotlin-native/pull/4618 WDYT?

runtime/src/main/cpp/KAssert.h
```diff
@@ -25,17 +25,17 @@
 #define STRINGIFY(x) #x
 #define TOSTRING(x) STRINGIFY(x)
 
-RUNTIME_NORETURN void RuntimeAssertFailed(const char* location, const char* message, ...);
+RUNTIME_NORETURN NO_INLINE void RuntimeAssertFailed(const char* location, const char* message, ...);
```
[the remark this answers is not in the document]
projedi: Tried it out over here: https://github.com/JetBrains/kotlin-native/pull/4624/files

ilmat192:
Closing in favor of a branch in the main Kotlin repo: https://github.com/JetBrains/kotlin/tree/rrn/matveev/stacktrace-in-asserts

PR #4610, opened 2020-12-21
[runtime] Old impl of RuntimeAssert for backward compatibility

homuroll:


runtime/src/main/cpp/KAssert.cpp
```diff
@@ -30,3 +30,14 @@ RUNTIME_NORETURN void RuntimeAssertFailed(const char* location, const char* form
     // TODO: Write the stacktrace.
     konan::abort();
 }
+
+// TODO: This is needed for backward compatibility.
```
SvyatoslavScherbina, line 34:
```suggestion
// TODO: this function is not used by runtime, but apparently there are
// third-party libraries that use it (despite the fact it is not a public API).
// Keeping the function here for now for backward compatibility, to be removed later.
```

ilmat192 (approved):

PR #4611, opened 2020-12-21
Use Alloc.h everywhere

projedi:


runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -212,7 +212,7 @@ class CycleDetector : private kotlin::Pinned {
 
   static CycleDetector& instance() {
     // Only store a pointer to CycleDetector in .bss
-    static CycleDetector* result = new CycleDetector();
+    static CycleDetector* result = new (konanAllocMemory(sizeof(CycleDetector))) CycleDetector();
```
projedi, line 215: An alternative to that would be something like:
```cpp
class KonanAllocatorAware {
public:
    static void* operator new(size_t size) noexcept {
        return konanAllocMemory(size);
    }
    static void operator delete(void* ptr) noexcept {
         konanFreeMemory(ptr);
    }
};

class CycleDetector : public KonanAllocatorAware {
    …
};

// Alternatively
class SomeOtherClass : private KonanAllocatorAware {
public:
    using KonanAllocatorAware::operator new;
    using KonanAllocatorAware::operator delete;
    …
};
```

projedi: Implemented the second variant.

runtime/src/main/cpp/Alloc.h
```diff
@@ -119,4 +119,48 @@ bool operator!=(
...
+template <class T>
+class KonanAllocatorAware {
+public:
+    static void* Alloc() noexcept { return konanAllocMemory(sizeof(T)); }
+
+    static void* operator new(size_t count) = delete;
+    static void* operator new[](size_t count) = delete;
+
+    static void* operator new(size_t count, void* ptr) noexcept { return ptr; }
+    static void* operator new[](size_t count, void* ptr) noexcept { return ptr; }
```
SvyatoslavScherbina, line 146: Would allocating an array work as expected?

runtime/src/main/cpp/Alloc.h
```diff
@@ -119,4 +119,48 @@ bool operator!=(
...
+};
+
+// Force a class to be heap-allocated using `konanAllocMemory`. Does not prevent stack allocation, or
+// allocation as part of another object.
+// Usage:
+// class A : public KonanAllocatorAware<A> {
+//     ...
+// };
+// To construct `A` on the heap:
+// new (A::Alloc()) A(...);
```
SvyatoslavScherbina, line 135: Can we make it more like `new (KonanAlloc) A(...)`?
This might be even more simple and less fragile.

projedi: That's much better, thanks!

runtime/src/main/cpp/Alloc.h
projedi: Nope. Fixed, added tests.

SvyatoslavScherbina (approved):

LepilkinaElena (approved):

SvyatoslavScherbina (approved):

runtime/src/main/cpp/SingleLockList.hpp
```diff
@@ -31,7 +32,7 @@ class SingleLockList : private Pinned {
         Node(Args... args) noexcept : value(args...) {}
 
         Value value;
-        std::unique_ptr<Node> next;
+        KStdUniquePtr<Node> next;
```
knebekaizer, line 35: Smart ptr in linked list may cause uncontrolled deep recursion on delete attempt so it's a sort of anti-pettern. Consider raw ptr, or prevent recursive destruction somehow

projedi: I kept smart pointer, but added manual destruction to avoid destructor recursion. This approach is advocated for here: https://youtu.be/JfmTagWcqoE?t=1581

knebekaizer (approved):

PR #4615, opened 2020-12-23
Fix reporting of filtered out test suites

egorzhdan:
This fixes the `--ktest_no_excluded_test_suites` command line option behavior to match its description in the help page.

It is a follow-up for https://github.com/JetBrains/kotlin-native/pull/4522, where I apparently misunderstood how test filtering works. Sorry for that, now the CLI flag is tested more carefully.


ilmat192:
@egorzhdan Hi! Sorry for the delay with the answer.

I've experimented a little and found out that skipping filtered out test suites is a default behaviour not only for JUnit but also for TestNG and Google Test. Thus it worth making such a behaviour a default for us too. I've cherry-picked your commit and made this change on top of it. Could you please take a look: https://github.com/JetBrains/kotlin-native/pull/4678?

egorzhdan:
@ilmat192 thanks, that change looks good to me 👍

ilmat192:
Ok, closing this PR in favor of https://github.com/JetBrains/kotlin-native/pull/4678.

PR #4618, opened 2020-12-24
Refactor stacktraces

projedi:


runtime/src/main/cpp/StackTrace.hpp
```diff
@@ -0,0 +1,117 @@
...
+    // TODO: This API is asking for a span.
+
+    size_t size() const noexcept { return skipFrames_ >= size_ ? 0 : size_ - skipFrames_; }
+
+    void* const* data() const noexcept { return buffer_.data() + skipFrames_; }
+
+private:
+    size_t size_ = 0;
+    size_t skipFrames_;
+    std::array<void*, 32> buffer_ = {nullptr};
```
ilmat192, line 44: This change limits the stacktrace depth available to user in exceptions.

We do has this limitation for the libc unwind in the current master. But for the GCC unwind, we lift this limitation by [counting](https://github.com/JetBrains/kotlin-native/blob/20b6f36af8d10b7f1877f9b83da0eea188114085/runtime/src/main/cpp/Exceptions.cpp#L130) stack frames and dynamically allocating a buffer for the stacktrace elements.

What if we keep using a dynamic storage (`std::vector`?) instead of `std::array`? It does prevent us from showing stacktraces in OOM situations, but they look like a rear case (and are not supported in the current master).

projedi: We want to use these stacktraces also from `RuntimeAssert`, where using heap allocation might be unwelcome. Also, do we really want to display stacktrace of unbounded length?

ilmat192: > We want to use these stacktraces also from `RuntimeAssert`, where using heap allocation might be unwelcome.

Just out of curiosity: what problems except OOM can we face trying to allocate a dynamic memory in `RuntimeAssert`? 

> Also, do we really want to display stacktrace of unbounded length?

I think that it is nice to have. But since current master already bounds the length for Apple targets, it doesn't seem to be a must.

OTOH, limiting stacktrace length for non-Apple targets looks like a regression. Is this regression is small enough to be acceptable? @SvyatoslavScherbina, what do you think?

SvyatoslavScherbina: > OTOH, limiting stacktrace length for non-Apple targets looks like a regression. Is this regression is small enough to be acceptable? @SvyatoslavScherbina, what do you think?

Actually I'm not sure that limiting stacktrace length on Apple platforms was a totally intentional design decision. I guess it was just the easiest way to quickly implement some stacktraces back then. So I don't think we should be guided by that decision. And new implementation shouldn't make it hard to change.

Also note that symbolication probably allocates memory anyway, and I'm not sure unsymbolicated stacktraces are totally useful.

projedi: > Just out of curiosity: what problems except OOM can we face trying to allocate a dynamic memory in RuntimeAssert?

`malloc` is not safe to use in signal handlers. It's the same reason we use gcc unwind instead of libc's `backtrace` on linux.

projedi: > Also note that symbolication probably allocates memory anyway

It does. Our own `AddressToSymbols` does it too. What's worse - it uses `RuntimeAssert` to signal that it's out of memory. However, I think, we can fix our implementation to avoid memory allocations.

> I'm not sure unsymbolicated stacktraces are totally useful.

Some stacktrace is better than no stacktrace. You can use an external tool (gdb/lldb, for example) to translate address into a symbol post factum.

SvyatoslavScherbina: > Some stacktrace is better than no stacktrace. You can use an external tool (gdb/lldb, for example) to translate address into a symbol post factum.

Yes, but there are extremely frequent cases where this is complicated or impossible:
1. When assertion happens on the CI
2. When assertion happens on user's side.

Anyway, I don't think we would encounter OOM during stack trace building often, in fact we've never encountered it yet, so this part of the discussion is probably irrelevant.

PR #4619, opened 2020-12-24
Fix NSUInteger size for Watchos x64

PavelPunegov:
Fix for the https://youtrack.jetbrains.com/issue/KT-44083

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
@@ -1563,34 +1563,22 @@ internal fun ObjCExportCodeGenerator.getEncoding(methodBridge: MethodBridge): St
         }
     }
 
-    val targetFamily = context.config.target.family
-    val returnTypeEncoding = methodBridge.returnBridge.getObjCEncoding(targetFamily)
+    val returnTypeEncoding = methodBridge.returnBridge.getObjCEncoding(context)
 
     val paramSize = paramOffset
     return "$returnTypeEncoding$paramSize$params"
 }
 
-// https://developer.apple.com/documentation/objectivec/nsuinteger?language=objc
-// `typedef unsigned long NSUInteger` on iOS, macOS, tvOS.
-// `typedef unsigned int NSInteger` on watchOS.
-private val Family.nsUIntegerEncoding: String get() = when (this) {
-    Family.OSX,
-    Family.IOS,
-    Family.TVOS -> "L"
-    Family.WATCHOS -> "I"
-    else -> error("Unexpected target platform: $this")
-}
-
-private fun MethodBridge.ReturnValue.getObjCEncoding(targetFamily: Family): String = when (this) {
+private fun MethodBridge.ReturnValue.getObjCEncoding(context: Context): String = when (this) {
     MethodBridge.ReturnValue.Suspend,
     MethodBridge.ReturnValue.Void -> "v"
-    MethodBridge.ReturnValue.HashCode -> targetFamily.nsUIntegerEncoding
+    MethodBridge.ReturnValue.HashCode -> if (context.is64BitNSInteger()) "L" else "I"
     is MethodBridge.ReturnValue.Mapped -> this.bridge.objCEncoding
     MethodBridge.ReturnValue.WithError.Success -> ObjCValueType.BOOL.encoding
 
     MethodBridge.ReturnValue.Instance.InitResult,
     MethodBridge.ReturnValue.Instance.FactoryResult -> ReferenceBridge.objCEncoding
-    is MethodBridge.ReturnValue.WithError.ZeroForError -> this.successBridge.getObjCEncoding(targetFamily)
+    is MethodBridge.ReturnValue.WithError.ZeroForError -> this.successBridge.getObjCEncoding(context)
 }
 
 private val MethodBridgeParameter.objCEncoding: String get() = when (this) {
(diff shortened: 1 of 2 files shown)


backend.native/tests/interop/objc/smoke.kt
```diff
@@ -67,11 +67,12 @@ fun run() {
 
     // hashCode (directly):
     // hash() returns value of NSUInteger type.
-    val hash = when (Platform.osFamily) {
+    val hash = if (Platform.osFamily == OsFamily.WATCHOS && Platform.cpuArchitecture.bitness == 32) {
```
sbogolepov, line 70: This check is not correct for `watchos_arm64` (its bitness is 64 and NSInteger is defined as int). Though we are not testing this target, this piece of code might be confusing for readers.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -1563,34 +1563,22 @@ internal fun ObjCExportCodeGenerator.getEncoding(methodBridge: MethodBridge): St
...
-    Family.WATCHOS -> "I"
-    else -> error("Unexpected target platform: $this")
-}
-
-private fun MethodBridge.ReturnValue.getObjCEncoding(targetFamily: Family): String = when (this) {
+private fun MethodBridge.ReturnValue.getObjCEncoding(context: Context): String = when (this) {
     MethodBridge.ReturnValue.Suspend,
     MethodBridge.ReturnValue.Void -> "v"
-    MethodBridge.ReturnValue.HashCode -> targetFamily.nsUIntegerEncoding
+    MethodBridge.ReturnValue.HashCode -> if (context.is64BitNSInteger()) "L" else "I"
```
sbogolepov, line 1575: Did you check it for `ios_arm32`?

PavelPunegov: We don't have any devices that support that target

backend.native/tests/interop/objc/smoke.kt
PavelPunegov: Changed the check

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina, line 1575: I'm not sure this is entirely correct. `L` doesn't mean "64-bit intenger", it is used for C `unsigned long` type. So `NSUInteger` should still be `L` on ios_arm32. See also https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html

backend.native/tests/interop/objc/smoke.kt
```diff
@@ -67,11 +67,12 @@ fun run() {
 
     // hashCode (directly):
     // hash() returns value of NSUInteger type.
-    val hash = when (Platform.osFamily) {
+    val hash = if (Platform.osFamily == OsFamily.WATCHOS && Platform.cpuArchitecture != CpuArchitecture.X64) {
```
SvyatoslavScherbina, line 70: Checking `sizeOf<NSUInteger>() == 4` instead also might make sense.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
SvyatoslavScherbina: JFTR: we've rechecked and apparently things are even more complicated here:
https://github.com/llvm/llvm-project/blob/c161775decddcc86fbbfefd7485a5d0ef5842aec/clang/lib/AST/ASTContext.cpp#L7163
(which kinda doesn't match the document referred above).

Current PR version still might be incorrect, probably needs more work.

sbogolepov: Yes, but we can compare compilers output. E.g. check what Clang does.

backend.native/tests/interop/objc/smoke.kt
```diff
@@ -67,11 +67,12 @@ fun run() {
 
     // hashCode (directly):
     // hash() returns value of NSUInteger type.
-    val hash = when (Platform.osFamily) {
+    val hash = if (sizeOf<NSUIntegerVar>() == 4L) {
```
SvyatoslavScherbina, line 70: The comments below are obsolete too.

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/objcexport/ObjCExportCodeGenerator.kt
```diff
@@ -1563,34 +1563,22 @@ internal fun ObjCExportCodeGenerator.getEncoding(methodBridge: MethodBridge): St
...
-    Family.WATCHOS -> "I"
-    else -> error("Unexpected target platform: $this")
-}
-
-private fun MethodBridge.ReturnValue.getObjCEncoding(targetFamily: Family): String = when (this) {
+private fun MethodBridge.ReturnValue.getObjCEncoding(context: Context): String = when (this) {
     MethodBridge.ReturnValue.Suspend,
     MethodBridge.ReturnValue.Void -> "v"
-    MethodBridge.ReturnValue.HashCode -> targetFamily.nsUIntegerEncoding
+    MethodBridge.ReturnValue.HashCode -> if (context.is64BitNSInteger()) "Q" else "I"
```
sbogolepov, line 1575: IIUC, It should be "L" for ios_arm32:
* `NSUInteger` is an alias for `unsigned long` on iOS.
* `unsigned long` is mapped to "L" on 32-bit platforms.
Am I missing something?

PavelPunegov: See [comment in the issue](https://youtrack.jetbrains.com/issue/KT-44083#focus=Comments-27-4644443.0-0) with details of the `clang` output. It shows that it encodes to `I` on 32-bit iOS.

sbogolepov (approved):

PR #4622, opened 2020-12-28
Support sanitizers in runtime tests

projedi:


PavelPunegov (approved):

runtime/build.gradle.kts
```diff
@@ -167,11 +170,16 @@ targetList.forEach { targetName ->
...
-        dependsOn("${targetName}ExperimentalMMStdAllocRuntimeTests")
-        dependsOn("${targetName}ExperimentalMMMimallocRuntimeTests")
+        val platformManager = project.rootProject.findProperty("platformManager") as PlatformManager
+        val target = platformManager.targetByName(targetName)
+        target.supportedSanitizers().forEach { sanitizer ->
+            dependsOn("${targetName}StdAllocRuntimeTests${CompileToBitcodeExtension.suffixForSanitizer(sanitizer)}")
+            dependsOn("${targetName}MimallocRuntimeTests${CompileToBitcodeExtension.suffixForSanitizer(sanitizer)}")
+            dependsOn("${targetName}ExperimentalMMStdAllocRuntimeTests${CompileToBitcodeExtension.suffixForSanitizer(sanitizer)}")
+            dependsOn("${targetName}ExperimentalMMMimallocRuntimeTests${CompileToBitcodeExtension.suffixForSanitizer(sanitizer)}")
+        }
```
SvyatoslavScherbina, line 182: This part is kinda too intrusive. Maybe make `createTestTasks` return list of tasks and use the results here?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Sanitizer.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.konan.target
+
+enum class SanitizerKind {
+    NONE,
```
sbogolepov, line 9: Maybe get rid of this entry in favor of nullability ?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
```diff
@@ -158,7 +159,8 @@ internal class Linker(val context: Context) {
                     kind = linkerOutput,
                     outputDsymBundle = context.config.outputFiles.symbolicInfoFile,
                     needsProfileLibrary = needsProfileLibrary,
-                    mimallocEnabled = mimallocEnabled)
+                    mimallocEnabled = mimallocEnabled,
+                    sanitizer = SanitizerKind.NONE)
```
sbogolepov, line 163: What if we want to use several sanitizers at the same time?

build-tools/src/main/kotlin/org/jetbrains/kotlin/testing/native/NativeTest.kt
```diff
@@ -13,24 +13,33 @@ import org.gradle.api.file.ConfigurableFileCollection
...
     @OutputFile
     var outputFile = project.buildDir.resolve("bin/test/$target/${inputFile.nameWithoutExtension}.o")
 
     @Input
     val clangArgs = mutableListOf<String>()
 
+    @Input
+    private val sanitizerFlags = when (sanitizer) {
+        SanitizerKind.NONE -> listOf()
+        SanitizerKind.ADDRESS -> listOf("-fsanitize=address")
```
sbogolepov, line 33: We can replace such `when`s with enum properties to avoid code duplication.

projedi: TBH I think every property would be uniquely used, so I don't see code duplication:
* postfixes for build task names
* flags for the clang frontend
* flags for the clang backend (which just turned out to be the same as frontend flags)
* n flags for n different linkers

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Linker.kt
projedi: For the sanitizers we want to use: asan, tsan - mixing is not even allowed. If in the future we'd want to combine several other sanitizers, we could either bundle them under one `SanitizerKind` entry, or refactor the code by passing a list of `SanitizerKind`s. For now, I think, a more simple implementation would do.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -123,7 +124,9 @@ class AndroidLinker(targetProperties: AndroidConfigurables)
                                    libraries: List<String>, linkerArgs: List<String>,
                                    optimize: Boolean, debug: Boolean,
                                    kind: LinkerOutputKind, outputDsymBundle: String,
-                                   needsProfileLibrary: Boolean, mimallocEnabled: Boolean): List<Command> {
+                                   needsProfileLibrary: Boolean, mimallocEnabled: Boolean,
+                                   sanitizer: SanitizerKind): List<Command> {
+        if (sanitizer != SanitizerKind.NONE) throw Error("Sanitizers are unsupported")
```
ilmat192, line 129: (minor) It's possible to use standard functions instead of plain `if`s: [`check`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html), [`require`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html)

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -387,20 +422,23 @@ class MingwLinker(targetProperties: MingwConfigurables)
...
 
-    override fun provideCompilerRtLibrary(libraryName: String): String? {
+    override fun provideCompilerRtLibrary(libraryName: String, isDynamic: Boolean): String? {
         val targetSuffix = when (target) {
             KonanTarget.MINGW_X64 -> "x86_64"
             else -> error("$target is not supported.")
         }
         val dir = File("$absoluteLlvmHome/lib/clang/").listFiles.firstOrNull()?.absolutePath
-        return if (dir != null) "$dir/lib/windows/libclang_rt.$libraryName-$targetSuffix.a" else null
+        val extension = if (isDynamic) "so" else "a"
```
ilmat192, line 431: Should it really be `so`, not `dll`?

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Sanitizer.kt
projedi: I had issues with that. First of all, gradle does not support calling task constructors with `null`: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:passing_arguments_to_a_task_constructor So I had to move `sanitizer` into a `public var`. Which broke `CompileToBitcode`<->`CompilationDatabase` interaction, because the latter reads target directory of the former during configuration phase, but the target directory depends on the sanitizer value which now can change during the configuration phase. This interaction is likely flawed in and of itself but will require some refactoring to get fixed.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
projedi: I just removed support for `isDynamic` for both mingw and linux, since it's unused (and thus untested).

ilmat192 (approved):

sbogolepov (approved):

SvyatoslavScherbina (approved):

PR #4624, opened 2020-12-29
Use clang format attribute

projedi:
`__attribute__(format(...))` forces the compiler to check call sites for valid usage of format strings.

### DIFF
--- a/runtime/src/main/cpp/StdCppStubs.cpp
+++ b/runtime/src/main/cpp/StdCppStubs.cpp
@@ -32,7 +32,8 @@ RUNTIME_USED RUNTIME_WEAK extern "C" char* Konan_cxa_demangle(
 
 namespace std {
 RUNTIME_WEAK void __throw_length_error(const char* __s __attribute__((unused))) {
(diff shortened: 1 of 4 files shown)


ilmat192 (approved):

runtime/src/main/cpp/StdCppStubs.cpp
```diff
@@ -32,7 +32,8 @@ RUNTIME_USED RUNTIME_WEAK extern "C" char* Konan_cxa_demangle(
 
 namespace std {
 RUNTIME_WEAK void __throw_length_error(const char* __s __attribute__((unused))) {
-  RuntimeAssert(false, __s);
+  // TODO: Don't we always need to abort here?
```
SvyatoslavScherbina, line 35: I guess we do. Probably this code was written when assertions were always enabled unconditionally. Could you please fix this?

SvyatoslavScherbina (approved):

SvyatoslavScherbina (approved):

PR #4625, opened 2020-12-29
Use common pattern for function-like macro

projedi:
```cpp
do {
   <stmts>
} while (false)
```
is a common pattern for function-like macro. It makes macro invocation look like a normal function call by requiring a trailing semicolon.

### DIFF
--- a/runtime/src/main/cpp/KAssert.h
+++ b/runtime/src/main/cpp/KAssert.h
@@ -67,10 +67,11 @@ extern "C" const int KonanNeedDebugInfo;
 
 // Use RuntimeCheck() in runtime checks that could fail due to external condition and shall lead
 // to program termination. Never compiled out.
+// TODO: Consider using `CURRENT_SOURCE_LOCATION` when `KonanNeedDebugInfo` is `true`.
 #define RuntimeCheck(condition, format, ...) \
     do { \
         if (!(condition)) { \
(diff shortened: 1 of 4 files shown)


runtime/src/main/cpp/KAssert.h
```diff
@@ -46,24 +53,30 @@ extern "C" const int KonanNeedDebugInfo;
...
 // Use RuntimeCheck() in runtime checks that could fail due to external condition and shall lead
 // to program termination. Never compiled out.
-#define RuntimeCheck(condition, format, ...)             \
-  if (!(condition)) {                                    \
-    RuntimeAssertFailed(nullptr, format, ##__VA_ARGS__); \
-  }
+#define RuntimeCheck(condition, format, ...) \
+    do { \
+        if (!(condition)) { \
+            RuntimeAssertFailed(CURRENT_SOURCE_LOCATION, format, ##__VA_ARGS__); \
```
SvyatoslavScherbina, line 73: So this enables source location info for `RuntimeCheck` even in release binaries, thus adds more strings to the binaries. Is this intended?

projedi: My bad, fixed.

ilmat192 (approved):

SvyatoslavScherbina (approved):

PR #4626, opened 2020-12-29
Locale-agnostic case conversions by default #KT-43023

qurbonzoda:


runtime/src/main/cpp/KString.cpp
```diff
@@ -773,7 +773,7 @@ OBJ_GETTER(Kotlin_String_plusImpl, KString thiz, KString other) {
   RETURN_OBJ(result->obj());
 }
 
-OBJ_GETTER(Kotlin_String_toUpperCase, KString thiz) {
+OBJ_GETTER(Kotlin_String_uppercase, KString thiz) {
```
ilya-g, line 776: It would be a more scoped commit if we don't change cpp symbol names for now, and just implement new functions in terms of the existing ones. It is possible?

runtime/src/main/kotlin/kotlin/text/regex/sets/CharSet.kt
```diff
@@ -28,13 +28,15 @@ package kotlin.text.regex
 open internal class CharSet(char: Char, val ignoreCase: Boolean = false) : LeafSet() {
 
     // We use only low case characters when working in case insensitive mode.
-    val char: Char = if (ignoreCase) char.toLowerCase() else char
+    @OptIn(ExperimentalStdlibApi::class)
+    val char: Char = if (ignoreCase) char.lowercaseChar() else char
```
ilya-g, line 32: Changing usages are also not strictly needed for this PR.

qurbonzoda: done

runtime/src/main/cpp/KString.cpp
qurbonzoda: done

ilya-g (approved):

qurbonzoda:
Merged manually

PR #4632, opened 2021-01-11
Use toolchain clang for runtime tests on mac

projedi:
This mirrors `BitcodeCompiler`: https://github.com/JetBrains/kotlin-native/blob/d3c164de78d5eb8088168d809ec70934ec2ddbf0/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeCompiler.kt#L81

### DIFF
--- a/build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecClang.kt
+++ b/build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecClang.kt
@@ -54,8 +54,13 @@ class ExecClang(private val project: Project) {
         val executable = executableOrNull ?: "clang"
 
         if (listOf("clang", "clang++").contains(executable)) {
+            // TODO: This is copied from `BitcodeCompiler`. Consider sharing the code instead.
             val platform = platformManager.platform(target)
-            return "${platform.absoluteTargetToolchain}/usr/bin/$executable"
+            return if (target.family.isAppleFamily) {
+                "${platform.absoluteTargetToolchain}/usr/bin/$executable"
+            } else {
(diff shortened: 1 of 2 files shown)


build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecClang.kt
```diff
@@ -50,6 +50,17 @@ class ExecClang(private val project: Project) {
         }
     }
 
+    fun resolveToolchainExecutable(target: KonanTarget, executableOrNull: String?): String {
+        val executable = executableOrNull ?: "clang"
+
+        if (listOf("clang", "clang++").contains(executable)) {
+            val platform = platformManager.platform(target)
+            return "${platform.absoluteTargetToolchain}/usr/bin/$executable"
```
ilmat192, line 58: (minor) May be support paths for non-Apple toolchains too, like it is done [here](https://github.com/JetBrains/kotlin-native/blob/d3c164de78d5eb8088168d809ec70934ec2ddbf0/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/BitcodeCompiler.kt#L37) for `BitcodeCompiler`? We could check `target.family` instead of `platform.configurables is AppleConfigurables`.

Not sure if all our toolchains include clang though.

ilmat192 (approved):

build-tools/src/main/kotlin/org/jetbrains/kotlin/ExecClang.kt
projedi: Fixed, thanks!

PR #4637, opened 2021-01-12
[IR] Fix for https://youtrack.jetbrains.com/issue/KT-44219 + test

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
```diff
@@ -5,35 +5,41 @@
...
+internal class PostInlineLowering(val context: Context) : BodyLoweringPass {
 
     private val symbols get() = context.ir.symbols
 
-    override fun lower(irFile: IrFile) {
-        irFile.transformChildrenVoid(object : IrBuildingTransformer(context) {
+    override fun lower(irBody: IrBody, container: IrDeclaration) {
+        val irFile = container.file
+        val irBuilder = context.createIrBuilder((container as IrSymbolOwner).symbol)
+        irBody.transformChildrenVoid(object : IrElementTransformerVoid() {
```
SvyatoslavScherbina, line 37: Would it traverse local function bodies multiple times?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
```diff
@@ -5,35 +5,41 @@
...
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 
 /**
  * This pass runs after inlining and performs the following additional transformations over some operations:
  *     - Convert immutableBlobOf() arguments to special IrConst.
  *     - Convert `obj::class` and `Class::class` to calls.
  */
-internal class PostInlineLowering(val context: Context) : FileLoweringPass {
+internal class PostInlineLowering(val context: Context) : BodyLoweringPass {
```
SvyatoslavScherbina, line 30: Why is `BodyLoweringPass` required here?

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
homuroll: No, by default `BodyLoweringPass` does not go into the bodies. (see https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/Lower.kt#L121)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
homuroll: We want to know the containing function.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/PostInlineLowering.kt
SvyatoslavScherbina: I see. But isn't `irBuilder`'s scope then wrong for local functions?

homuroll: Indeed, we have to the same trick as in `PreInlineLowering`.

SvyatoslavScherbina (approved):

PR #4640, opened 2021-01-14
Mark safe points in code generator

projedi:
**TODO**:

- [x] suspend functions
- [x] exceptions handling
- [ ] ~~tests~~
- [x] check that code generation hasn't changed in the legacy MM case.
- [ ] generated code around interops?


projedi:
I think, for stable tests on this, we need an infra to analyse generated llvm bitcode.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -1251,6 +1251,7 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
                 returnType == voidType -> {
                     releaseVars()
                     assert(returnSlot == null)
+                    call(context.llvm.Kotlin_mm_safePointFunctionEpilogue, emptyList())
```
SvyatoslavScherbina, line 1254: While these calls are probably inlined in release mode, I'm not sure this happens in debug mode.
If it isn't, we probably should emit the calls only for the new MM.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -1241,6 +1241,7 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
 
             functionGenerationContext.positionAtEnd(loopBody)
             loop.body?.generate()
+            call(context.llvm.Kotlin_mm_safePointWhileLoopBody, emptyList())
```
SvyatoslavScherbina, line 1244: I'm not sure this would work with `continue`. Same for `do-while`.

SvyatoslavScherbina: (I mean, it is possible that using `continue` would skip the safe point here)

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
sbogolepov: They are marked as `always_inline`, so they should.

projedi: > I'm not sure this happens in debug mode.

Huh, you're right. I thought `always_inline` would definitely help, but there are many instances (mostly in stdlib) where calls to `Kotlin_mm_safePointFunctionEpilogue` get left in place.

sbogolepov: That's strange. Is there any other `always_inline` functions that are not inlined?

sbogolepov: Oh, I see. We link object files in debug mode, thus these callees cannot be inlined.

projedi: > Oh, I see. We link object files instead of bitcode in debug mode, thus these callees cannot be inlined.

Do we also link object files when compiling via konanc cli?

sbogolepov: Not by default. Only if we pass `-g -Xcache-directory=path_to/macos_x64g-STATIC`.

projedi: Then we have 2 problems:
* `always_inline` may not inline even when compiling bitcode. And the [docs](https://clang.llvm.org/docs/AttributeReference.html#always-inline-force-inline) actually don't promise anything
* We may not inline with caches (i.e. actual user side case), because that requires LTO, which will slow down compilation.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/IrToBitcode.kt
```diff
@@ -534,9 +534,11 @@ internal class CodeGeneratorVisitor(val context: Context, val lifetimes: Map<IrE
         }
 
         override fun genContinue(destination: IrContinue) {
-            if (destination.loop == loop)
+            if (destination.loop == loop) {
+                if (context.memoryModel == MemoryModel.EXPERIMENTAL)
+                    call(context.llvm.Kotlin_mm_safePointWhileLoopContinue, emptyList())
```
SvyatoslavScherbina, line 539: Have you considered placing safe points to the beginning of `loopCheck`, to cover both cases -- with and without `continue`? I guess there are pros and cons, so we should at least compare both options.

projedi: This would make safe point fire even if the loop condition is false at the start. However, we could place safe point at the beginning of loop body instead.

SvyatoslavScherbina: Indeed. This would probably be the most efficient solution in terms of code size. Runtime performance is more complicated though.

projedi: Okay, done.

SvyatoslavScherbina (approved):
LGTM. (Ok to merge without tests for now).

ilmat192 (approved):

PR #4645, opened 2021-01-20
Lazy task configuration in Konan plugin and platform libs

PavelPunegov:
Use Gradle lazy API (Providers) in old Konan gradle plugin and platform libs to decrease amount of task created during confguration phase


build-tools/src/main/kotlin/org/jetbrains/kotlin/KLibInstall.kt
```diff
@@ -12,17 +13,16 @@ import java.io.File
...
 
     @Input
     var repo: File = project.rootDir
 
-    val installDir: File
-        @OutputDirectory
-        get() {
-            val klibName = klib.name.take(klib.name.lastIndexOf('.'))
-            return project.file("${repo.absolutePath}/$klibName")
+    val installDir: Provider<File>
```
vvlevchenko, line 22: May be @Output required?

vvlevchenko (approved):

build-tools/src/main/kotlin/org/jetbrains/kotlin/KLibInstall.kt
PavelPunegov: Only `@OutputDirectory` exists in gradle

backend.native/tests/build.gradle
```diff
@@ -3739,7 +3739,7 @@ createInterop("embedStaticLibraries") {
     // Note: also hardcoded in def file.
     final String libDir = "$buildDir/embedStaticLibraries/"
 
-    it.getByTarget(target.name).doFirst {
+    it.getByTarget(target.name).get().doFirst {
```
ilmat192, line 3742: May be use `configure { ... }` instead of `get()`?

build-tools/src/main/kotlin/org/jetbrains/kotlin/KLibInstall.kt
ilmat192: But it's still an output directory and we still need to tell about it to Gradle.
`@OutputDirectory` works well with providers too, so I think we can just keep this annotation here.

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanBuildingTask.kt
```diff
@@ -30,7 +29,7 @@ abstract class KonanBuildingTask: KonanArtifactWithLibrariesTask(), KonanBuildin
 
     @get:Internal
     internal abstract val toolRunner: KonanToolRunner
-    internal abstract fun toModelArtifact(): KonanModelArtifact
+//    internal abstract fun toModelArtifact(): KonanModelArtifact
```
ilmat192, line 32: May be remove it at all?

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanPlugin.kt
```diff
@@ -363,15 +363,36 @@ class KonanPlugin @Inject constructor(private val registry: ToolingModelBuilderR
             doLast { project.cleanKonan() }
         }
 
+        project.afterEvaluate {
+            project.tasks
+                .withType(KonanCompileProgramTask::class.java)
+                .forEach { task ->
```
ilmat192, line 369: Doesn't this call create all the tasks of the `KonanCompileProgramTask` class? 
Do we use the run tasks somehow? May be just drop them? :)

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanBuildingConfig.kt
```diff
@@ -143,31 +150,53 @@ abstract class KonanBuildingConfig<T: KonanBuildingTask>(private val name_: Stri
...
+        tasks().forEach {
+            it.configure { t ->
+                t.destinationDir(
+                    project.file(dir).targetSubdir(t.konanTarget)
+                )
+            }
+        }
+
+    override fun libraries(closure: Closure<Unit>) =
+        tasks().forEach { it.configure { t -> t.libraries(closure) } }
```
ilmat192, line 175: (minor) The construction `tasks().forEach { it.configure { ... } }` is very common now. May be move it to a separate function? E.g.

```kotlin
fun configureTasks(block: (T) -> Unit) = tasks().forEach { it.configure(block) }
```

backend.native/tests/build.gradle
PavelPunegov: Tnx, fixed

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/tasks/KonanBuildingTask.kt
PavelPunegov: Forgotten, fixed

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanPlugin.kt
PavelPunegov: Yes, this call will create tasks. As the task creation requires properties from the build task, this could not be avoided. It seems that we don't use these run tasks so it should not be an issue. I think it's ok to leave them

ilmat192: > It seems that we don't use these run tasks so it should not be an issue.

But we still execute this `afterEvaluate { ... }` on each configuration thus create all the `KonanCompileProgramTask`s, don't we?

ilmat192 (approved):

tools/kotlin-native-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/konan/KonanBuildingConfig.kt
PavelPunegov: I'll do that inside Kotlin repo with improvements to the test infra (making it configure lazily also)

PR #4650, opened 2021-01-22
Implement freezing in new MM

projedi:


projedi:
There are some problems with the approach (or in some other part of the new MM). In order to start running backend tests, I'll postpone this implementation.

projedi:
For now https://github.com/JetBrains/kotlin-native/pull/4669

projedi:
Moved over to https://github.com/JetBrains/kotlin/tree/rrn/alshabalin/mm-freeze

PR #4653, opened 2021-01-26
Implement ref updates in new MM

projedi:


runtime/src/mm/cpp/Memory.cpp
```diff
@@ -125,23 +127,81 @@ extern "C" OBJ_GETTER(AllocArrayInstance, const TypeInfo* typeInfo, int32_t elem
...
+    mm::CompareAndSwapHeapRef(location, nullptr, const_cast<ObjHeader*>(object));
+}
+
+extern "C" ALWAYS_INLINE RUNTIME_NOTHROW void UpdateReturnRef(ObjHeader** returnSlot, const ObjHeader* object) {
+    mm::SetStackRef(returnSlot, const_cast<ObjHeader*>(object));
+}
+
+extern "C" ALWAYS_INLINE RUNTIME_NOTHROW OBJ_GETTER(
+        SwapHeapRefLocked, ObjHeader** location, ObjHeader* expectedValue, ObjHeader* newValue, int32_t* spinlock, int32_t* cookie) {
+    auto* oldValue = mm::CompareAndSwapHeapRef(location, expectedValue, newValue);
```
SvyatoslavScherbina, line 192: It is possible that at this moment `oldValue` is not referred from roots or heap. So this implementation is safe-points-specific.

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -125,23 +127,81 @@ extern "C" OBJ_GETTER(AllocArrayInstance, const TypeInfo* typeInfo, int32_t elem
...
+    RETURN_OBJ(oldValue);
+}
+
+extern "C" ALWAYS_INLINE RUNTIME_NOTHROW void SetHeapRefLocked(
+        ObjHeader** location, ObjHeader* newValue, int32_t* spinlock, int32_t* cookie) {
+    mm::SetHeapRefAtomic(location, newValue);
+}
+
+extern "C" ALWAYS_INLINE RUNTIME_NOTHROW OBJ_GETTER(ReadHeapRefLocked, ObjHeader** location, int32_t* spinlock, int32_t* cookie) {
+    RETURN_OBJ(mm::ReadHeapRefAtomic(location));
```
SvyatoslavScherbina, line 202: Same as above.

runtime/src/mm/cpp/Memory.cpp
SvyatoslavScherbina: (Not sure if we should fix this, but maybe mark such places somehow).

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -214,19 +274,37 @@ extern "C" RUNTIME_NOTHROW OBJ_GETTER(AdoptStablePointer, void* pointer) {
...
     if (!obj->local() && pointee != nullptr && pointee->local()) {
         konan::consolePrintf("Attempt to store a stack object %p into a heap object %p\n", pointee, obj);
         konan::consolePrintf("This is a compiler bug, please report it to https://kotl.in/issue\n");
         konan::abort();
     }
 }
 
+extern "C" ALWAYS_INLINE bool TryAddHeapRef(const ObjHeader* object) {
+    // TODO: Remove when legacy MM is gone.
+    return true;
```
SvyatoslavScherbina, line 295: I'm not sure this implementation is correct. OTOH, I guess we can't provide a proper implementation at the moment. Can we leave it with `TODO`?

runtime/src/mm/cpp/RefUpdates.cpp
```diff
@@ -0,0 +1,106 @@
...
+ALWAYS_INLINE void mm::SetHeapRefAtomic(ObjHeader** location, ObjHeader* value) noexcept {
+    __atomic_store_n(location, value, __ATOMIC_RELEASE);
+}
+
+ALWAYS_INLINE ObjHeader* mm::ReadHeapRefAtomic(ObjHeader** location) noexcept {
+    return __atomic_load_n(location, __ATOMIC_ACQUIRE);
+}
+
+ALWAYS_INLINE ObjHeader* mm::CompareAndSwapHeapRef(ObjHeader** location, ObjHeader* expected, ObjHeader* value) noexcept {
+    if (expected == value) return expected;
```
SvyatoslavScherbina, line 38: Is this correct?

runtime/src/mm/cpp/RefUpdates.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#pragma clang diagnostic pop
+
+ObjHeader* mm::InitThreadLocalSingleton(ThreadData* threadData, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*)) {
+    if (auto* value = *location) {
+        // Initialized by someone else.
+        return value;
+    }
+    auto* value = threadData->objectFactoryThreadQueue().CreateObject(typeInfo);
+    // This places `value` in the root set.
```
SvyatoslavScherbina, line 50: Seems safe-points-specific too.

runtime/src/mm/cpp/RefUpdates.cpp
```diff
@@ -0,0 +1,106 @@
...
+        ctor(object);
+    } catch (...) {
+        mm::SetStackRef(OBJ_RESULT, nullptr);
+        mm::SetHeapRefAtomic(location, nullptr);
+        initializingSingletons.pop_back();
+        throw;
+    }
+#endif
+    mm::SetHeapRefAtomic(location, object);
+    mm::GlobalsRegistry::Instance().RegisterStorageForGlobal(threadData, location);
```
SvyatoslavScherbina, line 103: Maybe swap `SetHeapRefAtomic` and `RegisterStorageForGlobal`?

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -125,23 +127,81 @@ extern "C" OBJ_GETTER(AllocArrayInstance, const TypeInfo* typeInfo, int32_t elem
...
-extern "C" OBJ_GETTER(InitSingleton, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*)) {
+extern "C" ALWAYS_INLINE OBJ_GETTER(InitThreadLocalSingleton, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*)) {
     auto* threadData = mm::ThreadRegistry::Instance().CurrentThreadData();
-    // TODO: This should only be called if singleton is actually created here. It's possible that the
-    // singleton will be created on a different thread and here we should check that, instead of creating
-    // another one (and registering `location` twice).
-    mm::GlobalsRegistry::Instance().RegisterStorageForGlobal(threadData, location);
-    TODO();
+
+    auto* object = mm::InitThreadLocalSingleton(threadData, location, typeInfo, ctor);
```
SvyatoslavScherbina, line 133: Safe-points-specific?

runtime/src/mm/cpp/RefUpdatesTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+        InitSingleton(&location, 0, &stackLocation);
+        ASSERT_TRUE(false); // Cannot be reached.
+    } catch (int exception) {
+        EXPECT_THAT(exception, kException);
+    }
+    EXPECT_THAT(stackLocation, nullptr);
+    EXPECT_THAT(location, nullptr);
+}
+
+TEST_F(InitSingletonTest, InitSingletonRecursive) {
```
SvyatoslavScherbina, line 154: This test looks totally cryptic to me. Why do we use two different locations for same singleton?

runtime/src/mm/cpp/RefUpdates.cpp
projedi: That was silly of me, fixed. Thanks!

runtime/src/mm/cpp/RefUpdatesTest.cpp
projedi: It's not the same singleton - it's two different singletons. They share the constructor function due to technicalities, but the location is what identifies them. Added some comments to the test.

runtime/src/mm/cpp/Memory.cpp
projedi: Do you mean it's specific to a particular safe point implementation, or to safe points in general?

SvyatoslavScherbina: I mean that it doesn't expect GC to happen at this particular place, so it is incompatible with GCs that can stop threads at arbitrary points.

projedi: Refactored a bit. Concerns about thread stopping mechanism should be limited to `ObjectOps` now. Also extracted initialization schemes into a different file, and they should now be oblivious to thread stopping technicalities.

runtime/src/mm/cpp/ObjectOps.cpp
```diff
@@ -0,0 +1,60 @@
...
+// atomic operations. TODO: Consider using alternative ways of ordering memory operations if they
+// turn out to be more efficient on these platforms.
+#pragma clang diagnostic ignored "-Watomic-alignment"
+
+ALWAYS_INLINE void mm::SetHeapRefAtomic(ObjHeader** location, ObjHeader* value) noexcept {
+    __atomic_store_n(location, value, __ATOMIC_RELEASE);
+}
+
+ALWAYS_INLINE OBJ_GETTER(mm::ReadHeapRefAtomic, ObjHeader** location) noexcept {
+    __atomic_load(location, OBJ_RESULT, __ATOMIC_ACQUIRE);
```
SvyatoslavScherbina, line 34: I'm not sure this `__atomic_load` is actually _atomic_.
I would expect it to atomically load from `location` and _then_ store to `OBJ_RESULT`, which doesn't exclude the possibility to observe the state between these two actions.

runtime/src/mm/cpp/ObjectOps.cpp
```diff
@@ -0,0 +1,60 @@
...
+    return *OBJ_RESULT;
+}
+
+ALWAYS_INLINE OBJ_GETTER(mm::CompareAndSwapHeapRef, ObjHeader** location, ObjHeader* expected, ObjHeader* value) noexcept {
+    mm::SetStackRef(OBJ_RESULT, expected);
+    // TODO: Do we need this strong memory model? Do we need to use strong CAS?
+    __atomic_compare_exchange_n(location, OBJ_RESULT, value, false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST);
```
SvyatoslavScherbina, line 41: Same here.

runtime/src/mm/cpp/ObjectOps.cpp
SvyatoslavScherbina: If this is true, then it makes some sense to keep the implementation simple (the previous version), with a comment like one in `AllocateObject`.

projedi: > I would expect it to atomically load from location and then store to OBJ_RESULT

Yep, you're right: https://github.com/llvm/llvm-project/blob/00530dee5d1295dc20ebafdd9a8a79662f41513e/compiler-rt/lib/builtins/atomic.c#L174

projedi: > it makes some sense to keep the implementation simple (the previous version)

I think it'll have to stay `OBJ_GETTER`, because the implementation supporting signals will have to have this signature.

SvyatoslavScherbina: Yes, of course.
I meant that `__atomic_load` here is somewhat confusing (because it is not atomic 🤷‍♂️).

So, if it really makes the implementation simpler and more understandable, it is ok to keep it but add a comment regarding its atomicity. If not, then replacing with real atomics makes a lot of sense.

Similar for `__atomic_compare_exchange_n` below.

projedi: > I meant that __atomic_load here is somewhat confusing (because it is not atomic 🤷‍♂️).
> So, if it really makes the implementation simpler and more understandable, it is ok to keep it but add a comment regarding its atomicity. If not, then replacing with real atomics makes a lot of sense.

Makes sense, will do.

> Similar for __atomic_compare_exchange_n below.

Here, however, I think the implementation is fine: we write `expected` onto a stack, so it's definitely kept in the rootset; if CAS succeeds, `actual` is still in the rootset (because it equals `expected`, and `expected` is sitting on a stack); if CAS fails, `actual` is kept at the `location` and (eventually) copied onto a stack. So, it all relies on `value` and `*location` being kept alive.

SvyatoslavScherbina: > if CAS fails, `actual` is kept at the `location` and (eventually) copied onto a stack.

And before it is copied, another thread might change the `*location`.
So I don't see how it is different from the load: if CAS fails, it acts exactly like load, and it is not atomic exactly the same way.

projedi: Missed that, thanks! I'm keeping `__atomic_*` for consistency, though (and because it'll allow to tweak strong/weak guarantees + memory ordering; and in the future I'm thinking of using something like https://en.cppreference.com/w/cpp/atomic/atomic_ref instead of intrinsics anyway)

SvyatoslavScherbina (approved):

runtime/src/mm/cpp/InitializationScheme.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Spin lock.
+    ObjHeader* value = nullptr;
+    while ((value = __sync_val_compare_and_swap(location, nullptr, initializing)) == initializing) {
+    }
+    if (value != nullptr) {
+        // Initialized by someone else.
+        RETURN_OBJ(value);
+    }
+    auto* object = mm::AllocateObject(threadData, typeInfo, OBJ_RESULT);
+    initializingSingletons.push_back(std::make_pair(location, object));
```
ilmat192, line 56: I have two general questions here:

1. Is this correct that in the ideal world the allocator behind `KStdVector` should trigger GC when there is no free memory to increase the vector capacity?
2. When do we zero reference fields of a newly allocated object? It seems that I missed something but I don't see such zeroing netiher in the `mm::AllocateObject` call chain nor in the bitcode of the generated ctor.

runtime/src/mm/cpp/InitializationScheme.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "InitializationScheme.hpp"
+
+#include "Common.h"
+#include "ObjectOps.hpp"
+#include "ThreadData.hpp"
+
+using namespace kotlin;
+
+OBJ_GETTER(mm::InitThreadLocalSingleton, ThreadData* threadData, ObjHeader** location, const TypeInfo* typeInfo, void (*ctor)(ObjHeader*)) {
```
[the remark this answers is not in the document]
projedi: TLS registers its storage with GC in one go https://github.com/JetBrains/kotlin-native/blob/3f2d7db5b7afda0a59a453977451aef1f908c62d/runtime/src/main/cpp/Runtime.cpp#L125 Globals register the storage one by one.

runtime/src/mm/cpp/InitializationScheme.cpp
projedi: > Is this correct that in the ideal world the allocator behind KStdVector should trigger GC when there is no free memory to increase the vector capacity?

I'm not sure. On one hand it makes sense, on the other hand it means that every memory allocation in runtime is also a safepoint, for which our runtime is certainly not ready. In the ideal world we should probably keep all our "system" data in a separate arena with enough preallocated memory, so that we (practically) never run into OOM with "system" allocations between GC sweeps.

projedi: > When do we zero reference fields of a newly allocated object? It seems that I missed something but I don't see such zeroing netiher in the mm::AllocateObject call chain nor in the bitcode of the generated ctor.

We allocate zeroed memory: we call `calloc`. But I should've written a test on it for the object allocation. Also, there's a sanitizer, that would've fired if we were to access uninitialized data - MSAN. Unfortunately, it requires that all dependencies are built with it (even libc), and so would require us building yet another sysroot.

ilmat192: > keep all our "system" data in a separate arena with enough preallocated memory, so that we (practically) never run into OOM with "system" allocations between GC sweeps.

Ok, you're probably right.

> We allocate zeroed memory: we call `calloc`

Thank you, I've just missed that `calloc` zeros memory by itself.

ilmat192 (approved):

PR #4657, opened 2021-01-28
Switch thread states in codegen

ilmat192:
- [x] Implement state switching in codegen for C calls and C callbacks.
 - [x] Mark trivial stdlib calls with @GCCritical/add manual state switches.
 - [x] Add safe point checks for external calls
 - [ ] Add manual state switching in functions from:
    - Cleaner.kt, workers/internal.kt, GC.kt, NumberConvertor.kt, WeakPrivate.kt
 - [ ] Add manual state switching for `@ExportToCppRuntime` functions (when needed)


ilmat192:
Closing in favor of branch in the main Kotlin repo: https://github.com/JetBrains/kotlin/tree/rrn/matveev%2Fexecution-states.

PR #4661, opened 2021-01-28
kt-36639

vvlevchenko:


backend.native/tests/produce_dynamic/kt-36639/main.kt
```diff
@@ -0,0 +1,21 @@
+interface Foo {
+    fun Any.extfoo()
```
SvyatoslavScherbina, line 2: Is this function accessible from C?
If yes, then this should be checked in the test too.

vvlevchenko: right, added calls to this function too.

SvyatoslavScherbina (approved):

PR #4663, opened 2021-01-29
Fix for KT-43502

sbogolepov:


SvyatoslavScherbina:
There are actually more places where we compile sources with clang. For example, `produceCAdapterBitcode` and runtime compilation.
Should these be fixed too? If yes, then we should either get a methodology that would allow us to find all such places, or add `-fPIC` to some common place like `ClangArgs`.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/ClangArgs.kt
```diff
@@ -67,6 +67,11 @@ class ClangArgs(private val configurables: Configurables) : Configurables by con
             }
 
         }
+        if (configurables !is MingwConfigurables) {
```
SvyatoslavScherbina, line 70: A brief comment explaining MinGW specifics would be valuable.

SvyatoslavScherbina (approved):

PR #4667, opened 2021-01-29
Implement relaxed mode in ObjC backend for mobile

fkistner:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubBuilder.kt
```diff
@@ -5,21 +5,32 @@
 
 package org.jetbrains.kotlin.backend.konan.objcexport
 
-internal class StubBuilder {
-    private val children = mutableListOf<Stub<*>>()
+internal class StubBuilder<S : Stub<*>>(private val relaxedMode: Boolean) {
```
SvyatoslavScherbina, line 8: "Relaxed mode" is kinda too vague (especially in classes that just pass this flag down).
Maybe something like "ignore exceptions" or "ignore errors"?
Introducing a enum for exception handling policy is also a good choice here.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
```diff
@@ -52,79 +52,88 @@ internal class ObjCExportTranslatorImpl(
...
 
-    override fun generateBaseDeclarations(): List<ObjCTopLevel<*>> {
-        val stubs = mutableListOf<ObjCTopLevel<*>>()
-
-        stubs.add(objCInterface(namer.kotlinAnyName, superClass = "NSObject", members = buildMembers {
-            +ObjCMethod(null, true, ObjCInstanceType, listOf("init"), emptyList(), listOf("unavailable"))
-            +ObjCMethod(null, false, ObjCInstanceType, listOf("new"), emptyList(), listOf("unavailable"))
-            +ObjCMethod(null, false, ObjCVoidType, listOf("initialize"), emptyList(), listOf("objc_requires_super"))
-        }))
+    override fun generateBaseDeclarations(): List<ObjCTopLevel<*>> = buildTopLevel(relaxedMode) {
```
SvyatoslavScherbina, line 61: Adding wrappers (for `buildTopLevel` and `buildMembers`) that don't accept `relaxedMode` as a parameter value but obtain it from property instead would simplify the code and help to avoid affecting places like this by changes to the configuration.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/StubBuilder.kt
fkistner: Reworked the PR. As discussed, I suggest we extend the existing warning collector functionality to include handling of exceptions as well.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/objcexport/ObjCExportHeaderGenerator.kt
fkistner: ✅

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanCompilerFrontendServices.kt
```diff
@@ -19,7 +19,7 @@ internal fun StorageComponentContainer.initContainer(config: KonanConfig) {
 
     if (config.configuration.get(KonanConfigKeys.EMIT_LAZY_OBJC_HEADER_FILE) != null) {
         this.useImpl<ObjCExportLazyImpl>()
-        this.useInstance(ObjCExportWarningCollector.SILENT)
+        this.useInstance(ObjCExportProblemCollector.SILENT)
```
fkistner, line 22: Note that there is a semantic change here:
From now, `-Xemit-lazy-objc-header` would silently ignore exceptions. This is inline with silently ignoring warnings. Please advise, in case we should still throw exceptions here, while ignoring warnings, or in case it makes sense to report all problems similar to `ObjCExportHeaderGeneratorImpl#ProblemCollector`.

SvyatoslavScherbina: It makes a lot of sense to make `-Xemit-lazy-objc-header` throw exceptions instead of silently ignoring them. Motivation: `-Xemit-lazy-objc-header` exists only to test `ObjCExportLazy` implementation, it is not an actual feature of the compiler. If we ignore the exceptions in tests, we will probably break something in `ObjCExportLazy` without noticing this.

fkistner: Can you please clarify whether you would like warnings to continue being ignored?

SvyatoslavScherbina: Yes, warning should continue being ignored. So `-Xemit-lazy-objc-header` behaviour shouldn't change.

SvyatoslavScherbina (approved):

PR #4682, opened 2021-02-03
Sketch GC interface

projedi:


runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -250,25 +251,74 @@ class ObjectFactoryStorage : private Pinned {
...
+            auto* object = static_cast<ObjHeader*>(node.Data());
+            object->typeInfoOrMeta_ = const_cast<TypeInfo*>(typeInfo);
+            return object;
+        }
 
-        ObjHeader* CreateObject(const TypeInfo* typeInfo) noexcept;
-        ArrayHeader* CreateArray(const TypeInfo* typeInfo, uint32_t count) noexcept;
+        ArrayHeader* CreateArray(const TypeInfo* typeInfo, uint32_t count) noexcept {
+            RuntimeAssert(typeInfo->IsArray(), "Must be an array");
+            uint32_t arraySize = static_cast<uint32_t>(-typeInfo->instanceSize_) * count;
```
RUSshy, line 306: ``-typeInfo->instanceSize_`` why negative?

projedi: It's negative for the arrays: https://github.com/JetBrains/kotlin-native/blob/bb397ed8cc31f8f755ccaef93aa0f92ca61296d6/runtime/src/main/cpp/TypeInfo.h#L112

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -81,15 +78,18 @@ class ObjectFactoryStorage : private Pinned {
         // with C++ members.
     };
 
+    template <typename Allocator>
     class Producer : private MoveOnly {
     public:
+        explicit Producer(ObjectFactoryStorage& owner, Allocator&& allocator) noexcept : owner_(owner), allocator_(std::move(allocator)) {}
+
```
SvyatoslavScherbina, line 85: Do we need the constructor below then?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -59,19 +60,15 @@ class ObjectFactoryStorage : private Pinned {
...
             RuntimeAssert(
                     DataOffset() + dataSize <= totalSize, "totalSize %zu is not enough to fit data %zu at offset %zu", totalSize, dataSize,
                     DataOffset());
-            void* ptr = konanAllocAlignedMemory(totalSize, totalAlignment);
-            if (!ptr) {
-                // TODO: Try doing GC first.
-                konan::consoleErrorf("Out of memory trying to allocate %zu. Aborting.\n", totalSize);
-                konan::abort();
-            }
+            void* ptr = allocator.Alloc(totalSize, totalAlignment);
```
SvyatoslavScherbina, line 71: So the memory is allocated using `allocator.Alloc`, but to be deallocated with `KStdUniquePtr` delegating to `konanDestructInstance`? Looks like different levels of abstraction mixed.

runtime/src/mm/cpp/gc/LeakGC.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <cstddef>
+
+#include "Utils.hpp"
+
+namespace kotlin {
+namespace mm {
+
+// LeakGC (aka no-op GC) is a GC that does not free memory.
+// TODO: It can be made more efficient.
+class LeakGC : private Pinned {
```
SvyatoslavScherbina, line 18: `NoGC` or `NoOpGC` or similar kinda look more understandable than `LeakGC`.

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -366,7 +375,22 @@ extern "C" void AdoptReferenceFromSharedVariable(ObjHeader* object) {
...
 }
 
-void CheckGlobalsAccessible() {
+extern "C" void CheckGlobalsAccessible() {
     // TODO: Remove when legacy MM is gone.
     // Always accessible
 }
+
+extern "C" RUNTIME_NOTHROW void Kotlin_mm_safePointFunctionEpilogue() {
+    auto* threadData = mm::ThreadRegistry::Instance().CurrentThreadData();
```
SvyatoslavScherbina, line 384: I guess eventually it would make a lot of sense (for performance) to take `threadData` as a parameter.
For example, almost each non-trivial function would have at least a couple of runtime calls requiring thread data -- `EnterFrame`, `LeaveFrame`, `Kotlin_mm_safePointFunctionEpilogue`.

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -245,35 +254,104 @@ class ObjectFactoryStorage : private Pinned {
...
+};
+
+template <typename BaseAllocator, typename GC>
+class AllocatorWithGC {
+public:
+    AllocatorWithGC(BaseAllocator base, GC& gc) noexcept : base_(std::move(base)), gc_(gc) {}
+
+    void* Alloc(size_t size, size_t alignment) noexcept {
+        gc_.SafePointAllocation(size);
+        void* ptr = base_.Alloc(size, alignment);
```
SvyatoslavScherbina, line 283: Both `AllocatorWithGC` and `BaseAllocator` use the same name for `Alloc` methods with clearly different semantics -- the former aborts if can't alloc, while the latter returns `nullptr` in this case.
It makes some sense to distinguish these methods (using different names or whatever).

runtime/src/mm/cpp/Memory.cpp
projedi: I think so too. However, it might turn out that we'll have to split `ThreadData` into parts and pass them separately (or acquire them via TLS) to avoid an extra indirection from dereferencing `ThreadData`.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Good point, thanks! Moved panicking back into `ObjectFactoryStorage`.

SvyatoslavScherbina (approved):

runtime/src/mm/cpp/ObjectFactory.cpp
```diff
@@ -1,61 +0,0 @@
...
-#include "ObjectFactory.hpp"
-
-#include "Alignment.hpp"
-#include "Alloc.h"
-#include "GlobalData.hpp"
-#include "Types.h"
-
-using namespace kotlin;
-
-ObjHeader* mm::ObjectFactory::ThreadQueue::CreateObject(const TypeInfo* typeInfo) noexcept {
```
ilmat192, line 15: Why do we need these methods in a header instead of a cpp? Due to performance reasons?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -245,35 +265,155 @@ class ObjectFactoryStorage : private Pinned {
...
+
+        static NodeRef From(ObjHeader* object) noexcept {
+            RuntimeAssert(object->heap(), "Must be a heap object");
+            auto* heapObject = reinterpret_cast<HeapObjHeader*>(reinterpret_cast<uintptr_t>(object) - offsetof(HeapObjHeader, object));
+            RuntimeAssert(&heapObject->object == object, "HeapObjHeader layout has broken");
+            return NodeRef(Storage::Node::FromData(heapObject));
+        }
+
+        static NodeRef From(ArrayHeader* array) noexcept {
+            // `ArrayHeader` and `ObjHeader` are kept compatible, so the former can
```
ilmat192, line 336: It seems that `HeapObjHeader` and `HeapArrayHeader` also must be compatible for this method and the methods below to work. May be add a corresponding comment to the declaration of `HeapObjHeader` and `HeapArrayHeader`?

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -245,35 +265,155 @@ class ObjectFactoryStorage : private Pinned {
...
+            // be always casted to the other.
+            return From(reinterpret_cast<ObjHeader*>(array));
+        }
+
+        NodeRef* operator->() noexcept { return this; }
+
+        GCObjectData& GCObjectData() noexcept {
+            // `ArrayHeader` and `ObjHeader` are kept compatible, so the former can
```
ilmat192, line 344: Shouldn't they be heap headers here and below?

```suggestion
            // `HeapArrayHeader` and `HeapObjHeader` are kept compatible, so the former can
```

runtime/src/mm/cpp/ObjectFactory.hpp
```diff
@@ -245,35 +265,155 @@ class ObjectFactoryStorage : private Pinned {
...
+            // be always casted to the other.
+            return static_cast<HeapObjHeader*>(node_.Data())->gcData;
+        }
+
+        bool IsArray() const noexcept {
+            // `ArrayHeader` and `ObjHeader` are kept compatible, so the former can
```
ilmat192, line 350:
```suggestion
            // `HeapArrayHeader` and `HeapObjHeader` are kept compatible, so the former can
```

runtime/src/mm/cpp/ObjectFactory.cpp
projedi: The entire `ObjectFactory` class has become a template (with a `typename GC` parameter), so all the definitions need to become visible.

runtime/src/mm/cpp/ObjectFactory.hpp
projedi: Good point, thanks!

runtime/src/mm/cpp/ObjectFactory.cpp
ilmat192: I see, thanks!

ilmat192 (approved):

PR #4685, opened 2021-02-05
Add RootSet to unify iterating across the entire rootset

projedi:


runtime/src/mm/cpp/ThreadData.hpp
```diff
@@ -59,6 +59,13 @@ class ThreadData final : private Pinned {
 
     GC::ThreadData& gc() noexcept { return gc_; }
 
+    void Publish() noexcept {
```
projedi, line 62: This is not the most fitting place for the method. However, it has high visibility when adding new per-thread collections, that need to be published into their global counterparts.

ilmat192 (approved):

PR #4691, opened 2021-02-09
[ABI Compatibility] Check that given path is actually a Kotlin/Native distribution

sbogolepov:


build-tools/src/main/kotlin/org/jetbrains/kotlin/CompareDistributionSignatures.kt
```diff
@@ -137,4 +161,12 @@ open class CompareDistributionSignatures : DefaultTask() {
                 oldKlibSignatures - newKlibSignatures,
         )
     }
+
+    private fun isLooksLikeKotlinNativeDistribution(directory: Path): Boolean {
```
SvyatoslavScherbina, line 165: Minor: is this a correct phrase in Engish? Maybe `looksLikeKotlinNativeDistribution`?

SvyatoslavScherbina (approved):

PR #4696, opened 2021-02-10
[IR] Fix for https://youtrack.jetbrains.com/issue/KT-44799

homuroll:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -282,7 +282,8 @@ internal class FunctionReferenceLowering(val context: Context): FileLoweringPass
                 else
                     kFunctionImplSymbol.typeWith(referencedFunction.returnType)
                 functionClass = (if (isKFunction) symbols.kFunctionN(numberOfParameters) else symbols.functionN(numberOfParameters)).owner
-                superTypes += functionClass.typeWith(functionParameterTypes + referencedFunction.returnType)
+                if (samSuperType == null)
```
SvyatoslavScherbina, line 285: invoke method generation below probably needs an update too.
I guess current version generates a bogus `overriddenSymbols` entry now, which is inconsistent with actual super types.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -292,19 +293,20 @@ internal class FunctionReferenceLowering(val context: Context): FileLoweringPass
...
+                        superTypes += suspendFunctionClass.symbol.typeWith(suspendFunctionClassTypeParameters)
                 }
             }
 
             val constructor = buildConstructor()
             val functionInvoke = if (isKSuspendFunction)
                 null
             else
-                buildInvokeMethod(functionClass.getInvokeFunction())
+                buildInvokeMethod(functionClass.getInvokeFunction(), putToOverriddenSymbols = samSuperType == null)
```
SvyatoslavScherbina, line 305: And now this probably just builds redundant method (e.g. if SAM has extension receiver).

I guess this code is quite complicated, so optimizing it would make it even worse.
On the other hand, building a specialized implementation for SAM might help to make it more precise and simplify the implementation of both (SAM and regular function references/lambdas). WDYT?

homuroll: >building a specialized implementation for SAM might help to make it more precise and simplify the implementation of both
Makes sense, I don't remember why I hadn't implemented it that way in the first place.

SvyatoslavScherbina (approved):
Let's also add a test for the original ObjCExport problem.
I've made a commit: https://github.com/JetBrains/kotlin-native/commit/d344f8a770af066e568ecacef9bb54c37347f19b
The test crashes without this fix and passes with it.
It makes some sense to merge it as a part of this PR :)

PR #4699, opened 2021-02-12
Misc changes for GC in new MM

projedi:


backend.native/tests/build.gradle
```diff
@@ -2396,6 +2396,11 @@ task hash_map0(type: KonanLocalTest) {
     source = "runtime/collections/hash_map0.kt"
 }
 
+task hash_map1(type: KonanLocalTest) {
+    goldValue = "OK\n"
+    source = "runtime/collections/hash_map1.kt"
```
projedi, line 2401: Extracted long tests here.

backend.native/tests/runtime/collections/hash_map1.kt
```diff
@@ -0,0 +1,100 @@
...
+        for (i in 1..n) {
+            assertFalse(m.containsKey(i.toString()))
+        }
+    }
+}
+
+@Test fun runTest() {
+    // TODO: Do not manually control this.
+    if (Platform.memoryModel == MemoryModel.EXPERIMENTAL) {
+        GC.threshold = 1000000
```
projedi, line 94: In these tests there're a lot of safepoints, so until we have a proper GC scheduling mechanism, we have to manually tune thresholds here.

runtime/src/mm/cpp/GlobalData.hpp
```diff
@@ -29,8 +29,9 @@ class GlobalData : private Pinned {
 
 private:
     GlobalData();
-    ~GlobalData();
+    ~GlobalData() = delete;
 
+    // This `GlobalData` is never destroyed.
```
projedi, line 34: With the new MM we don't shutdown the runtime at the program finish, so no need to destroy `GlobalData` too.

runtime/src/mm/cpp/InitializationSchemeTest.cpp
```diff
@@ -39,6 +39,7 @@ class InitSingletonTest : public testing::Test {
         // Make sure to clean everything allocated by the tests.
         for (auto& threadData : threadDatas_) {
             threadData->objectFactoryThreadQueue().ClearForTests();
+            threadData->globalsThreadQueue().ClearForTests();
```
projedi, line 42: `globalsThreadQueue()` will publish into a global (inside `GlobalData`) when `threadData` is destroyed, so we need to clean them up before it happens. In the future I'll look into decoupling tests from accessing globals.

runtime/src/mm/cpp/Memory.cpp
```diff
@@ -263,6 +263,48 @@ extern "C" void Kotlin_native_internal_GC_collectCyclic(ObjHeader*) {
     ThrowIllegalArgumentException();
 }
 
+extern "C" void Kotlin_native_internal_GC_setThreshold(ObjHeader*, int32_t value) {
```
projedi, line 266: Reusing existing `kotlin.native.internal.GC` API for tuning GC. It's a temporary solution.

ilmat192 (approved):

PR #4700, opened 2021-02-12
Do not generate safepoints for initializer functions

projedi:


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
```diff
@@ -1221,13 +1228,15 @@ internal class FunctionGenerationContext(val function: LLVMValueRef,
     internal fun epilogue() {
         appendingTo(prologueBb) {
             if (needsRuntimeInit) {
+                assert(!forbidRuntime)
```
SvyatoslavScherbina, line 1231: Please use `require` or `check` instead of `assert`. Assertions can be disabled and somewhat don't match the semantics required here (~precondition).

SvyatoslavScherbina: (same for the other new assertions)

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/llvm/CodeGenerator.kt
projedi: Right, I forgot those existed, thanks!

ilmat192 (approved):

PR #4707, opened 2021-02-17
[IR] Erase non-trivial type projections off of super types

homuroll:
This fixes https://youtrack.jetbrains.com/issue/KT-44826


backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
```diff
@@ -89,6 +92,20 @@ internal class FunctionReferenceLowering(val context: Context): FileLoweringPass
                 return result
             }
 
+            // TODO: Move to common IR utils.
+            fun IrType.eraseProjections(): IrType {
+                if (this !is IrSimpleType) return this
+                return buildSimpleType {
```
SvyatoslavScherbina, line 98: Shouldn't we set `hasQuestionMark` too?

SvyatoslavScherbina (approved):

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/FunctionReferenceLowering.kt
homuroll: Indeed.

SvyatoslavScherbina (approved):

PR #4708, opened 2021-02-17
Optimize CPointer.toKStringFromUtf8 on Native

SvyatoslavScherbina:
Use strlen and stop allocating intermediate ByteArray
to make the performance comparable to ByteArray.decodeToString().

Partially fix KT-44357.

### DIFF
--- a/backend.native/tests/interop/basics/ctoKString.def
+++ b/backend.native/tests/interop/basics/ctoKString.def
@@ -1,0 +1,6 @@
+---
+const char* empty() { return ""; }
+const char* foo() { return "foo"; }
+const char* kuku() { return "куку"; }
+const char* invalid_utf8() { return "\x85\xAF"; }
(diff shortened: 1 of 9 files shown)


backend.native/tests/interop/basics/ctoKString.def
```diff
@@ -0,0 +1,5 @@
+---
+const char* empty() { return ""; }
+const char* foo() { return "foo"; }
+const char* kuku() { return "куку"; }
+const char* invalid_utf8() { return "\x85\xAF"; }
```
projedi, line 5: Maybe also a test with a string that has a `\0` in the middle?

projedi (approved):

backend.native/tests/interop/basics/ctoKString.def
SvyatoslavScherbina: Nice idea, thanks! Done.

projedi (approved):

PR #4710, opened 2021-02-18
[IR][cache] Fix for https://youtrack.jetbrains.com/issue/KT-44764

homuroll:
The .konanLibrary is cached for each declaration for performance purposes, but it is only ok to do it
after IR has been lowered to its final state (or at least when declarations aren't being moved around),
so the fix respects that by only taking .konanLibrary of a IrFile (it is assumed that files stay on their place
during entire backend lowering procedure).

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LlvmModuleSpecification.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/LlvmModuleSpecification.kt
@@ -7,7 +7,9 @@ package org.jetbrains.kotlin.backend.konan
 
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.ir.declarations.IrDeclaration
+import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
+import org.jetbrains.kotlin.ir.declarations.IrPackageFragment
 import org.jetbrains.kotlin.library.KotlinLibrary
 
 /**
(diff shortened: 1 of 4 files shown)


SvyatoslavScherbina (approved):

backend.native/tests/build.gradle
```diff
@@ -3726,6 +3748,11 @@ linkTest("inline_inlineCtor_linkTest") {
     lib = "codegen/inline/inlineCtor_linkTest_lib.kt"
 }
 
+linkTest("inline_kt44764") {
+    source = "codegen/inline/kt44764_main.kt"
+    lib = "codegen/inline/kt44764_lib.kt"
```
SvyatoslavScherbina, line 3753: Minor: since the test prints a value, it makes some sense to check it here (with `goldValue`).

backend.native/tests/build.gradle
```diff
@@ -402,13 +402,35 @@ Task linkTest(String name, Closure<KonanLinkTest> configureClosure) {
...
+                def dist = project.kotlinNativeDist
+                def stdlib = "$dist/klib/common/stdlib"
+                def cacheDir = project.file("${project.buildDir}/cache")
+                def stdlibCache = "$cacheDir/stdlib-cache"
+                def libPath = konanArtifacts[lib].getArtifactByTarget(targetName).absolutePath
+                def libCache = "${cacheDir.absolutePath}/${name}-cache"
+                def cacheKind = project.findProperty("test_with_cache_kind")
                 // Build an executable with library
                 program(name, targets: [targetName]) {
+                    doFirst {
```
SvyatoslavScherbina, line 414: Is it possible that this `doFirst` is applied to the task, not to the `program`?
In this case it will be called too late.
That also might explain why the added test failes with `No cache found for library` for me.

Also a review of this commit from @PavelPunegov might be useful.

homuroll: Yes, that's true, the test is unstable now.

PR #4711, opened 2021-02-18
Fix -Xoverride-konan-properties for Apple strip

idstein:
When building an optimised release dynamic framework we see a lot of local symbols in Mach-O. When calling `strip -x` this can be dramatically reduced whilst still having a usable artefact.

Current logic seems to keep symbols for DSYM extraction in debug light mode. But only strips hardcoded debug symbols (`strip -S`) afterwards.


shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
```diff
@@ -265,7 +265,7 @@ class MacOSBasedLinker(targetProperties: AppleConfigurables)
         if (debug) {
             result += dsymUtilCommand(executable, outputDsymBundle)
             if (optimize) {
-                result += Command(strip, "-S", executable)
+                result += Command(strip, linkerNoDebugFlags, executable)
```
sbogolepov, line 268: I'm not sure that sharing the same set of flags between `ld` and `strip` is a good idea. For example, changing `linkerNoDebugFlags` with something `ld`-specific will break compilation.
A better approach would be an addition of a separate `stripFlags` property. [This commit](https://github.com/JetBrains/kotlin-native/commit/6137389c757ec8ca5298b533d1044e4955d1adc2) will likely help you. There we add a single `llvmInlineThreshold` that is shared across all targets.

sbogolepov (changes requested):

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Linker.kt
idstein: I absolutely agree and will adjust it accordingly by introducing a new konan parameter to tune it.

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Configurables.kt
```diff
@@ -98,6 +98,7 @@ interface AppleConfigurables : Configurables, ClangFlags {
     val arch get() = targetString("arch")!!
     val osVersionMin get() = targetString("osVersionMin")!!
     val osVersionMinFlagLd get() = targetString("osVersionMinFlagLd")!!
+    val stripFlags get() = targetString("stripFlags")!!
```
sbogolepov, line 101: Generally everything is OK. The last two things:
1. Use `targetList` instead of `targetString`. Otherwise things might break in case of several flags. Note that `Command` construction should also be changed to something like `Command(strip, *stripFlags.toTypedArray(), executable)`.
2. Squash your commits, please :)

sbogolepov (changes requested):

shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/Configurables.kt
idstein: Thanks a lot for the feedback ;) It was a bit in hurray. But you're absolute right about those two points.

sbogolepov (approved):
Thanks!

PR #4714, opened 2021-02-19
Update runtimeDefinitions after toolchain upgrade

projedi:
Addressing

```cpp
// Some libstdc++-based targets has limited support for std::current_exception and other C++11 functions.
// This restriction can be lifted later when toolchains will be updated.
```
https://github.com/JetBrains/kotlin-native/blob/ce946a9679087e6c72ad0d2c076dfa1f3cfc062c/runtime/src/main/cpp/Exceptions.cpp#L255

### DIFF
--- a/konan/konan.properties
+++ b/konan/konan.properties
@@ -510,7 +510,8 @@ abiSpecificLibraries.linux_arm32_hfp = lib usr/lib
 # targetSysRoot relative
 crtFilesLocation.linux_arm32_hfp = usr/lib
 runtimeDefinitions.linux_arm32_hfp = USE_GCC_UNWIND=1 KONAN_LINUX=1 \
-  KONAN_ARM32=1 USE_ELF_SYMBOLS=1 ELFSIZE=32 KONAN_NO_UNALIGNED_ACCESS=1
+  KONAN_ARM32=1 USE_ELF_SYMBOLS=1 ELFSIZE=32 KONAN_NO_UNALIGNED_ACCESS=1 \
+  KONAN_HAS_CXX11_EXCEPTION_FUNCTIONS=1
 
 # Linux arm64
 gccToolchain.linux_arm64 = aarch64-unknown-linux-gnu-gcc-8.3.0-glibc-2.25-kernel-4.9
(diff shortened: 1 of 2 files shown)


konan/konan.properties
```diff
@@ -672,7 +675,7 @@ clangNooptFlags.android_arm32 = -O1
 targetSysRoot.android_arm32 = target-sysroot-1-android_ndk
 linkerKonanFlags.android_arm32 = -lm -lc++_static -lc++abi -landroid -llog -latomic
 runtimeDefinitions.android_arm32 = __ANDROID__ USE_GCC_UNWIND=1 USE_ELF_SYMBOLS=1 ELFSIZE=32 \
-  KONAN_ANDROID=1 KONAN_ARM32=1 KONAN_NO_UNALIGNED_ACCESS=1
+  KONAN_ANDROID=1 KONAN_ARM32=1 KONAN_NO_UNALIGNED_ACCESS=1 KONAN_HAS_CXX11_EXCEPTION_FUNCTIONS=1
```
projedi, line 678: Not sure about this one.

konan/konan.properties
```diff
@@ -510,7 +510,8 @@ abiSpecificLibraries.linux_arm32_hfp = lib usr/lib
 # targetSysRoot relative
 crtFilesLocation.linux_arm32_hfp = usr/lib
 runtimeDefinitions.linux_arm32_hfp = USE_GCC_UNWIND=1 KONAN_LINUX=1 \
-  KONAN_ARM32=1 USE_ELF_SYMBOLS=1 ELFSIZE=32 KONAN_NO_UNALIGNED_ACCESS=1
+  KONAN_ARM32=1 USE_ELF_SYMBOLS=1 ELFSIZE=32 KONAN_NO_UNALIGNED_ACCESS=1 \
+  KONAN_HAS_CXX11_EXCEPTION_FUNCTIONS=1
```
projedi, line 514: Maybe it's more prudent to use a reverse property

sbogolepov: Or just get rid of this obsolete property at all.

sbogolepov (approved):

PR #4715, opened 2021-02-19
Remove .git directory when downloading googletest sources

ilmat192:
Keeping the .git folder in the googletest directory may cause
changing .idea/vcs.xml during IDEA import. .idea/vcs.xml is under
source control, thus changing it litters a current diff.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -16,6 +16,5 @@
   </component>
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
-    <mapping directory="$PROJECT_DIR$/runtime/googletest" vcs="Git" />
   </component>
 </project>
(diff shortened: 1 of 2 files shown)


projedi (approved):

vvlevchenko (approved):

vvlevchenko (commented):
maybe git archive better solution here?

ilmat192:
> maybe git archive better solution here?

It seems interesting but I haven't managed to make it work with github repos:

```bash
$ git archive -o googletest.zip -0 --remote=https://github.com/google/googletest.git master
fatal: operation not supported by protocol
```

```bash
$ git archive -o googletest.zip -0  --remote=git@github.com:google/googletest.git master
Invalid command: 'git-upload-archive 'google/googletest.git''
  You appear to be using ssh to clone a git:// URL.
  Make sure your core.gitProxy config option and the
  GIT_PROXY_COMMAND environment variable are NOT set.
fatal: the remote end hung up unexpectedly
```

It [seems](https://stackoverflow.com/a/43079188/9422800) that GitHub doesn't support `git archive`. Discussed offline and decided to merge the change as is.

PR #4723, opened 2021-03-01
Extract finalizer hooks

projedi:


runtime/src/main/cpp/FinalizerHooksTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+    type.flags_ |= TF_HAS_FINALIZER;
+    ObjHeader obj = {&type};
+    ASSERT_FALSE(obj.has_meta_object());
+
+    EXPECT_TRUE(HasFinalizers(&obj));
+    EXPECT_CALL(finalizerHook(), Call(&obj));
+    RunFinalizers(&obj);
+    EXPECT_FALSE(obj.has_meta_object());
+
+    testing::Mock::VerifyAndClearExpectations(&finalizerHook());
```
SvyatoslavScherbina, line 42: Btw, this looks kinda fragile. What happens if one forgets to add this line?
Can we make this implicitly?

SvyatoslavScherbina (approved):

runtime/src/main/cpp/FinalizerHooksTest.cpp
projedi: It's not really needed: the verification should happen automatically during mock's destructor. I'm placing them by the force of habbit.

SvyatoslavScherbina: Well, then this might be confusing, and also probably will be copied by other developers to new tests.
So if this happens automatically at the end of the test, then it makes some sense to remove these lines.

projedi: Good point, cleaned them up.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/FinalizerHooksTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+using ::testing::_;
+
+namespace {
+
+class FinalizerHooksTest : public testing::Test {
+public:
+    testing::MockFunction<void(ObjHeader*)>& finalizerHook() { return finalizerHooks_.finalizerHook(); }
+
+private:
+    FinalizerHooksTestSupport finalizerHooks_;
```
ilmat192, line 25: (minor) Why not include the logic from `FinalizerHooksTestSupport` directly into `FinalizerHooksTest`? Will we need this logic in other tests?

ilmat192 (approved):

runtime/src/main/cpp/FinalizerHooksTest.cpp
projedi: > Will we need this logic in other tests?

Yep, I think tests on GC might need that.

PR #4724, opened 2021-03-01
Extract object traversals

projedi:


runtime/src/main/cpp/ObjectTraversal.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "Memory.h"
+#include "Natives.h"
+#include "Types.h"
+
+namespace kotlin {
+
+// TODO: Consider an iterator/ranges based approaches for traversals.
+
+template <typename F>
+void traverseObjectFields(ObjHeader* object, F process) noexcept(std::is_nothrow_invocable_v<F, ObjHeader**>) {
```
projedi, line 21: So, apparently our wasm toolchain is not 17 enough. I don't really want to implement the entire `std::is_nothrow_invocable`, so I'll use a simpler check here.

SvyatoslavScherbina (approved):

runtime/src/main/cpp/ObjectTraversal.hpp
```diff
@@ -0,0 +1,46 @@
...
+        ArrayHeader* array = object->array();
+        for (uint32_t index = 0; index < array->count_; index++) {
+            process(ArrayAddressOfElementAt(array, index));
+        }
+    }
+}
+
+template <typename F>
+void traverseReferredObjects(ObjHeader* object, F process) noexcept(noexcept(process(std::declval<ObjHeader*>()))) {
+    traverseObjectFields(object, [&process](ObjHeader** location) noexcept(noexcept(process(std::declval<ObjHeader*>()))) {
```
SvyatoslavScherbina, line 37: Is `noexcept(noexcept(` here a typo?

projedi: Unfortunately, not.
The first one is a [`noexcept` specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec) marking the function conditionally `noexcept`
The second on is a [`noexcept` operator](https://en.cppreference.com/w/cpp/language/noexcept) – a compile-time check returning true iff the given expression is non-throwing.

SvyatoslavScherbina: Ah, I see. Ok.

runtime/src/main/cpp/ObjectTraversalTest.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+} // namespace
+
+TEST(ObjectTraversalTest, TraverseFieldsExceptions) {
+    static_assert(
+            noexcept(traverseObjectFields(std::declval<ObjHeader*>(), std::declval<CallableWithoutExceptions>())),
+            "Callable is noexcept, so traverse is noexcept");
+    static_assert(
+            !noexcept(traverseObjectFields(std::declval<ObjHeader*>(), std::declval<CallableWithExceptions>())),
+            "Callabl is noexcept(false), so traverse is noexcept(false)");
```
ilmat192, line 79: Typo:
```suggestion
            "Callable is noexcept(false), so traverse is noexcept(false)");
```

runtime/src/main/cpp/ObjectTraversalTest.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+TEST(ObjectTraversalTest, TraverseRefsExceptions) {
+    static_assert(
+            noexcept(traverseReferredObjects(std::declval<ObjHeader*>(), std::declval<CallableWithoutExceptions>())),
+            "Callable is noexcept, so traverse is noexcept");
+    static_assert(
+            !noexcept(traverseReferredObjects(std::declval<ObjHeader*>(), std::declval<CallableWithExceptions>())),
+            "Callabl is noexcept(false), so traverse is noexcept(false)");
```
ilmat192, line 180:
```suggestion
            "Callable is noexcept(false), so traverse is noexcept(false)");
```

ilmat192 (approved):

PR #4728, opened 2021-03-02
Thread state switching preparations

ilmat192:


runtime/src/main/cpp/Runtime.cpp
```diff
@@ -371,3 +371,17 @@ void Kotlin_Debugging_setForceCheckedShutdown(KBoolean value) {
 }
 
 }  // extern "C"
+
+MemoryState* kotlin::mm::GetMemoryState() {
```
projedi, line 375: Maybe implement it in each memory model instead? It'll require changing `thread_local` in `ThreadRegistry.hpp` to `Node*`, but I don't think it's a problem.

runtime/src/main/cpp/TestSupport.hpp
```diff
@@ -15,4 +21,46 @@ constexpr int kDefaultThreadCount = 10;
...
+// Deinitiliazes memory subsystem used for C++ stdlib tests.
+//  - For the new MM: unregisteres the current thread, nullify current thread data.
+//  - For the legacy MM: does nothing.
+void DeinitMemoryForTests(MemoryState* state);
+
+// Scopely initializes the runtime for tests.
+class ScopedRuntimeInit : private kotlin::Pinned {
+public:
+    ScopedRuntimeInit() : memoryState_(InitMemoryForTests()) {
+        test_support::InitRuntimeForTests(memoryState());
```
projedi, line 39: If we move `GetMemoryState` into each mm, could we use real `InitMemory` and `DeinitMemory` instead?

runtime/src/main/cpp/Memory.h
```diff
@@ -369,4 +369,66 @@ class ExceptionObjHolder {
...
+        SwitchThreadState(thread_, oldState_);
+    }
+private:
+    MemoryState* thread_;
+    ThreadState oldState_;
+};
+
+// Calls the given function in the `Runnable` thread state.
+template <typename... Args, typename R>
+ALWAYS_INLINE inline R CallKotlin(R(*kotlinFunction)(Args...), Args... args) {
```
projedi, line 419: Let's use perfect forwarding just in case?
```cpp
template <typename R, typename... Args>
ALWAYS_INLINE inline R CallKotlin(R(*kotlinFunction)(Args...), Args... args) {
    ThreadStateGuard guard(ThreadState::kRunnable);
    return kotlinFunction(std::forward<Args>(args)...);
}
```

SvyatoslavScherbina (approved):

runtime/src/main/cpp/TestSupport.hpp
```diff
@@ -15,4 +21,46 @@ constexpr int kDefaultThreadCount = 10;
...
+        DeinitMemoryForTests(memoryState_);
+    }
+
+    MemoryState* memoryState() { return memoryState_; }
+private:
+    MemoryState* memoryState_;
+};
+
+// Runs the given function in a separate thread with minimally initialized runtime.
+inline void RunInNewThread(std::function<void(MemoryState*)> f) {
```
projedi, line 52: Do we need `std::function` here instead of a usual template? If it's for the overload below, we can just ignore the argument at each use site.

runtime/src/mm/cpp/ThreadState.cpp
```diff
@@ -35,26 +34,26 @@ ALWAYS_INLINE mm::ThreadState mm::SwitchThreadState(ThreadData* threadData, Thre
...
-    oldState_ = SwitchThreadState(threadData, state);
-}
-
-mm::ThreadStateGuard::~ThreadStateGuard() noexcept {
-    SwitchThreadState(threadData_, oldState_);
+ALWAYS_INLINE void kotlin::AssertThreadState(MemoryState* thread, ThreadState expected) noexcept {
+    AssertThreadState(thread->GetThreadData(), expected);
 }
 
 extern "C" ALWAYS_INLINE RUNTIME_NOTHROW void Kotlin_mm_switchThreadStateNative() {
```
projedi, line 52: What do you think about moving these two inside `Memory.cpp`? That's the place riddled with `extern "C"` stuff anyway, and pretty much serving as the API surface for the module.

runtime/src/main/cpp/Memory.h
```diff
@@ -369,4 +369,66 @@ class ExceptionObjHolder {
...
+    ThreadStateGuard guard(ThreadState::kRunnable);
+    return kotlinFunction(args...);
+}
+
+// Calls the given function in the `Runnable` thread state. The function must be marked as RUNTIME_NORETURN.
+// If the function returns, behaviour is undefined.
+template <typename... Args>
+ALWAYS_INLINE RUNTIME_NORETURN inline void CallKotlinNoReturn(void(*noreturnKotlinFunction)(Args...), Args... args) {
+    CallKotlin(noreturnKotlinFunction, args...);
+    __builtin_unreachable();
```
projedi, line 429: It's a shame we can't require `noreturn` annotation on an argument, so maybe add `RuntimeFail(…)` here instead? I don't think it'll hurt performance, but will aid debugging if the called function in fact returns.

runtime/src/main/cpp/TestSupport.hpp
ilmat192: It's rather for [this](https://github.com/JetBrains/kotlin-native/blob/5778e42fec07ef2fd37856203febe94e0a89eff0/runtime/src/mm/cpp/TestSupport.hpp#L13) overload because new MM tests usually work with `ThreadData`, not `MemoryState`. Of course we could convert `MemoryState` to `ThreadData` at use sites but it looks like an unneeded boilerplate. 
I haven't find a way to keep both this overload and usage of a usual template.

runtime/src/main/cpp/Runtime.cpp
ilmat192: Ok, done.

runtime/src/main/cpp/TestSupport.hpp
ilmat192: Yes. My original intent was to make this "initialization for tests" as minimal as possible. But it seems that using the regular `InitMemory` will be more clear. Fixed.

runtime/src/main/cpp/Memory.h
ilmat192: Missed it, thank you!

runtime/src/mm/cpp/ThreadState.cpp
ilmat192: Sounds reasonable. Moved.

runtime/src/main/cpp/Memory.h
ilmat192: Done.

projedi (approved):

SvyatoslavScherbina (approved):

runtime/src/main/cpp/TestSupport.hpp
ilmat192: I thought about it but decided that having functions with the same name and almost the same behaviour in different namespaces will be less clear and less convenient.

projedi (approved):

PR #4729, opened 2021-03-02
[LEGACY MM] Fixed types in messages for tracing GC

LepilkinaElena:


runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -1394,7 +1394,7 @@ inline void decrementRC(ContainerHeader* container) {
         container->setBuffered();
         if (state->toFree != nullptr) {
           state->toFree->push_back(container);
-          MEMORY_LOG("toFree is now %d\n", state->toFree->size())
+          MEMORY_LOG("toFree is now %lu\n", state->toFree->size())
```
projedi, line 1397: Shouldn't everything `…->size()` should be `%zu`? Since `size()` returns `size_t` for std containers.

projedi (approved):

runtime/src/legacymm/cpp/Memory.cpp
LepilkinaElena: It was compiler suggestion to use here `%lu`, changed to `%zu`

PR #4731, opened 2021-03-02
[LEGACY MM] Escape making extra increments and decrements during coping to the same array

LepilkinaElena:


runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -2192,6 +2192,41 @@ void updateHeapRef(ObjHeader** location, const ObjHeader* object) {
...
+  }
+  *const_cast<const ObjHeader**>(ArrayAddressOfElementAt(destinationArray, toIndex + index)) = object;
+  // Release rewritten elements.
+  if (releaseHeapRefCondition && reinterpret_cast<uintptr_t>(old) > 1) {
+    releaseHeapRef<Strict>(old);
+  }
+}
+
+template <bool Strict>
+void updateHeapRefsInsideOneArray(const ArrayHeader* array, const ArrayHeader* destinationArray,
```
ilmat192, line 2214: Why do we need both array and destinationArray? I mean it is the same array, isn't it?

runtime/src/legacymm/cpp/Memory.cpp
```diff
@@ -2192,6 +2192,41 @@ void updateHeapRef(ObjHeader** location, const ObjHeader* object) {
   }
 }
 
+template <bool Strict>
+void updateHeapRefInsideArray(const ArrayHeader* array, const ArrayHeader* destinationArray, int index,
```
ilmat192, line 2196: Please correct me if I'm wrong, but it seems that we can simplify the logic. 

It seems that in both cases (`from >= to` and `from < to`) we need to:
1. Copy first `abs(from - to)` elements with old ref releasing.
2. Copy middle elements
3. Copy last `abs(from - to)` elements with ref adding.

The only difference is the direction of array traversing.

What if split each of the loops in `updateHeapRefsInsideOneArray` into 3 loops as stated above? It's just an IMHO, but this approach seems more readable to me than calculating the release/add condition for each element.

runtime/src/main/cpp/Arrays.cpp
```diff
@@ -133,15 +133,30 @@ void Kotlin_Array_copyImpl(KConstRef thiz, KInt fromIndex,
...
+      // Relocated elements.
+      for (int index = count - 1; index >= 0; index--) {
+        UpdateHeapRefsInsideOneArray(array, destinationArray, index, fromIndex, toIndex, count);
+      }
     }
   } else {
-    for (int index = count - 1; index >= 0; index--) {
-      UpdateHeapRef(ArrayAddressOfElementAt(destinationArray, toIndex + index),
-                      *ArrayAddressOfElementAt(array, fromIndex + index));
+    if (fromIndex >= toIndex) {
```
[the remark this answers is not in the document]
LepilkinaElena: > Doesn't this if become redundant because we always copy elements between different arrays here?

Didn't get it, this if for order in loop.

runtime/src/legacymm/cpp/Memory.cpp
LepilkinaElena: I also started to do like this, but the order of each of this 3 loops are different, but order is important for coping. And making releaseRef earlier isn't possible, because it calls GC and we can lose these objects before coping is made.

runtime/src/legacymm/cpp/Memory.cpp
LepilkinaElena: True

runtime/src/main/cpp/Arrays.cpp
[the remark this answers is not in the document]
LepilkinaElena: > CurrentMemoryModel != MemoryModel::kExperimental

Rewrote using this

runtime/src/legacymm/cpp/Memory.cpp
LepilkinaElena: I rewrote and clang better optimizes this code, it becomes much faster.

runtime/src/main/cpp/Arrays.cpp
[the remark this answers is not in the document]
ilmat192: >Didn't get it, this if for order in loop.

As far as I understand, this order matters only if we copy elements inside the same array. Because otherwise we could rewrite elements that have not been copied yet. But for different arrays we can always use the same order because there is no danger to rewrite something. So this `if` looked redundant in the previous version of the patch.

But with the Alexander's suggestion we still use this logic in the new MM. So we do need this `if` in the current version of the patch.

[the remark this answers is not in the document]
LepilkinaElena: Ok, I got your idea, it still can be one array if we copy parts without intersection `abs(fromIndex - toIndex) > count`, but the order really doesn't matter. But yes, now it's part for new MM

ilmat192 (approved):
Looks good to me. BTW, how difficult is adding a couple of tests for this optimization? I mean creating a proper array container and updating its RC without fully initialized legacy runtime.

projedi (approved):

PR #4733, opened 2021-03-03
[IR] Fixed problems with nested inline classes

homuroll:
Fixes https://youtrack.jetbrains.com/issue/KT-45139 as well

### DIFF
--- a/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
+++ b/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
@@ -42,8 +42,14 @@ private fun KonanSymbols.getTypeConversionImpl(
 internal object DECLARATION_ORIGIN_INLINE_CLASS_SPECIAL_FUNCTION : IrDeclarationOriginImpl("INLINE_CLASS_SPECIAL_FUNCTION")
 
 internal val Context.getBoxFunction: (IrClass) -> IrSimpleFunction by Context.lazyMapMember { inlinedClass ->
-    assert(inlinedClass.isUsedAsBoxClass())
-    assert(inlinedClass.parent is IrFile) { "Expected top level inline class" }
+    require(inlinedClass.isUsedAsBoxClass())
+    val classes = mutableListOf<IrClass>(inlinedClass)
+    var parent = inlinedClass.parent
+    while (parent is IrClass) {
+        classes.add(parent)
+        parent = parent.parent
+    }
+    require(parent is IrFile) { "Local inline classes are not supported" }
 
     val symbols = ir.symbols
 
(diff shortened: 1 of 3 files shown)


backend.native/tests/codegen/inlineClass/nestedInlineClasses.kt
```diff
@@ -0,0 +1,21 @@
...
+    inline class IC(val s: String)
+}
+
+interface I2 {
+    inline class IC(val s: String)
+}
+
+@Test fun runTest() {
+    assertEquals("zzz", I.IC("zzz").s)
+    assertEquals("zzz", I2.IC("zzz").s)
```
SvyatoslavScherbina, line 20: The test doesn't involve boxing, so it might leave a box/unbox functions clash undetected.

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
```diff
@@ -42,8 +42,7 @@ private fun KonanSymbols.getTypeConversionImpl(
 internal object DECLARATION_ORIGIN_INLINE_CLASS_SPECIAL_FUNCTION : IrDeclarationOriginImpl("INLINE_CLASS_SPECIAL_FUNCTION")
 
 internal val Context.getBoxFunction: (IrClass) -> IrSimpleFunction by Context.lazyMapMember { inlinedClass ->
-    assert(inlinedClass.isUsedAsBoxClass())
-    assert(inlinedClass.parent is IrFile) { "Expected top level inline class" }
+    require(inlinedClass.isUsedAsBoxClass())
```
SvyatoslavScherbina, line 45: Btw, we still rely on inline classes being non-local (more specifically, we expect that inline class is nested to the chain of classes, last of which is top-level).

backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/Boxing.kt
```diff
@@ -61,7 +60,7 @@ internal val Context.getBoxFunction: (IrClass) -> IrSimpleFunction by Context.la
             startOffset, endOffset,
             DECLARATION_ORIGIN_INLINE_CLASS_SPECIAL_FUNCTION,
             IrSimpleFunctionSymbolImpl(),
-            Name.special("<${inlinedClass.name}-box>"),
+            Name.special("<${inlinedClass.fqNameForIrSerialization}-box>"),
```
SvyatoslavScherbina, line 63: It makes some sense to make this less verbose, by using a "relative" class name (i.e. "fq name" that doesn't include package). This also might help to implement a missing pre-condition ([see above](https://github.com/JetBrains/kotlin-native/pull/4733/files#r586175920)), by extracting something like `IrClass.relativeNameFromTopLevel: String?` or whatever.

SvyatoslavScherbina (approved):

